PR #1, opened 2025-02-05
Support convenient global platforms for shims

jaybosamiya-ms:
Due to Rust language-design reasons, supporting a **global** runtime-parametric platform is not quite directly feasible. Thus, _either_ each shim above needs to hard-code which particular platform it works with, _or_ we need conditional compilation at each shim, _or_ we need to pass a parametric platform in each function, _or_ we have this particular PR. 

This PR localizes the impacts of the conditional compilation to a single multiplexing crate. Thus, the final application that is using LiteBox can pick the platform as a conditional compilation _independent_ of the particular shim it is using. Higher-level shims (e.g., POSIX shim) only needs to use the `litebox_platform_multiplex::platform()` function to get global access to the platform, without needing to pass it as arguments.


jaybosamiya-ms:
Note: ignore the semver checks CI check, since I was not using PRs before, the CI for it is not fully working. The PR itself should stand just fine without it (and is ready to review). I will have the semver checks CI thing fixed soon.

litebox_platform_multiplex/src/lib.rs
```diff
@@ -0,0 +1,53 @@
+//! A multiplexer for [LiteBox platforms](../litebox/platform/index.html), to simplify access to a
+//! global platform for shims "above" LiteBox.
+//!
+//! At a high level, due to Rust language design decisions, platform-dependent functionality either
+//! needs to all sit in a single crate
```
wdcui: Missing "or ..."?

jaybosamiya-ms: Oh oops, that got cut off when reshuffling some things. Will add that back, good catch.

jaybosamiya-ms:
Fixed the above comment. Also, rebased to work on top of the semver checks CI fixes, so semver checks _actually_ has meaning now. In this particular case, since this PR introduces a new crate, that _is_ semver breaking, so (as expected) the CI complains about this. We are safe to ignore the CI and merge in this case, since this is an expected one-time breakage (whenever we introduce a new crate).

PR #2, opened 2025-02-05
Begin initial implementation of the POSIX shim

jaybosamiya-ms:
This PR introduces the basic shape of the POSIX shim layer on top of LiteBox.  Currently, only `open` and `close` have been implemented, but should give an indication of how the rest of the code will end up looking like.

This PR is stacked on top of #1

### DIFF
--- a/litebox_shim_posix/src/lib.rs
+++ b/litebox_shim_posix/src/lib.rs
@@ -165,6 +165,9 @@ pub(crate) fn file_descriptors<'a>() -> &'a RwLock<'static, Platform, Descriptor
 ///
 /// `pathname` must point to a valid nul-terminated C string
 pub unsafe extern "C" fn open(pathname: *const i8, flags: u32, mode: u32) -> i32 {
+    if pathname.is_null() {
+        return -errno::constants::EFAULT;
+    }
     let path = unsafe { CStr::from_ptr(pathname) };
     match litebox_fs().open(
         path,


jaybosamiya-ms:
Note: ignore the semver checks CI check, since I was not using PRs before, the CI for it is not fully working. The PR itself should stand just fine without it (and is ready to review). I will have the semver checks CI thing fixed soon.

jaybosamiya-ms:
Note: semver checks CI has now been fixed, _but_ it will still complain on this PR since we are introducing a new crate. This is expected (for this PR), and should not be considered to actually be a CI failure.

litebox_shim_posix/src/errno.rs
```diff
@@ -0,0 +1,326 @@
+//! Error handling
+
+use constants::*;
+
+extern "C" {
+    fn __errno_location() -> *mut i32;
```
CvvT: Does this crate suppose to run in user mode with libc? For vsbox, does it require a different shim layer or it can build on top of this one?

jaybosamiya-ms: The shim does NOT assume that there is libc, but instead is _providing_ a libc-like interface (i.e., with a global errno). 

This can be confusing indeed, so probably making it more syscall-like is the right move. I'll remove the unnecessary indirection that is created here.

Vsbox is supposed to build using this shim layer and should NOT require a different shim.

jaybosamiya-ms: Updated to remove the `__errno_location()` indirection, and instead use negative values as error numbers, as would be expected from a syscall interface. This should keep things closer to what vsbox expects, and also removes the unnecessary global-errno libc-ism.  Thanks for catching this!

litebox_shim_posix/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+        } else {
+            idx as u32
+        }
+    }
+    fn remove(&mut self, fd: u32) -> Option<Descriptor> {
+        if fd >= (2 << 30) {
+            return None;
+        }
+        let fd = fd as usize;
+        self.descriptors[fd].take()
```
CvvT: Do we assume `fd` is always valid (in range)?

jaybosamiya-ms: Good catch. Within the main LiteBox code, we have a closed (unforgeable) newtype for FDs, which gives us strong guarantees on fds being valid. For the shim, since it goes externally as a regular integer, we actually need to check validity. Without the checks, while it would not have led to memory corruption, it would have definitely led to panics rather than properly giving a EBADF. I've updated this to do the relevant checks in the shim.

jaybosamiya-ms:
Marked @wdcui as reviewer too, since I believe you (Weidong) wanted the first few sets of PRs on this repo to have more than one reviewer (GitHub-free prevents us from marking more than one at the same time; but accepted reviews _can_ be maintained in parallel to a review request).

litebox_shim_posix/src/errno.rs
```diff
@@ -0,0 +1,315 @@
+//! Error handling
+
+use constants::*;
+
+pub(crate) trait AsErrno {
+    /// Returns the **positive** error number associated with `self`
+    fn as_errno(&self) -> i32;
```
wdcui, line 7: If it returns a positive error no, why do we use i32 not u32?

litebox_shim_posix/src/errno.rs
```diff
@@ -0,0 +1,315 @@
...
+    }
+}
+
+/// Error number constants, generated using
+///
+/// ```sh
+/// for i in $(seq 1 200); do echo $(errno $i) ; done | \
+///    awk '/#define/{n=$2; c=$3; $1=""; $2=""; $3=""; print "/// " substr($0,4) "\npub const " n ": i32 = " c ";"}'
+/// ```
+pub mod constants {
```
wdcui, line 54: Are these error numbers consistent with Linux?

litebox_shim_posix/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+            litebox_platform_multiplex::platform(),
+        ))
+    })
+}
+
+struct Descriptors {
+    descriptors: Vec<Option<Descriptor>>,
+}
+
+impl Descriptors {
```
wdcui, line 51: We need to populate the descriptors with stdin/stdout/stderr (anything else) first?

litebox_shim_posix/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+        .get_or_init(|| alloc::boxed::Box::new(litebox_sync().new_rwlock(Descriptors::new())))
+}
+
+/// Open a file
+///
+/// # Safety
+///
+/// `pathname` must point to a valid nul-terminated C string
+pub unsafe extern "C" fn open(pathname: *const i8, flags: u32, mode: u32) -> i32 {
+    let path = unsafe { CStr::from_ptr(pathname) };
```
wdcui, line 171: What's the implication of this unsafe call?

litebox_shim_posix/src/errno.rs
jaybosamiya-ms: Mostly out of ergonomics. Otherwise, we would need to write `-i32::try_from(err.as_errno()).unwrap()` or `-(err.as_errno() as i32)` everywhere in the syscalls themselves. With it being an `i32`, we are able to do `-err.as_errno()` and get the ergonomic convenience. The comment there is not a safety comment, but a reminder for users that things like `EBADF` and such are positive values, even though they sit in `i32`.

litebox_shim_posix/src/errno.rs
jaybosamiya-ms: Yes, these are the exact constants used in Linux, taken by running the script that I have documented.

litebox_shim_posix/src/lib.rs
jaybosamiya-ms: Yes, we will need to do that; I am leaving that for future PRs.

wdcui: It might help to add a comment/todo to track it.

litebox_shim_posix/src/lib.rs
jaybosamiya-ms: I already have a `# Safety` comment right above it for the function documenting the requirements. The expectations are exactly the same as that of the `const char* pathname` that the `open` syscall has on POSIX, in that it needs to actually be a valid nul-terminated string. As long as it _is_ such a valid string, this "unsafe" call is safe. Maybe I am not understanding your question here?

litebox_shim_posix/src/lib.rs
jaybosamiya-ms: Added TODO comment for it

litebox_shim_posix/src/lib.rs
wdcui: The input is given by the untrusted code. If we don't check it, would that be a problem?

jaybosamiya-ms: As the POSIX shim, we cannot change the interface (ideally, a safe interface would be an `&str` here), thus we are forced to use it. With the current interface, I don't believe it is possible for us to do any checks on the input in general (except a non-null check---I will add this shortly). Other than that, we can't know the length and whether every element in that space is valid without first doing a traversal starting from the pointer until a `\0`, but that is _exactly_ what `CStr::from_ptr` does.

wdcui: I guess the question is who should check it: the shim or litebox or the platform. I will resolve this one but we need to think about this question.

PR #8, opened 2025-02-11
Vsbox

CvvT:
Initial draft of platform for vsbox.

Run `cargo doc --features platform_snp --open` and navigate to the `host` module to see the interface design.

`HostPunchthroughProvider`: Interface for punchthrough requests
`HostPunchthroughToken`: A wrapper for PunchthroughToken that has lifetime
`HyperCallInterface`: Interface for hypercalls (See hypercall::HyperVInterface for example)

### DIFF
--- a/litebox-platform-linux-kernel/src/error.rs
+++ b/litebox-platform-linux-kernel/src/error.rs
@@ -1,0 +1,547 @@
+/// Operation not permitted
+pub const EPERM: i32 = 1;
+/// No such file or directory
+pub const ENOENT: i32 = 2;
+/// No such process
+pub const ESRCH: i32 = 3;
+/// Interrupted system call
+pub const EINTR: i32 = 4;
+/// Input/output error
+pub const EIO: i32 = 5;
+/// No such device or address
+pub const ENXIO: i32 = 6;
+/// Argument list too long
+pub const E2BIG: i32 = 7;
+/// Exec format error
+pub const ENOEXEC: i32 = 8;
+/// Bad file descriptor
+pub const EBADF: i32 = 9;
+/// No child processes
+pub const ECHILD: i32 = 10;
+/// Resource temporarily unavailable
+pub const EAGAIN: i32 = 11;
+/// Cannot allocate memory
+pub const ENOMEM: i32 = 12;
+/// Permission denied
+pub const EACCES: i32 = 13;
+/// Bad address
+pub const EFAULT: i32 = 14;
+/// Block device required
+pub const ENOTBLK: i32 = 15;
+/// Device or resource busy
+pub const EBUSY: i32 = 16;
+/// File exists
+pub const EEXIST: i32 = 17;
+/// Invalid cross-device link
+pub const EXDEV: i32 = 18;
+/// No such device
+pub const ENODEV: i32 = 19;
+/// Not a directory
+pub const ENOTDIR: i32 = 20;
+/// Is a directory
+pub const EISDIR: i32 = 21;
+/// Invalid argument
+pub const EINVAL: i32 = 22;
+/// Too many open files in system
+pub const ENFILE: i32 = 23;
+/// Too many open files
+pub const EMFILE: i32 = 24;
+/// Inappropriate ioctl for device
+pub const ENOTTY: i32 = 25;
+/// Text file busy
+pub const ETXTBSY: i32 = 26;
+/// File too large
+pub const EFBIG: i32 = 27;
+/// No space left on device
+pub const ENOSPC: i32 = 28;
+/// Illegal seek
+pub const ESPIPE: i32 = 29;
+/// Read-only file system
+pub const EROFS: i32 = 30;
+/// Too many links
+pub const EMLINK: i32 = 31;
+/// Broken pipe
+pub const EPIPE: i32 = 32;
+/// Numerical argument out of domain
+pub const EDOM: i32 = 33;
+/// Numerical result out of range
+pub const ERANGE: i32 = 34;
+/// Resource deadlock avoided
+pub const EDEADLOCK: i32 = 35;
+/// File name too long
+pub const ENAMETOOLONG: i32 = 36;
+/// No locks available
+pub const ENOLCK: i32 = 37;
+/// Function not implemented
+pub const ENOSYS: i32 = 38;
+/// Directory not empty
+pub const ENOTEMPTY: i32 = 39;
+/// Too many levels of symbolic links
+pub const ELOOP: i32 = 40;
+/// No message of desired type
+pub const ENOMSG: i32 = 42;
+/// Identifier removed
+pub const EIDRM: i32 = 43;
+/// Channel number out of range
+pub const ECHRNG: i32 = 44;
+/// Level 2 not synchronized
+pub const EL2NSYNC: i32 = 45;
+/// Level 3 halted
+pub const EL3HLT: i32 = 46;
+/// Level 3 reset
+pub const EL3RST: i32 = 47;
+/// Link number out of range
+pub const ELNRNG: i32 = 48;
+/// Protocol driver not attached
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -2,6 +2,7 @@
 resolver = "2"
 members = [
     "litebox",
+    "litebox-platform-linux-kernel",
```
wdcui, line 5: Shall we keep the names consistent with "_"?

litebox-platform-linux-kernel/src/error.rs
wdcui, line 1: I remember we have a similar error file for the posix shim. Can we use a single error file?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "litebox-platform-linux-kernel/sandbox_driver"]
+	path = litebox-platform-linux-kernel/sandbox_driver
```
wdcui, line 2: Indexing by 4 spaces or 8 spaces?

litebox-platform-linux-kernel/src/host/hypercall.rs
wdcui, line 1: Is the hypercall interface the same for SNP and VBS?

litebox-platform-linux-kernel/src/host/snp.rs
```diff
@@ -0,0 +1,474 @@
...
+            OtherPunchthrough::Syscall2(ref v) => v.into(),
+            OtherPunchthrough::Syscall3(ref v) => v.into(),
+            OtherPunchthrough::Syscall4(ref v) => v.into(),
+            OtherPunchthrough::Syscall5(ref v) => v.into(),
+            OtherPunchthrough::Syscall6(ref v) => v.into(),
+        }
+    }
+}
+
+const NR_SYSCALL_KILL: u64 = 62;
```
wdcui, line 152: Why do we define these two syscalls here? Do we need to define other syscall numbers as well?

litebox-platform-linux-kernel/src/host/snp.rs
```diff
@@ -0,0 +1,474 @@
...
+        $v.saved_args[5] = $v.pt_regs.r9;
+        $v.pt_regs.r9 = $v.args[5];
+    }};
+}
+
+pub struct SnpPunchthroughToken<'a> {
+    punchthrough: SnpPunchthrough<'a>,
+}
+
+impl SnpPunchthroughToken<'_> {
```
wdcui, line 276: Why do we define both HostPunchThrough and SnpPunchthrough?

litebox-platform-linux-kernel/src/host/hypercall.rs
wdcui: It's not clear to me why we need a host subfolder here.

litebox-platform-linux-kernel/src/host/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+
+/// Common punchthrough requests
+/// `Other` is for additional punchthrough requests (See [`snp::SnpPunchthrough`] for example)
+pub enum HostPunchthrough<'a, Other> {
+    /// Allocate pages ((2 ^ order) * PAGE_SIZE)
+    Alloc {
+        order: u64,
+    },
+    RecvPacket(&'a mut [u8]),
+    SendPacket(&'a [u8]),
```
jaybosamiya-ms, line 23: As I understand it, these are intended for the IP interface stuff, yes? In that case, it should not be part of punchthrough since we already have the `IPInterfaceProvider` trait.

litebox-platform-linux-kernel/src/error.rs
jaybosamiya-ms: I think this should be done in a separate PR to merge things into a shared crate `litebox_linux_common` or similar. I agree that currently this is essentially being duplicated.

litebox-platform-linux-kernel/src/host/snp.rs
```diff
@@ -0,0 +1,474 @@
...
+                    0,
+                ],
+            ),
+            OtherPunchthrough::Syscall0(ref v) => v.into(),
+            OtherPunchthrough::Syscall1(ref v) => v.into(),
+            OtherPunchthrough::Syscall2(ref v) => v.into(),
+            OtherPunchthrough::Syscall3(ref v) => v.into(),
+            OtherPunchthrough::Syscall4(ref v) => v.into(),
+            OtherPunchthrough::Syscall5(ref v) => v.into(),
+            OtherPunchthrough::Syscall6(ref v) => v.into(),
```
jaybosamiya-ms, line 147: Instead of `match *req`, if you use `match req` then I think you don't need these `ref`s (modern-ish Rust tends to use `ref`s _very_ rarely).

jaybosamiya-ms:
Quick question regarding the git-submodule for `sandbox_driver`; do we only need the `.h` files from there to set up bindings? If so, it _may_ be a better idea for us to copy in the `.h` files, rather than deal with the complexity that comes from git-submodules _especially_ for CI purposes (it is less of a complexity when things are public, but even then, it would lead to _every_ user of litebox automatically cloning every submodule, even if not used, simply because of how Cargo does things conservatively). Having a `sync_deps.sh` or such script would make it easier for us to use a recent version, but it would allow us to decouple things better.

litebox-platform-linux-kernel/src/error.rs
jaybosamiya-ms: (Fwiw, my suggestion for `litebox_linux_common` is a crate that also defines syscall punchthrough layer too, can explain more IRL if needed)

CvvT:
> Quick question regarding the git-submodule for `sandbox_driver`; do we only need the `.h` files from there to set up bindings? If so, it _may_ be a better idea for us to copy in the `.h` files, rather than deal with the complexity that comes from git-submodules _especially_ for CI purposes (it is less of a complexity when things are public, but even then, it would lead to _every_ user of litebox automatically cloning every submodule, even if not used, simply because of how Cargo does things conservatively). Having a `sync_deps.sh` or such script would make it easier for us to use a recent version, but it would allow us to decouple things better.

Thanks for the suggestion! Less complexity is better.

Cargo.toml
jaybosamiya-ms: Sorry, I am to blame for this; I used to use `-` before deciding to make everything consistent in https://github.com/MSRSSP/litebox/commit/1e6568d43721fe61907e9b38ffb989e96586423d by switching to underscores (which turns out to be handled better by tooling and also make more sense in `use ...` declarations when specifying crates). Going forward, just using `_` for all crate names should keep things nice, but yeah I believe @CvvT used `-` because I was using `-` at the point where he originally started working on this PR.

litebox-platform-linux-kernel/src/host/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+pub trait HostPunchthroughToken<'a, InOut>
+where
+    Self: PunchthroughToken,
+{
+    type HyperCallInterface: HyperCallInterface<InOut>;
+}
+
+/// Interface for punchthrough requests
+pub trait HostPunchthroughProvider<'a, InOut, Other> {
+    type Token: HostPunchthroughToken<'a, InOut>;
```
CvvT, line 47: @jaybosamiya-ms I was trying to use `PunchthroughProvider` but found that the associated `PunchthroughToken` does not have lifetime. The token may have shorter lifetime than the provider's.

jaybosamiya-ms: Discussed this IRL, we need to de-restrict the situations that are allowed here. Fixed by #9 which also demonstrates how a (borrowing) punchthrough and/or punchthroughtoken can be implemented.

CvvT:
Move to a new one: https://github.com/MSRSSP/litebox/pull/10

PR #9, opened 2025-02-12
Support punchthrough tokens that have lifetime restrictions

jaybosamiya-ms:
This PR adds a generic lifetime bound to the associated type for the `PunchthroughToken`, which allows us to more precisely capture lifetimes, and thereby de-restrict the conditions under which punchthrough tokens can be implemented (such as those with explicit lifetimes within them).

### DIFF
--- a/litebox/src/platform/mock.rs
+++ b/litebox/src/platform/mock.rs
@@ -115,13 +115,49 @@ impl TimeProvider for MockPlatform {
     }
 }
 
+pub(crate) struct MockPunchthrough<'a> {
+    msg: &'a str,
+}
+
+pub(crate) struct MockPunchthroughToken<'a> {
+    // Just to demonstrate that it can get access to everything necessary, both the platform, as
+    // well as the punchthrough object itself (which _itself_ has an associated lifetime).
+    platform: &'a mut MockPlatform,
+    punchthrough: MockPunchthrough<'a>,
+}
+
+impl Punchthrough for MockPunchthrough<'_> {
+    type ReturnSuccess = ();
+    type ReturnFailure = core::convert::Infallible;
+}
+
+impl<'a> PunchthroughToken for MockPunchthroughToken<'a> {
+    type Punchthrough = MockPunchthrough<'a>;
+
+    fn execute(
+        self,
+    ) -> Result<
+        <Self::Punchthrough as Punchthrough>::ReturnSuccess,
+        PunchthroughError<<Self::Punchthrough as Punchthrough>::ReturnFailure>,
+    > {
+        std::eprintln!("Punched through with {:?}", self.punchthrough.msg);
+        Ok(())
+    }
+}
+
 impl PunchthroughProvider for MockPlatform {
-    type PunchthroughToken = trivial_providers::ImpossiblePunchthroughToken;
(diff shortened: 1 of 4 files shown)


jaybosamiya-ms:
Ah, it looks like this particular feature breaks dyn-compatibility (thus the semver checks complaint).  I'll have to think a little bit if it is possible to maintain dyn compatibility while also supporting the restricted lifetimes we want to support.  I suspect it is not possible (we need to give up one or the other).  

In general, I don't expect us to ever want to have multiple runtime-generic platforms that are fully-runtime-dynamic-dispatched. Thus, I don't think we actually _need_ dyn-compatibility. Indeed, we would still be able to support multiple _static_ platforms at the same time if we wanted to, so this is not a major restriction. Thus, I feel we can give up on dyn compatibility here.

CvvT:
As discussed with Jay this afternoon, the current design assumes that punchthrough does not block. One example of blocking punchthrough is futex (might be the only one). Either we remove `mut` from `get_punchthrough_token_for` so we can get multiple tokens from the provider concurrently, or we consider futex is a special case that should be handled separately.

wdcui:
What's the downside of just removing `mut`?

> As discussed with Jay this afternoon, the current design assumes that punchthrough does not block. One example of blocking punchthrough is futex (might be the only one). Either we remove `mut` from `get_punchthrough_token_for` so we can get multiple tokens from the provider concurrently, or we consider futex is a special case that should be handled separately.

CvvT:
> Ah, it looks like this particular feature breaks dyn-compatibility (thus the semver checks complaint). I'll have to think a little bit if it is possible to maintain dyn compatibility while also supporting the restricted lifetimes we want to support. I suspect it is not possible (we need to give up one or the other).
> 
> In general, I don't expect us to ever want to have multiple runtime-generic platforms that are fully-runtime-dynamic-dispatched. Thus, I don't think we actually _need_ dyn-compatibility. Indeed, we would still be able to support multiple _static_ platforms at the same time if we wanted to, so this is not a major restriction. Thus, I feel we can give up on dyn compatibility here.

I drafted a new PR: https://github.com/MSRSSP/litebox/pull/10. I removed most punchthroughs and only kept syscalls, now it does not need to specify lifetime.

CvvT:
> What's the downside of just removing `mut`?
> 
> > As discussed with Jay this afternoon, the current design assumes that punchthrough does not block. One example of blocking punchthrough is futex (might be the only one). Either we remove `mut` from `get_punchthrough_token_for` so we can get multiple tokens from the provider concurrently, or we consider futex is a special case that should be handled separately.

Jay just told me futex could be supported in litebox given RawMutexProvider implementation and our single process assumption. Probably we won't have any blocking punchthroughs.

jaybosamiya-ms:
I'm closing this PR, since it appears that we don't need it. If necessary, we can resurrect it.

Keeping the `mut` around helps us be aware of situations where we might be using punchthroughs "badly". It is more of a "if you can get this to work with the `mut` then it is more likely you got the implementation correct; if we remove the `mut` then it is easier to make mistakes". Both with and without mut _could_ be correct (thus the more expressive approach would be _without_ `mut`), but I personally feel it is better to keep things restrictive until we have a good reason to derestrict it.

Issue #7, opened 2025-02-10
User-pointers support


jaybosamiya-ms:
Consider adding something to the platform for separate user-space pointers, which are allowed to be transparent for user-space platforms, but kernel-space platforms would ensure that copy-to-kernel/copy-to-user is happening at the right points.

Things like the POSIX shim would use only user-space pointers in their interface (thus they need to be `repr(C)` pointers in the first place), but this would transparently support our kernel and user space variants of litebox safely.

PR #10, opened 2025-02-12
Platform for Kernel

CvvT:
Per our discussion, `LinuxKernel` is the provider/platform which depends on `HostInterface` (host specific interface) and `Task` (for per task storage).

### DIFF
--- a/litebox_platform_linux_kernel/src/host/mod.rs
+++ b/litebox_platform_linux_kernel/src/host/mod.rs
@@ -1,4 +1,4 @@
 //! Different host implementations of [`super::HostInterface`]
 pub mod linux;
 #[cfg(feature = "platform_snp")]
-pub mod snp;
+mod snp;
--- a/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
+++ b/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
@@ -10,17 +10,20 @@ use crate::{
     HostInterface,
 };
 
-include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+#[allow(dead_code)]
+mod bindings {
+    include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+}
 
 const MAX_ARGS_SIZE: usize = 6;
 type ArgsArray = [u64; MAX_ARGS_SIZE];
 
-impl SnpVmplRequestArgs {
+impl bindings::SnpVmplRequestArgs {
     #[inline]
     fn new_request(code: u32, size: u32, args: ArgsArray) -> Self {
-        SnpVmplRequestArgs {
+        bindings::SnpVmplRequestArgs {
             code,
-            status: SNP_VMPL_REQ_INCOMPLETE,
+            status: bindings::SNP_VMPL_REQ_INCOMPLETE,
             size,
             padding: 0,
             args,
@@ -29,7 +32,11 @@ impl SnpVmplRequestArgs {
     }
 
     pub fn new_exit_request() -> Self {
-        SnpVmplRequestArgs::new_request(SNP_VMPL_EXIT_REQ, 0, ArgsArray::default())
+        bindings::SnpVmplRequestArgs::new_request(
+            bindings::SNP_VMPL_EXIT_REQ,
+            0,
+            ArgsArray::default(),
+        )
     }
 }
 
@@ -63,7 +70,7 @@ pub struct SyscallN<const N: usize, const ID: u32> {
 
 impl HostSnpInterface {
     /// [VTL CALL](https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/tlfs/vsm#vtl-call) via VMMCALL
-    fn request(arg: &mut SnpVmplRequestArgs) {
+    fn request(arg: &mut bindings::SnpVmplRequestArgs) {
         unsafe {
             asm!("vmmcall",
                 in("rcx") HVCALL_VTL_CALL,
@@ -77,9 +84,9 @@ impl HostSnpInterface {
     ) -> Result<usize, crate::error::Errno> {
         let mut args = [0; MAX_ARGS_SIZE];
         args[..N].copy_from_slice(&arg.args);
-        let mut req = SnpVmplRequestArgs::new_request(
+        let mut req = bindings::SnpVmplRequestArgs::new_request(
             // FIXME: need to update sandbox driver for the change of this interface
-            SNP_VMPL_SYSCALL_REQ,
+            bindings::SNP_VMPL_SYSCALL_REQ,
             ID, // repurpose size field to syscall id
             args,
         );
@@ -90,8 +97,11 @@ impl HostSnpInterface {
     /// To be used by [`Self::alloc_raw_mutex`]
     #[allow(dead_code)]
     fn alloc_futex_page() -> Result<u64, error::Errno> {
-        let mut req =
-            SnpVmplRequestArgs::new_request(SNP_VMPL_ALLOC_FUTEX_REQ, 0, [0, 0, 0, 0, 0, 0]);
+        let mut req = bindings::SnpVmplRequestArgs::new_request(
+            bindings::SNP_VMPL_ALLOC_FUTEX_REQ,
+            0,
+            [0, 0, 0, 0, 0, 0],
+        );
         Self::request(&mut req);
         Self::parse_alloc_result(0, req.ret)
     }
@@ -107,7 +117,7 @@ impl HostSnpInterface {
 
     fn parse_alloc_result(order: u32, addr: u64) -> Result<u64, crate::error::Errno> {
         if addr == 0 {
-            if order > SNP_VMPL_ALLOC_MAX_ORDER {
+            if order > bindings::SNP_VMPL_ALLOC_MAX_ORDER {
                 Err(error::Errno::EINVAL)
             } else {
                 Err(error::Errno::ENOMEM)
@@ -123,8 +133,8 @@ impl HostSnpInterface {
 
 impl HostInterface for HostSnpInterface {
     fn send_ip_packet(packet: &[u8]) -> Result<usize, crate::error::Errno> {
-        let mut req = SnpVmplRequestArgs::new_request(
-            SNP_VMPL_TUN_WRITE_REQ,
+        let mut req = bindings::SnpVmplRequestArgs::new_request(
+            bindings::SNP_VMPL_TUN_WRITE_REQ,
             3,
             [packet.as_ptr() as u64, packet.len() as u64, 0, 0, 0, 0],
         );
@@ -133,8 +143,8 @@ impl HostInterface for HostSnpInterface {
     }
 
     fn receive_ip_packet(packet: &mut [u8]) -> Result<usize, crate::error::Errno> {
-        let mut req = SnpVmplRequestArgs::new_request(
-            SNP_VMPL_TUN_READ_REQ,
+        let mut req = bindings::SnpVmplRequestArgs::new_request(
+            bindings::SNP_VMPL_TUN_READ_REQ,
             3,
             [packet.as_ptr() as u64, packet.len() as u64, 0, 0, 0, 0],
         );
@@ -147,14 +157,17 @@ impl HostInterface for HostSnpInterface {
     }
 
     fn alloc(order: u32) -> Result<u64, error::Errno> {
-        let mut req =
-            SnpVmplRequestArgs::new_request(SNP_VMPL_ALLOC_REQ, 1, [order as u64, 0, 0, 0, 0, 0]);
+        let mut req = bindings::SnpVmplRequestArgs::new_request(
+            bindings::SNP_VMPL_ALLOC_REQ,
+            1,
+            [order as u64, 0, 0, 0, 0, 0],
+        );
         Self::request(&mut req);
         Self::parse_alloc_result(order, req.ret)
     }
 
     fn exit() -> ! {
-        let mut req = SnpVmplRequestArgs::new_exit_request();
+        let mut req = bindings::SnpVmplRequestArgs::new_exit_request();
         Self::request(&mut req);
         loop {
             unsafe { asm!("hlt") }
@@ -162,8 +175,8 @@ impl HostInterface for HostSnpInterface {
     }
 
     fn terminate(reason_set: u64, reason_code: u64) -> ! {
-        let mut req = SnpVmplRequestArgs::new_request(
-            SNP_VMPL_TERMINATE_REQ,
+        let mut req = bindings::SnpVmplRequestArgs::new_request(
+            bindings::SNP_VMPL_TERMINATE_REQ,
             2,
             [reason_set, reason_code, 0, 0, 0, 0],
         );
--- a/litebox_platform_linux_kernel/src/host/snp/mod.rs
+++ b/litebox_platform_linux_kernel/src/host/snp/mod.rs
@@ -1,4 +1,4 @@
 //! An implementation of [`crate::HostInterface`] for SNP VMM
 
 mod ghcb;
-pub mod snp_impl;
+mod snp_impl;


CvvT:
As we discussed https://github.com/MSRSSP/litebox/pull/9, removed blocking punchthrough futex (which will be reimplemented in Litebox)

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [workspace]
 resolver = "2"
 members = [
-    "litebox",
+    "litebox", "litebox_platform_linux_kernel",
```
wdcui: Should this be on a separate line?

litebox_platform_linux_kernel/snp-sandbox.h
wdcui, line 1: Should this file be put inside the host subfolder?

litebox_platform_linux_kernel/snp-sandbox.h
```diff
@@ -0,0 +1,166 @@
...
+	uint32_t size;
+	/// padding
+	uint32_t padding;
+	/// arguments up to 6
+	uint64_t args[6];
+	/// return value
+	uint64_t ret;
+} __attribute__((__packed__));
+
+#define SNP_SANDBOX_IOCTL_DEBUG 0x6300 // _IO('c', 0)
```
wdcui: TODO: maybe split this header file to two parts and only copy the part related to litebox here? For instance, litebox platform doesn't need to know the driver's IOCTL code.

litebox_platform_linux_kernel/src/error.rs
wdcui, line 1: This file will be merged with the other error file and be put in a linux_common crate?

litebox_platform_linux_kernel/src/host/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! Different host implementations of [`super::HostInterface`]
+#[cfg(feature = "platform_snp")]
+pub(crate) mod ghcb;
+pub mod linux;
```
wdcui, line 2: This file is potentially snp specific since VBS and Hyperlight may not use these structures?

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+        &self,
+        val: u32,
+        time: core::time::Duration,
+    ) -> Result<litebox::platform::UnblockedOrTimedOut, ImmediatelyWokenUp> {
+        self.block_or_maybe_timeout(val, Some(time))
+    }
+}
+
+impl<Host: HostInterface, T: Task> RawMutex<Host, T> {
+    fn block_or_maybe_timeout(
```
wdcui, line 217: Why is this function not included in the "impl" above?

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+    /// For exit/terminate
+    ///
+    /// Exit allows to come back to handle some requests from host,
+    /// but it should not return back to the caller.
+    /// TODO: add a callback func as argument
+    fn exit();
+    fn terminate(reason_set: u64, reason_code: u64) -> !;
+
+    /// For Punchthrough
+    fn syscalls<const N: usize, const ID: u32>(arg: SyscallN<N, ID>)
```
wdcui: Do we expect these syscalls to be temporary (for instance, Hyperlight or VBS will not have these syscalls).

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+        -> Result<usize, error::Errno>;
+
+    /// For RawMutex
+    fn alloc_raw_mutex() -> *mut AtomicU32;
+
+    fn release_raw_mutex(mutex: *mut AtomicU32);
+
+    fn wake_many<T: Task>(mutex: *mut AtomicU32, n: usize) -> Result<usize, error::Errno> {
+        let mutex = T::current().unwrap().convert_mut_ptr_to_host(mutex);
+        Self::syscalls(crate::SyscallN::<6, NR_SYSCALL_FUTEX> {
```
wdcui: Hyperlight or VBS will not have this syscall support. What should be the host interface for the RawMutex support?

litebox_platform_linux_kernel/src/lib.rs
wdcui: I thought about this. I think we should have a specified list of syscalls or APIs we want the host to support.

litebox_platform_linux_kernel/snp-sandbox.h
CvvT: I don't think this file can be moved to a different folder.

litebox_platform_linux_kernel/src/error.rs
CvvT: I think so.

litebox_platform_linux_kernel/src/lib.rs
CvvT: The above "impl" is to implement the trait "RawMutex", no new methods can be implemented.

litebox_platform_linux_kernel/src/lib.rs
CvvT: The interface can call any syscalls, but "LinuxPunchthrough" specifies all allowed syscalls. Internally, the platform can use this interface to call any syscalls, but the upper user have to use "LinuxPunchthrough".

litebox_platform_linux_kernel/src/lib.rs
CvvT: My understanding is the interface is a superset of all host interface. One specific host implementation can override this default implementation. If it does not need it, it can just leave a "unreachable!".

litebox_platform_linux_kernel/src/host/mod.rs
CvvT: Some structs like pt_regs might be snp specific while others like "sigset_t" are used by the platform for syscall punchthrough https://github.com/MSRSSP/litebox/blob/7b50a1c7889a2e2233337fa730623dd640c400c5/litebox_platform_linux_kernel/src/lib.rs#L109

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+
+const NR_SYSCALL_FUTEX: u32 = 202;
+const NR_SYSCALL_RT_SIGPROCMASK: u32 = 14;
+
+/// Punchthrough for syscalls
+/// Note we assume all punchthroughs are non-blocking
+pub enum LinuxPunchthrough {
+    RtSigprocmask(SyscallN<4, NR_SYSCALL_RT_SIGPROCMASK>),
+    // TODO: Add more syscalls
+}
```
jaybosamiya-ms: I would feel a little more comfortable if the specific arguments were written out here, e.g.,
```rs
RtSigprocmask {
  how: i32,
  set: *const sigset_t,
  oldset: *mut sigset_t,
  sigsetsize: usize,
}
```
Generally speaking, we can afford to be explicit with things, so that we don't have to look at the manpages to make sure things are correct and in the right order.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+    /// For memory allocation
+    fn alloc(order: u32) -> Result<u64, error::Errno>;
+
+    /// For exit/terminate
+    ///
+    /// Exit allows to come back to handle some requests from host,
+    /// but it should not return back to the caller.
+    /// TODO: add a callback func as argument
+    fn exit();
+    fn terminate(reason_set: u64, reason_code: u64) -> !;
```
jaybosamiya-ms: I'm not sure I understand this comment. I agree that `exit` can cause unwinding, but I don't understand the callback function argument.

Also, `exit()` should have the signature `exit() -> !`, right?

Finally, the semantics of doc-comments is that it would be associated only with `exit` and not for `terminate`. The `terminate` function is left without any documentation comments. I don't think that is what you intended here.

litebox_platform_linux_kernel/src/lib.rs
jaybosamiya-ms: I think I agree with Weidong, I would feel more comfortable if there was an explicit set of syscalls being punched through. I don't think we should block merging this on that (esp since we'll hopefully soon have `litebox_linux_common` crate). I think it is fine if it is a single function (if you think that is simpler), but the syscalls should be in an explicit enum, rather than the generic "any syscalls" portion.

litebox_platform_linux_kernel/src/lib.rs
jaybosamiya-ms: I don't know if the implementations of `wake_many`/`block_or_maybe_timeout` should be overridable-defaults, or defined purely at the host.

Currently, we have two host implementations in mind: will both use the same code (without overriding)? If so, we should not even have it be part of the host interface. If at least one will override it, then we shouldn't define it as a default, and instead place it directly into the hands of each of the hosts.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+            );
+        }
+        ((hi as u64) << 32) | (lo as u64)
+    }
+
+    fn now(cpu_mhz: u64) -> Self {
+        let tsc = Self::rdtsc();
+        let ms = tsc / cpu_mhz;
+        Instant(ms)
+    }
```
jaybosamiya-ms: If we are running on the machine with varying frequency, this could break monotonicity, right? I don't see why we need to bother dividing by `cpu_mhz` except when doing a `checked_duration_since`. That way, we don't break monotonicity of `Instant`s, although durations (on a varying-frequency-processor) would be "stretchy" (as expected).

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+
+/// Analogous to `task struct` in Linux
+pub trait Task {
+    fn current<'a>() -> Option<&'a mut Self>;
+
+    /// Shared memory may be mapped to different address spaces in host and guest.
+    /// This function is to convert the pointer to host address space.
+    fn convert_ptr_to_host<T>(&self, ptr: *const T) -> *const T;
+    /// Similar to [`Self::convert_ptr_to_host`], but for mutable pointers
+    fn convert_mut_ptr_to_host<T>(&self, ptr: *mut T) -> *mut T;
```
jaybosamiya-ms: This will be superseded by an upcoming PR from me, related to #7. For now, I think it is fine to have in this PR, but just mentioning that this will need an update soon-ish.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+/// It requires a host that implements the [`HostInterface`]
+/// and a task that implements the [`Task`] trait.
+pub struct LinuxKernel<Host: HostInterface, T: Task> {
+    cpu_mhz: u64,
+    host_and_task: core::marker::PhantomData<(Host, T)>,
+}
+
+/// Analogous to `task struct` in Linux
+pub trait Task {
+    fn current<'a>() -> Option<&'a mut Self>;
```
jaybosamiya-ms: This function might cause undefined behavior, and thus is unsafe to use. In particular, nothing prevents a function from saying:
```rs
let a = current();
let b = current();
a.whatever();
b.whatever();
```
Specifically, if `a` and `b` are the same task struct (which is what you seem to be attempting to do) then this is a double mutable borrow, which is immediately UB.

I don't think you need it to actually be mutable, with the usage that I am seeing. If you made it a read-only reference, then it should be fine.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -0,0 +1,208 @@
...
+impl Task for vsbox_task {
+    fn current<'a>() -> Option<&'a mut Self> {
+        let task: u64;
+        unsafe {
+            asm!("rdgsbase {}", out(reg) task, options(nostack, preserves_flags));
+
+            if task == 0 {
+                return None;
+            }
+            Some(&mut *(task as *mut Self))
```
jaybosamiya-ms: See my comment on `current` in the definition of the trait, but essentially this line is what introduces the UB. It is sitting inside `unsafe` so it is able to cast the `*mut` to a `&mut`, but there is nothing that prevents/locks this behavior to prevent this happening twice at once.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+        Self::RawMutex {
+            inner: Host::alloc_raw_mutex(),
+            host: core::marker::PhantomData,
+        }
+    }
+}
+
+/// An implementation of [`litebox::platform::RawMutex`]
+pub struct RawMutex<Host: HostInterface, T: Task> {
+    inner: *mut AtomicU32,
```
jaybosamiya-ms: The design of the litebox `RawMutex` trait does not require us to actually hold on to a pointer to an inner variable for things to be safe/correct. I believe it would be fine to change this to `AtomicU32` to remove the alloc/free of raw mutexes. If you don't believe this to be the case, we should chat IRL to make sure we've got the details right.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+impl<Host: HostInterface, T: Task> Drop for RawMutex<Host, T> {
+    fn drop(&mut self) {
+        Host::release_raw_mutex(self.inner);
+    }
+}
+
+/// TODO: common mutex implementation could be moved to a shared crate
+impl<Host: HostInterface, T: Task> litebox::platform::RawMutex for RawMutex<Host, T> {
+    fn underlying_atomic(&self) -> &core::sync::atomic::AtomicU32 {
+        unsafe { &mut *self.inner }
```
jaybosamiya-ms: Why is this doing a `&mut` here? We just need a shared ref to it.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
jaybosamiya-ms: Generally speaking, this is why I prefer having `unsafe` blocks be _incredibly_ tiny each time I use them. It would have made this particular issue easier to identify.

litebox_platform_linux_kernel/snp-sandbox.h
jaybosamiya-ms: Presumably, this file would sit within the (future) `litebox_host_snp` crate/folder, right? I am not sure why it could not be moved.

jaybosamiya-ms: Sidenote: I don't know if it the intention is that all the hosts would sit within this same `litebox_platform_linux_kernel` crate, or would each be their own crate.

CvvT: ops, I thought you're talking about a different file. Moved to the snp folder.

litebox_platform_linux_kernel/src/lib.rs
CvvT: I agree. For now I can change it to read-only. In vsbox, we do have cases where I need to modify it. This might be fine for vsbox as it runs in kernel mode with no preemption and it is per-cpu (no concurrency issue).

litebox_platform_linux_kernel/src/lib.rs
CvvT: I'm not sure if it is the same. The purpose of these two is to convert a kernel pointer to a pointer that host can access as they may have different page tables. In the future, I may also need to convert a user pointer to one that host can access.

litebox_platform_linux_kernel/src/lib.rs
CvvT: It might be possible to just use AtomicU32. In vsbox, we use *mut AtomicU32 because we have a special allocator which may not be necessary. For simplicity, changed it to `AtomicU32` unless I found it does not work in the future.

litebox_platform_linux_kernel/src/lib.rs
CvvT: Good point!

litebox_platform_linux_kernel/src/lib.rs
CvvT: I was thinking about letting the upper user to decide how to handle the request from host after exit by passing a callback function. Removed it for now for simplicity. When I have an actual use case, it might be clearer.

litebox_platform_linux_kernel/src/lib.rs
CvvT: VBS may not need this interface. Moved the implementation to snp.

litebox_platform_linux_kernel/src/lib.rs
jaybosamiya-ms: Ah, I see what you mean. The user-pointers thing converts user<->kernel, while this pair of functions converts kernel<->host. If that understanding is correct, then maybe it would help updating the comment above it to say `... in host and guest kernel`?

litebox_platform_linux_kernel/src/lib.rs
jaybosamiya-ms: The particular situation I pointed out earlier does not depend on concurrency to cause the UB. If we do need mutable access, then we need to place it behind a lock (or at minimum, a refcell) to prevent issues.

Since it has been changed to read-only, I will mark this as resolved, but in the future, if we want to make it mutable, we'd need to discuss things a bit more.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+    PunchthroughError, PunchthroughProvider, PunchthroughToken, RawMutexProvider, TimeProvider,
+    UnblockedOrTimedOut,
+};
+
+extern crate alloc;
+
+pub mod error;
+pub mod host;
+
+static CPU_MHZ: once_cell::race::OnceBox<u64> = once_cell::race::OnceBox::new();
```
jaybosamiya-ms: This seems like an overkill use for `once_cell`. We could replace this with an `AtomicU64` and remove the dependency on `once_cell`. 

Additionally, we can actually stop having this be a global, place it inside the `LinuxKernel` struct (thus it doesn't even need to be an `AtomicU64` and can be a `u64` directly, which can be stored inside the `Instant`, thereby allowing for entirely removing the global). I think this step is overkill, and the global is likely fine. But yeah, replacing it with a `AtomicU64` would probably be a good idea.

CvvT: Good idea! I'll go with AtomicU64 for now.

litebox_platform_linux_kernel/src/lib.rs
CvvT: Sure.

litebox_platform_linux_kernel/src/lib.rs
CvvT: Another similar example is `pt_regs` which is shared and modified by both host kernel (linux) and vsbox kernel. I'm not sure if lock or refcell in Rust is compatible with C. Essentially, we need a container that has `from_raw_pointer` where the pointer is from C. If there is not any, we can implement our own.

CvvT: One solution might be we get the Task once at the entry and pass it to all functions that require it.

CvvT:
Removed Task trait from this PR. My initial thought was that we need it to help convert a kernel pointer to a pointer accessible to the host. However, I can also change the sandbox driver to do it.

litebox_platform_linux_kernel/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[features]
+platform_snp = []
+
+[build-dependencies]
+bindgen = "0.71.0"
+
+[dependencies]
+litebox = { path = "../litebox/", version = "0.1.0" }
+once_cell = { version = "1.20.3", default-features = false, features = ["alloc", "race"] }
```
jaybosamiya-ms: Minor:
```suggestion
```

CvvT:
Rebase on the main branch and add implementation for `RawPointerProvider` (See https://github.com/MSRSSP/litebox/blob/8ad1533d8031851fa42f5438701c0c458dea3d6c/litebox_platform_linux_kernel/src/ptr.rs). To avoid complicate this PR, left `todo`s in the function body.

Another change I did due to `RawPointerProvider` is the following:
```diff
     fn rt_sigprocmask(
         how: i32,
-        set: Option<*const crate::host::linux::sigset_t>,
-        oldset: Option<*mut crate::host::linux::sigset_t>,
+        set: UserConstPtr<sigset_t>,
+        oldset: UserMutPtr<sigset_t>,
         sigsetsize: usize,
     ) -> Result<usize, error::Errno> {
```

Also, instead of passing user space pointer to host, I believe it is preferable to do some extra copy and pass kernel memory instead. See more comments here: https://github.com/MSRSSP/litebox/blob/8ad1533d8031851fa42f5438701c0c458dea3d6c/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs#L185-L195

PR #11, opened 2025-02-13
Add a pointers interface to transparently handle user/kernel split

### LINKED ISSUE #7: User-pointers support

jaybosamiya-ms:
This PR introduces a `RawPointerProvider` trait interface to the platform which the platform can use to support a user/kernel pointer split (i.e., supporting a copy-to-user/copy-to-kernel functionality basically transparently). The PR also implements "transparent pointers" for use in pure user-land platforms that don't have a meaningful user/kernel split. Finally, it updates relevant parts of the codebase to account for this new API.

Morally, a `RawConstPointer<T>` is a `*const T` _in userland_, while a bare `*const T` (without anything else specified) is intended to be a kernel pointer (again, only in platforms where this distinction makes sense; in platforms where it doesn't, both are identical).

Closes #7

### DIFF
--- a/litebox_shim_posix/src/lib.rs
+++ b/litebox_shim_posix/src/lib.rs
@@ -11,8 +11,6 @@
 
 extern crate alloc;
 
-use core::ffi::CStr;
-
 use alloc::vec;
 use alloc::vec::Vec;
 
@@ -20,7 +18,7 @@ use alloc::vec::Vec;
 // platform-specific things within it.
 use once_cell::race::OnceBox;
 
-use litebox::{fs::FileSystem as _, sync::RwLock};
+use litebox::{fs::FileSystem as _, platform::RawConstPointer as _, sync::RwLock};
 use litebox_platform_multiplex::Platform;
 
 pub mod errno;
@@ -43,6 +41,10 @@ pub(crate) fn litebox_sync<'a>() -> &'a litebox::sync::Synchronization<'static, 
         ))
     })
 }
+
+// Convenience type aliases
+type ConstPtr<T> = <Platform as litebox::platform::RawPointerProvider>::RawConstPointer<T>;
+type MutPtr<T> = <Platform as litebox::platform::RawPointerProvider>::RawMutPointer<T>;
 
 struct Descriptors {
     descriptors: Vec<Option<Descriptor>>,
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
Maybe I should update the SemverChecks CI to be "warn but don't mark an `X`" for now? I think it is a very useful thing to have in the CI, since it points out important possible breakages, but currently since we are working on the absolutely early pre-release, we don't actually want it to be complaining so much. EDIT: I have updated the semver checks with this in #12

litebox/src/platform/mod.rs
```diff
@@ -260,3 +265,165 @@ pub trait DebugLogProvider {
     /// `debug_log_print` call.
     fn debug_log_print(&self, msg: &str);
 }
+
+/// A common interface for raw pointers, aimed at usage in shims _above_ LiteBox.
+///
+/// Essentially, these types indicate "user" pointers (which are allowed to be null). Platforms with
+/// no meaningful user-kernel separation can use [`trivial_providers::TransparentConstPtr`] and
+/// [`trivial_providers::TransparentMutPtr`]. Platforms with meaningful user-kernal separation
+/// should define their own `repr(C)` newtype wrappers that perform relevant copying between user
```
wdcui, line 274: Minor: newtype => new type?

jaybosamiya-ms: "Newtype" (without the space) is a standard name of a specific pattern for defining types: https://doc.rust-lang.org/book/ch19-04-advanced-types.html#using-the-newtype-pattern-for-type-safety-and-abstraction

litebox/src/platform/mod.rs
```diff
@@ -260,3 +265,165 @@ pub trait DebugLogProvider {
...
+    /// Read the value of the pointer at signed offset from it.
+    ///
+    /// Returns `None` if the provided pointer is invalid, or such an offset is known (in advance)
+    /// to be invalid.
+    ///
+    /// # Safety
+    ///
+    /// The pointer (and underlying memory for the value at the offset) should remain valid and
+    /// unchanged for the entirety of the lifetime that the borrow (if any) is made.
+    unsafe fn read_at_offset<'a>(self, count: isize) -> Option<alloc::borrow::Cow<'a, T>>;
```
CvvT, line 297: @jaybosamiya-ms Should we add an interface `is_null`? Sometimes we need to distinguish between null pointer and unmapped memory.

jaybosamiya-ms: This interface is meant to be used mostly within the shim layer. Maybe I am missing a specific use case where the shim would need to be aware of this difference. Happy to add an `is_null` to the trait if there is such a use case, but at least right now I am unable to think of a case where this matters (at the shim layer; I would _completely_ understand it mattering at the platform or lower layers, but those layers are fully aware of the pointer details, so can do extra checks internally if needed).

PR #13, opened 2025-02-19
Allocator

CvvT:
Add global allocator using buddy system allocator and slabmalloc.

<del>The upstream slabmalloc is broken if using the latest stable rust: https://github.com/gz/rust-slabmalloc/pull/14/files. Currently, I used a forked version with the fix.</del>

### DIFF
--- a/litebox_platform_linux_kernel/src/host/mock.rs
+++ b/litebox_platform_linux_kernel/src/host/mock.rs
@@ -26,7 +26,7 @@ impl HostInterface for MockHostInterface {
         Ok((addr as usize, size))
     }
 
-    fn free(addr: usize) {
+    unsafe fn free(addr: usize) {
         unsafe { libc::free(addr as *mut _) };
     }
 
--- a/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
+++ b/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
@@ -40,8 +40,8 @@ impl crate::mm::MemoryProvider for SnpLinuxKenrel {
         HostSnpInterface::alloc(layout)
     }
 
-    fn free(addr: usize) {
-        HostSnpInterface::free(addr)
+    unsafe fn free(addr: usize) {
+        unsafe { HostSnpInterface::free(addr) }
     }
 }
 
@@ -191,7 +191,7 @@ impl HostInterface for HostSnpInterface {
         })
     }
 
-    fn free(_addr: usize) {
+    unsafe fn free(_addr: usize) {
         unimplemented!()
     }
 
--- a/litebox_platform_linux_kernel/src/lib.rs
+++ b/litebox_platform_linux_kernel/src/lib.rs
@@ -162,7 +162,7 @@ impl<Host: HostInterface> LinuxKernel<Host> {
         token.execute()
     }
 
-    pub fn exit(&self) {
+    pub fn exit(&self) -> ! {
         Host::exit();
     }
 
@@ -354,7 +354,11 @@ pub trait HostInterface {
     ///
     /// Note host should know the size of allocated memory and needs to check the validity
     /// of the given address.
-    fn free(addr: usize);
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the `addr` is valid and was allocated by this [`Self::alloc`].
+    unsafe fn free(addr: usize);
 
     /// Exit
     ///
--- a/litebox_platform_linux_kernel/src/mm/mod.rs
+++ b/litebox_platform_linux_kernel/src/mm/mod.rs
@@ -19,7 +19,11 @@ pub trait MemoryProvider {
     fn alloc(layout: &core::alloc::Layout) -> Result<(usize, usize), crate::error::Errno>;
 
     /// Returns the memory back to host.
-    fn free(addr: usize);
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the `addr` is valid and was allocated by [`Self::alloc`].
+    unsafe fn free(addr: usize);
 
     /// Called to refill the buddy allocator when OOM occurs.
     fn rescue_heap<const ORDER: usize>(heap: &mut Heap<ORDER>, layout: &core::alloc::Layout) {
--- a/litebox_platform_linux_kernel/src/mm/tests.rs
+++ b/litebox_platform_linux_kernel/src/mm/tests.rs
@@ -23,8 +23,8 @@ impl super::MemoryProvider for MockKernel {
         unsafe { ALLOCATOR.free_pages(ptr, order) }
     }
 
-    fn free(addr: usize) {
-        MockHostInterface::free(addr);
+    unsafe fn free(addr: usize) {
+        unsafe { MockHostInterface::free(addr) };
     }
 }
 


litebox_platform_linux_kernel/Cargo.toml
```diff
@@ -10,4 +10,9 @@ platform_snp = []
 bindgen = "0.71.0"
 
 [dependencies]
+buddy_system_allocator = { version = "0.11.0", default-features = false }
+slabmalloc = { git = "https://github.com/MSRSSP/rust-slabmalloc.git", branch = "master" }
```
wdcui: If we use the master branch, do we need to specify it?

litebox_platform_linux_kernel/src/host/mock.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    fn block_or_maybe_timeout(
+        _mutex: &AtomicU32,
+        _val: u32,
+        _timeout: Option<core::time::Duration>,
+    ) -> Result<(), crate::error::Errno> {
+        todo!()
+    }
+
+    fn rt_sigprocmask(
```
wdcui, line 65: Just a side thought: we shouldn't have signal handling as part of the host interface?

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -339,8 +346,11 @@ impl<Host: HostInterface> IPInterfaceProvider for LinuxKernel<Host> {
 
 /// Platform-Host Interface
 pub trait HostInterface {
-    /// For memory allocation
-    fn alloc(order: u32) -> Result<u64, error::Errno>;
+    /// For page allocation.
+    ///
+    /// It can return more than requested size. On success, it returns the start address
+    /// and the size of the allocated memory.
+    fn alloc(layout: &core::alloc::Layout) -> Result<(usize, usize), error::Errno>;
```
wdcui, line 351: Do we need to allow free() as well?

litebox_platform_linux_kernel/src/mm/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    ///
+    /// Note this is only called by [`Self::rescue_heap`] when the buddy allocator is out of memory.
+    ///
+    /// It can return more than requested size. On success, it returns the start address
+    /// and the size of the allocated memory.
+    fn alloc(layout: &core::alloc::Layout) -> Result<(usize, usize), crate::error::Errno>;
+
+    /// Called to refill the buddy allocator when OOM occurs.
+    fn rescue_heap<const ORDER: usize>(heap: &mut Heap<ORDER>, layout: &core::alloc::Layout) {
+        match Self::alloc(layout) {
```
wdcui, line 30: In the VSBox code, we do a fixed allocation from the host instead of using the layout to avoid allocating many small pages.

litebox_platform_linux_kernel/src/mm/slab.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub(crate) unsafe fn free_pages(&self, ptr: *mut u8, order: usize) {
+        self.buddy_allocator.dealloc(
+            ptr,
+            Layout::from_size_align(Self::BASE_PAGE_SIZE << order, Self::BASE_PAGE_SIZE << order)
+                .unwrap(),
+        );
+    }
+}
+
+unsafe impl<const ORDER: usize, Platform: MemoryProvider> GlobalAlloc
```
wdcui: I found it a little confusing to use Platform for MemoryProvider. I thought Platform is for the platform layer.

litebox_platform_linux_kernel/src/mm/slab.rs
```diff
@@ -0,0 +1,145 @@
...
+                    Layout::from_size_align(Self::BASE_PAGE_SIZE, Self::BASE_PAGE_SIZE).unwrap(),
+                )
+                .expect("allocate page"),
+            Self::LARGE_PAGE_SIZE => self
+                .buddy_allocator
+                .alloc_pages(
+                    Layout::from_size_align(Self::LARGE_PAGE_SIZE, Self::LARGE_PAGE_SIZE).unwrap(),
+                )
+                .expect("allocate large page"),
+            0..=super::zone::ZoneAllocator::MAX_ALLOC_SIZE => {
```
wdcui: What's the assumption of the relationship between LAGE_PAGE_SIZE and MAX_ALLOC_SIZE?

litebox_platform_linux_kernel/src/mm/zone.rs
wdcui, line 1: Why do we need to copy this file? I don't see where the code uses the ZoneAllocator.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -156,14 +167,23 @@ impl HostInterface for HostSnpInterface {
         ghcb_prints(msg);
     }
 
-    fn alloc(order: u32) -> Result<u64, error::Errno> {
+    fn alloc(layout: &core::alloc::Layout) -> Result<(usize, usize), error::Errno> {
+        // allocate at least one page
+        let size = core::cmp::max(layout.size().next_power_of_two(), PAGE_SIZE as usize);
```
CvvT: Add some comments

jaybosamiya-ms:
I haven't yet done a full review of the code, but wrt the fork, the following diff removes dependency on maintaining a fork (both `cargo check` and `cargo clippy` succeed with this):
```diff
diff --git a/litebox_platform_linux_kernel/Cargo.toml b/litebox_platform_linux_kernel/Cargo.toml
index 8f885e0d6c..c9522876f4 100644
--- a/litebox_platform_linux_kernel/Cargo.toml
+++ b/litebox_platform_linux_kernel/Cargo.toml
@@ -11,7 +11,7 @@

 [dependencies]
 buddy_system_allocator = { version = "0.11.0", default-features = false }
-slabmalloc = { git = "https://github.com/MSRSSP/rust-slabmalloc.git", branch = "master" }
+slabmalloc = { version = "0.11.0", default-features = false }
 litebox = { path = "../litebox/", version = "0.1.0" }
 spin = { version = "0.9.8", default-features = false, features = ["spin_mutex"] }
 libc = "0.2.169"
diff --git a/litebox_platform_linux_kernel/src/mm/slab.rs b/litebox_platform_linux_kernel/src/mm/slab.rs
index 2ec3a2dba2..9e5d5a9621 100644
--- a/litebox_platform_linux_kernel/src/mm/slab.rs
+++ b/litebox_platform_linux_kernel/src/mm/slab.rs
@@ -28,7 +28,7 @@
     const BASE_PAGE_SIZE: usize = 4096;
     const LARGE_PAGE_SIZE: usize = 2 * 1024 * 1024;

-    pub const fn new() -> Self {
+    pub fn new() -> Self {
         LockedSlabAllocator {
             buddy_allocator: LockedHeapWithRescue::new(),
             slab_allocator: super::zone::ZoneAllocator::new(),
diff --git a/litebox_platform_linux_kernel/src/mm/zone.rs b/litebox_platform_linux_kernel/src/mm/zone.rs
index 507417e6d0..39d30d7921 100644
--- a/litebox_platform_linux_kernel/src/mm/zone.rs
+++ b/litebox_platform_linux_kernel/src/mm/zone.rs
@@ -39,7 +39,7 @@
     /// How many allocators of type [`SCAllocator<LargeObjectPage>`] we have.
     const MAX_LARGE_SIZE_CLASSES: usize = 9;

-    pub(super) const fn new() -> Self {
+    pub(super) fn new() -> Self {
         ZoneAllocator {
             small_slabs: [
                 SpinMutex::new(SCAllocator::new(8)),
```

CvvT:
> I haven't yet done a full review of the code, but wrt the fork, the following diff removes dependency on maintaining a fork (both `cargo check` and `cargo clippy` succeed with this):
> 
> ```diff
> diff --git a/litebox_platform_linux_kernel/Cargo.toml b/litebox_platform_linux_kernel/Cargo.toml
> index 8f885e0d6c..c9522876f4 100644
> --- a/litebox_platform_linux_kernel/Cargo.toml
> +++ b/litebox_platform_linux_kernel/Cargo.toml
> @@ -11,7 +11,7 @@
> 
>  [dependencies]
>  buddy_system_allocator = { version = "0.11.0", default-features = false }
> -slabmalloc = { git = "https://github.com/MSRSSP/rust-slabmalloc.git", branch = "master" }
> +slabmalloc = { version = "0.11.0", default-features = false }
>  litebox = { path = "../litebox/", version = "0.1.0" }
>  spin = { version = "0.9.8", default-features = false, features = ["spin_mutex"] }
>  libc = "0.2.169"
> diff --git a/litebox_platform_linux_kernel/src/mm/slab.rs b/litebox_platform_linux_kernel/src/mm/slab.rs
> index 2ec3a2dba2..9e5d5a9621 100644
> --- a/litebox_platform_linux_kernel/src/mm/slab.rs
> +++ b/litebox_platform_linux_kernel/src/mm/slab.rs
> @@ -28,7 +28,7 @@
>      const BASE_PAGE_SIZE: usize = 4096;
>      const LARGE_PAGE_SIZE: usize = 2 * 1024 * 1024;
> 
> -    pub const fn new() -> Self {
> +    pub fn new() -> Self {
>          LockedSlabAllocator {
>              buddy_allocator: LockedHeapWithRescue::new(),
>              slab_allocator: super::zone::ZoneAllocator::new(),
> diff --git a/litebox_platform_linux_kernel/src/mm/zone.rs b/litebox_platform_linux_kernel/src/mm/zone.rs
> index 507417e6d0..39d30d7921 100644
> --- a/litebox_platform_linux_kernel/src/mm/zone.rs
> +++ b/litebox_platform_linux_kernel/src/mm/zone.rs
> @@ -39,7 +39,7 @@
>      /// How many allocators of type [`SCAllocator<LargeObjectPage>`] we have.
>      const MAX_LARGE_SIZE_CLASSES: usize = 9;
> 
> -    pub(super) const fn new() -> Self {
> +    pub(super) fn new() -> Self {
>          ZoneAllocator {
>              small_slabs: [
>                  SpinMutex::new(SCAllocator::new(8)),
> ```

I made these functions const to initialize the global allocator: https://github.com/MSRSSP/litebox/blob/9fe54c2d359f4b585b166c838052a6c9db9350f6/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs#L26
My understanding is that with `const` all objects are pre-allocated so there is no circular dependency on allocator.

jaybosamiya-ms:
Even without `const`, it won't cause a circular dependency on the allocator _unless_ one of the functions internally invokes the allocator (i.e., `const` or not does not _by itself_ define whether it allocates or doesn't); I do agree that it being `const` makes it easier to see that there is no allocation happening. Anyways, this point is moot since https://github.com/gz/rust-slabmalloc/pull/14 has now been merged.

litebox_platform_linux_kernel/src/mm/alloc.rs
```diff
@@ -0,0 +1,162 @@
...
+    unsafe fn alloc(&self, layout: core::alloc::Layout) -> *mut u8 {
+        match layout.size() {
+            Self::BASE_PAGE_SIZE => {
+                // Best to use the underlying backend directly to allocate pages
+                // to avoid fragmentation
+                self.allocate_pages(Self::BASE_PAGE_SIZE_ORDER)
+                    .expect("allocate page")
+            }
+            Self::LARGE_PAGE_SIZE => {
+                // Best to use the underlying backend directly to allocate large
```
wdcui: fix the comment ("to allocate large page")?

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -149,19 +162,11 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
 
-    /// Allocate 2^order pages from host
-    #[allow(dead_code)]
-    fn alloc(&self, order: u32) -> Result<u64, error::Errno> {
-        Host::alloc(order)
-    }
-
-    #[allow(dead_code)]
-    fn exit(&self) {
+    pub fn exit(&self) {
```
jaybosamiya-ms: Minor:
```suggestion
    pub fn exit(&self) -> ! {
```

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -339,8 +344,17 @@ impl<Host: HostInterface> IPInterfaceProvider for LinuxKernel<Host> {
...
+    ///
+    /// It can return more than requested size. On success, it returns the start address
+    /// and the size of the allocated memory.
+    fn alloc(layout: &core::alloc::Layout) -> Result<(usize, usize), error::Errno>;
+
+    /// Returns the memory back to host.
+    ///
+    /// Note host should know the size of allocated memory and needs to check the validity
+    /// of the given address.
+    fn free(addr: usize);
```
jaybosamiya-ms: This interface technically hides away the unsafety that can occur due to `free`. Specifically, purely safe code would be allowed to simply invoke `free(1337)` and break memory safety without any indication that this can cause a problem. Thus, this interface should be marked `unsafe`:
```suggestion
    unsafe fn free(addr: usize);
```

litebox_platform_linux_kernel/src/mm/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// For page allocation from host.
+    ///
+    /// Note this is only called by [`Self::rescue_heap`] when the buddy allocator is out of memory.
+    ///
+    /// It can return more than requested size. On success, it returns the start address
+    /// and the size of the allocated memory.
+    fn alloc(layout: &core::alloc::Layout) -> Result<(usize, usize), crate::error::Errno>;
+
+    /// Returns the memory back to host.
+    fn free(addr: usize);
```
jaybosamiya-ms: Similar to the other scenario:
```suggestion
    unsafe fn free(addr: usize);
```

litebox_platform_linux_kernel/src/lib.rs
CvvT: I see. I was thinking host should add checks to prevent that so the guest is allowed to call `free(1337)`; but from development perspective, it is better to be marked as unsafe. Thanks!

PR #15, opened 2025-02-21
Migrate to Rust 2024 edition

jaybosamiya-ms:
Rust stable updated today to 1.85.0, which finally stabilized a new edition of Rust (the 2024 edition).  For the most part, this doesn't really impact us (feel free to read [the blogpost announcement](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0.html) for details if you are curious).  However, there are some pieces we need to update to be able to take advantage of the niceties of the newer edition. 

One major such change that impacts us is that Rust 2024-edition onwards requires unsafe blocks even inside unsafe functions (i.e., just because a function _itself_ is marked `unsafe`, its body cannot arbitrarily call unsafe functions _without_ marking the usages as `unsafe` too); this allows us to more clearly understand the usage of `unsafe`.

The newest version of Rust comes with `cargo fix --edition` which performs an automatic migration for us; however, the automatic migration is overly-conservative in its choices, and can sometimes lead to stylistically suboptimal choices. Thus, I've also manually looked over every automatic change, and fixed it up manually to be more friendly to the style we've been doing things.

This PR contains both the automatic, and then manual fixes, thereby migrating us over to the 2024 edition.  As a quick note, you _may_ need to run `rustup update` on your machine to get the latest stable (i.e., I don't know if rustup will automatically pick up 2024 edition and pull it for you).

### DIFF
--- a/litebox/src/platform/mod.rs
+++ b/litebox/src/platform/mod.rs
@@ -338,43 +338,41 @@ where
         T: core::cmp::PartialEq<core::ffi::c_char>,
         Self: RawConstPointer<core::ffi::c_char>,
     {
-        unsafe {
-            use alloc::borrow::Cow;
-            use alloc::boxed::Box;
-            use alloc::vec::Vec;
-            use core::ffi::c_char;
-            let nul_position = {
-                let mut i = 0isize;
-                while *<Self as RawConstPointer<T>>::read_at_offset(self, i)? != 0 {
-                    i = i.checked_add(1)?;
-                }
-                i
-            };
-            let len = nul_position.checked_add(1)?.try_into().ok()?;
-            let slice: Cow<[c_char]> = self.to_cow_slice(len)?;
-            match slice {
-                Cow::Borrowed(bytes) => {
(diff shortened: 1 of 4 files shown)


litebox/src/fs/errors.rs
```diff
@@ -88,7 +88,9 @@ pub enum MkdirError {
 pub enum RmdirError {
     #[error("the parent directory does not allow write permission")]
     NoWritePerms,
-    #[error("currently in use by the system, or something prevents its removal (e.g., is the root directory)")]
+    #[error(
+        "currently in use by the system, or something prevents its removal (e.g., is the root directory)"
```
wdcui, line 92: Did you break this into two lines intentionally?

jaybosamiya-ms: The 2024 edition of rustfmt has this just above the limit line width limit, and thus it automatically broke it into multiple lines. I would ideally prefer it on the same line, but I didn't immediately find a shorter version of the message, thus left is as-is.

PR #16, opened 2025-02-24
Page Table

CvvT:
Utilize [x86_64](https://crates.io/crates/x86_64) for page table implementation (`PageTableImpl`):
* init
* translate va to pa
* handle page fault
* unmap_pages
* remap_pages
* mprotect_pages

With limited APIs x86_64 provides (translate and map/unmap one page at a time), the current implementation is super slow when operating on a large range of pages. VSBox has better implementations (with more code), I can port it if needed.

### DIFF
--- a/litebox_platform_linux_kernel/src/arch/mod.rs
+++ b/litebox_platform_linux_kernel/src/arch/mod.rs
@@ -1,10 +1,9 @@
 //! Arch-specific code
 
+#[cfg(target_arch = "x86_64")]
 mod x86;
 
-pub use x86::{
-    Page, PageFaultErrorCode, PageSize, PageTableFlags, PhysAddr, PhysFrame, Size4KiB,
-    TranslateResult, VirtAddr, X64PageTable,
-};
+#[cfg(target_arch = "x86_64")]
+pub use x86::*;
 
 pub const PAGE_SIZE: usize = Size4KiB::SIZE as usize;
--- a/litebox_platform_linux_kernel/src/arch/x86/mm/mod.rs
+++ b/litebox_platform_linux_kernel/src/arch/x86/mm/mod.rs
@@ -1,3 +1,1 @@
-mod paging;
-
-pub use paging::X64PageTable;
+pub mod paging;
--- a/litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
+++ b/litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
@@ -154,8 +154,7 @@ impl<M: MemoryProvider> PageTableImpl for X64PageTable<'_, M> {
     ///
     /// Note it does not free the allocated frames for page table itself (only those allocated to
     /// user space).
-    fn unmap_pages(&mut self, va: VirtAddr, len: usize, free_page: bool, flush: bool) {
-        assert!(len != 0);
+    unsafe fn unmap_pages(&mut self, va: VirtAddr, len: usize, free_page: bool, flush: bool) {
         assert!(va.is_aligned(Size4KiB::SIZE));
         assert!(len % Size4KiB::SIZE as usize == 0);
 
--- a/litebox_platform_linux_kernel/src/mm/pgtable.rs
+++ b/litebox_platform_linux_kernel/src/mm/pgtable.rs
@@ -67,9 +67,17 @@ pub trait PageTableImpl {
     /// Unmap 4KiB pages from the page table
     ///
     /// `flush` indicates whether the TLB should be flushed after the pages are unmapped.
+    /// `free_page` indicates whether the unmapped pages should be freed (This may be helpful
+    /// when implementing [`Self::remap_pages`]).
+    ///
+    /// # Safety
     ///
     /// `start` and `len` must be aligned to 4KiB.
-    fn unmap_pages(&mut self, start: VirtAddr, len: usize, free_page: bool, flush: bool);
+    ///
+    /// # Panics
+    ///
+    /// panic if `start` or `len` is misaligned.
+    unsafe fn unmap_pages(&mut self, start: VirtAddr, len: usize, free_page: bool, flush: bool);
 
     /// Remap 4KiB pages in the page table from `old_addr` to `new_addr`
     ///
--- a/litebox_platform_linux_kernel/src/arch/x86/mod.rs
+++ b/litebox_platform_linux_kernel/src/arch/x86/mod.rs
@@ -1,4 +1,4 @@
-mod mm;
+pub mod mm;
 
 pub use x86_64::{
     addr::{PhysAddr, VirtAddr},
@@ -7,5 +7,3 @@ pub use x86_64::{
         paging::{Page, PageSize, PageTableFlags, PhysFrame, Size4KiB, mapper::TranslateResult},
     },
 };
-
-pub use mm::X64PageTable;
--- a/litebox_platform_linux_kernel/src/mm/tests.rs
+++ b/litebox_platform_linux_kernel/src/mm/tests.rs
@@ -16,7 +16,7 @@ use x86_64::{
 
 use crate::{
     HostInterface, LinuxKernel,
-    arch::{PAGE_SIZE, X64PageTable},
+    arch::{PAGE_SIZE, mm::paging::X64PageTable},
     host::mock::MockHostInterface,
     mm::{MemoryProvider, pgtable::PageTableAllocator},
     mock_log_println,
@@ -195,12 +195,14 @@ fn test_page_table() {
     }
 
     // unmap all pages
-    pgtable.unmap_pages(
-        start_addr,
-        (new_addr - start_addr) as usize + 4 * PAGE_SIZE,
-        true,
-        false,
-    );
+    unsafe {
+        pgtable.unmap_pages(
+            start_addr,
+            (new_addr - start_addr) as usize + 4 * PAGE_SIZE,
+            true,
+            false,
+        )
+    };
     for page in Page::range(start_page, new_page + 4) {
         assert!(matches!(
             pgtable.translate(page.start_address()),


litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
```diff
@@ -0,0 +1,307 @@
...
+    _provider: core::marker::PhantomData<M>,
+}
+
+unsafe impl<M: MemoryProvider> PageTableFrameMapping for FrameMapping<M> {
+    fn frame_to_pointer(&self, frame: PhysFrame) -> *mut PageTable {
+        frame_to_pointer::<M>(frame)
+    }
+}
+
+unsafe impl<M: MemoryProvider> FrameAllocator<Size4KiB> for PageTableAllocator<M> {
```
wdcui, line 50: Out of my curiosity, why do we mark this impl as unsafe instead of marking the function?

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn unmap_pages(&mut self, va: VirtAddr, len: usize, free_page: bool, flush: bool) {
+        assert!(len != 0);
+        assert!(va.is_aligned(Size4KiB::SIZE));
+        assert!(len % Size4KiB::SIZE as usize == 0);
+
+        let start = Page::<Size4KiB>::from_start_address(va).unwrap();
+        let end = Page::<Size4KiB>::from_start_address(va + len as _).unwrap();
+        let mut allocator = PageTableAllocator::<M>::new();
+
+        // Note this implementation is slow as each page requires a full page table walk.
```
wdcui, line 165: (No need to implement this now): A faster alternative is to handle all frames in a single page table page together?

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
```diff
@@ -0,0 +1,307 @@
...
+        // TODO: this implementation is slow as each page requires two full page table walks.
+        // If we have N pages, it will be 2N times slower.
+        for page in Page::range(begin, end + 1) {
+            match self.inner.translate(page.start_address()) {
+                TranslateResult::Mapped {
+                    frame: _,
+                    offset: _,
+                    flags,
+                } => {
+                    // If it is changed to writable, we leave it to page fault handler (COW)
```
wdcui, line 273: Why can we assume it can always be handled as COW?

litebox_platform_linux_kernel/src/mm/pgtable.rs
```diff
@@ -0,0 +1,117 @@
...
+        })
+    }
+}
+
+pub trait PageTableImpl {
+    /// Initialize the page table
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the `p4` is valid and properly aligned.
```
wdcui: The comment says "p4" but the argument is "p"?

litebox_platform_linux_kernel/src/mm/pgtable.rs
```diff
@@ -0,0 +1,117 @@
...
+    unsafe fn init(p: PhysAddr) -> Self;
+
+    fn translate(&self, addr: VirtAddr) -> TranslateResult;
```
wdcui, line 48: Should we add a comment for translate() since it's a function in the trait?

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
```diff
@@ -0,0 +1,307 @@
...
+                    return Ok(());
+                }
+
+                panic!("Page fault on present page: {:#x}", page.start_address());
+            }
+            TranslateResult::NotMapped => {
+                let mut allocator = PageTableAllocator::<M>::new();
+                // TODO: if it is file-backed, we need to read the page from file
+                let frame = allocator.allocate_frame(true).unwrap();
+                let table_flags = PageTableFlags::PRESENT
```
wdcui, line 115: What's this table_flags for?

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
CvvT: I feel there is only subtle difference (see https://doc.rust-lang.org/reference/unsafe-keyword.html). In this case, FrameAllocator is defined as `unsafe trait` in `x86_64`, so the implementation needs `unsafe impl` prefix.

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
CvvT: During page walk, if there is any page table missing (before it reaches PTE, it would create one with `table_flag` permissions.

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
CvvT: Roughly speaking, yes. We have that implementation in [VSBox](https://github.com/MSRSSP/snp-sandbox/blob/6b111447f3637fdaad10d8fb3de5c1322b359966/sandbox_kernel/src/mem/pgtable.rs#L681) (and it can be simplified for litebox scenarios).

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
CvvT: Here it tries to update the flags but leave `WRITABLE` bit unchanged. Changes to other bits would be reflected to PTE. COW is not implemented yet because I'm not sure if we have any shared memory (COW pages).

litebox_platform_linux_kernel/src/mm/pgtable.rs
CvvT: Fixed

litebox_platform_linux_kernel/src/mm/pgtable.rs
CvvT: Added

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
wdcui: If we don't update the flags now, how do we differentiate between the two cases: a page shouldn't be written to at all vs. it is COW?

CvvT: Page table itself does not have that information, this function is marked as unsafe and caller needs to make sure users have access to the page (See docstring of the trait function: https://github.com/MSRSSP/litebox/blob/bc084c54ddb4088cd9198c992feb866b8d3fb39d/litebox_platform_linux_kernel/src/mm/pgtable.rs#L55-L58)

wdcui: Who is the caller here? I don't follow your reasoning here.

CvvT: The caller is not implemented yet. The check of write permission should be implemented in caller, so this function only handles case where write access is allowed.

wdcui: Do you assume this unimplemented caller will also manage the COW information?

CvvT: Sorry, I forgot to give the reference to mprotect_pages:
https://github.com/MSRSSP/litebox/blob/bc084c54ddb4088cd9198c992feb866b8d3fb39d/litebox_platform_linux_kernel/src/mm/pgtable.rs#L90-L96

Essentially, both mprotect and page fault handler needs to check whether the write to a page is legal or not. But, how we maintain that information and perform the check is unclear now.

wdcui: I see.

litebox_platform_linux_kernel/src/arch/x86/mm/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod paging;
+
+pub use paging::X64PageTable;
```
jaybosamiya-ms: I'm not sure why this module has a submodule if the only one submodule has itself re-exposed (there is only one `pub` member inside `paging`, and it is re-exposed immediately by using a `pub use` for it). Seems like unnecessary indirection.

litebox_platform_linux_kernel/src/arch/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! Arch-specific code
+
+mod x86;
+
+pub use x86::{
+    Page, PageFaultErrorCode, PageSize, PageTableFlags, PhysAddr, PhysFrame, Size4KiB,
+    TranslateResult, VirtAddr, X64PageTable,
+};
```
jaybosamiya-ms: As arch-specific code, an unrestricted `pub use` is probably not ideal. Imho, either this should not be a `pub use` _or_ it should be gated behind a `#[cfg(target_arch = ...)]` configuration.

litebox_platform_linux_kernel/src/mm/pgtable.rs
```diff
@@ -0,0 +1,118 @@
...
+        error_code: PageFaultErrorCode,
+        flush: bool,
+    ) -> Result<(), PageFaultError>;
+
+    /// Unmap 4KiB pages from the page table
+    ///
+    /// `flush` indicates whether the TLB should be flushed after the pages are unmapped.
+    ///
+    /// `start` and `len` must be aligned to 4KiB.
+    fn unmap_pages(&mut self, start: VirtAddr, len: usize, free_page: bool, flush: bool);
```
jaybosamiya-ms: Mostly for documentation purposes: what happens if they are not aligned? 

Either this function needs to have a `Result<(), UnmapError>` _or_ this must be an `unsafe` function with a `# Safety` saying that the caller should ensure this _or_ there should be a `# Panics` saying that there will be a panic in case of misalignment.

Given the implementation for it in this PR, the `# Panics` route is likely most correct for this comment.

jaybosamiya-ms: Sidenote: generally speaking, unmapping pages should be marked `unsafe` anyways, no?

CvvT: Right! Thanks for the comments. Fixed all of them.

PR #17, opened 2025-02-24
Tar-based read-only file system

### LINKED ISSUE #4: Implement a `tar` backed FileSystem
See https://lib.rs/crates/tar

jaybosamiya-ms:
This PR adds a tar-based read only (tar-ro) file system, and closes #4.

Turns out the crate mentioned in #4 (https://lib.rs/crates/tar) requires `std`, so instead I am using the slightly simpler https://lib.rs/crates/tar-no-std (which supports a smaller number of tar files, e.g., no GNU extensions, or long filename support). However, I think for a significant number of use cases, this should work out quite well.

### DIFF
--- a/litebox/src/fs/tar_ro.rs
+++ b/litebox/src/fs/tar_ro.rs
@@ -94,6 +94,11 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::private::Sealed
 {
 }
 
+fn contains_dir(haystack: &str, needle: &str) -> bool {
+    assert!(!needle.ends_with('/'));
+    haystack.starts_with(needle) && haystack.as_bytes().get(needle.len()) == Some(&b'/')
+}
+
 impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem<'_, Platform> {
     fn open(
         &self,
@@ -115,7 +120,7 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
             // direct hashmap lookup of relevant information and data.
             self.tar_data.entries().enumerate().find(|(_, entry)| {
                 match entry.filename().as_str() {
-                    Ok(p) => p.starts_with(path),
+                    Ok(p) => p == path || contains_dir(p, path),
                     Err(_) => false,
                 }
             })
@@ -175,7 +180,7 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
             .tar_data
             .entries()
             .any(|entry| match entry.filename().as_str() {
-                Ok(p) => p.starts_with(path),
+                Ok(p) => p == path || contains_dir(p, path),
                 Err(_) => false,
             })
         {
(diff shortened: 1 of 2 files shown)


litebox/src/fs/errors.rs
```diff
@@ -16,6 +16,8 @@ pub enum OpenError {
     AccessNotAllowed,
     #[error("the parent directory does not allow write permission")]
     NoWritePerms,
+    #[error("write access requested for a file on a read-only filesystem")]
```
wdcui, line 19: Minor: should this comment be consistent with other error message of the same error code?

litebox/src/fs/tar_ro.rs
```diff
@@ -0,0 +1,221 @@
...
+        use super::OFlags;
+        let currently_supported_oflags: OFlags = OFlags::RDONLY | OFlags::WRONLY | OFlags::RDWR;
+        if flags.contains(currently_supported_oflags.complement()) {
+            unimplemented!()
+        }
+        let path = self.absolute_path(path)?;
+        assert!(path.starts_with('/'));
+        let path = &path[1..];
+        let Some((idx, entry)) =
+            self.tar_data.entries().enumerate().find(|(_, entry)| {
```
wdcui, line 121: How efficient is this "find"?

litebox/src/fs/errors.rs
jaybosamiya-ms: Technically, the other ones complain just because it is a read-only file system, not because a write access is restricted; here, it is specifically when a write access is requested. Not a strong opinion on whether it should be made consistent or not, but having it be more precise doesn't hurt.

litebox/src/fs/tar_ro.rs
jaybosamiya-ms: It is a linear scan based on the implementation of `tar_no_std`. We can speed it up via a hashmap to cache out the scan (essentially at initialization time, do a full scan, and store the entries into a single lookup table), but it would lead to a self-referential struct, thus would lead to some complications. Thus, I opted to delay implementing that caching layer for now, keeping the implementation simpler to understand. If you believe this would be a significant bottleneck, I could add the caching layer (either as part of this PR, or a separate PR).

wdcui: Or we can contribute a PR to tar_no_std in the future. I suggest we add a TODO here to keep a reminder of this potential performance issue.

jaybosamiya-ms: Added a comment with TODO here to document it.

litebox/src/fs/tar_ro.rs
```diff
@@ -0,0 +1,224 @@
...
+        let path = self.absolute_path(path)?;
+        assert!(path.starts_with('/'));
+        let path = &path[1..];
+        let Some((idx, entry)) =
+            // TODO: this might be slow for large tar files, due to a linear scan. If better perf is
+            // needed, we can add a hashmap layer after doing one scan (in `new()`) that allows a
+            // direct hashmap lookup of relevant information and data.
+            self.tar_data.entries().enumerate().find(|(_, entry)| {
+                match entry.filename().as_str() {
+                    Ok(p) => p.starts_with(path),
```
CvvT: What if p is `a/b.txt` and path is `a/b`?

jaybosamiya-ms: Good catch, added a fix, and a test case to account for this.

PR #25, opened 2025-03-05
impl virtual memory management

CvvT:
Implement Virtual memory management similar to what we have in [VSBox with MMStruct](https://github.com/MSRSSP/snp-sandbox/blob/6b111447f3637fdaad10d8fb3de5c1322b359966/sandbox_kernel/src/mem/task.rs#L60). The difference is that I used [rangemap](https://docs.rs/rangemap/latest/rangemap/) to maintain non-overlapping intervals (which internally uses BTreeMap) instead of linked list used in MMStruct.

It provides the following APIs:
* insert_mapping
* remove_mapping
* resize_mapping (shrink or enlarge one mapping)
* move_mapping (move from old address to new address)
* protect_mapping
* handle_page_fault

Whenever it tries to update the page table, it performs permission check to make sure the user is allowed to do so.

### DIFF
--- a/litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
+++ b/litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
@@ -39,7 +39,7 @@ unsafe impl<M: MemoryProvider> PageTableFrameMapping for FrameMapping<M> {
 
 unsafe impl<M: MemoryProvider> FrameAllocator<Size4KiB> for PageTableAllocator<M> {
     fn allocate_frame(&mut self) -> Option<PhysFrame<Size4KiB>> {
-        self.allocate_frame(true)
+        Self::allocate_frame(true)
     }
 }
 
@@ -101,7 +101,7 @@ impl<M: MemoryProvider> PageTableImpl for X64PageTable<'_, M> {
             TranslateResult::NotMapped => {
                 let mut allocator = PageTableAllocator::<M>::new();
                 // TODO: if it is file-backed, we need to read the page from file
-                let frame = allocator.allocate_frame(true).unwrap();
+                let frame = PageTableAllocator::<M>::allocate_frame(true).unwrap();
                 let table_flags = PageTableFlags::PRESENT
                     | PageTableFlags::WRITABLE
                     | PageTableFlags::USER_ACCESSIBLE;
--- a/litebox_platform_linux_kernel/src/mm/pgtable.rs
+++ b/litebox_platform_linux_kernel/src/mm/pgtable.rs
@@ -27,7 +27,7 @@ impl<M: super::MemoryProvider> PageTableAllocator<M> {
     /// # Panics
     ///
     /// Panics if the address is not correctly aligned (i.e. is not a valid frame start)
-    pub fn allocate_frame(&mut self, clear: bool) -> Option<PhysFrame<Size4KiB>> {
+    pub fn allocate_frame(clear: bool) -> Option<PhysFrame<Size4KiB>> {
         M::mem_allocate_pages(0).map(|addr| {
             if clear {
                 unsafe {
--- a/litebox_platform_linux_kernel/src/mm/tests.rs
+++ b/litebox_platform_linux_kernel/src/mm/tests.rs
@@ -132,11 +132,10 @@ fn get_test_pgtable<'a>(
     range: PageRange,
     fault_flags: PageTableFlags,
 ) -> X64PageTable<'a, MockKernel> {
-    let mut allocator = PageTableAllocator::<MockKernel>::new();
-    let p4 = allocator.allocate_frame(true).unwrap();
+    let p4 = PageTableAllocator::<MockKernel>::allocate_frame(true).unwrap();
     let mut pgtable = unsafe { X64PageTable::<MockKernel>::init(p4.start_address()) };
 
-    for page in range {
+    for page in range.clone() {
         unsafe {
             pgtable
                 .handle_page_fault(page, fault_flags, PageFaultErrorCode::USER_MODE, false)
@@ -224,13 +223,14 @@ fn collect_mappings(vmm: &Vmem<X64PageTable<MockKernel>>) -> Vec<Range<u64>> {
 }
 
 #[test]
+#[allow(clippy::too_many_lines)]
 fn test_vmm_mapping() {
     let start_addr = VirtAddr::new(0x1000);
     let start_page = Page::from_start_address(start_addr).unwrap();
     let range = PageRange::new(start_page, start_page + 12);
     let fault_flags =
         PageTableFlags::PRESENT | PageTableFlags::USER_ACCESSIBLE | PageTableFlags::WRITABLE;
-    let pt = get_test_pgtable(range, fault_flags);
+    let pt = get_test_pgtable(range.clone(), fault_flags);
     let mut vmm = Vmem::new(pt);
 
     // []
@@ -307,7 +307,7 @@ fn test_vmm_mapping() {
     // try to remap [0x3000, 0x5000)
     let r = PageRange::new(start_page + 2, start_page + 4);
     assert!(matches!(
-        vmm.resize_mapping(r, NonZeroPageSize::new(PAGE_SIZE * 4)),
+        vmm.resize_mapping(r.clone(), NonZeroPageSize::new(PAGE_SIZE * 4)),
         Err(VmemResizeError::RangeOccupied(_))
     ));
     assert!(
--- a/litebox_platform_linux_kernel/src/mm/vm.rs
+++ b/litebox_platform_linux_kernel/src/mm/vm.rs
@@ -66,7 +66,7 @@ impl From<VmFlags> for PageTableFlags {
 
 /// A page range that guarantees the `start` and `end` addresses
 /// are page aligned and `start` < `end`.
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 pub(super) struct PageRange {
     /// Start page of the range.
     start: Page,
@@ -108,7 +108,7 @@ impl NonZeroPageSize {
     }
 
     #[inline]
-    fn as_u64(&self) -> u64 {
+    fn as_u64(self) -> u64 {
         self.size as u64
     }
 }
@@ -167,8 +167,12 @@ impl<PT: PageTableImpl> Vmem<PT> {
         let (start, end) = (range.start.start_address(), range.end.start_address());
         self.vmas.remove(start..end);
         unsafe {
-            self.pt
-                .unmap_pages(start, (end - start) as usize, true, FLUSH_TLB)
+            self.pt.unmap_pages(
+                start,
+                usize::try_from(end - start).unwrap(),
+                true,
+                FLUSH_TLB,
+            );
         };
     }
 
@@ -197,10 +201,10 @@ impl<PT: PageTableImpl> Vmem<PT> {
                     (intersection.end - intersection.start).try_into().unwrap(),
                     true,
                     FLUSH_TLB,
-                )
+                );
             };
         }
-        self.vmas.insert(start..end, VmArea { flags })
+        self.vmas.insert(start..end, VmArea { flags });
     }
 
     /// Create a new mapping in the virtual address space.
@@ -222,7 +226,7 @@ impl<PT: PageTableImpl> Vmem<PT> {
         let len = suggested_end - suggested_start;
         let new_addr = self.get_unmmaped_area(
             suggested_start,
-            NonZeroPageSize::new(len as usize),
+            NonZeroPageSize::new(usize::try_from(len).unwrap()),
             fixed_addr,
         )?;
         self.vmas.insert(new_addr..new_addr + len, VmArea { flags });
@@ -314,7 +318,12 @@ impl<PT: PageTableImpl> Vmem<PT> {
             .insert(new_addr..new_addr + new_size.as_u64(), *vma);
         unsafe {
             self.pt
-                .remap_pages(start, new_addr, (end - start) as usize, FLUSH_TLB)
+                .remap_pages(
+                    start,
+                    new_addr,
+                    usize::try_from(end - start).unwrap(),
+                    FLUSH_TLB,
+                )
                 .ok()
         };
         self.vmas.remove(start..end);
@@ -364,12 +373,12 @@ impl<PT: PageTableImpl> Vmem<PT> {
             match unsafe {
                 self.pt.mprotect_pages(
                     intersection.start,
-                    (intersection.end - intersection.start) as usize,
+                    usize::try_from(intersection.end - intersection.start).unwrap(),
                     new_flags.into(),
                     FLUSH_TLB,
                 )
             } {
-                Ok(_) => {}
+                Ok(()) => {}
                 Err(PageTableWalkError::MappedToHugePage) => {
                     unreachable!("VMEM: We don't support huge pages")
                 }
@@ -480,11 +489,7 @@ impl<PT: PageTableImpl> Vmem<PT> {
         // top down
         // 1. check [last_end, TASK_SIZE_MAX)
         let (low_limit, high_limit) = (Self::TASK_ADDR_MIN, Self::TASK_ADDR_MAX - size.as_u64());
-        let last_end = self
-            .vmas
-            .last_range_value()
-            .map(|r| r.0.end)
-            .unwrap_or(low_limit);
+        let last_end = self.vmas.last_range_value().map_or(low_limit, |r| r.0.end);
         if last_end <= high_limit {
             return Some(last_end);
         }
@@ -500,8 +505,7 @@ impl<PT: PageTableImpl> Vmem<PT> {
         let first_start = self
             .vmas
             .first_range_value()
-            .map(|r| r.0.start)
-            .unwrap_or(high_limit);
+            .map_or(high_limit, |r| r.0.start);
         if low_limit + size.as_u64() <= first_start {
             return Some(first_start - size.as_u64());
         }


litebox_platform_linux_kernel/src/mm/pgtable.rs
```diff
@@ -37,19 +39,27 @@ impl<M: super::MemoryProvider> PageTableAllocator<M> {
...
+    const MPROTECT_PTE_MASK: u64 = PageTableFlags::WRITABLE.bits()
+        | PageTableFlags::USER_ACCESSIBLE.bits()
+        | PageTableFlags::NO_EXECUTE.bits();
+
     /// Initialize the page table with the physical address of the top-level page table.
     ///
     /// # Safety
     ///
     /// The caller must ensure that the `p` is valid and properly aligned.
+    #[allow(dead_code)]
```
jaybosamiya-ms: I assume this is an `allow` lint to suppress the error temporarily (and that this code would be used later), yes? If so, my preference would be to use an `expect` lint instead of `allow`:
```suggestion
    #[expect(dead_code, reason = "temporarily unused; will be used in foobar")]
```
Specifically, an expect lint will _itself_ error out as soon as it stops applying, reminding you to remove it. Placing a reason in place will also make the (eventual) error message and such easier to understand. Similar with all other `allow`s in this PR.

litebox_platform_linux_kernel/src/mm/pgtable.rs
```diff
@@ -111,16 +122,20 @@ pub trait PageTableImpl {
     ) -> Result<(), PageTableWalkError>;
 }
 
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum PageTableWalkError {
-    NotMapped,
+    #[error("Given page is part of an already mapped huge page")]
     MappedToHugePage,
+    #[error("Page table allocation failed")]
```
jaybosamiya-ms: General convention for error strings is to have them start with a lowercase letter, so that when displayed in context (wrapped by other errors), it shows up correctly (e.g., `Foo: Bar: Baz` looks weird, but `foo: bar: baz` looks fine):
```suggestion
    #[error("given page is part of an already mapped huge page")]
    MappedToHugePage,
    #[error("page table allocation failed")]
```

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,465 @@
...
+
+#[allow(dead_code)]
+/// Virtual Address Space Manager
+pub struct Vmem<PT: PageTableImpl> {
+    /// Virtual memory areas. The ranges need to be page aligned.
+    vmas: RangeMap<VirtAddr, VmArea>,
+    pt: PT,
+}
+
+#[allow(dead_code)]
```
jaybosamiya-ms: I prefer more restricted set of `allow` lints, whenever possible (i.e., at function level, rather than at full `impl` level) since otherwise it is disabling an entire set of lints at once.

(Sidenote: also see other comment on changing it to `expect`).

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,465 @@
...
+    /// existing range in the map, then the existing range (or ranges) will be
+    /// partially or completely replaced by the inserted range.
+    ///
+    /// If the inserted range either overlaps or is immediately adjacent
+    /// any existing range _mapping to the same value_, then the ranges
+    /// will be coalesced into a single contiguous range.
+    ///
+    /// # Panics
+    ///
+    /// Panics if range `start >= end`.
```
jaybosamiya-ms: I notice a large number of these functions all have this same panic. It might be a good idea to have a newtype that _upon construction_ ensures that it has the `start < end` property (in one single spot); it should have a _private_ `x86::...::PageRange` field in it. This would ensure that you don't have to remember to validate everywhere, but can validate it once, and then be ensured that your validation is correct everywhere. This obviously has some (minor) perf benefit, but most importantly, you don't need to remember to check things everywhere.

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,465 @@
...
+
+            let new_flags = (vma.flags & !VmFlags::VM_ACCESS_FLAGS) | flags;
+            unsafe {
+                // `intersection` is page aligned.
+                match self.pt.mprotect_pages(
+                    intersection.start,
+                    (intersection.end - intersection.start) as usize,
+                    new_flags.into(),
+                    flush,
+                ) {
```
jaybosamiya-ms: Possibly slightly more restricted `unsafe` invocation?
```suggestion
            let new_flags = (vma.flags & !VmFlags::VM_ACCESS_FLAGS) | flags;
                // `intersection` is page aligned.
                match unsafe { self.pt.mprotect_pages(
                    intersection.start,
                    (intersection.end - intersection.start) as usize,
                    new_flags.into(),
                    flush,
                ) } {
```

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,465 @@
...
+    const STACK_GUARD_GAP: u64 = 256u64 << 12;
+
+    pub const fn new(pt: PT) -> Self {
+        Self {
+            vmas: RangeMap::new(),
+            pt,
+        }
+    }
+
+    pub fn iter(&self) -> rangemap::map::Iter<'_, VirtAddr, VmArea> {
```
jaybosamiya-ms: Generally speaking, I prefer that iterators do not reveal the specific internal implementation (i.e., if we switch from rangemap to something else, we would need to change this particular type, causing all of its callers to also change). Something like this would probably work better:
```suggestion
    pub fn iter(&self) -> impl Iterator<Item=(&VirtAddr, &VmArea)> {
```
This essentially makes the return type a restricted existential type, rather than a concrete type, thereby allowing future updates to usually not need to break compatibility.

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,465 @@
...
+use thiserror::Error;
+
+use crate::arch::{PAGE_SIZE, Page, PageFaultErrorCode, PageRange, PageTableFlags, VirtAddr};
+
+use super::pgtable::{PageFaultError, PageTableImpl, PageTableWalkError};
+
+bitflags::bitflags! {
+    /// Flags to describe the properties of a memory region.
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    pub struct VmFlags: u32 {
```
jaybosamiya-ms: Making things more restricted at the module level (via the `pub(crate) mod vm;`) for example, is great! It would additionally be helpful for the items _within_ the module to also be `pub(crate)` rather than `pub`.

Technically `pub` items within `pub(crate)` modules are different from `pub(crate)` items within `pub(crate)` modules. However, from what I can tell from this implementation, we mostly only need `pub(crate)` and don't want to make things _really_ pub-within-non-pub (which carries with it subtle possibility of future complications).

(Obviously, this applies to the other `pub` members within `pub(crate)` modules, not just this one line)

litebox_platform_linux_kernel/src/mm/pgtable.rs
CvvT: Awesome! I didn't know this nice trick.

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: I see. Thanks!

litebox_platform_linux_kernel/src/mm/pgtable.rs
CvvT: Got it!

litebox_platform_linux_kernel/src/mm/pgtable.rs
CvvT: These are actually not dead code because they are used by testing code. For now, I just annotate them with `cfg(test)`.

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
```diff
@@ -119,7 +109,7 @@ impl<M: MemoryProvider> PageTableImpl for X64PageTable<'_, M> {
                     self.inner.map_to_with_table_flags(
                         page,
                         frame,
-                        flags,
+                        flags | PageTableFlags::PRESENT,
```
wdcui, line 112: It's a little weird to add the PRESENT bit here directly. What about other bits? What's the assumption of the flags passed from the caller?

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
```diff
@@ -258,6 +250,8 @@ impl<M: MemoryProvider> PageTableImpl for X64PageTable<'_, M> {
         new_flags: PageTableFlags,
         flush: bool,
     ) -> Result<(), PageTableWalkError> {
+        let mask = PageTableFlags::from_bits(Self::MPROTECT_PTE_MASK).unwrap();
```
wdcui: Minor: does it make sense to make this mask a constant?

litebox_platform_linux_kernel/src/mm/tests.rs
```diff
@@ -210,3 +222,178 @@ fn test_page_table() {
...
+
+    assert!(matches!(
+        vmm.resize_mapping(
+            PageRange::new(start_page + 2, start_page + 3),
+            NonZeroPageSize::new(PAGE_SIZE * 2),
+            false
+        ),
+        Err(VmemResizeError::NotExist(_))
+    ));
+    // Failed to resize, [(0x1000, 0x3000), (0x5000, 0xd000)]
```
wdcui: Minor: If this comment is about the code below, maybe remove the empty line?

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,513 @@
...
+            | Self::VM_MAYWRITE.bits()
+            | Self::VM_MAYEXEC.bits();
+    }
+}
+
+impl From<VmFlags> for PageTableFlags {
+    fn from(values: VmFlags) -> Self {
+        let mut flags = PageTableFlags::empty();
+        if values.intersects(VmFlags::VM_READ | VmFlags::VM_WRITE) {
+            flags |= PageTableFlags::USER_ACCESSIBLE;
```
wdcui, line 55: Do we use VMA for kernel page table? If so, this may be incorrect for that case.

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,513 @@
...
+    }
+}
+
+/// A non-zero page-aligned size in bytes.
+#[derive(Clone, Copy)]
+pub(super) struct NonZeroPageSize {
+    size: usize,
+}
+
+impl NonZeroPageSize {
```
wdcui, line 98: Does the alignment need to be consistent with the size? E.g., if the size is 8KB, then the address needs to be aligned with 8K?

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,513 @@
...
+    /// Panics if `start` >= `end`.
+    pub(super) fn new(start: Page, end: Page) -> Self {
+        assert!(start < end);
+        Self { start, end }
+    }
+}
+
+/// Virtual memory area
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(super) struct VmArea {
```
wdcui, line 130: We don't need fields to specify the start/end of the VmArea?

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,513 @@
...
+    /// Flags describing the properties of the memory region.
+    flags: VmFlags,
+}
+
+/// Virtual Address Space Manager
+pub(super) struct Vmem<PT: PageTableImpl> {
+    /// Virtual memory areas. The ranges need to be page aligned.
+    vmas: RangeMap<VirtAddr, VmArea>,
```
wdcui, line 138: We don't need to specify the size?

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,513 @@
...
+    ///
+    /// See https://elixir.bootlin.com/linux/v5.19.17/source/mm/mremap.c#L886 for reference.
+    pub(super) fn resize_mapping(
+        &mut self,
+        range: PageRange,
+        new_size: NonZeroPageSize,
+        flush: bool,
+    ) -> Result<(), VmemResizeError> {
+        let range = range.start.start_address()..range.end.start_address();
+        let (cur_range, cur_vma) = self
```
wdcui, line 249: What's the invariant between range.start and cur_range.start?

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,513 @@
...
+                    ),
+                    flush,
+                );
+                return Ok(());
+            }
+            core::cmp::Ordering::Greater => {}
+        }
+
+        // grow
+        if range.end > cur_range.end {
```
wdcui, line 272: What's the invariant between range and cur_range?

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,513 @@
...
+        &mut self,
+        range: PageRange,
+        new_size: NonZeroPageSize,
+        suggested_addr: VirtAddr,
+        flush: bool,
+    ) -> Option<VirtAddr> {
+        let (start, end) = (range.start.start_address(), range.end.start_address());
+        assert!(new_size.as_u64() >= (end - start));
+
+        let (target_range, vma) = self
```
wdcui: Why is this called target_range? The previous function calls it cur_range. And target_range should be at the suggtested_addr?

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,513 @@
...
+                return Err(PageFaultError::AccessError("no mapping"));
+            }
+
+            if !self
+                .vmas
+                .overlapping(Self::TASK_ADDR_MIN..fault_addr)
+                .next_back()
+                .is_none_or(|(prev_range, prev_vma)| {
+                    // enforce gap between stack and other preceding non-stack mappings
+                    // Either the previous mapping is also a stack mapping w/ some access flags
```
wdcui, line 436: Can we merge Either and Or into a single comment?

litebox_platform_linux_kernel/src/mm/vm.rs
```diff
@@ -0,0 +1,513 @@
...
+        if low_limit + size.as_u64() <= first_start {
+            return Some(first_start - size.as_u64());
+        }
+
+        None
+    }
+
+    /// Check if it has access to the fault address.
+    fn access_error(error_code: PageFaultErrorCode, flags: VmFlags) -> bool {
+        if error_code.intersects(PageFaultErrorCode::PROTECTION_KEY | PageFaultErrorCode::SGX) {
```
wdcui: Why do we need to worry about SGX in a general case?

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
CvvT: This function would directly overwrite the entry flag with the given value, so the assumption is that the given flag must be correct except the `PRESENT` bit as it is implied (all mapped pages are present).

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
CvvT: Fixed.

litebox_platform_linux_kernel/src/mm/tests.rs
CvvT: Moved the comments

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: No, it only manages user space address. Kernel address shouldn't accessible to users. When creating new mappings, it would ensure it only has user space address.

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: No, just 4KiB page aligned.

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: start/end is used as the key.

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: Range<VirtAddr, VirtAddr> is the key to the map.

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: `cur_range` must contain `range.start` so `cur_range` <= `range.start`.

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: `cur_range` must contain `range.start` but not necessarily `range.end` (i.e., the entire `range`).

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: Right!

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: Sure

litebox_platform_linux_kernel/src/mm/vm.rs
CvvT: We don't have protection key either. This is copied from Linux, I can just remove it.

PR #27, opened 2025-03-06
Layered file system

### LINKED ISSUE #5: Support a parametric layered `FileSystem`
Roughly, `layered::FileSystem<UPPER: fs::FileSystem, LOWER: fs::FileSystem>` such that the upper layer is read/write, the lower layer is read-only. Any reading calls, if they fail to find something in the upper layer, go down to lower layer; any writing calls will only hit upper layer. 

This can be useful with #1. Specifically, `layered::FileSystem<in_mem::FileSystem, tar::FileSystem>` would allow for an easy way to have the initial files.

jaybosamiya-ms:
This PR implements a layered file system, roughly `layered::FileSystem<Upper, Lower>` (closes #5).

image

Essentially, a layered filesystem itself doesn't carry or store any of the files, but delegates to each of the the layers. Specifically, this implementation will look for and work with files in the upper layer, unless they don't exist, in which case the lower layer is looked at.

The current design of layering treats the lower layer as read-only and performs copy-on-write semantics for moving to upper layer.

The biggest complexity of this implementation comes from needing to support the deletion of files from the lower layer (without actually changing the lower layer), as well as supporting aliasing (same file opened twice) under the presence of writing. There are tests added for ensuring that the behavior of these are as expected.

Currently, the implementation's largest downside is in its handling of directories and permissions, each of which would be improved by having support for `stat` in the core `FileSystem` trait, but I decided that should be a separate PR and not be part of this one. There are relevant TODOs in the code for this. We also would need to handle correct positioning of the "migrated" CoW files, but that would require having support for `lseek` in the core `FileSystem` trait; similarly, future update.

### DIFF
--- a/litebox/src/fs/layered.rs
+++ b/litebox/src/fs/layered.rs
@@ -108,13 +108,17 @@ impl<
         let mut upper_fd = None;
         let mut temp_buf = [0u8; 4096];
         loop {
-            match self.lower.read(&lower_fd, &mut temp_buf) {
+            match self.lower.read(&lower_fd, &mut temp_buf, None) {
                 Ok(size) => {
                     if upper_fd.is_none() {
                         // We are here the first time around, and did not error out, yay! We can
                         // actually open up the file.
                         //
                         // TODO: We might need to make all the parent directories?
+                        //
+                        // TODO: We need to migrate the file offsets (for read/write with offset:
+                        // `None`), but we currently don't have `lseek` support yet, so we
+                        // effectively end up resetting position upon a move. (╯°□°）╯︵ ┻━┻
                         //
                         // TODO: We need to `stat` the mode from the lower level, and migrate it
                         // over, but we currently don't have `stat` support yet, so we're just
@@ -127,7 +131,7 @@ impl<
                     }
                     let upper_fd = upper_fd.as_ref().unwrap();
                     if size > 0 {
-                        self.upper.write(upper_fd, &temp_buf[..size]);
+                        self.upper.write(upper_fd, &temp_buf[..size], None);
                     } else {
                         // EOF
                         break;
@@ -459,7 +463,12 @@ impl<Platform: sync::RawSyncPrimitivesProvider, Upper: super::FileSystem, Lower:
         }
     }
 
-    fn read(&self, fd: &crate::fd::FileFd, buf: &mut [u8]) -> Result<usize, ReadError> {
+    fn read(
+        &self,
+        fd: &crate::fd::FileFd,
+        buf: &mut [u8],
+        offset: Option<usize>,
+    ) -> Result<usize, ReadError> {
         // Since a write to a lower-level file upgrades the underlying entry out completely to an
         // upper-level file, we don't actually need to worry about a desync; a write to lower-level
         // file will successfully be seen as just being an upper level file. Thus, it is sufficient
@@ -470,13 +479,18 @@ impl<Platform: sync::RawSyncPrimitivesProvider, Upper: super::FileSystem, Lower:
             return Err(ReadError::NotForReading);
         }
         match descriptor.entry.as_ref() {
-            EntryX::Upper { fd } => self.upper.read(fd, buf),
-            EntryX::Lower { fd } => self.lower.read(fd, buf),
+            EntryX::Upper { fd } => self.upper.read(fd, buf, offset),
+            EntryX::Lower { fd } => self.lower.read(fd, buf, offset),
             EntryX::Tombstone => unreachable!(),
         }
     }
 
-    fn write(&self, fd: &crate::fd::FileFd, buf: &[u8]) -> Result<usize, WriteError> {
+    fn write(
+        &self,
+        fd: &crate::fd::FileFd,
+        buf: &[u8],
+        offset: Option<usize>,
+    ) -> Result<usize, WriteError> {
         // Writing needs to be careful of how it is performing the write. Any upper-level file can
         // instantly be written to; but a lower-level file must become a upper-level file, before
         // actually being written to.
@@ -486,7 +500,7 @@ impl<Platform: sync::RawSyncPrimitivesProvider, Upper: super::FileSystem, Lower:
             return Err(WriteError::NotForWriting);
         }
         match descriptor.entry.as_ref() {
-            EntryX::Upper { fd } => return self.upper.write(fd, buf),
+            EntryX::Upper { fd } => return self.upper.write(fd, buf, offset),
             EntryX::Lower { fd } => {
                 // fallthrough
             }
@@ -506,7 +520,7 @@ impl<Platform: sync::RawSyncPrimitivesProvider, Upper: super::FileSystem, Lower:
         }
         // Since it has been migrated, we can just re-trigger, causing it to apply to the
         // upper layer
-        self.write(fd, buf)
+        self.write(fd, buf, offset)
     }
 
     fn chmod(&self, path: impl crate::path::Arg, mode: Mode) -> Result<(), ChmodError> {
@@ -689,6 +703,8 @@ struct Descriptor {
     path: String,
     flags: OFlags,
     entry: Entry,
+    // TODO: We also need to track file offsets, and perform the relevant `lseek` whenever we
+    // migrate an FD from lower to upper.
 }
 
 struct RootDir {
--- a/litebox/src/fs/tests.rs
+++ b/litebox/src/fs/tests.rs
@@ -233,14 +233,18 @@ mod layered {
             .open("foo", OFlags::RDONLY, Mode::RWXU)
             .expect("Failed to open file");
         let mut buffer = vec![0; 1024];
-        let bytes_read = fs.read(&fd, &mut buffer).expect("Failed to read from file");
+        let bytes_read = fs
+            .read(&fd, &mut buffer, None)
+            .expect("Failed to read from file");
         assert_eq!(&buffer[..bytes_read], b"testfoo\n");
         fs.close(fd).expect("Failed to close file");
         let fd = fs
             .open("bar/baz", OFlags::RDONLY, Mode::empty())
             .expect("Failed to open file");
         let mut buffer = vec![0; 1024];
-        let bytes_read = fs.read(&fd, &mut buffer).expect("Failed to read from file");
+        let bytes_read = fs
+            .read(&fd, &mut buffer, None)
+            .expect("Failed to read from file");
         assert_eq!(&buffer[..bytes_read], b"test bar baz\n");
         fs.close(fd).expect("Failed to close file");
     }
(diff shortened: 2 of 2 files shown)


litebox/src/path.rs
```diff
@@ -91,6 +91,39 @@ pub trait Arg: private::Sealed {
...
+            if let Some(posn) = path.rfind('/') {
+                path = &path[..posn];
+            } else {
+                res.push("");
+                break;
+            }
+        }
+
+        if res.last().unwrap().len() > 1 {
+            res.push("/");
```
wdcui, line 128: This logic doesn't match with the comment above?

litebox/src/fs/layered.rs
```diff
@@ -0,0 +1,712 @@
...
+                        // making it have every bit in the mode flags. ┻━┻︵ \(°□°)/ ︵ ┻━┻
+                        upper_fd = Some(
+                            self.upper
+                                .open(path, OFlags::CREAT | OFlags::WRONLY, Mode::all())
+                                .unwrap(),
+                        );
+                    }
+                    let upper_fd = upper_fd.as_ref().unwrap();
+                    self.upper.write(upper_fd, &temp_buf[..size]);
+                    if size == 0 {
```
wdcui: Why do we check size after the write()?

litebox/src/fs/layered.rs
```diff
@@ -0,0 +1,712 @@
...
+                    }
+                    ReadError::NotForReading => unreachable!(),
+                },
+            }
+        }
+        // Now that we've migrated the data over, we can close out both of the file descriptors.
+        self.upper.close(upper_fd.unwrap()).unwrap();
+        self.lower.close(lower_fd).unwrap();
+
+        // Now we need to migrate all the descriptor entries over.
```
wdcui, line 155: Why do we need to migrate all the descriptor entries (of the file that's being migrated)?

litebox/src/fs/layered.rs
```diff
@@ -0,0 +1,712 @@
...
+    // Note: does NOT account for symlinks.
+    fn absolute_path(&self, path: impl crate::path::Arg) -> Result<String, PathError> {
+        assert!(self.current_working_dir.ends_with('/'));
+        let path = path.as_rust_str()?;
+        if path.starts_with('/') {
+            // Absolute path
+            Ok(path.normalized()?)
+        } else {
+            // Relative path
+            Ok((self.current_working_dir.clone() + path.as_rust_str()?).normalized()?)
```
wdcui, line 231: Is current_working_dir guaranteed to end with '/'?

litebox/src/fs/layered.rs
```diff
@@ -0,0 +1,712 @@
...
+                        tombstone_removal = true;
+                    } else {
+                        return Err(PathError::NoSuchFileOrDirectory)?;
+                    }
+                }
+                EntryX::Upper { .. } => {
+                    // fallthrough since we might not have the descriptor open with the correct flags
+                }
+                EntryX::Lower { .. } => {
+                    // As an optimization, since a lower-level file entry is always is opened with
```
wdcui: Typo in comment: is always is opened

litebox/src/fs/layered.rs
```diff
@@ -0,0 +1,712 @@
...
+                        path,
+                        flags,
+                        entry: Arc::clone(entry),
+                    }));
+                }
+            }
+        }
+        if tombstone_removal {
+            let entry = self.root.write().entries.remove(&path).unwrap();
+            let EntryX::Tombstone = *entry else {
```
wdcui: is this check necessary?

litebox/src/fs/layered.rs
```diff
@@ -0,0 +1,712 @@
...
+                            None => {
+                                // Someone else's job. We can quit successfully.
+                                return Ok(());
+                            }
+                        }
+                    }
+                }
+                // Pull out the root entry, and perform a quick sanity check, and drop it out
+                // entirely, which should lead us to become the sole owner.
+                let root_entry = root_entries.remove(&path).unwrap();
```
wdcui, line 451: What is a root_entry?

litebox/src/fs/layered.rs
```diff
@@ -0,0 +1,712 @@
...
+            .insert(path, Arc::new(EntryX::Tombstone));
+        Ok(())
+    }
+
+    fn mkdir(&self, path: impl crate::path::Arg, mode: Mode) -> Result<(), MkdirError> {
+        let path = self.absolute_path(path)?;
+        match self.upper.mkdir(path.as_str(), mode) {
+            Ok(()) => {
+                // If we could successfully make the directory, we know that things are "sane" at
+                // the upper level, but we must also check the lower level to make sure that this
```
wdcui, line 608: Why do we make a directory in the upper first before we check the lower?

litebox/src/fs/layered.rs
CvvT: should it break if `size` < 4096 (which indicates EOF) ?

litebox/src/fs/layered.rs
```diff
@@ -0,0 +1,712 @@
...
+            }
+        }
+        if tombstone_removal {
+            let entry = self.root.write().entries.remove(&path).unwrap();
+            let EntryX::Tombstone = *entry else {
+                unreachable!()
+            };
+        }
+        // Otherwise, we first check the upper level, creating an entry if needed
+        match self.upper.open(&*path, flags, mode) {
```
CvvT, line 311: If `tombstone_removal` is true, does this return `Ok(fd)` or `Err(NoSuchFileOrDirectory)`?

litebox/src/path.rs
jaybosamiya-ms: The comment above specifically is talking about _relative_ paths of the form `a/b/c` (which is handled by the `else` case, which places the empty string as the last element). 

This particular case is handling absolute paths (i.e., `/a/b/c`) in which case the last element of `res` would be `a`. The specifics of this are not _immediately_ obvious, which is why I have a detailed set of tests for this function (which it successfully passes).

litebox/src/fs/layered.rs
jaybosamiya-ms: I don't break when `size` < 4096, because I don't think we currently guarantee `read` to mean "fill the buffer as far as possible", instead only requiring the slightly less strict "return 0 if EOF". Thus _technically_ we'll end up doing two `read`s to find out that we've reached EOF.

The `write` (even in case of 0) was to handle the case where the `O_CREAT` for some reason is using lazy-semantics (biggest reason was for 0-sized files). I think that is overthinking things, so I will update things and will swap the order of operations.

litebox/src/fs/layered.rs
jaybosamiya-ms: Otherwise we would have a desync between doubly-opened files. Specifically, see the `file_read_write_sync_up` test which looks at exactly this scenario.

litebox/src/fs/layered.rs
jaybosamiya-ms: Yes, see line 40 of this file which already has this comment.

litebox/src/fs/layered.rs
jaybosamiya-ms: Not _necessary_ (this is why the else clause is an `unreachable!()`, rather than a `panic!()`/`unimplemented!()`---I am stating that I believe the else clause to be entirely unreachable). 

At a high level, I have written this assertion here in the form of a move.  This is mostly to account for possible future code changes---if we ever decide to store additional information in tombstones (I don't have any specific plans, but can imagine cases where we might want to store some information in them), the compiler will force us to look at this line to see if anything special needs to be done (if we didn't have this line and simply did a `let _ = entry;`, then the compiler won't be able to inform us at that time, and we'd manually need to remember that we have `tombstone_removal` as a relevant part of the code).

Fwiw, the `unreachable!()` is a relatively good hint to the compiler to attempt to optimize this assertion (if it is unable to optimize the check, it will keep an actual runtime check, but fairly often, such checks get optimized out entirely. Even if it is not optimized out, this would be a single integer comparison, which is imho a trivial cost to improve clarity.

litebox/src/fs/layered.rs
jaybosamiya-ms: `tombstone_removal` can only occur upon `OFlags::CREAT`, which will cause the upper layer to create the file, and return an `Ok(fd)`

litebox/src/fs/layered.rs
jaybosamiya-ms: `self.root` holds the underlying tombstones and lower-level FDs, to support both deletion and aliasing. The `root_entry` is just the entry that is in `self.root`.

litebox/src/fs/layered.rs
jaybosamiya-ms: It mostly doesn't matter which order we check/make things in. Making an upper directory, even if a lower one exists for it, doesn't hurt us (other than costing a tiny bit of extra memory). I could swap the order of check/make but I don't think it particularly matters. 

The semantics were easier to reason about in the current order (at least to me) because I assume the "upper" level to be "the truth", which is only "supported" by the lower layer. Keeping this informal notion in mind helped make most of the implementations consistent: do things in upper layer and confirm/check/see lower layer if/when necessary.

litebox/src/fs/layered.rs
```diff
@@ -0,0 +1,713 @@
...
+                    return Ok(self.descriptors.write().insert(Descriptor {
+                        path,
+                        flags,
+                        entry: Arc::clone(entry),
+                    }));
+                }
+            }
+        }
+        if tombstone_removal {
+            let entry = self.root.write().entries.remove(&path).unwrap();
```
CvvT: Is it possible for two concurrent threads to reach this line?

litebox/src/path.rs
wdcui: It might help to provide an example for "/a/b/c" in the comment. It wasn't obvious to me that this function also handles the case of absolute paths.

litebox/src/fs/layered.rs
jaybosamiya-ms: If two threads at the same time are attempting to `CREAT` a file at the exact same time that has been previously deleted, then yes. I think it is a rather unlikely scenario; in any case, if this happens, then one will be able to grab the `write` lock and remove it, and the other will wait till the lock is released, and then the `unwrap` would fail. This is a rare set of circumstances, but yes, it is technically feasible.

I've added a fix for this, and written a comment explaining it. Thanks for catching this!

litebox/src/path.rs
jaybosamiya-ms: Added the comment, as well as made it into a doc-test, so that the comment cannot get out of sync with the implementation.

jaybosamiya-ms:
Note: CI failures are due to #29 being merged into `main` which changed the interfaces. For now, the _migration_ from lower to upper does not account for offsets, which will be fixed up in a future PR. I'm adding a TODO item for this as a comment.

PS: reads/writes on either lower or upper layer would handle offsets correctly in this implementation: the issue only shows up for _migrated_ files, which is a rarer scenario _and_ requires `lseek` support, thus I am postponing to a separate PR. 

Will merge once CI goes green again.

PR #28, opened 2025-03-06
Loader PoC

CvvT:
This is just a PoC to show a few issues I found. See more comments in code.

### DIFF
--- a/litebox/src/fs/mod.rs
+++ b/litebox/src/fs/mod.rs
@@ -43,8 +43,8 @@ pub trait FileSystem: private::Sealed {
     fn open(&self, path: impl path::Arg, flags: OFlags, mode: Mode) -> Result<FileFd, OpenError>;
     /// Close the file at `fd`
     fn close(&self, fd: FileFd) -> Result<(), CloseError>;
-    /// Read from a file descriptor into a buffer
-    fn read(&self, fd: &FileFd, buf: &mut [u8]) -> Result<usize, ReadError>;
+    /// Read from a file descriptor into a buffer at a specific offset if given
+    fn read(&self, fd: &FileFd, buf: &mut [u8], offset: Option<usize>) -> Result<usize, ReadError>;
     /// Write from a buffer to a file descriptor
     fn write(&self, fd: &FileFd, buf: &[u8]) -> Result<usize, WriteError>;
     /// Change the permissions of a file
--- a/litebox_platform_linux_kernel/src/mm/mapping.rs
+++ b/litebox_platform_linux_kernel/src/mm/mapping.rs
@@ -1,13 +1,67 @@
+//! Memory mapping functions for the Linux kernel.
+//!
+//! Here is an example of mmap sequences when loading a simple statically linked binary:
+//! ```
+//! mmap addr: 0x400000, len: 0xcd000, prot: PROT_READ, flags: MAP_PRIVATE, offset: 0x0, fd: 3
+//! mmap addr: 0x401000, len: 0x97000, prot: PROT_READ | PROT_EXEC, flags: MAP_PRIVATE | MAP_FIXED, offset: 0x1000, fd: 3
+//! mmap addr: 0x498000, len: 0x29000, prot: PROT_READ, flags: MAP_PRIVATE | MAP_FIXED, offset: 0x98000, fd: 3
+//! mmap addr: 0x4c1000, len: 0xc000,  prot: PROT_READ | PROT_WRITE, flags: MAP_PRIVATE | MAP_FIXED, offset: 0xc0000, fd: 3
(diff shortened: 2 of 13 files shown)


litebox_platform_linux_kernel/src/mm/mapping.rs
```diff
@@ -0,0 +1,123 @@
...
+//! ```
+//! mmap addr: 0x400000, len: 0xcd000, prot: PROT_READ, flags: MAP_PRIVATE, offset: 0x0, fd: 3
+//! mmap addr: 0x401000, len: 0x97000, prot: PROT_READ | PROT_EXEC, flags: MAP_PRIVATE | MAP_FIXED, offset: 0x1000, fd: 3
+//! mmap addr: 0x498000, len: 0x29000, prot: PROT_READ, flags: MAP_PRIVATE | MAP_FIXED, offset: 0x98000, fd: 3
+//! mmap addr: 0x4c1000, len: 0xc000,  prot: PROT_READ | PROT_WRITE, flags: MAP_PRIVATE | MAP_FIXED, offset: 0xc0000, fd: 3
+//! mmap addr: 0x4c8000, len: 0x5000,  prot: PROT_READ | PROT_WRITE, flags: MAP_PRIVATE | MAP_FIXED | MAP_ANONYMOUS
+//! ```
+//!
+//! Our current design would result in copying the entire file into the memory twice.
+//!
```
CvvT, line 13: This is sequence I got when trying to use `rust_elfloader`. Instead of calling `mprotect`. It just call `mmap` with fixed address (`MAP_FIXED` flag) which would remap the file with new permissions. Thus, we need to copying the entire file twice. This may not be a big issue as we don't focus on it for now.

CvvT: As we discussed earlier, we may want to use heap (global allocator) instead of native syscall `mmap` to support `mmap`. However, allocator does not accept a suggested address to support MAP_FIXED. In this case, it might be fine as the fixed addresses are all already allocated (within a existing mapped memory region). Or we just rely on syscall `mmap`.

litebox_shim_linux/src/syscall/mmap.rs
```diff
@@ -0,0 +1,163 @@
...
+                user_buf.copy_from_slice(&buffer[..size]);
+            });
+            copied += size;
+            file_offset += size;
+            if size < PAGE_SIZE {
+                break;
+            }
+        }
+        Ok(copied)
+    };
```
CvvT, line 104: Here I'm trying to read content from a file and write it to user space. fs `read` does not accept a user pointer as it does not handle page faults, so I have to provide a valid buffer to have extra copying.

litebox/src/fs/mod.rs
```diff
@@ -43,8 +43,8 @@ pub trait FileSystem: private::Sealed {
     fn open(&self, path: impl path::Arg, flags: OFlags, mode: Mode) -> Result<FileFd, OpenError>;
     /// Close the file at `fd`
     fn close(&self, fd: FileFd) -> Result<(), CloseError>;
-    /// Read from a file descriptor into a buffer
-    fn read(&self, fd: &FileFd, buf: &mut [u8]) -> Result<usize, ReadError>;
+    /// Read from a file descriptor into a buffer at a specific offset if given
+    fn read(&self, fd: &FileFd, buf: &mut [u8], offset: Option<usize>) -> Result<usize, ReadError>;
     /// Write from a buffer to a file descriptor
     fn write(&self, fd: &FileFd, buf: &[u8]) -> Result<usize, WriteError>;
```
jaybosamiya-ms, line 49: Just making a quick note here to say that rather than adding an `offset` field to read (in which case we would need to add similar to `write`, and _also_ specify what the expected behavior _after_ it completes is (i.e., what should be the semantics for `read(...., offset: Some(5))` followed by `read(..., offset: None)`). It also would complicate the Linux shim, for imho not a significant enough gain. Instead, adding an `lseek` like interface is probably the right move?

litebox_shim_linux/src/syscall/mmap.rs
CvvT: MutPtr does not have an API `write_slice`. Instead, it accepts a function that could access the raw memory directly. I can either call `write_at_offset` to copy one byte at a time or handle it differently depending on whether it runs in kernel mode or not.

litebox/src/fs/mod.rs
CvvT: I would like to have something similar to `pread` which does not change the internal file position. `lseek` may also be necessary.

litebox_shim_linux/src/syscall/mmap.rs
CvvT: Both litebox `FileSystem`  and linux shim layer maintain a file table. Is that redundant? Also, it seems to assume we only run a process. What about shared FileSystem for multiple processes (each process has its own file table)?

PR #29, opened 2025-03-07
support pread/pwrite

CvvT:
Support [pread/pwrite](https://www.man7.org/linux/man-pages/man2/pread.2.html). Compared to `read/write`, the only difference is whether it needs to update the file offset.

### DIFF
--- a/litebox/src/fs/in_mem.rs
+++ b/litebox/src/fs/in_mem.rs
@@ -181,7 +181,12 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
         Ok(())
     }
 
-    fn read(&self, fd: &FileFd, buf: &mut [u8]) -> Result<usize, ReadError> {
+    fn read(
+        &self,
+        fd: &FileFd,
+        buf: &mut [u8],
+        mut offset: Option<usize>,
+    ) -> Result<usize, ReadError> {
         let mut descriptors = self.descriptors.write();
         let Descriptor::File {
             file,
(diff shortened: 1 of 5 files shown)


litebox/src/fs/in_mem.rs
```diff
@@ -208,7 +214,12 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
         Ok(retlen)
     }
 
-    fn write(&self, fd: &FileFd, buf: &[u8]) -> Result<usize, WriteError> {
+    fn write(
+        &self,
+        fd: &FileFd,
+        buf: &[u8],
+        mut offset: Option<usize>,
```
wdcui, line 221: The definition in the trait is not `mut`.

litebox/src/fs/in_mem.rs
```diff
@@ -181,7 +181,12 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
         Ok(())
     }
 
-    fn read(&self, fd: &FileFd, buf: &mut [u8]) -> Result<usize, ReadError> {
+    fn read(
+        &self,
+        fd: &FileFd,
+        buf: &mut [u8],
+        mut offset: Option<usize>,
```
wdcui, line 188: The definition in the trait is not `mut`.

litebox/src/fs/in_mem.rs
jaybosamiya-ms: Essentially:
```rs
fn foo(mut x: u32) {
   ...
}
```
is equivalent to
```rs
fn foo(x: u32) {
   let mut x = x;
   ...
}
```
Roughly speaking, it is rebinding the variable itself to be mutable (notice that it is not a reference in either case). To a user of the function, both of the above are _exactly_ equivalent (the user is moving in a value into `foo`, what foo does with internally after the move is `foo`'s business).

_Personally_ I prefer the `let mut x = x;` approach, since it makes the case clearer, but also I know that for some people it is not the most preferred, so I won't take a particularly strong stance one way or another.

wdcui: Thanks!

wdcui:
Feel free to merge, @CvvT

PR #30, opened 2025-03-08
Add seek functionality to the various file systems

jaybosamiya-ms:
This is roughly analogous to `lseek`

### DIFF
--- a/litebox/src/fs/tests.rs
+++ b/litebox/src/fs/tests.rs
@@ -309,10 +309,60 @@ mod layered {
         fs.write(&fd2, b"share", None)
             .expect("Failed to write to file");
 
+        fs.seek(&fd1, 0, crate::fs::SeekWhence::RelativeToBeginning)
+            .expect("Failed to seek to start");
         let bytes_read = fs
             .read(&fd1, &mut buffer, None)
             .expect("Failed to read from file");
         assert_eq!(&buffer[..bytes_read], b"shareoo\n");
+
+        fs.close(fd1).expect("Failed to close file");
+        fs.close(fd2).expect("Failed to close file");
+    }
+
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
Interesting thing to note here: the SemverChecks CI gave a 🎉 rather than an :eyes: _even though_ the trait interface changed. This is because we had thought ahead and made it into a "sealed trait", thus users of the trait could not directly depend on the _non-existence_ of a particular method. Thus, adding a new method to this trait does not require a version bump (i.e., it does not violate semantic versioning to keep the version number the same), thus it gave the :tada:. Again worrying about semantic versioning of interfaces is not particularly helpful at the current moment in the project, but getting that 🎉 by the CI means that we are (successfully) thinking ahead to non-breaking changes as we build this version of LiteBox out, which is great :)

litebox/src/fs/in_mem.rs
```diff
@@ -253,6 +253,34 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
+        let file_len = file.read().data.len();
+        let base = match whence {
+            SeekWhence::RelativeToBeginning => 0,
+            SeekWhence::RelativeToCurrentOffset => *position,
+            SeekWhence::RelativeToEnd => file_len,
+        };
+        let new_posn = base
+            .checked_add_signed(offset)
+            .ok_or(SeekError::InvalidOffset)?;
+        if new_posn > file_len {
```
wdcui, line 276: Do we need to check if new_posn is negative?

jaybosamiya-ms: `new_posn` is a `usize`, thus is guaranteed to be non-negative.

PR #33, opened 2025-03-10
Split vmm

CvvT:
This PR does not add any new functionality but rather factor out the common part of `Vmem` useful for both user and kernel mode (i.e., virtual memory management). Now it is splitted into two parts: 1. `Vmem` resides in litebox, responsible for virtual memory management (i.e., permission checks) and 2. `VmemBackend` that does the actual jobs for memory allocation and etc (e.g., in kernel mode it would be the page table code).

### DIFF
--- a/litebox/src/mm/linux.rs
+++ b/litebox/src/mm/linux.rs
@@ -489,7 +489,17 @@ impl<Backend: VmemBackend<ALIGN>, const ALIGN: usize> Vmem<Backend, ALIGN> {
             unsafe { self.create_mapping(suggested_range, VmArea::new(before_flags), fixed_addr) }
                 .ok_or(MappingError::OutOfMemory)?;
         // call the user function with the pages
-        let _ = op(addr)?;
+        if let Err(e) = op(addr) {
+            // remove the mapping if the user function fails
+            unsafe {
+                self.remove_mapping(
+                    PageRange::new(addr.as_usize(), addr.as_usize() + suggested_range.len())
+                        .unwrap(),
+                )
+            }
+            .unwrap();
+            return Err(e);
+        }
         if before_flags != after_flags {
             // `protect` should succeed, as we just created the mapping.
             unsafe {


litebox/src/mm/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod vm;
```
jaybosamiya-ms: Minor: adding module-level docs
```suggestion
//! Memory management related functionality

pub mod vm;
```

litebox/src/mm/tests.rs
```diff
@@ -0,0 +1,194 @@
...
+use alloc::vec;
+use alloc::vec::Vec;
+
+use super::vm::{
+    NonZeroPageSize, PAGE_SIZE, PageRange, ProtectError, RemapError, VmArea, VmFlags, Vmem,
+    VmemBackend, VmemProtectError, VmemResizeError,
+};
+
+/// A dummy implementation of [`VmemBackend`] that does nothing.
+impl VmemBackend for () {
```
jaybosamiya-ms: Rather than implementing on `()` it might be better to implement it on top of an empty struct defined in this module.

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
+//! Implement [`Vmem`] that manages virtual address space backed by a memory [`VmemBackend`].
+//!
+
+use core::ops::Range;
```
jaybosamiya-ms: Remove unnecessary blank comment line, and improve comment a little for readability:
```suggestion
//! Manage [virtual address space](Vmem) backed by a memory [backend](VmemBackend).

use core::ops::Range;
```

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
...
+            | Self::VM_WRITE.bits()
+            | Self::VM_EXEC.bits();
+        const VM_MAY_ACCESS_FLAGS = Self::VM_MAYREAD.bits()
+            | Self::VM_MAYWRITE.bits()
+            | Self::VM_MAYEXEC.bits();
+    }
+}
+
+/// A page range that guarantees the `start` and `end` addresses
+/// are page aligned and `start` < `end`.
```
jaybosamiya-ms:
```suggestion
/// A non-empty range of page-aligned addresses
```

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
...
+}
+
+impl<const ALIGN: usize> Iterator for PageRange<ALIGN> {
+    type Item = usize;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.start < self.end {
+            let page = self.start;
+            self.start += ALIGN;
```
jaybosamiya-ms: This line breaks the guarantees that `PageRange` usually provides. In particular, an invariant of `PageRange` is that `start < end`. With this particular line, it is possible for `start == end`, leading to an empty `PageRange`. This is essentially why you need the `start < end` check in this function in the first place.

My suggestion for this is to store `end` using `Option<NonZeroUsize>`, and set it to `None` for if `start + ALIGN` would become equal to `end`. This would use the exact same amount of space but would allow for a more "correct" implementation.

The other alternative would be to implement a _separate_ iterator for it, for example, making `PageRange` an `IntoIterator` which you could use to simply return `(0..(end - start) / ALIGN).map(|i| start + ALIGN * i)` which would take care of the iteration for you.

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
...
+    }
+}
+
+impl<const ALIGN: usize> PageRange<ALIGN> {
+    /// Create a new page range.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `start` >= `end`.
+    pub fn new(start: usize, end: usize) -> Self {
```
jaybosamiya-ms: Rather than have this be a panic, consider returning `Option<Self>` which allows the user of this function to decide if they want it to be a panic or not.

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
...
+        assert!(start < end);
+        assert!((start % ALIGN) == 0);
+        assert!((end % ALIGN) == 0);
+        Self { start, end }
+    }
+}
+
+/// A non-zero 4KiB-page-aligned size in bytes.
+#[derive(Clone, Copy)]
+pub(super) struct NonZeroPageSize<const ALIGN: usize = PAGE_SIZE> {
```
jaybosamiya-ms: `pub(super)` in this module makes it available only for other modules in the `mm` space. This is a question more about _everything_ here that is marked `pub(super)`: do we expect these to become available to users in the future (i.e., `pub`)? Or are they only intended for internal use? If they are only for internal use, then generally speaking, I prefer internal use tests to sit within the same module (so they would not be `pub(super)`, and the tests that sit in `tests` module would only depend on actually `pub` functionality).

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// need this).
+    ///
+    /// # Safety
+    ///
+    /// `start` and `len` must be aligned to 4KiB.
+    ///
+    /// # Panics
+    ///
+    /// panic if `start` or `len` is misaligned.
+    unsafe fn unmap_pages(&mut self, start: usize, len: usize, free_page: bool);
```
jaybosamiya-ms: Rather than a panic, can we have this be a `Result<(), UnmapError>`?

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
...
+        None
+    }
+}
+
+pub trait VmemBackend {
+    /// Unmap 4KiB pages
+    ///
+    /// `free_page` indicates whether the unmapped pages should be freed (This may be helpful
+    /// when implementing [`Self::remap_pages`]. If we maintain refcnt for pages, we may not
+    /// need this).
```
jaybosamiya-ms: This comment could be improved. This comment seems specific to one particular implementation, and it is unclear how an "upper level" function should make decisions on whether to pass `free_page` or not.

Additionally, if `free_page`'s main usage is for `remap_pages`, then that is an implementation detail, and should not be included in the interface; instead, whichever type is implementing `VmemBackend` can choose to make its own helper function (which has the flag) and then have a thin call to the interface version without the boolean.

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
...
+    ///
+    /// # Safety
+    ///
+    /// `start` and `len` must be aligned to 4KiB.
```
jaybosamiya-ms: This safety section seems to talk about things included in the panic; however, it doesn't really talk about when `unmap_pages` is safe to execute or not safe to execute. With the current comment, it appears that it would be safe to _arbitrarily_ invoke `unmap_pages(0, 1024, true)` (just because both are aligned to 4KiB) which seems incorrect. We need to be more precise about safety requirements being specified.

jaybosamiya-ms: Additionally, some parts of this module are parametric in page size, while this is specific to 4KiB. It is probably better to make it parametric, or simply lock everything down to 4KiB, rather than have this inconsistent. Personally, I prefer the parametric one.

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// partially or completely replaced by the inserted range.
+    ///
+    /// If the inserted range either overlaps or is immediately adjacent
+    /// any existing range _mapping to the same value_, then the ranges
+    /// will be coalesced into a single contiguous range.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the range is beyond the task address space [[`Self::TASK_ADDR_MIN`], [`Self::TASK_ADDR_MAX`]]).
+    pub fn insert_mapping(&mut self, range: PageRange, vma: VmArea) {
```
wdcui: Why is this one not "pub(super)" as other methods?

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,531 @@
...
+            .map_or(high_limit, |r| r.0.start);
+        if low_limit + size <= first_start {
+            return Some(first_start - size);
+        }
+
+        None
+    }
+}
+
+pub trait VmemBackend {
```
wdcui: We don't need the trait to implement "map_pages"?

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,547 @@
...
+    size: usize,
+}
+
+impl<const ALIGN: usize> NonZeroPageSize<ALIGN> {
+    /// Create a new non-zero page-aligned size.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `size` is zero or not a multiple of [`PAGE_SIZE`].
+    pub fn new(size: usize) -> Self {
```
jaybosamiya-ms: Similarly, return `Option<Self>` here?

litebox/src/mm/vm.rs
CvvT: This function is used externally by another crate: https://github.com/MSRSSP/litebox/blob/0498e7f274be806a509b0e603a6ae604f013fd94/litebox_platform_linux_kernel/src/mm/vm.rs#L88
Other functions in this file will become `pub` in my next PR in which they are also exported to be used externally.

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,547 @@
...
+///
+/// This struct mantains the virtual memory areas of a process.
+pub struct Vmem<Backend: VmemBackend> {
+    /// Virtual memory areas. The ranges need to be page aligned.
+    vmas: RangeMap<usize, VmArea>,
+    /// Memory backend that provides the actual memory.
+    backend: Backend,
+}
+
+pub(super) const PAGE_SIZE: usize = 4096;
```
jaybosamiya-ms: Should this be `pub`? Why is this just `pub(super)`?

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,547 @@
...
+    pub const TASK_ADDR_MAX: usize = 0x7FFF_FFFF_F000; // (1 << 47) - PAGE_SIZE;
+
+    pub const fn new(backend: Backend) -> Self {
+        Self {
+            vmas: RangeMap::new(),
+            backend,
+        }
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = (&Range<usize>, &VmArea)> {
```
jaybosamiya-ms: Please include docs for every `pub` item.

litebox/src/mm/vm.rs
CvvT: Good catch! It was in my next PR, ported it into this one.

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,547 @@
...
+        self.vmas.iter()
+    }
+
+    pub fn get_backend(&self) -> &Backend {
+        &self.backend
+    }
+
+    pub fn get_inner_mut(&mut self) -> &mut Backend {
+        &mut self.backend
+    }
```
jaybosamiya-ms: These two together seem to imply that you can just make `backend` be a `pub` field, and remove the unnecessary indirection here.

litebox/src/mm/vm.rs
```diff
@@ -0,0 +1,547 @@
+//! Implement [`Vmem`] that manages virtual address space backed by a memory [`VmemBackend`].
```
jaybosamiya-ms: Maybe this module should be under `litebox::mm::linux` instead or directly as `litebox::mm::vm`? I say this because at least with the current design, there seem to be a large number of hardcoded bits that make me think that this is fairly Linux specific. It could be made general, but the fact that even constants like min/max are hardcoded and unchangeable, I find it a little difficult to believe that this is generally useful outside of Linux systems.

litebox_platform_linux_kernel/src/mm/mod.rs
```diff
@@ -3,19 +3,19 @@
...
 pub(crate) mod pgtable;
-#[cfg(test)]
-pub(crate) mod vm;
+pub mod vm;
 
 #[cfg(test)]
 pub mod tests;
 
 /// Memory provider trait for global allocator.
+#[sealed(pub(crate))]
```
jaybosamiya-ms: Sidenote: do we expect every host for the Linux kernel platform (even in the future) to sit within the host? If so, adding these `sealed` bits are perfectly fine, but otherwise `sealed` will have to be unsealed again in the future. I'm mentioning this mostly because I don't recall us making this decision to not allow others to build hosts for the kernel platform.

litebox/src/mm/vm.rs
CvvT: How other systems are different from Linux? Besides the min/max constants (which I can make them become variables/fields to this struct or parameters), any other parts that are not general? maybe the `VmFlags` or the behaviors of `resize_mapping` we discussed earlier?

jaybosamiya-ms: I didn't have any particular other system in mind, but mostly this comment came up due those exact parts that you mention. If we keep it within `linux`, then we don't need to worry about making a more general version, which (imho) makes it easier for us to focus on Linux-y things. Yes, we _can_ make it more general, but it is more a question of "do we need/want to"? We can afford to keep things more Linux-specific if we call it as a Linux-y subsystem. Otherwise, (imho) we need to worry more about whether the interface is general enough.

We can/should discuss this more offline, it is definitely possible I am being overly cautious with what becomes "top level" in the core LiteBox crate itself.

CvvT:
New code added in this update:
* All memory management APIs should be unsafe
* Add a new `MappingProvider` trait for creating readable/writable/executable pages
* Implement `MappingProvider` for our virtual memory mamager `Vmem`

litebox/src/mm/vm.rs
CvvT: They all became `pub`.

litebox/src/mm/vm.rs
CvvT: I see. This argument is removed.

litebox/src/mm/vm.rs
CvvT: Moved it to `litebox::mm::linux`. We can worry about other systems when we have one.

litebox/src/mm/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// # Safety
+    ///
+    /// The caller must ensure that the `data` is a valid initialization item for the backend.
+    /// See [`VmemBackend::new`] for more details.
+    pub unsafe fn new(
+        sync: &Synchronization<'platform, Platform>,
+        data: <<Platform as PageManagementProvider<ALIGN>>::Backend as VmemBackend<ALIGN>>::InitItem,
+    ) -> Self {
+        let backend = unsafe { <Platform as PageManagementProvider<ALIGN>>::Backend::new(data) };
+        let vmem = sync.new_rwlock(linux::Vmem::new(backend));
```
wdcui, line 50: Minor: add a comment here about that one can use other Vmem in the future?

litebox/src/mm/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+        Self { vmem }
+    }
+
+    /// Create readable and executable pages.
+    ///
+    /// `suggested_addr` is the hint address where to create the pages. Provide `0` to let the kernel
```
wdcui, line 56: Minor: to let the "kernel" => to let the page manager?

litebox/src/mm/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+    ///
+    /// Set `fixed_addr` to `true` to force the mapping to be created at the given address, resulting in any
+    /// existing overlapping mappings being removed.
+    ///
+    /// `op` is a callback for caller to initialize the created pages.
+    ///
+    /// # Safety
+    ///
+    /// If the suggested start address is given (i.e., not zero) and `fixed_addr` is set to `true`,
+    /// the kernel uses it directly without checking if it is available, causing overlapping
```
wdcui, line 67: Change the kernel to the page manager?

litebox/src/mm/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// The caller must ensure that the memory region is no longer used by any other.
+    pub unsafe fn remove_pages(
+        &self,
+        ptr: <<Platform as PageManagementProvider<ALIGN>>::Backend as VmemBackend<ALIGN>>::RawMutPointer,
+        len: usize,
+    ) -> Result<(), VmemUnmapError> {
+        let mut vmem = self.vmem.write();
+        let start = ptr.as_usize();
+        let range = PageRange::new(start, start + len).ok_or(VmemUnmapError::MisAligned)?;
+        unsafe { vmem.remove_mapping(range) }
```
wdcui, line 210: To match the function name "create_pages", shall we change "remove_mapping" to "remove_pages"?

litebox/src/mm/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /// Returns all mappings in a vector.
+    pub fn mappings(&self) -> Vec<(Range<usize>, VmFlags)> {
```
wdcui, line 214: Should this function be only used for testing?

litebox/src/mm/linux.rs
```diff
@@ -0,0 +1,729 @@
...
+            return None;
+        }
+        Some(unsafe { Self::new_unchecked(start, end) })
+    }
+
+    /// Create a new [`PageRange`] without checking for alignment or emptiness.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the range is `ALIGN`-aligned and not empty.
```
wdcui, line 92: What if the caller is in the user mode and litebox runs in the kernel mode? If the range is not aligned or it's empty, would it crash/affect litebox?

wdcui: Is this un internal function? If so, we should remove "pub".

litebox/src/mm/linux.rs
```diff
@@ -0,0 +1,729 @@
...
+            | Self::VM_EXEC.bits();
+        const VM_MAY_ACCESS_FLAGS = Self::VM_MAYREAD.bits()
+            | Self::VM_MAYWRITE.bits()
+            | Self::VM_MAYEXEC.bits();
+    }
+}
+
+/// A non-empty range of page-aligned addresses
+#[derive(Clone, Copy)]
+pub struct PageRange<const ALIGN: usize> {
```
wdcui, line 53: Should everything in this file be only visible to code in this crate? Do we expect code outside litebox to call/access them directly?

litebox/src/mm/linux.rs
```diff
@@ -0,0 +1,729 @@
...
+        op: F,
+    ) -> Result<<Backend as VmemBackend<ALIGN>>::RawMutPointer, MappingError>
+    where
+        F: FnOnce(<Backend as VmemBackend<ALIGN>>::RawMutPointer) -> Result<usize, MappingError>,
+    {
+        let addr =
+            unsafe { self.create_mapping(suggested_range, VmArea::new(before_flags), fixed_addr) }
+                .ok_or(MappingError::OutOfMemory)?;
+        // call the user function with the pages
+        let _ = op(addr)?;
```
wdcui: Should we handle the error first before returning it to the caller? For instance, should we remove the mapping first?

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
```diff
@@ -6,24 +10,33 @@ use x86_64::{
...
-    pgtable::{PageFaultError, PageTableAllocator, PageTableImpl, PageTableWalkError},
+use crate::{
+    mm::{
+        MemoryProvider,
+        pgtable::{PageTableAllocator, PageTableImpl},
+    },
+    ptr::UserMutPtr,
 };
 
+#[cfg(not(test))]
```
wdcui, line 28: Why do we need to treat FLUSH_TLB differently with/without test?

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
```diff
@@ -50,150 +63,87 @@ impl<M: MemoryProvider> FrameDeallocator<Size4KiB> for PageTableAllocator<M> {
...
-                panic!("Invalid frame address: {:#x}", pa);
-            }
-        }
-        Ok(())
+    unsafe fn map_pages(
+        &mut self,
+        range: PageRange<ALIGN>,
+        _flags: VmFlags,
+    ) -> Result<UserMutPtr<u8>, MmapError> {
+        // leave it to page fault handler
```
wdcui, line 93: Is this how Linux kernel does?

jaybosamiya-ms:
The biggest interface-level change I would like to be made is a fairly non-trivial refactor of this PR, so I've opened #36 for it. If you folks prefer, we can merge this current PR as-is and then consider #36 separately, since it is a non-trivial change to the current PR.

litebox/src/mm/linux.rs
CvvT: I should remove this function.

litebox/src/mm/linux.rs
CvvT: You're right!

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
CvvT: because it is a privileged instruction only kernel can use.

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
CvvT: No, just for simplicity. Linux has advanced optimizations.

CvvT:
> The biggest interface-level change I would like to be made is a fairly non-trivial refactor of this PR, so I've opened #36 for it. If you folks prefer, we can merge this current PR as-is and then consider #36 separately, since it is a non-trivial change to the current PR.

Sure. I just made one small change to handle an error case.

jaybosamiya-ms:
Yeah no worries, feel free to continue making updates to this PR as you feel necessary until you think it is ready to merge. I'll take care of the rebase of changes on #36.

CvvT:
> Yeah no worries, feel free to continue making updates to this PR as you feel necessary until you think it is ready to merge. I'll take care of the rebase of changes on #36.

I won't make any changes. I only addressed some comments that were from Weidong and do not relate to #36

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
wdcui: Please add a comment here to explain this choice. This choice potentially introduces slowdowns due to future page fault handling.

CvvT: I just checked again. Linux by default does not allocate any physical memory unless user specify the flag `MAP_POPULATE`, in which case we should handle it in the `mmap` syscall not here.

CvvT: Other comments should be address by #36

CvvT:
rebase on main branch and merge it as we discussed.

Issue #22, opened 2025-02-25
Factor out Linux common


jaybosamiya-ms:
We need to factor out the common elements from Linux kernel platform, and Linux shim into a separate `litebox_common_linux` crate.

For example, https://github.com/MSRSSP/litebox/blob/b90d32dc976ca8553fdedef708f7bc843173ce8c/litebox_platform_linux_kernel/src/error.rs#L1-L6 and https://github.com/MSRSSP/litebox/blob/b90d32dc976ca8553fdedef708f7bc843173ce8c/litebox_shim_posix/src/errno.rs#L55-L60 have significant commonalities, unsurprisingly.

jaybosamiya-ms:
The other bit that should sit in the common crate is the _definition_ (and not implementation) of the punchthrough for Linux syscalls.

jaybosamiya-ms:
This is effectively complete. Closing.

PR #34, opened 2025-03-14
Create the `litebox_common_linux` crate, starting with `errno`

jaybosamiya-ms:
This PR creates the `litebox_common_linux` crate (see #22), which will house bits that are common across platforms/shims on Linux systems. This PR only focuses on error numbers for now, since those are a large chunk of duplication across two existing crates.

This is not just a simple moving of code, but a refactor of the design of `Errno` to make it both more ergonomic and safe to use, compared to the `i32` based design used earlier.

The generated functionality (both for associated constants, as well as `as_str` for better human-readability for display/debugging purposes) in `generated.rs` relies on `/usr/bin/errno` and custom `awk` scripts that are included as doc strings in `generated.rs`. I don't expect us to need to re-generate these often (or even at all), but they exist there nonetheless.

### DIFF
--- a/litebox_common_linux/src/errno/mod.rs
+++ b/litebox_common_linux/src/errno/mod.rs
@@ -21,7 +21,7 @@ mod generated;
 ///
 /// This is a transparent wrapper around Linux error numbers (i.e., `i32`s) intended
 /// to provide some type safety by expecting explicit conversions to/from `i32`s.
-#[derive(PartialEq, Eq, Clone, Copy, Debug, Error)]
+#[derive(PartialEq, Eq, Clone, Copy, Error)]
 pub struct Errno {
     value: core::num::NonZeroU8,
 }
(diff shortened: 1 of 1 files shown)


litebox_common_linux/src/errno/generated.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// Owner died
+    pub const EOWNERDEAD: Self = Self::from_const(130);
+    /// State not recoverable
+    pub const ENOTRECOVERABLE: Self = Self::from_const(131);
+    /// Operation not possible due to RF-kill
+    pub const ERFKILL: Self = Self::from_const(132);
+    /// Memory page has hardware error
+    pub const EHWPOISON: Self = Self::from_const(133);
+    /// Operation not supported
+    pub const ENOTSUP: Self = Self::from_const(95);
```
wdcui, line 437: Just to double check: this is supposed to be another constant of value 95?

jaybosamiya-ms: Yes, there are a few constants that have multiple names. The full list:
```
EAGAIN 11 Resource temporarily unavailable
EWOULDBLOCK 11 Resource temporarily unavailable

EDEADLK 35 Resource deadlock avoided
EDEADLOCK 35 Resource deadlock avoided

EOPNOTSUPP 95 Operation not supported
ENOTSUP 95 Operation not supported
```

Issue #24, opened 2025-02-28
Consider a component registry pattern


jaybosamiya-ms:
At a high level, the idea is that all components `fs`, `net`, ... would sit within a top-level registry, thereby allowing convenient inter-dependency (rather than forcing a tree-like shape), without imposing the cost of full arc-rwlocks for _every_ use case. 

Importantly, this would allow single-process single-threaded contexts to be able to have almost zero-cost access to things, while the threaded cases will look similar.

More importantly however, it would ensure that users of LiteBox do not have to come up with a storage discipline for the components, and instead simply are registering components into the services.

Roughly speaking, the user would then simply be able to say something like:

```rs
let litebox = LiteBox::new::<litebox_platform_linux_user::LinuxUserland>()
  .register::<litebox::net::Network>()
  .register::<litebox::fs::in_mem::FileSystem>()
  .finish();
```
and then be able to access things via (as example):
```rs
litebox.get::<litebox::net::Network>().socket(...)
```

Obviously, we can provide convenience functions like `litebox.get_network()` or such, and can even provide nicer wrappers to the above builder functionality. But at its core, this would then allow each component to simply hold on to a handle (internally) to the registry, from which it can access other `RegisterableComponent`s.

This would be a win over the current approach which has the user manually needing to store and manage access to the various components (which can lead to some awkwardness, in cases like the 9p client fs, where network access would either need to be stored within the fs component, or there would need to be careful management of arc-rwlocks).

---

Sidenote: I _believe_ the above design is implementable, _however_ there might be some lifetime issues to deal with if we are using [anymap](https://docs.rs/anymap/latest/anymap/index.html) as the underlying idea for it, so I am not 100% sure on the implementation just yet. However, I do believe that the overall effect (if we can provide the above interface) would be a significant improvement.

PR #35, opened 2025-03-15
Elf loader

CvvT:
Implemented an elf loader for both statically and dynamically linked executables.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -19,7 +19,6 @@ use alloc::vec::Vec;
 use once_cell::race::OnceBox;
 
 use litebox::{
-    fs::FileSystem as _,
     mm::{PageManager, linux::PAGE_SIZE},
     platform::RawConstPointer as _,
     sync::RwLock,
@@ -212,15 +211,5 @@ pub unsafe extern "C" fn open(pathname: ConstPtr<i8>, flags: u32, mode: u32) -> 
 
 /// Closes the file
 pub extern "C" fn close(fd: i32) -> i32 {
-    let Ok(fd) = u32::try_from(fd) else {
-        return Errno::EBADF.as_neg();
-    };
-    match file_descriptors().write().remove(fd) {
-        Some(Descriptor::File(file_fd)) => match litebox_fs().close(file_fd) {
-            Ok(()) => 0,
-            Err(err) => Errno::from(err).as_neg(),
-        },
-        Some(Descriptor::Socket(socket_fd)) => todo!(),
-        None => Errno::EBADF.as_neg(),
-    }
+    syscalls::file::sys_close(fd).map_or_else(Errno::as_neg, |()| 0)
 }
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -38,3 +38,15 @@ pub(crate) fn sys_read(fd: i32, buf: &mut [u8], offset: Option<usize>) -> Result
         None => Err(Errno::EBADF),
     }
 }
+
+/// Close a file
+pub(crate) fn sys_close(fd: i32) -> Result<(), Errno> {
+    let Ok(fd) = u32::try_from(fd) else {
+        return Err(Errno::EBADF);
+    };
+    match file_descriptors().write().remove(fd) {
+        Some(Descriptor::File(file_fd)) => litebox_fs().close(file_fd).map_err(Errno::from),
+        Some(Descriptor::Socket(socket_fd)) => todo!(),
+        None => Err(Errno::EBADF),
+    }
+}


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -75,6 +75,18 @@ impl<PunchthroughProvider: litebox::platform::PunchthroughProvider>
             punchthrough_provider,
         }
     }
+
+    /// Create a new userland-Linux platform for use in `LiteBox` for testing.
+    ///
+    /// # Safety
+    ///
+    /// This is for testing purposes only.
+    pub unsafe fn new_for_test(punchthrough_provider: PunchthroughProvider) -> Self {
```
wdcui, line 86: If this function is only for testing, should we add a config flag for it?

litebox_shim_linux/src/lib.rs
```diff
@@ -18,18 +18,31 @@ use alloc::vec::Vec;
...
     FS.get_or_init(|| {
-        alloc::boxed::Box::new(litebox::fs::in_mem::FileSystem::new(
-            litebox_platform_multiplex::platform(),
-        ))
+        let mut in_mem_fs =
+            litebox::fs::in_mem::FileSystem::new(litebox_platform_multiplex::platform());
+        #[cfg(test)]
+        in_mem_fs.with_root_privileges(|fs| {
+            use litebox::fs::Mode;
+            fs.chmod("/", Mode::RWXU | Mode::RWXG | Mode::RWXO)
```
wdcui: Why do we need to change mode for testing?

litebox_shim_linux/src/loader/hello.c
wdcui, line 1: Should we check in this file into some test folder?

litebox_shim_linux/src/loader/stack.rs
wdcui, line 1: Should the stack management in a module that does process/thread management? We will need it when we start a new thread?

litebox_shim_linux/src/loader/stack.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    /* 28...30 not used */
+    AT_EXECFN = 31, /* filename of program */
+    AT_SYSINFO = 32,
+    AT_SYSINFO_EHDR = 33, /* the start address of the page containing the VDSO */
+}
+
+/// The stack layout for the user process. This is used to set up the stack
+/// for the new process. The stack is set up in a way that is compatible with
+/// the Linux ABI.
```
wdcui: Is this the ABI for how to pass arguments in syscalls?

litebox_shim_linux/src/loader/stack.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// Get the current stack pointer.
+    pub(super) fn get_cur_stack_top(&self) -> usize {
+        debug_assert!(self.pos >= 0);
+        self.stack_top.as_usize() + self.pos.unsigned_abs()
+    }
+
+    /// Push `bytes` to the stack.
+    ///
+    /// Returns `None` if stack has no enough space.
+    fn write_bytes(&mut self, bytes: &[u8]) -> Option<()> {
```
wdcui: Shall we change "write" to "push" to make it more clear what this function does?

litebox_shim_linux/src/loader/stack.rs
```diff
@@ -0,0 +1,239 @@
...
+    ///
+    /// `offsets` are the offsets of the pointers in the stack.
+    ///
+    /// Returns `None` if stack has no enough space.
+    fn write_pointers(&mut self, offsets: Vec<isize>) -> Option<()> {
+        // write end marker
+        self.write_usize(0)?;
+        let len = offsets.len();
+        let old_pos = self.pos;
+        self.pos = self.pos.checked_sub_unsigned(len * size_of::<isize>())?;
```
wdcui: Do we need to first do a checked multiplication of len * sizeof()?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -0,0 +1,410 @@
...
+
+// An opened elf file
+struct ElfFile {
+    name: CString,
+    fd: u32,
+}
+
+impl ElfFile {
+    pub(super) fn from_path(path: &str) -> Result<Self, OpenError> {
+        let file = litebox_fs().open(path, OFlags::RDWR, Mode::empty())?;
```
wdcui: If we don't write to the file, should we open it read-only?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -0,0 +1,410 @@
...
+        let fd = file_descriptors()
```
wdcui: I'm not sure if this is the right way to interact with both fs and file descriptors here? If fd is required, should we use the fs function in the shim which returns a fd?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+}
+
+impl ElfObject for ElfFile {
+    fn file_name(&self) -> &core::ffi::CStr {
+        &self.name
+    }
+
+    fn read(&mut self, buf: &mut [u8], offset: usize) -> elf_loader::Result<()> {
+        if let Some(file) = file_descriptors().read().get_file_fd(self.fd) {
```
wdcui: Should we just save the "file" returned from litebox_fs.open() into self.file?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -0,0 +1,410 @@
...
+        }
+    }
+
+    fn as_fd(&self) -> Option<i32> {
+        Some(self.fd.try_into().unwrap())
+    }
+}
+
+/// [`elf_loader::mmap::Mmap`] implementation for ELF loader
+/// using [`litebox::mm::mapping::MappingProvider`].
```
wdcui: Update the comment (no MappingProvider anymore)?

litebox_platform_linux_userland/src/lib.rs
CvvT: Sure

litebox_shim_linux/src/lib.rs
CvvT: We need to install our test executables to the in memory filesystem.

litebox_shim_linux/src/loader/elf.rs
CvvT: I cannot save a reference or copy because file_descriptors().write().insert would take the ownership of the `file`. Making `file` sharable may require significant change.

litebox_shim_linux/src/loader/elf.rs
CvvT: Right, it should be read-only. There seems to be an issue.

https://github.com/MSRSSP/litebox/blob/5d233e7f5a3ab7546fcbf0b72674c61ff390c07b/litebox/src/fs/in_mem.rs#L157-L164

@jaybosamiya-ms  The check at 160 should be moved to the if branch?

litebox_shim_linux/src/loader/elf.rs
CvvT: You mean `open` from `lib.rs`? `open` takes a ConstPtr<u8> which may represent a user-space pointer, but I guess I can split `open` into two as follows:

```Rust
pub unsafe extern "C" fn open(pathname: ConstPtr<i8>, flags: u32, mode: u32) -> i32 {
    let Some(path) = pathname.to_cstring() else {
        return -errno::constants::EFAULT;
    };

    sys_open(path, litebox::fs::OFlags::from_bits(flags).unwrap(), litebox::fs::Mode::from_bits(mode).unwrap())
}

pub sys_open(path: impl path::Arg, flags: OFlags, mode: Mode) -> i32 {
    match litebox_fs().open(
        path,
        flags,
        mode,
    ) {
        Ok(fd) => file_descriptors()
            .write()
            .insert(Descriptor::File(fd))
            .try_into()
            .unwrap(),
        Err(err) => -err.as_errno(),
    }
}
```

Then call `sys_open` directly.

litebox_shim_linux/src/loader/stack.rs
CvvT: Sure

litebox_shim_linux/src/loader/stack.rs
CvvT: Sure

litebox_shim_linux/src/loader/stack.rs
CvvT: I moved the stack layout here. it might be more clear.

litebox_shim_linux/src/loader/stack.rs
CvvT: yes, not just stack, loader shoud also be part of process management.

litebox_shim_linux/src/loader/elf.rs
CvvT: done

litebox_shim_linux/src/loader/hello.c
CvvT: Moved all tests to intergration test folder: https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html so that each test can run separately.

litebox_shim_linux/src/loader/stack.rs
CvvT: I may restructure it once we have process management.

litebox_shim_linux/src/loader/elf.rs
jaybosamiya-ms: Oops, yes #37

CvvT:
This PR is not ready for reviewing yet as it relies on #36 and CI change for Nextest.

jaybosamiya-ms:
Nextest CI change is at #38. Changes/improvements to #36 coming soon.

CvvT:
[like]  Weiteng Chen reacted to your message:
________________________________

litebox_shim_linux/src/loader/elf.rs
CvvT: added `sys_open` and `sys_read` to `syscalls` folder

litebox_shim_linux/src/loader/hello.c
CvvT: Moved the tests back to unit tests

litebox_shim_linux/src/syscalls/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! Syscalls Handlers
+
+pub(crate) mod open;
+pub(crate) mod read;
```
jaybosamiya-ms: Any particular reason we have this as an indirection, and not just directly implemented in this file?

litebox_platform_linux_userland/Cargo.toml
```diff
@@ -3,6 +3,9 @@ name = "litebox_platform_linux_userland"
 version = "0.1.0"
 edition = "2024"
 
+[features]
+testing = []
```
jaybosamiya-ms: I recommend using the feature name `unstable-testing` which explicitly lets tooling (including our semver checks tooling) know that this feature is meant to be unstable, and thus changes to things within this particular feature should not be considered a "breaking change", but is instead _intentionally_ unstable. Also, adding a brief comment linking to https://github.com/rust-lang/cargo/issues/8379 might be good.

litebox_platform_multiplex/Cargo.toml
```diff
@@ -10,6 +10,7 @@ once_cell = { version = "1.20.2", default-features = false, features = ["alloc",
 
 [features]
 default = ["platform_linux_userland"]
+testing = ["litebox_platform_linux_userland?/testing"]
```
jaybosamiya-ms: Similar `unstable-testing`

litebox_shim_linux/src/lib.rs
```diff
@@ -18,12 +18,33 @@ use alloc::vec::Vec;
...
+};
 use litebox_common_linux::errno::Errno;
 use litebox_platform_multiplex::Platform;
 
-pub(crate) fn litebox_fs<'a>() -> &'a impl litebox::fs::FileSystem {
-    static FS: OnceBox<litebox::fs::in_mem::FileSystem<Platform>> = OnceBox::new();
+pub mod loader;
+pub mod syscalls;
+
+use syscalls::open::sys_open;
```
jaybosamiya-ms: Nit: generally speaking in Rust, it is not very common to do a `use` for functions. Use-declarations are normally restricted for types, traits, and modules. This helps reduce the indirection needed.

litebox_shim_linux/src/lib.rs
```diff
@@ -18,12 +18,33 @@ use alloc::vec::Vec;
...
+/// Set the global file system
+///
+/// # Panics
+///
+/// Panics if this is called more than once or `litebox_fs` is called before this
+pub fn set_fs(fs: litebox::fs::in_mem::FileSystem<'static, Platform>) {
+    FS.set(alloc::boxed::Box::new(fs))
+        .map_err(|_| {})
+        .expect("fs is already set");
+}
```
jaybosamiya-ms: I think this should be gated under test/unstable-testing. The shape of how FS is set up will change based on #24 which will obviate this. It being under a flag would mean that it is easier to reason about the change later.

litebox_shim_linux/src/lib.rs
```diff
@@ -40,6 +61,14 @@ pub(crate) fn litebox_sync<'a>() -> &'a litebox::sync::Synchronization<'static,
     })
 }
 
+pub(crate) fn litebox_vmm<'a>() -> &'a PageManager<'static, Platform, PAGE_SIZE> {
```
jaybosamiya-ms: Call this page manager instead of vmm, for naming consistency?

litebox_shim_linux/src/syscalls/read.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use crate::{file_descriptors, litebox_fs};
+
+/// Read from a file
+///
+/// `offset` is an optional offset to read from. If `None`, it will read from the current file position.
+/// If `Some`, it will read from the specified offset without changing the current file position.
+pub(crate) fn sys_read(fd: i32, buf: &mut [u8], offset: Option<usize>) -> Result<usize, ReadError> {
+    let Ok(fd) = u32::try_from(fd) else {
+        return Err(ReadError::NotAFile);
```
jaybosamiya-ms: I don't think `NotAFile` is a valid error here. I think it should be something equivalent to `EBADF` (currently `NotAFile` refers to `EISDIR`).

litebox_shim_linux/src/syscalls/read.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use litebox::fs::{FileSystem, errors::ReadError};
+
+use crate::{file_descriptors, litebox_fs};
+
+/// Read from a file
+///
+/// `offset` is an optional offset to read from. If `None`, it will read from the current file position.
+/// If `Some`, it will read from the specified offset without changing the current file position.
+pub(crate) fn sys_read(fd: i32, buf: &mut [u8], offset: Option<usize>) -> Result<usize, ReadError> {
```
jaybosamiya-ms: I think the better signature here is to return a `Result<usize, Errno>`

litebox_shim_linux/src/syscalls/open.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use litebox::{
+    fs::{FileSystem as _, Mode, OFlags, errors::OpenError},
+    path,
+};
+
+use crate::{Descriptor, file_descriptors, litebox_fs};
+
+/// Open a file
+pub(crate) fn sys_open(path: impl path::Arg, flags: OFlags, mode: Mode) -> Result<i32, OpenError> {
```
jaybosamiya-ms: I think the better signature here is to return a `Result<i32, Errno>`

litebox_shim_linux/src/loader/stack.rs
```diff
@@ -0,0 +1,241 @@
+//! This module contains the stack layout for the user process.
```
jaybosamiya-ms:
```suggestion
//! This module manages the stack layout for the user process.
```

litebox_shim_linux/src/loader/stack.rs
```diff
@@ -0,0 +1,241 @@
...
+use alloc::{collections::btree_map::BTreeMap, ffi::CString, vec::Vec};
+use litebox::platform::{RawConstPointer, RawMutPointer};
+
+use crate::MutPtr;
+
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[repr(u8)]
+pub enum AuxKey {
+    AT_NULL = 0,      /* end of vector */
```
jaybosamiya-ms: These comments can be made into doc comments

litebox_shim_linux/src/loader/stack.rs
```diff
@@ -0,0 +1,241 @@
...
+
+use alloc::{collections::btree_map::BTreeMap, ffi::CString, vec::Vec};
+use litebox::platform::{RawConstPointer, RawMutPointer};
+
+use crate::MutPtr;
+
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[repr(u8)]
+pub enum AuxKey {
```
jaybosamiya-ms: I'm surprised at the use of `pub` inside a non-pub module. Is this intentional?

litebox_shim_linux/src/loader/stack.rs
```diff
@@ -0,0 +1,241 @@
...
+            len,
+            pos: isize::try_from(len).ok()?,
+        })
+    }
+
+    /// Get the current stack pointer.
+    pub(super) fn get_cur_stack_top(&self) -> usize {
+        debug_assert!(self.pos >= 0);
+        self.stack_top.as_usize() + self.pos.unsigned_abs()
+    }
```
jaybosamiya-ms: Why not just have `pos` be a `usize` rather than an `isize`? Currently this seems confusing. Are you intentionally letting `pos` be negative whenever this function is not called? Even so, you can use [`usize::checked_add_signed`](https://doc.rust-lang.org/std/primitive.usize.html#method.checked_add_signed) to make things safer than suddenly taking the absolute value.

litebox_shim_linux/src/loader/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! This module contains the loader for the LiteBox shim.
```
jaybosamiya-ms, line 1: Things in this module seem 64-bit specific, might be good to place a `#![cfg(target_arch = ...)]` here?

litebox_shim_linux/src/loader/stack.rs
```diff
@@ -0,0 +1,241 @@
...
+            self.push_usize(val)?;
+            self.push_usize(key as usize)?;
+        }
+        Some(())
+    }
+
+    fn align_down(pos: isize, alignment: usize) -> isize {
+        assert!(alignment.is_power_of_two());
+        pos & !isize::try_from(alignment - 1).unwrap()
+    }
```
jaybosamiya-ms: I'm surprised by this being an associated function. I see it only being used inside `init`; in that case, you can just move it inside `init`'s body, so you don't have to do `Self::align_down`)

litebox_shim_linux/src/loader/stack.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    fn get_random_value() -> [u8; 16] {
+        // TODO: generate a random value
+        [
+            0xDE, 0xAD, 0xBE, 0xEF, 0xDE, 0xAD, 0xBE, 0xEF, 0xDE, 0xAD, 0xBE, 0xEF, 0xDE, 0xAD,
+            0xBE, 0xEF,
+        ]
+    }
```
jaybosamiya-ms: Similarly, should probably be moved into `init`.

Additionally (not as part of this PR) we likely should expose a randomness interface from litebox, which is where this should pull in randomness from. I've opened #41 to track this, but we should probably link to it in this TODO comment.

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -0,0 +1,395 @@
...
+use super::stack::{AuxKey, UserStack};
+
+// An opened elf file
+struct ElfFile {
+    name: CString,
+    fd: i32,
+}
+
+impl ElfFile {
+    fn from_path(path: &str) -> Result<Self, OpenError> {
```
jaybosamiya-ms: This function should probably be called `new`?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -0,0 +1,395 @@
...
+        let fd = sys_open(path, OFlags::RDONLY, Mode::empty())?;
+
+        Ok(Self {
+            name: CString::from_str(path).unwrap(),
```
jaybosamiya-ms: Nit: mildly safer to perform these two calls in the opposite order

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -0,0 +1,395 @@
...
+}
+
+impl ElfObject for ElfFile {
+    fn file_name(&self) -> &core::ffi::CStr {
+        &self.name
+    }
+
+    fn read(&mut self, buf: &mut [u8], offset: usize) -> elf_loader::Result<()> {
+        match sys_read(self.fd, buf, Some(offset)) {
+            Ok(_) => Ok(()),
```
jaybosamiya-ms: Given the signature of the function here, I suspect you need to do a loop of repeatedly running `sys_read` until sufficient amount has been read in, otherwise we might be left with a partially read file.

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -0,0 +1,395 @@
...
+    }
+}
+
+/// [`elf_loader::mmap::Mmap`] implementation for ELF loader
+struct ElfLoaderMmap;
+
+impl ElfLoaderMmap {
+    const PROT_EXECUTABLE: c_int = ProtFlags::PROT_READ.bits() | ProtFlags::PROT_EXEC.bits();
+    const PROT_WRITABLE: c_int = ProtFlags::PROT_READ.bits() | ProtFlags::PROT_WRITE.bits();
+    const PROT_READABLE: c_int = ProtFlags::PROT_READ.bits();
```
jaybosamiya-ms, line 84: I don't think the indirection by these constants is helping with readability; you can include them in the match statements directly (if necessary, you can use const expressions, e.g., `const { ProtFlags::PROT_READ.bits() | ProtFlags::PROT_EXEC.bits() }`)

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -0,0 +1,395 @@
...
+    fn do_mmap_file(
+        addr: Option<usize>,
+        len: usize,
+        prot: ProtFlags,
+        flags: MapFlags,
+        fd: i32,
+        offset: usize,
+    ) -> elf_loader::Result<usize> {
+        // TODO: we copy the file to the memory to support file-backed mmap.
+        // Loader may rely on `mmap`` instead of `mprotect` to change the memory protection,
```
jaybosamiya-ms: Typo
```suggestion
        // Loader may rely on `mmap` instead of `mprotect` to change the memory protection,
```

litebox_shim_linux/hello_dylib
jaybosamiya-ms, line 1: This seems to have been included accidentally? I believe this should be added to the `.gitignore` since it is generated as part of the tests.

litebox_shim_linux/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    // Just to make sure cargo sets up the `OUT_DIR` environment variable
```
jaybosamiya-ms, line 2: Maybe `CARGO_TARGET_DIR` is a better place to place the test executables? Alternatively, if we are doing it only for integration tests, then `CARGO_TARGET_TMPDIR` is a better place?

litebox_shim_linux/src/syscalls/mod.rs
CvvT: I don't want to implement all syscalls in just one file (in vsbox it became a very long file), but I guess could group related syscalls, e.g., put `sys_open` and `sys_read` in `file.rs`.

litebox_platform_linux_userland/Cargo.toml
CvvT: Good point!

litebox_shim_linux/build.rs
CvvT: I just tried and that variable was not present. I checked the [doc](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads) again, looks like that are the env variables cargo reads but not set for crates. For env variables set by cargo, see section [Environment variables Cargo sets for crates](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates). Alternatively, I can config the target dir (https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir) and hardcode the path, but it does not feel right either.

litebox_shim_linux/hello_dylib
CvvT: I forgot to delete this one.

litebox_shim_linux/src/lib.rs
CvvT: I see

litebox_shim_linux/src/lib.rs
CvvT: Added `#cfg(test)`

litebox_shim_linux/src/lib.rs
CvvT: Agree!

litebox_shim_linux/src/loader/elf.rs
CvvT: Yes, `new` might be better (or `from_str`)

litebox_shim_linux/src/loader/elf.rs
CvvT: Got it

litebox_shim_linux/src/loader/elf.rs
CvvT: Good catch!

litebox_shim_linux/src/loader/elf.rs
CvvT: Looks like they are going to remove this feature: https://github.com/rust-lang/rust/issues/76001/

litebox_shim_linux/src/loader/elf.rs
CvvT: Fixed!

litebox_shim_linux/src/loader/mod.rs
CvvT: Good point. The crate I used only supports 64 bits: https://github.com/weizhiao/rust-elfloader?tab=readme-ov-file#architecture-support

litebox_shim_linux/src/loader/stack.rs
CvvT: No, I forgot to change it.

litebox_shim_linux/src/loader/stack.rs
CvvT: Sure

litebox_shim_linux/src/loader/stack.rs
CvvT: Mostly because `mutate_subslice_with` takes isize, but now I feel using usize makes more sense

litebox_shim_linux/src/loader/stack.rs
CvvT: Why is that better? I thought it would be inlined anyway.

litebox_shim_linux/src/syscalls/open.rs
CvvT: Right

litebox_shim_linux/src/syscalls/read.rs
CvvT: Right!

litebox_shim_linux/src/syscalls/read.rs
CvvT: I see. Thanks for clarifying it.

litebox/src/mm/linux.rs
```diff
@@ -620,8 +620,8 @@ pub enum MappingError {
     MisAligned,
     #[error("not enough memory")]
     OutOfMemory,
-    #[error("failed to read from file")]
-    ReadError(#[from] crate::fs::errors::ReadError),
+    #[error("failed to read from file: {0}")]
+    ReadError(i32),
```
jaybosamiya-ms: I'm not sure changing this is the right move, but I also see why you've done it. While this _may_ be solvable with a wrapper enum, it might be better if this was a new error code, rather than replacing out the `ReadError`, possibly add a:
```
/// For errors that do not fit into the regular hierarchy, but require user-defined understanding of the error value
#[error("Error code {0}")]
ErrorCode(i32),
```

I am not too thrilled about this (it feels like adding "magic numbers" even if we all agree on what the magic numbers are), but I would completely understand if you took this route since it is the path of least friction for now, and we can clean it up separately. We can just open a `priority-low` issue to track this in that case.

litebox_shim_linux/src/syscalls/mod.rs
jaybosamiya-ms: That sounds reasonable; indeed grouping related syscalls sounds like a good idea, thanks!

litebox_shim_linux/src/loader/stack.rs
jaybosamiya-ms: Ah, you are right about `mutate_subslice_with`, and I understand now. I still think using `usize` here is safer. The main reason `mutate_subslice_with` exists with `isize` is that _technically_ you may sometimes have buffers where you want to access negative indexes into it. It is quite rare though, so maybe the interface there needs to change a little bit to improve it (i.e., give a nicer default which uses `usize`s, and a rarely-used-but-more-general `isize` based version). I'll give it some thought on how that interface can be improved.

litebox_shim_linux/src/loader/stack.rs
jaybosamiya-ms: Yes, it is inlined-away anyways, so would not impact perf in any way. It is more so just from a stylistic perspective: `align_down` has nothing to do with the the type it is associated with, so having it being referred to as `Self::align_down` feels weird. If this was a separate function in the same file, that'd be fine too, but stylistically it felt _slightly_ better to place it inside `init` as an indicator of "this is the only usage of it".

The actually ideal situation of course would be if Rust itself gave us a nicer thing. Currently Rust only gives us [`next_power_of_two`](https://doc.rust-lang.org/std/primitive.usize.html#method.next_power_of_two), which is mildly unfortunate.

litebox_shim_linux/src/loader/elf.rs
jaybosamiya-ms: Ah, I didn't realize this was feature-gated and _also_ potentially under removal. Unfortunate. I guess we're sticking with the constants here then. Thanks for checking that and letting me know.

litebox_shim_linux/build.rs
jaybosamiya-ms: Ah, unfortunate. Good point with keeping it as `OUT_DIR` in that case. Thanks for checking through the details on that.

litebox_shim_linux/src/loader/mod.rs
```diff
@@ -1,4 +1,5 @@
 //! This module contains the loader for the LiteBox shim.
 
+#[cfg(target_arch = "x86_64")]
```
jaybosamiya-ms: Should this be
```suggestion
#![cfg(target_arch = "x86_64")]
```
instead, so as to cover the stack too?

CvvT: This is nice. I didn't know it before.

litebox/src/mm/linux.rs
CvvT: I see. I added more error codes here.

PR #42, opened 2025-03-20
Syscall rewriter

wdcui:
This PR adds the support of static binary rewriting that hooks syscalls in an ELF file. We need this functionality because we would like to redirect syscalls made by an app to litebox when the latter runs in the user mode. We modify each syscall to add an indirect jmp to a trampoline . All trampolines are stored in a new .trampoline section. The trampoline section starts with 8 bytes which will be used by the litebox loader to store the syscall entry in litebox.

### DIFF
--- a/litebox_syscall_rewriter/src/main.rs
+++ b/litebox_syscall_rewriter/src/main.rs
@@ -1,7 +1,7 @@
 use anyhow::{Context, Result};
 use capstone::prelude::*;
 use object::build::elf;
-use std::{env, fs, path::Path};
+use std::{env, fs, path::Path, vec};
 
 fn hook_syscalls(
     text_data: &mut [u8],
@@ -145,7 +145,9 @@ fn rewrite_elf(in_path: &Path) -> Result<()> {
         text_segment.append_section(trampoline_section);
         (trampoline_section.id(), trampoline_section.sh_addr)
     };
-    let mut trampoline_data = vec![0x0; 0x8];
+
+    // We use the first 8 bytes to store the syscall hook address
+    let mut trampoline_data = vec![0; 8];
 
     // Hook syscalls
     hook_syscalls(


litebox_syscall_rewriter/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "litebox_syscall_rewriter"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+capstone = { version = "0.13", features = ["full"] }
+object = { version = "0.36.7", features = ["build"] }
```
jaybosamiya-ms: Slightly more restricted feature set
```suggestion
object = { version = "0.36.7", default-features = false, features = ["build_core", "elf", "std"] }
```

litebox_syscall_rewriter/src/main.rs
```diff
@@ -0,0 +1,217 @@
+use anyhow::{Context, Result};
+use capstone::prelude::*;
+use object::build::elf;
+use std::{env, fs, path::Path, vec::Vec};
```
jaybosamiya-ms: `Vec` is already part of the `std` prelude, so you don't need to re-import it.
```suggestion
use std::{env, fs, path::Path};
```

litebox_syscall_rewriter/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "litebox_syscall_rewriter"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+capstone = { version = "0.13", features = ["full"] }
```
jaybosamiya-ms, line 7: I don't expect this to be a bottleneck, but capstone is not particularly fast at disassembling. Iced is 10~30x faster, so that might be something worth considering if this is ever a bottleneck for us (which again, I don't expect it to be; just making a note here for future reference).

litebox_syscall_rewriter/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+
+fn hook_syscalls(
+    text_data: &mut Vec<u8>,
```
jaybosamiya-ms: This makes it clearer that the size of the `text_data` won't change:
```suggestion
    text_data: &mut [u8],
```
To do this, you also need to remove the `.as_slice()` on line 19, but otherwise it works.

litebox_syscall_rewriter/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+                text_data[replace_offset + idx] = 0x90;
+            }
+        }
+    }
+
+    Ok(())
+}
+
+#[allow(clippy::too_many_lines)]
+fn rewrite_elf(in_path: &Path) -> Result<()> {
```
jaybosamiya-ms, line 80: Automatically picking foo -> foo.hooked is not ideal for the CLI. I'm happy to write a quick separate PR once this is merged to give a better interface.

Additionally, we may wish to library-ify this, for use with convenience wrappers (that do a compile+run thing). I'm happy to do that in a separate PR too.

litebox_syscall_rewriter/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+    let in_data = &*in_data;
+
+    let file_kind =
+        object::FileKind::parse(in_data).with_context(|| "Failed to parse file kind")?;
+    let arch_mode = match file_kind {
+        object::FileKind::Elf32 => capstone::arch::x86::ArchMode::Mode32,
+        object::FileKind::Elf64 => capstone::arch::x86::ArchMode::Mode64,
+        _ => {
+            return Err(anyhow::anyhow!("Unsupported file format"));
+        }
```
jaybosamiya-ms: 32-bit ELFs don't actually use the `syscall` instruction, thus they would need a different rewriter. For now, it might be better to restrict ourselves to 64-bit ELFs.

Also, `anyhow` has a nicer macro for the `return Err(anyhow::anyhow!(...);`.
```suggestion
        object::FileKind::Elf64 => capstone::arch::x86::ArchMode::Mode64,
        _ => anyhow::bail!("Unsupported file format"),
```

litebox_syscall_rewriter/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+    };
+
+    let mut builder = elf::Builder::read(in_data)?;
+
+    // Locate the .text section and its segment
+    let mut text_section_id: Option<elf::SectionId> = None;
+    let mut text_segment_id: Option<elf::SegmentId> = None;
+    for section in &builder.sections {
+        if section.name.eq_ignore_ascii_case(".text".as_bytes()) {
```
jaybosamiya-ms, line 100: Not sure why the case is ignored here. Additionally, probably the right move here is `section.sh_type == elf::SHT_PROGBITS` instead? This will require restructuring the rest of the code a little bit though, since there can (and usually are) multiple PROGBITS sections.

litebox_syscall_rewriter/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+        trampoline_name.extend_from_slice(".trampoline".as_bytes());
+        trampoline_section.sh_type = sh_type;
+        trampoline_section.sh_flags = sh_flags;
+        trampoline_section.sh_addralign = sh_addralign;
+
+        let text_segment = builder.segments.get_mut(text_segment_id);
+        text_segment.append_section(trampoline_section);
+        (trampoline_section.id(), trampoline_section.sh_addr)
+    };
+    let mut trampoline_data = vec![0x0; 0x8];
```
jaybosamiya-ms: Needs a comment on why this starts with 8 zeros.

jaybosamiya-ms:
The libc that I tried, for quick reference: [libc.zip](https://github.com/user-attachments/files/19375791/libc.zip)

litebox_syscall_rewriter/src/main.rs
wdcui: Fixed.

litebox_syscall_rewriter/src/main.rs
wdcui: Let's do this in another PR.

litebox_syscall_rewriter/Cargo.toml
wdcui: Let's do this in another PR when needed.

litebox_syscall_rewriter/src/main.rs
wdcui: Will address this in another PR.

litebox_syscall_rewriter/src/main.rs
```diff
@@ -0,0 +1,216 @@
...
+        (
+            text_data,
+            text_addr,
+            text_section.sh_type,
+            text_section.sh_flags,
+            text_section.sh_addralign,
+        )
+    };
+
+    // Create a new loadable segment and section for the trampoline
```
CvvT, line 134: Looks like it does not create a new segment. Instead, append the new section to the existing segment.

litebox_syscall_rewriter/src/main.rs
```diff
@@ -0,0 +1,216 @@
...
+        .build()?;
+    let instructions = cs.disasm_all(text_data, text_addr)?;
+
+    for (i, inst) in instructions.iter().enumerate() {
+        if inst.mnemonic().unwrap_or_default() == "syscall" {
+            let mut replace_len = inst.bytes().len();
+            let mut replace_start = inst.address();
+            let prev_idx = i;
+
+            while replace_len < 5 && prev_idx.checked_sub(1).is_some() {
```
CvvT, line 27: `checked_sub` does not do sub and assign like "-=". In fact, `prev_idx` is declared as immutable.

litebox_syscall_rewriter/src/main.rs
wdcui: Good catch! Need to update the comment in a future PR.

litebox_syscall_rewriter/src/main.rs
wdcui: Good catch of this bug! Need to update it in a future PR. Right now it works only if the previous instruction has at least 3 bytes.

PR #49, opened 2025-03-21
Systrap

CvvT:
* Use seccomp filter to implement syscall hooks for Linux user land.
* Move some unit tests to integration tests because cargo test does not forward signals to unit tests.
* implement `sys_mmap`

More details about systrap:
1. https://gvisor.dev/blog/2023/04/28/systrap-release/
2. https://github.com/google/gvisor/blob/master/pkg/sentry/platform/systrap/README.md

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -32,7 +32,6 @@ impl<PunchthroughProvider: litebox::platform::PunchthroughProvider>
     ///
     /// Takes a tun device name (such as `"tun0"` or `"tun99"`) to connect networking.
     /// Registers `syscall_handler` to handle all intercepted syscalls.
-    /// for details.
     ///
     /// # Panics
     ///


litebox_platform_linux_userland/src/platform/mod.rs
wdcui, line 1: Why do we need a platform folder inside the platform crate?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -29,11 +31,19 @@ impl<PunchthroughProvider: litebox::platform::PunchthroughProvider>
     /// Create a new userland-Linux platform for use in `LiteBox`.
     ///
     /// Takes a tun device name (such as `"tun0"` or `"tun99"`) to connect networking.
+    /// Registers `sys_handler` to handle all intercepted syscalls. See `platform::systrap::syscall_dispatcher`
```
wdcui: Nit: shall we change all sys_handler to syscall_handler (and SysHandler to SyscallHandler)?

litebox_platform_linux_userland/src/platform/systrap.rs
```diff
@@ -0,0 +1,177 @@
...
+struct SyscallSiginfo {
+    signo: c_int,
+    errno: c_int,
+    code: c_int,
+    call_addr: *mut libc::c_void,
+    syscall: c_int,
+    arch: c_uint,
+}
+
+type SysHandler = dyn Fn(i64, &[usize]) -> i64 + Send + Sync;
```
wdcui: Nit: SysHandler => SyscallHandler?

litebox_platform_linux_userland/src/platform/systrap.rs
```diff
@@ -0,0 +1,177 @@
...
+    let syscall_args = unsafe { std::slice::from_raw_parts(args, 6) };
+    SYSCALL_HANDLER.get().unwrap()(syscall_number, syscall_args)
+}
+
+extern "C" fn sigsys_handler(sig: c_int, info: *mut libc::siginfo_t, context: *mut libc::c_void) {
+    unsafe {
+        assert!(sig == libc::SIGSYS);
+        let custom_info = &*info.cast::<SyscallSiginfo>();
+        let syscall_number = custom_info.syscall;
+        let addr = custom_info.call_addr;
```
wdcui: What's in this call_addr?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -29,11 +31,19 @@ impl<PunchthroughProvider: litebox::platform::PunchthroughProvider>
...
+    /// for details.
```
jaybosamiya-ms: The `platform::systrap::syscall_dispatcher` is a non-public member of this crate, thus the "see" doesn't quite work here in the docs.  Also, I am not sure what I am supposed to be seeing regarding this code when I look there.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -29,11 +31,19 @@ impl<PunchthroughProvider: litebox::platform::PunchthroughProvider>
...
     ///
     /// # Panics
     ///
     /// Panics if the tun device could not be successfully opened.
-    pub fn new(tun_device_name: &str, punchthrough_provider: PunchthroughProvider) -> Self {
+    pub fn new(
+        tun_device_name: &str,
+        punchthrough_provider: PunchthroughProvider,
+        sys_handler: impl Fn(i64, &[usize]) -> i64 + Send + Sync + 'static,
```
jaybosamiya-ms: Might be good to leave a comment related to this, that the signature would change due to #22 

In particular, we would have stricter types once we have definitions of syscalls with #22 rather than converting everything to `usize`s.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -13,6 +13,8 @@ use litebox::platform::ImmediatelyWokenUp;
 use litebox::platform::UnblockedOrTimedOut;
 use litebox::platform::page_mgmt::MemoryRegionPermissions;
 
+pub mod platform;
```
jaybosamiya-ms: I'm not sure why this is `pub`; nothing within it is made public to the outside world, thus the docs end up looking bizarre:
image

litebox_platform_linux_userland/src/platform/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! Implementation of syscall interception for Linux userland.
```
jaybosamiya-ms: This module should be called `syscall_interception`. Our meaning of platform is quite clear (e.g., `LinuxKernel` is a platform, `LinuxUserland` is a platform, ...), so it is surprising to have a module under it called platform that is dealing with one specific component of platforms.

litebox_platform_linux_userland/src/platform/mod.rs
```diff
@@ -0,0 +1,5 @@
...
+
+mod systrap;
+
+pub(crate) use systrap::*;
```
jaybosamiya-ms: Generally `pub(..) use foo::*` declarations are usually overly permissive, and lead to longer term maintainability issues. If a specific list of things is mentioned it could be fine, but I think the main reason you have this is to set up some sort of multiplexing around different forms of syscall interception, correct? If so, we should actually have separate multiplexing for this, rather than doing so directly here. We can do that multiplexing in a separate PR, but for now, I think removing this line is better.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+    if addr & !PAGE_MASK != 0 {
+        return Err(Errno::EINVAL);
+    }
+    if len == 0 {
+        return Err(Errno::EINVAL);
+    }
+
+    let prot = ProtFlags::from_bits_truncate(prot);
+    let flags = MapFlags::from_bits(flags).expect("Unsupported flags");
```
jaybosamiya-ms, line 148: Similar to `flags`, I think we should do the `.expect` variant for `prot`, so that we are not silently throwing away things

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: Also, if we have #42, this argument would be unused, yes? We might want to briefly discuss how to set up the interface for this in that case.

litebox_platform_linux_userland/src/platform/systrap.rs
```diff
@@ -0,0 +1,177 @@
...
+        signal::sigaction(Signal::SIGSYS, &sig_action).expect("Failed to register SIGSYS handler");
+    }
+}
+
+#[cfg(not(test))]
+fn register_seccomp_filter() {
+    use seccompiler::{BpfProgram, SeccompAction, SeccompFilter, SeccompRule};
+
+    let rules = vec![];
+    let rule_map: std::collections::BTreeMap<i64, Vec<SeccompRule>> = rules.into_iter().collect();
```
jaybosamiya-ms:
```suggestion
    let rules = vec![];
    let rule_map = std::collections::BTreeMap::<i64, Vec<SeccompRule>>::new();
```
You can go a step further by simply saying:
```suggestion
    let rule_map = Default::default();
```

litebox_platform_linux_userland/src/lib.rs
wdcui: I didn't see where the syscall handler is set.

litebox_platform_linux_userland/src/platform/mod.rs
wdcui: I had a similar comment.

litebox_platform_linux_userland/src/platform/mod.rs
CvvT: Got it. Removed

litebox_platform_linux_userland/src/platform/systrap.rs
CvvT: Sure

litebox_platform_linux_userland/src/platform/systrap.rs
CvvT: The address of the instruction that calls syscall

litebox_platform_linux_userland/src/platform/systrap.rs
CvvT: Changed.

litebox_platform_linux_userland/src/platform/systrap.rs
```diff
@@ -0,0 +1,177 @@
...
+extern "C" fn sigsys_handler(sig: c_int, info: *mut libc::siginfo_t, context: *mut libc::c_void) {
+    unsafe {
+        assert!(sig == libc::SIGSYS);
+        let custom_info = &*info.cast::<SyscallSiginfo>();
+        let syscall_number = custom_info.syscall;
+        let addr = custom_info.call_addr;
+        eprintln!("Caught SIGSYS: syscall={syscall_number} addr={addr:?}");
+
+        // Ensure the address is valid
+        if addr.is_null() {
```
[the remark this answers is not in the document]
CvvT: Removed all print and use abort instead of panic

litebox_platform_linux_userland/src/lib.rs
CvvT: Changed to private.

litebox_platform_linux_userland/src/lib.rs
CvvT: I planned to have some doc to explain the arguments of this handler. Then I realized it is likely to change in the future. It might be easier to understand it by looking at the function that calls this handler.

litebox_platform_linux_userland/src/lib.rs
CvvT: It is set by the function `init_sys_intercept`

CvvT: I left a TODO. With #42, we may still want to keep it to handle corner cases our binary rewriting tool cannot handle.

litebox_shim_linux/src/lib.rs
```diff
@@ -213,3 +213,7 @@ pub unsafe extern "C" fn open(pathname: ConstPtr<i8>, flags: u32, mode: u32) ->
 pub extern "C" fn close(fd: i32) -> i32 {
     syscalls::file::sys_close(fd).map_or_else(Errno::as_neg, |()| 0)
 }
+
+pub fn syscall_entry(sysno: i64, args: &[usize]) -> i64 {
+    unimplemented!()
```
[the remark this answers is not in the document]
CvvT: maybe `todo` is better?

litebox_platform_linux_userland/src/platform/mod.rs
CvvT: Changed

litebox_shim_linux/src/syscalls/mm.rs
CvvT: `ProtFlags` only has 4 flags: https://www.man7.org/linux/man-pages/man2/mmap.2.html so it does not throwing away anything. Other bits are not used.

litebox_platform_linux_userland/src/lib.rs
CvvT: Changed.

litebox_platform_linux_userland/src/platform/mod.rs
CvvT: Changed to `syscall_intercept`

litebox_platform_linux_userland/src/lib.rs
CvvT: I believe we still need it because even if we could rewrite all the syscall instructions from the target binaries, litebox itself also contains a lot of syscalls that could be used by the malicious workload through ROP. With systrap, we could gurantee the workload can never call syscalls we don't allow.

litebox_shim_linux/src/lib.rs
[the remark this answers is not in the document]
jaybosamiya-ms: `unimplemented!()` or `todo!()` here is fine. Both `todo!()` and `unimplemented!()` cause a panic. The Rust [doc for `unimplemented`](https://doc.rust-lang.org/std/macro.unimplemented.html) states:
> The difference between unimplemented! and [todo!](https://doc.rust-lang.org/std/macro.todo.html) is that while todo! conveys an intent of implementing the functionality later and the message is “not yet implemented”, unimplemented! makes no such claims. Its message is “not implemented”.

Personally, I use `unimplemented!()` to mark "will implement at some undefined point in the future", while I use `todo!()` to mark "will implement at some point soon"; so yeah, definitely just a stylistic difference, and nothing crucial.

Basically, ignore Copilot here :)

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: That's a fair point, agreed. The TODO is helpful, thanks!

litebox_shim_linux/src/syscalls/mm.rs
jaybosamiya-ms: Fair point, I was probably being overly cautious (i.e., "what if the user passes a flag outside the allowed bits"), but at least from a quick glance through https://elixir.bootlin.com/linux/v5.19.17/source/mm/mmap.c#L1416 it does appear that Linux throws away any bits that are not part of the expected set, so truncate is correct here.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -29,11 +31,20 @@ impl<PunchthroughProvider: litebox::platform::PunchthroughProvider>
     /// Create a new userland-Linux platform for use in `LiteBox`.
     ///
     /// Takes a tun device name (such as `"tun0"` or `"tun99"`) to connect networking.
+    /// Registers `syscall_handler` to handle all intercepted syscalls.
+    /// for details.
     ///
```
jaybosamiya-ms: Accidentally left in a line here
```suggestion
    /// Takes a tun device name (such as `"tun0"` or `"tun99"`) to connect networking.
    /// Registers `syscall_handler` to handle all intercepted syscalls.
    ///
```

PR #50, opened 2025-03-21
Add file status to all the LiteBox file systems

jaybosamiya-ms:
This PR adds `file_status` and `fd_file_status`, roughly analogous to `stat` and `fstat`. It currently restricts itself to providing relatively little information (type, permissions, size), with the intention of future PRs making use of this information, and possibly even extending it with inodes/nlinks/...

### DIFF
--- a/litebox/src/fs/tar_ro.rs
+++ b/litebox/src/fs/tar_ro.rs
@@ -276,10 +276,12 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
             Some(p) if p.filename().as_str().unwrap() != path => Ok(super::FileStatus {
                 file_type: super::FileType::Directory,
                 mode: DEFAULT_DIR_MODE,
+                size: super::DEFAULT_DIRECTORY_SIZE,
             }),
             Some(p) => Ok(super::FileStatus {
                 file_type: super::FileType::RegularFile,
                 mode: mode_of_modeflags(p.posix_header().mode.to_flags().unwrap()),
+                size: p.size(),
             }),
         }
     }
@@ -289,22 +291,18 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
         fd: &crate::fd::FileFd,
     ) -> Result<super::FileStatus, super::errors::FileStatusError> {
         match self.descriptors.read().get(fd) {
-            Descriptor::File { idx, .. } => Ok(super::FileStatus {
-                file_type: super::FileType::RegularFile,
-                mode: mode_of_modeflags(
-                    self.tar_data
-                        .entries()
-                        .nth(*idx)
-                        .unwrap()
-                        .posix_header()
-                        .mode
-                        .to_flags()
-                        .unwrap(),
-                ),
-            }),
+            Descriptor::File { idx, .. } => {
+                let entry = self.tar_data.entries().nth(*idx).unwrap();
+                Ok(super::FileStatus {
+                    file_type: super::FileType::RegularFile,
+                    mode: mode_of_modeflags(entry.posix_header().mode.to_flags().unwrap()),
+                    size: entry.size(),
+                })
+            }
             Descriptor::Dir { .. } => Ok(super::FileStatus {
                 file_type: super::FileType::Directory,
                 mode: DEFAULT_DIR_MODE,
+                size: super::DEFAULT_DIRECTORY_SIZE,
             }),
         }
     }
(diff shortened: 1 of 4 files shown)


litebox/src/fs/tar_ro.rs
```diff
@@ -257,6 +257,73 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
+        let path = &path[1..];
+        let entry = self
+            .tar_data
+            .entries()
+            .find(|entry| match entry.filename().as_str() {
+                Ok(p) => p == path || contains_dir(p, path),
+                Err(_) => false,
+            });
+        match entry {
+            None => Err(PathError::NoSuchFileOrDirectory)?,
```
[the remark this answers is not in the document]
jaybosamiya-ms: We're consistently using this pattern to upcast errors based on their `#[from]`, so this comment doesn't apply (and indeed would be incorrect to apply). Ignoring this.

jaybosamiya-ms:
Heh, technically this _does_ change the API, but it is a safe change. I should probably update the text for the 🤖 to stop saying "unchanged" but instead say "No breaking changes detected" (EDIT: Pushed df2224a30f1122d5892512bbf2de0c67e71136dc to improve clarity)

litebox/src/fs/tar_ro.rs
```diff
@@ -257,6 +257,73 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
+            Some(p) if p.filename().as_str().unwrap() != path => Ok(super::FileStatus {
```
wdcui, line 276: I don't understand this line. Why is it a dir when the paths are not equal?

litebox/src/fs/tar_ro.rs
```diff
@@ -257,6 +257,73 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
         Err(RmdirError::ReadOnlyFileSystem)
     }
+
+    fn file_status(
+        &self,
+        path: impl crate::path::Arg,
+    ) -> Result<super::FileStatus, super::errors::FileStatusError> {
+        let path = self.absolute_path(path)?;
+        let path = &path[1..];
+        let entry = self
```
wdcui, line 267: So right now we need to check every file in the tar to find the file we are looking for? I understand we can build a hashmap to speed it up. Just to confirm my understanding.

litebox/src/fs/tar_ro.rs
```diff
@@ -257,6 +257,73 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
+            .entries()
+            .find(|entry| match entry.filename().as_str() {
+                Ok(p) => p == path || contains_dir(p, path),
+                Err(_) => false,
+            });
+        match entry {
+            None => Err(PathError::NoSuchFileOrDirectory)?,
+            Some(p) if p.filename().as_str().unwrap() != path => Ok(super::FileStatus {
+                file_type: super::FileType::Directory,
+                mode: DEFAULT_DIR_MODE,
```
CvvT, line 278: Why dir has the same mode?

litebox/src/fs/tar_ro.rs
jaybosamiya-ms: This is a pattern we use regularly in `tar_ro`, e.g.:
https://github.com/MSRSSP/litebox/blob/df2224a30f1122d5892512bbf2de0c67e71136dc/litebox/src/fs/tar_ro.rs#L134-L145
or
https://github.com/MSRSSP/litebox/blob/df2224a30f1122d5892512bbf2de0c67e71136dc/litebox/src/fs/tar_ro.rs#L242-L246

Essentially, the `tar_no_std` crate only gives us files, and does not give us the directories in the `.tar` file, so we need to infer them. We do this by checking for prefixes.

litebox/src/fs/tar_ro.rs
jaybosamiya-ms: Yes, exactly. We have a TODO in the code to speed this up:
https://github.com/MSRSSP/litebox/blob/df2224a30f1122d5892512bbf2de0c67e71136dc/litebox/src/fs/tar_ro.rs#L118-L120

litebox/src/fs/tar_ro.rs
jaybosamiya-ms: `tar_no_std` (the crate we depend upon) does not give us any directories, and thus we must infer them from the files that exist within it. Thus, we can't really have different modes for different directories, and must pick a single default for each of them.

wdcui:
Please feel free to merge it (I hope you are not waiting for my approval).

jaybosamiya-ms:
It isn't a blocker to anything currently being worked on (thus not a rush to merge) but thanks :)

PR #54, opened 2025-03-25
Systrap allow list

CvvT:
This PR addresses some of the TODOs from previous PR #49 

* Switch from deny list to allow list for seccomp filter
* Handle some syscalls in the shim layer (see `syscall_entry`). Note this is not complete yet, there are more syscalls needed to be implemented, e.g., `readlink /proc/self/exe`, `getpid`, etc.

### DIFF
--- a/litebox_shim_linux/src/loader/elf.rs
+++ b/litebox_shim_linux/src/loader/elf.rs
@@ -95,7 +95,7 @@ impl ElfLoaderMmap {
             addr.unwrap_or(0),
             len,
             litebox_common_linux::ProtFlags::from_bits_truncate(prot.bits()),
-            litebox_common_linux::MapFlags::from_bits_truncate(flags.bits()),
+            litebox_common_linux::MapFlags::from_bits(flags.bits()).expect("unsupported flags"),
             fd,
             offset,
         ) {
@@ -114,9 +114,10 @@ impl ElfLoaderMmap {
             addr.unwrap_or(0),
             len,
             litebox_common_linux::ProtFlags::from_bits_truncate(prot.bits()),
-            litebox_common_linux::MapFlags::from_bits_truncate(
+            litebox_common_linux::MapFlags::from_bits(
                 flags.bits() | MapFlags::MAP_ANONYMOUS.bits(),
-            ),
+            )
+            .expect("unsupported flags"),
             -1,
             0,
         ) {


litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
```diff
@@ -1,5 +1,7 @@
 //! The systrap platform relies on seccomp’s `SECCOMP_RET_TRAP` feature to intercept system calls.
 
+use litebox::platform::trivial_providers::{TransparentConstPtr, TransparentMutPtr};
```
wdcui, line 3: Is this a temporary choice to use the trivial_providers here?

litebox_shim_linux/src/lib.rs
```diff
@@ -214,6 +214,59 @@ pub extern "C" fn close(fd: i32) -> i32 {
...
+
+/// Entry point for the syscall handler
+pub fn syscall_entry(request: SyscallRequest<Platform>) -> isize {
+    match request {
+        SyscallRequest::Read(fd, buf, count) => {
+            let Ok(count) = isize::try_from(count) else {
+                return Errno::EINVAL.as_neg() as isize;
+            };
+            buf.mutate_subslice_with(..count, |user_buf| {
+                // TODO: user kernel buffer to avoid page faults
```
wdcui: NIT: "user kernel buffer" => "use kernel buffer"?

litebox_shim_linux/src/lib.rs
```diff
@@ -214,6 +214,59 @@ pub extern "C" fn close(fd: i32) -> i32 {
...
+                    #[allow(clippy::cast_possible_wrap)]
+                    |size| size as isize,
+                )
+            })
+            .unwrap_or(Errno::EFAULT.as_neg() as isize)
+        }
+        SyscallRequest::Mmap(addr, len, prot, flags, fd, offset) => {
+            syscalls::mm::sys_mmap(addr, len, prot, flags, fd, offset).map_or_else(
+                |e| e.as_neg() as isize,
+                #[allow(clippy::cast_possible_wrap)]
```
wdcui: Is this safe? Would a better solution be to do this cast with try_from()?

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
jaybosamiya-ms: For userland linux platform, using the transparent pointers _is_ the right move. It is not a temporary choice. The module is named `trivial_providers` mostly because they have a trivial implementation, not to imply "do not use them".

litebox_shim_linux/src/lib.rs

litebox_shim_linux/src/lib.rs
CvvT: If use `try_from` instead, it assumes a user space address can always fit into a `isize`, which I guess is okay for Linux. Changed.

litebox_common_linux/src/lib.rs
```diff
@@ -5,3 +5,50 @@
...
+        /// Changes are private
+        const MAP_PRIVATE = 1 << 1;
+        /// Interpret addr exactly
+        const MAP_FIXED = 1 << 4;
+        /// don't use a file
+        const MAP_ANONYMOUS = 1 << 5;
+        /// This flag is ignored.
+        const MAP_DENYWRITE = 1 << 11;
+    }
+}
```
jaybosamiya-ms: You might wish to add something like https://github.com/MSRSSP/litebox/blob/eaefc448a490d498c993deb6b2140fcbc74236af/litebox/src/fs/mod.rs#L166-L167
to this bitflags. In that case, you can drop `MAP_DENYWRITE` (and also don't need to mention `MAP_EXECUTABLE` and such which are ignored by Linux. The `non_exhaustive` doesn't quite work here for what you mean since it is not an enum.

litebox_common_linux/src/lib.rs
```diff
@@ -5,3 +5,50 @@
 pub mod errno;
 
 // TODO(jayb): Should errno::Errno be publicly re-exported?
+
+bitflags::bitflags! {
+    /// Desired memory protection of a memory mapping.
+    #[non_exhaustive]
+    #[derive(PartialEq, Debug)]
+    pub struct ProtFlags: core::ffi::c_int {
```
jaybosamiya-ms: Similar to `MapFlags`

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
```diff
@@ -81,15 +85,60 @@ sigsys_callback:
...
 unsafe extern "C" {
-    fn sigsys_callback() -> i64;
+    fn sigsys_callback() -> isize;
 }
 
+#[allow(
+    clippy::cast_possible_truncation,
+    clippy::cast_possible_wrap,
+    reason = "reinterprete arguments directly"
+)]
```
jaybosamiya-ms: I personally think these particular lints should not be disabled since it allows easily losing info or leading to bad info that might be hard to debug. Alternatively, if you can update the `reason` to be "TODO: this will be cleaned up later", that is fine too.

litebox_shim_linux/src/lib.rs
```diff
@@ -20,10 +20,10 @@ use once_cell::race::OnceBox;
 
 use litebox::{
     mm::{PageManager, linux::PAGE_SIZE},
-    platform::RawConstPointer as _,
+    platform::{RawConstPointer as _, RawMutPointer},
```
jaybosamiya-ms: Nit:
```suggestion
    platform::{RawConstPointer as _, RawMutPointer as _},
```

Placing the `as _` allows the trait to be in-scope (to allow it being used) but also says "this will not be referenced by name" (i.e., there won't be an `impl` showing up and such). Essentially is a slightly stricter import.

litebox_common_linux/src/lib.rs
```diff
@@ -5,3 +5,50 @@
...
+        /// Interpret addr exactly
+        const MAP_FIXED = 1 << 4;
+        /// don't use a file
+        const MAP_ANONYMOUS = 1 << 5;
+        /// This flag is ignored.
+        const MAP_DENYWRITE = 1 << 11;
+    }
+}
+
+pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
```
jaybosamiya-ms, line 84: I think this should be marked non_exhaustive, and also have a doc comment.

Additionally, I would appreciate it if you could add a `// TODO: change to struct-like enum rather than tuple-like enum`, that'd be great. It doesn't need to actually change in this PR, but essentially, we need a reminder to do the nicer version.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -39,7 +39,10 @@ impl<PunchthroughProvider: litebox::platform::PunchthroughProvider>
     pub fn new(
         tun_device_name: &str,
         punchthrough_provider: PunchthroughProvider,
-        syscall_handler: impl Fn(i64, &[usize]) -> i64 + Send + Sync + 'static,
+        syscall_handler: impl Fn(litebox_common_linux::SyscallRequest<LinuxUserland>) -> isize
```
jaybosamiya-ms: I'm not quite sure why things changed from i64 to isize everywhere for the return types of things

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -25,6 +25,19 @@ pub(crate) fn sys_open(path: impl path::Arg, flags: OFlags, mode: Mode) -> Resul
         .map_err(Errno::from)
 }
 
+const AT_FDCWD: i32 = -100;
```
jaybosamiya-ms: This probably needs to be updated to have a comment to it saying that this constant is taken from libc, otherwise it seems like a magic number (which I know it _is_ a magic number, but still).

Also, it should likely be the same visibility as `sys_openat`.

litebox_common_linux/src/lib.rs

litebox_common_linux/src/lib.rs
CvvT: Got it, thanks! I added all other bits.

litebox_common_linux/src/lib.rs
CvvT: struct-like enum is clearer. Fixed.

litebox_platform_linux_userland/src/lib.rs
CvvT: In C, the return type is `long`, `isize` seems to better match it. Also, since we only support 64 bits, it is same as i64.

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
CvvT: Sure, added the lint info to individual statement. The only concern I have now is this function is doubled in LOC and less readable, but it might be fine as I don't expect this function to do anything but converting types.

litebox_shim_linux/src/lib.rs

litebox_shim_linux/src/syscalls/file.rs

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
jaybosamiya-ms: Oops, sorry, I was not recommending copying it over to every individual statement (I definitely agree with your concern there). If you don't mind, I can push a new commit directly to this branch to clean it up? (EDIT: I can just make a new PR pointing at this branch, will do that shortly)

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: I see, in that case wouldn't `i64` be better? Rust maps `std::ffi::c_long` to `i64` (https://doc.rust-lang.org/std/ffi/type.c_long.html)

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
jaybosamiya-ms: #59

litebox_platform_linux_userland/src/lib.rs
CvvT: I see. It doesn't explain why it chose `i64` over `isize`. I don't have preference, so I will just follow it.

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
CvvT: Cool, thanks! I assume all the code would be optimized out by the compiler.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -143,9 +112,23 @@ pub(crate) fn sys_mmap(
...
+            | MapFlags::MAP_POPULATE
+            | MapFlags::MAP_NONBLOCK
+            | MapFlags::MAP_SYNC
+            | MapFlags::MAP_HUGETLB
+            | MapFlags::MAP_HUGE_2MB
+            | MapFlags::MAP_HUGE_1GB
+            | MapFlags::MAP_FIXED_NOREPLACE,
+    ) {
+        todo!("Unsupported flags {:?}", flags);
+    }
```
CvvT, line 130: Now that MapFlags includes more flags we don't support, added todo.

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
jaybosamiya-ms: Yep, it will all be optimized out; it is mostly just a way for us to ensure we are not using `as` (which can more hide issues when things change and make debugging harder).

jaybosamiya-ms: Running on godbolt: https://rust.godbolt.org/z/xjM7nxs31

```rs
pub fn foo(x: i64, y: u64) -> (u8, i8) {
    (x.reinterpret_as_unsigned().truncate(), y.reinterpret_as_signed().truncate())
}
```
compiles to
```asm
example::foo::h3defd21e5aa8b36d:
        mov     rdx, rsi
        mov     rax, rdi
        ret
```
(Note: here it is doing a full 64-bit move, but that's mostly an unrelated behavior of LLVM of keeping that extra data around unless there is a good reason to zero it out; only the lower order bits are interpreted anyways)

wdcui:
Please feel free to merge it once Jay approves it.

jaybosamiya-ms:
The conflict is due to https://github.com/MSRSSP/litebox/pull/63; essentially you'll be able to just remove the usage of `new_for_test` by using `new(None, ...` instead

CvvT:
@jaybosamiya-ms Looks like `objdump` running on my machine generates slightly different assembly code. It somehow appends some spaces to the end of certain instructions like ret/syscall/endbr64 (those do not have operands I guess?)

PR #55, opened 2025-03-25
Refactor syscall rewriter to library-ify it

### LINKED ISSUE #45: Handle all code sections in syscall rewriter

### LINKED ISSUE #48: Syscall rewriter lib+CLI
> Automatically picking foo -> foo.hooked is not ideal for the CLI. I'm happy to write a quick separate PR once this is merged to give a better interface.
>
> Additionally, we may wish to library-ify this, for use with convenience wrappers (that do a compile+run thing). I'm happy to do that in a separate PR too.

_Originally posted by @jaybosamiya-ms in https://github.com/MSRSSP/litebox/pull/42#discussion_r2006452038_

### LINKED ISSUE #52: Fix the comment in syscall rewriter.
Good catch! Need to update the comment in a future PR.

_Originally posted by @wdcui in https://github.com/MSRSSP/litebox/pull/42#discussion_r2008826104_

### LINKED ISSUE #53: Fix prev_idx in syscall rewriter.
Good catch of this bug! Need to update it in a future PR. Right now it works only if the previous instruction has at least 3 bytes.

_Originally posted by @wdcui in https://github.com/MSRSSP/litebox/pull/42#discussion_r2008826280_

jaybosamiya-ms:
This PR does the following things:
- sets up syscall rewriter as a library, allowing for more convenient writing of integration tests, as well as setting up a nicer step for the userland-wrapper crate that I'll be working on soon
- updates the `.text` manipulation to instead do it on all executable segments
- uses `thiserror` to set up more precise errors
- nicer CLI that supports `-o` and such
- sets up a version numbering (as well as checks using it) to make sure we can test if a binary has been rewritten already (and if we update the rewriter to change some semantic behavior, then we can bump the version number to better detect rewriting)
- sets up a snapshot test, to make sure we are not breaking functionality without noticing it
- when adding tests, I found what I think is a (imho) bug in `object` that we need to work around. Ideally, we'd upstream a fix, but I didn't find an immediate/clean fix for it, so for now, we just have a workaround here.
- minor: switches us to the "diet" version of capstone

Helping GitHub connect the relevant issues: closes #52, closes #53, closes #48, closes #45

### DIFF
--- a/litebox_syscall_rewriter/src/main.rs
+++ b/litebox_syscall_rewriter/src/main.rs
@@ -25,7 +25,7 @@ fn main() -> anyhow::Result<()> {
     let mut input_binary = std::fs::File::open(&cli_args.input_binary)?;
     let mut input_binary_bytes = vec![];
     input_binary.read_to_end(&mut input_binary_bytes)?;
-    let res = litebox_syscall_rewriter::hook_syscalls_in_elf(
+    let output_binary = litebox_syscall_rewriter::hook_syscalls_in_elf(
         &input_binary_bytes,
         cli_args.trampoline_addr,
     )?;
@@ -44,6 +44,6 @@ fn main() -> anyhow::Result<()> {
     file.set_permissions(std::fs::Permissions::from_mode(
         input_binary.metadata()?.mode(),
     ))?;
-    file.write_all(&res)?;
+    file.write_all(&output_binary)?;
     Ok(())
 }


litebox_syscall_rewriter/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+    #[error("failed to generate object file: {0}")]
+    GenerateObjFileError(String),
+    #[error("unsupported executable")]
+    UnsupportedObjectFile,
+    #[error("executable is already hooked with trampoline")]
+    AlreadyHooked,
+    #[error("no .text section found")]
+    NoTextSectionFound,
+    #[error("no syscall instructions found")]
+    NoSyscallInstructionsFound,
```
wdcui, line 31: Why would this be an error? It's legitimate for some files to not have syscalls, right?

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        match hook_syscalls_in_section(
+            s.sh_addr,
+            data.to_mut(),
+            trampoline_base_addr,
+            &mut trampoline_data,
+        ) {
+            Ok(()) => {
+                syscall_insns_found = true;
+            }
+            Err(Error::NoSyscallInstructionsFound) => {}
```
wdcui, line 148: Some sections may not have syscalls. Why should this be an error?

litebox_syscall_rewriter/src/main.rs
```diff
@@ -1,216 +1,49 @@
...
-            text_addr,
-            text_section.sh_type,
-            text_section.sh_flags,
-            text_section.sh_addralign,
+fn main() -> anyhow::Result<()> {
+    let cli_args = CliArgs::parse();
+    let mut input_binary = std::fs::File::open(&cli_args.input_binary)?;
+    let mut input_binary_bytes = vec![];
+    input_binary.read_to_end(&mut input_binary_bytes)?;
+    let res = litebox_syscall_rewriter::hook_syscalls_in_elf(
```
wdcui: NIT: change res to something like output_binary?

wdcui:
Thank you @jaybosamiya-ms for cleaning up my code and fixing the issues I left!

litebox_syscall_rewriter/src/lib.rs
jaybosamiya-ms: Yes, agreed it is legitimate for some files to not have syscalls, but this allows a user of the crate to decide if they want that to be an error or silence this. I would prefer it is not silent _at this crate_ but instead a decision made by user of this crate.

litebox_syscall_rewriter/src/lib.rs
jaybosamiya-ms: This is not an error, this explicitly _prevents_ this from being thrown as an error, and waits till end of whole binary to decide if no syscalls are found.

PR #56, opened 2025-03-25
save and restore fs_base

CvvT:
Both litebox and loaded target binary set up their own fs register for thread local storage. Thus, whenever we switch between these two (i.e., in our `syscall_dispatcher`), we need to save and restore the current fs base (which introduces two syscall overhead). We could avoid it if know for sure our syscall handler does not use any per-thread data (but I haven't figure out a good way to detect it).

### DIFF
--- a/litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
+++ b/litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
@@ -340,6 +340,8 @@ fn register_seccomp_filter() {
 }
 
 /// Save the current thread's fs base to thread local storage.
+///
+/// TODO: make FS_BASE per-thread.
 fn save_current_fs_base() {
     FS_BASE.store(
         GET_FS_BASE.get().unwrap()(),


jaybosamiya-ms:
As discussed IRL, we probably want to gate this under a `cfg(feature)` flag, and attempt to make sure that we don't have any thread local usage in LiteBox itself. This would require making sure we don't use `println` and such, but I _think_ this is definitely possible to maintain, given that 90+% of our codebase is `no_std` already.

Another thought that came to mind is that we may be able to do a rewrite of all the `mov %fs:0x0,%rax` instructions that show up in the guest binary (as part of the syscall rewriter or loader)---we can set up our own thread control block, change out the 0x0 (there are 4 bytes used for the constant `0x0`) so that we can have a separate space (essentially "by convention") for the guest program, that is different from the space for LiteBox.  I don't think this is ideal (esp since it assumes that compilers are generating only `mov %fs:0x0,%rax`), and we'd need to make sure that the constant we set up is actually reasonable, but at the same time, we _should_ be able to alert whenever there is a `%fs` that we don't handle, so I _think_ this is actually statically manageable. I don't yet know how complicated setting up the block would be ([link to reference doc that might help](https://www.akkadia.org/drepper/tls.pdf)), but yeah, it is a possibility if we want to support thread locals in both.

wdcui:
Please feel free to merge it once Jay approves it.

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
```diff
@@ -25,7 +31,7 @@ global_asm!(
     .globl  sigsys_callback
     .type   sigsys_callback,@function
 sigsys_callback:
-    /* TODO: save floating point registers if needed */
+    /* TODO: save float and vector registers (xsave or fxsave) */
```
CvvT: We could potentially disable SIMD to avoid the overhead: https://os.phil-opp.com/disable-simd/

CvvT:
As we discussed, using `rdfsbase` and `wrfsbase` is much more efficient and they are enabled by default (on my machine).

CvvT:
oops, I tried to use thread_local! to store fs register, which itself relies on fs register. Need to figure out how to fix it.

jaybosamiya-ms:
Briefly noting an idea we discussed IRL: since our main usage for storing fs register value is to support a swap, we can maintain a (global, rwlocked) hashmap that tells us the swaps. We would need to set up a state-machine-like thing to leave values "unlinked" until we find the link (during libc initialization), but generally speaking, the idea seems feasible.

@CvvT is also exploring some pthread internals to see if they might be helpful as an alternative, just in case that ends up simpler, but yeah, we have ways to solve this :)

CvvT:
> Briefly noting an idea we discussed IRL: since our main usage for storing fs register value is to support a swap, we can maintain a (global, rwlocked) hashmap that tells us the swaps. We would need to set up a state-machine-like thing to leave values "unlinked" until we find the link (during libc initialization), but generally speaking, the idea seems feasible.
> 
> @CvvT is also exploring some pthread internals to see if they might be helpful as an alternative, just in case that ends up simpler, but yeah, we have ways to solve this :)

Thanks @jaybosamiya-ms for the discussion. Will do a separate PR to address it.

PR #65, opened 2025-03-27
fix objdump

CvvT:
The `snapshot_test_hello_world` test was broken on my machine because of a different version of objdump (which appends some spaces to instructions with no operands). This PR fixes it.


CvvT:
> Thanks, looks reasonable to me. I don't quite see why the regex itself is necessary, it would be perfectly reasonable to just trim all ending whitespace always, right? Either way is fine fwiw.

You're right. I use regex because there are also lines like "476e48:	00 00 ", but I guess I just need to update the snapshot.

CvvT:
I updated the snapshot to remove all spaces at the end of each line.

jaybosamiya-ms:
Sounds good, feel free to squash-merge, thanks!

PR #67, opened 2025-03-31
Implement file related syscall

CvvT:
Remove all file related syscalls from the seccomp filter (except for `write` as we still forward write to STDOUT/STDERR to the host) and add our own implementation. The implementation of these syscalls is not complete yet. I left a lot of TODOs. It just meant to make our `hello world` program run. Once we have enough support from our file system, I will revisit them.

The syscalls I added in this PR include:
* `write/pwrite64`: for regular file only
* `fstat/newfstatat`: only for regular file only with same fake values
* `readv/writev`: for regular file only with no guarantee of atomic read/write
* `access`: incomplete check as we don't have `uid` 
* `getcwd`: always returns "/"
* `readlink/readlinkat`: always return ENOSYS as we don't support symbolic links

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -114,8 +114,6 @@ bitflags::bitflags! {
         /// Used with `faccessat`, the checks for accessibility are performed using the
         /// effective user and group IDs instead of the real user and group ID
         const AT_EACCESS = 0x200;
-        /// Remove directory instead of unlinking file.
-        const AT_REMOVEDIR = 0x200;
         /// Do not follow symbolic links.
         const AT_SYMLINK_NOFOLLOW = 0x100;
 
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -101,16 +101,10 @@ impl Descriptors {
         }
     }
     fn remove(&mut self, fd: u32) -> Option<Descriptor> {
-        if fd >= (2 << 30) {
-            return None;
-        }
         let fd = fd as usize;
         self.descriptors.get_mut(fd)?.take()
     }
     fn remove_file(&mut self, fd: u32) -> Option<litebox::fd::FileFd> {
-        if fd >= (2 << 30) {
-            return None;
-        }
         let fd = fd as usize;
         if let Some(Descriptor::File(file_fd)) = self
             .descriptors
@@ -123,9 +117,6 @@ impl Descriptors {
         }
     }
     fn remove_socket(&mut self, fd: u32) -> Option<litebox::fd::SocketFd> {
-        if fd >= (2 << 30) {
-            return None;
-        }
         let fd = fd as usize;
         if let Some(Descriptor::Socket(socket_fd)) = self
             .descriptors
@@ -138,42 +129,27 @@ impl Descriptors {
         }
     }
     fn get_fd(&self, fd: u32) -> Option<&Descriptor> {
-        if fd >= (2 << 30) {
-            return None;
-        }
         self.descriptors.get(fd as usize)?.as_ref()
     }
     fn get_file_fd(&self, fd: u32) -> Option<&litebox::fd::FileFd> {
-        if fd >= (2 << 30) {
-            return None;
-        }
         match self.descriptors.get(fd as usize)?.as_ref()? {
             Descriptor::File(file_fd) => Some(file_fd),
             Descriptor::Socket(_) => None,
         }
     }
     fn get_file_fd_mut(&mut self, fd: u32) -> Option<&mut litebox::fd::FileFd> {
-        if fd >= (2 << 30) {
-            return None;
-        }
         match self.descriptors.get_mut(fd as usize)?.as_mut()? {
             Descriptor::File(file_fd) => Some(file_fd),
             Descriptor::Socket(_) => None,
         }
     }
     fn get_socket_fd(&self, fd: u32) -> Option<&litebox::fd::SocketFd> {
-        if fd >= (2 << 30) {
-            return None;
-        }
         match self.descriptors.get(fd as usize)?.as_ref()? {
             Descriptor::File(_) => None,
             Descriptor::Socket(socket_fd) => Some(socket_fd),
         }
     }
     fn get_socket_fd_mut(&mut self, fd: u32) -> Option<&mut litebox::fd::SocketFd> {
-        if fd >= (2 << 30) {
-            return None;
-        }
         match self.descriptors.get_mut(fd as usize)?.as_mut()? {
             Descriptor::File(_) => None,
             Descriptor::Socket(socket_fd) => Some(socket_fd),


litebox_common_linux/src/lib.rs
```diff
@@ -79,6 +81,177 @@ bitflags::bitflags! {
...
+        /// Test for existence of file.
+        const F_OK = 0;
+        /// Test for read permission.
+        const R_OK = 4;
+        /// Test for write permission.
+        const W_OK = 2;
+        /// Test for execute (search) permission.
+        const X_OK = 1;
+    }
+}
```
jaybosamiya-ms: Generally bitflags that come from the user likely need the
```rs
/// <https://docs.rs/bitflags/*/bitflags/#externally-defined-flags>
const _ = !0;
```
part to them

litebox_common_linux/src/lib.rs
```diff
@@ -79,6 +81,177 @@ bitflags::bitflags! {
...
+        /// Type of synchronisation required from statx(), used to control what sort of
+        /// synchronization the kernel will do when querying a file on a remote filesystem
+        const AT_STATX_SYNC_TYPE = 0x6000;
+        /// Do whatever stat() does
+        const AT_STATX_SYNC_AS_STAT = 0x0;
+        /// Force the attributes to be sync'd with the server
+        const AT_STATX_FORCE_SYNC = 0x2000;
+        /// Don't sync attributes with the server
+        const AT_STATX_DONT_SYNC = 0x4000;
+    }
```
jaybosamiya-ms: Similarly, here too

litebox_common_linux/src/lib.rs
```diff
@@ -79,6 +81,177 @@ bitflags::bitflags! {
...
+        match value {
+            litebox::fs::FileType::RegularFile => InodeType::File,
+            litebox::fs::FileType::Directory => InodeType::Dir,
+            _ => unimplemented!(),
+        }
+    }
+}
+
+/// Linux's `stat` struct
+#[repr(C)]
```
jaybosamiya-ms: If we are mantioning the padding and unused parts, then we should also include the packed modifier on repr c

litebox_common_linux/src/lib.rs
```diff
@@ -79,6 +81,177 @@ bitflags::bitflags! {
...
+            mode,
+            size,
+            ..
+        } = value;
+        unsafe {
+            INO += 1;
+        }
+        Self {
+            // TODO: st_dev and st_ino are used by ld.so to unique identify
+            // shared libraries. Give a random value for now.
```
jaybosamiya-ms, line 234: I've opened https://github.com/MSRSSP/litebox/issues/68 to track this. I had `ino` on my TODO list as a possible thing to add for `stat`, but I wasn't sure how important it is; seems like it is quite important for here. Is the `INO += 1` good enough for now? Or does it need to be high priority to fix?

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1,68 +1,139 @@
...
-) -> Result<i32, Errno> {
-    if dirfd == AT_FDCWD {
-        return sys_open(pathname, flags, mode);
+) -> Result<u32, Errno> {
+    let fs_path = FsPath::new(dirfd, pathname)?;
+    match fs_path {
+        FsPath::Absolute { path } | FsPath::CwdRelative { path } => sys_open(path, flags, mode),
+        FsPath::Cwd => sys_open("", flags, mode),
+        FsPath::Fd(fd) => todo!(),
+        FsPath::FdRelative { fd, path } => todo!(),
```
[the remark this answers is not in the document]
jaybosamiya-ms: Copilot complaining about an explicit `todo!()` is quite annoying, isn't it. Marking it as resolved since this `todo!()` is intentional, and reviewing code with the Copilot box showing up within it for this particular reason is quite annoying. Please unresolve it if you actually intend to update this in the same PR, thanks!

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1,68 +1,139 @@
...
     }
```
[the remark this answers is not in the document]
jaybosamiya-ms: Same

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1,68 +1,139 @@
...
 /// If `Some`, it will read from the specified offset without changing the current file position.
-pub(crate) fn sys_read(fd: i32, buf: &mut [u8], offset: Option<usize>) -> Result<usize, Errno> {
+pub fn sys_read(fd: i32, buf: &mut [u8], offset: Option<usize>) -> Result<usize, Errno> {
+    let Ok(fd) = u32::try_from(fd) else {
+        return Err(Errno::EBADF);
+    };
+    match file_descriptors().read().get_fd(fd) {
+        Some(desc) => match desc {
+            Descriptor::File(file) => litebox_fs().read(file, buf, offset).map_err(Errno::from),
+            Descriptor::Socket(socket) => todo!(),
```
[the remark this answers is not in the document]
jaybosamiya-ms: Same

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1,68 +1,139 @@
...
+pub fn sys_write(fd: i32, buf: &[u8], offset: Option<usize>) -> Result<usize, Errno> {
     let Ok(fd) = u32::try_from(fd) else {
         return Err(Errno::EBADF);
     };
-    match file_descriptors().read().get_file_fd(fd) {
-        Some(file) => litebox_fs().read(file, buf, offset).map_err(Errno::from),
+    match file_descriptors().read().get_fd(fd) {
+        Some(desc) => match desc {
+            Descriptor::File(file) => litebox_fs().write(file, buf, offset).map_err(Errno::from),
+            Descriptor::Socket(socket) => todo!(),
```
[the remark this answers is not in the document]
jaybosamiya-ms: Same

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -72,3 +143,197 @@ pub(crate) fn sys_close(fd: i32) -> Result<(), Errno> {
...
+                    .iov_base
+                    .mutate_subslice_with(..iov_len, |user_buf| {
+                        // TODO: The data transfers performed by readv() and writev() are atomic: the data
+                        // written by writev() is written as a single block that is not intermingled with
+                        // output from writes in other processes
+                        match desc {
+                            Descriptor::File(file) => {
+                                litebox_fs().read(file, user_buf, None).map_err(Errno::from)
+                            }
+                            Descriptor::Socket(socket) => todo!(),
```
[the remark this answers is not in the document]
jaybosamiya-ms: Same

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -72,3 +143,197 @@ pub(crate) fn sys_close(fd: i32) -> Result<(), Errno> {
...
+        Some(desc) => {
+            let mut total_read = 0;
+            for iov in iovs {
+                if iov.iov_len == 0 {
+                    continue;
+                }
+                let Ok(iov_len) = isize::try_from(iov.iov_len) else {
+                    return Err(Errno::EINVAL);
+                };
+                // TODO: use kernel buffer to avoid page faults
```
jaybosamiya-ms: Maybe the nicer interface that could be provided for `RawMutPointer` is a `copy_from_slice`? It would make such cases easier to see/fix. If that sounds reasonable to you, I am happy to push a PR that does that.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -72,3 +143,197 @@ pub(crate) fn sys_close(fd: i32) -> Result<(), Errno> {
...
+                let size = iov
+                    .iov_base
+                    .mutate_subslice_with(..iov_len, |user_buf| {
+                        // TODO: The data transfers performed by readv() and writev() are atomic: the data
+                        // written by writev() is written as a single block that is not intermingled with
+                        // output from writes in other processes
```
jaybosamiya-ms: Do you think this is better provided as a primitive by `litebox::fs::FileSystem` or do you think it is better managed as a lock within the Linux shim? I am leaning towards possibly adding a `fn lock_fd(fd: &Fd) -> ...` (and unlock) to the `FileSystem` trait, so that you can take a lock at the start of readv and thus get atomicity. Please open an issue to track this if so.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -72,3 +143,197 @@ pub(crate) fn sys_close(fd: i32) -> Result<(), Errno> {
...
+                let slice =
+                    unsafe { iov.iov_base.to_cow_slice(iov.iov_len) }.ok_or(Errno::EFAULT)?;
+                // TODO: The data transfers performed by readv() and writev() are atomic: the data
+                // written by writev() is written as a single block that is not intermingled with
+                // output from writes in other processes
+                let size = match desc {
+                    Descriptor::File(file) => litebox_fs()
+                        .write(file, &slice, None)
+                        .map_err(Errno::from)?,
+                    Descriptor::Socket(socket) => todo!(),
```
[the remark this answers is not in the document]
jaybosamiya-ms: Same

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -72,3 +143,197 @@ pub(crate) fn sys_close(fd: i32) -> Result<(), Errno> {
...
+                                litebox_fs().read(file, user_buf, None).map_err(Errno::from)
+                            }
+                            Descriptor::Socket(socket) => todo!(),
+                        }
+                    })
+                    .ok_or(Errno::EFAULT)??;
+
+                total_read += size;
+                if size != iov.iov_len {
+                    // Okay to transfer fewer bytes than requested
```
jaybosamiya-ms:
```suggestion
                if size < iov.iov_len {
                    // Okay to transfer fewer bytes than requested
```
would improve readability; similarly in writev.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -72,3 +143,197 @@ pub(crate) fn sys_close(fd: i32) -> Result<(), Errno> {
...
+    iovec: ConstPtr<IoReadVec<MutPtr<u8>>>,
+    iovcnt: usize,
+) -> Result<usize, Errno> {
+    let Ok(fd) = u32::try_from(fd) else {
+        return Err(Errno::EBADF);
+    };
+    let iovs: &[IoReadVec<MutPtr<u8>>] =
+        unsafe { &iovec.to_cow_slice(iovcnt).ok_or(Errno::EFAULT)? };
+    match file_descriptors().read().get_fd(fd) {
+        Some(desc) => {
```
jaybosamiya-ms: Minor: the rightward shift of nesting can be improved by using something similar to:
```rs
let desc = file_descriptors....get_fd(fd).ok_or(Errno::EBADF)?;
```

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -72,3 +143,197 @@ pub(crate) fn sys_close(fd: i32) -> Result<(), Errno> {
...
+pub fn sys_access(
+    pathname: impl path::Arg,
+    mode: litebox_common_linux::AccessFlags,
+) -> Result<(), Errno> {
+    let status = litebox_fs().file_status(pathname)?;
+    if mode == litebox_common_linux::AccessFlags::F_OK {
+        return Ok(());
+    }
+    // TODO: the check is done using the calling process's real UID and GID.
+    // Here we assume the caller owns the file.
```
jaybosamiya-ms, line 241: The `in_mem` file system already has an internal notion of uid/gid. We should expose this to be able to check stuff better. I added that to https://github.com/MSRSSP/litebox/issues/68 for tracking this

litebox_common_linux/src/lib.rs
CvvT: Got it!

litebox_common_linux/src/lib.rs
CvvT: Fixed!

litebox_common_linux/src/lib.rs
CvvT: I see, thanks for catching that!

litebox_common_linux/src/lib.rs
CvvT: I think it is good enough for now. I'll let you know if I find cases where it does not work.

litebox_shim_linux/src/syscalls/file.rs

litebox_shim_linux/src/syscalls/file.rs
CvvT: Yes, similar to `write_at_offset`, we can have another once for slice. Thanks!

litebox_shim_linux/src/syscalls/file.rs
CvvT: Thanks!

litebox_shim_linux/src/syscalls/file.rs
CvvT: I feel `lock_fd` is more general.

litebox_shim_linux/src/syscalls/file.rs

litebox_common_linux/src/lib.rs
```diff
@@ -79,6 +81,182 @@ bitflags::bitflags! {
...
+        /// descriptor instead.
+        const AT_EMPTY_PATH = 0x1000;
+        /// Don't automount the terminal ("basename") component of pathname if it is a directory
+        /// that is an automount point.
+        const AT_NO_AUTOMOUNT = 0x800;
+        /// Follow symbolic links.
+        const AT_SYMLINK_FOLLOW = 0x400;
+        /// Used with `faccessat`, the checks for accessibility are performed using the
+        /// effective user and group IDs instead of the real user and group ID
+        const AT_EACCESS = 0x200;
```
wdcui, line 116: These two flags have the same value?

litebox_shim_linux/src/lib.rs
```diff
@@ -136,6 +137,12 @@ impl Descriptors {
             None
         }
     }
+    fn get_fd(&self, fd: u32) -> Option<&Descriptor> {
+        if fd >= (2 << 30) {
```
wdcui: Do you use 2<<30 to check if fd is negative?

litebox_common_linux/src/lib.rs
CvvT: Yes, I directly copied it from nix crate: https://github.com/nix-rust/nix/blob/ea012bec465a6d9341d8e4bd8b7c16100393c2aa/src/fcntl.rs#L89
Removed it for now. I believe this flag should only be used by `unlinkat`.

litebox_shim_linux/src/lib.rs
CvvT: Now I feel we don't need this check.

PR #73, opened 2025-04-07
Support fcntl

CvvT:
Add some support for `fcntl`:
*  F_GETFL: each file now is associated with a `status` in the file table.
*  F_GETFD/F_SETFD: each file table entry now is associated with a `close_on_exec`.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -62,7 +62,6 @@ impl<P: path::Arg> FsPath<P> {
 
 /// Handle syscall `open`
 pub fn sys_open(path: impl path::Arg, flags: OFlags, mode: Mode) -> Result<u32, Errno> {
-    let close_on_exec = flags.contains(OFlags::CLOEXEC);
     litebox_fs()
         .open(path, flags, mode)
         .map(|file| {


litebox_common_linux/src/lib.rs
```diff
@@ -255,6 +255,44 @@ impl From<litebox::fs::FileStatus> for FileStat {
...
+        const _ = !0;
+    }
+}
+
+impl FcntlArg {
+    pub fn from(cmd: i32, arg: usize) -> Self {
+        match cmd {
+            F_GETFD => Self::GETFD,
+            #[allow(clippy::cast_possible_truncation)]
+            F_SETFD => Self::SETFD(FileDescriptorFlags::from_bits_truncate(arg as u32)),
```
jaybosamiya-ms, line 289: Maybe I should move https://github.com/MSRSSP/litebox/pull/59 into `litebox` itself so that we can have the nicer version here too. Doesn't need to be in this PR, just making a quick note here.

litebox_shim_linux/src/lib.rs
```diff
@@ -72,8 +72,17 @@ pub(crate) fn litebox_page_manager<'a>() -> &'a PageManager<'static, Platform, P
 type ConstPtr<T> = <Platform as litebox::platform::RawPointerProvider>::RawConstPointer<T>;
 type MutPtr<T> = <Platform as litebox::platform::RawPointerProvider>::RawMutPointer<T>;
 
+/// A file table entry containing a file descriptor and its status
+///
+/// Note `close_on_exec` is per-entry while the `status` is per-file.
```
jaybosamiya-ms: Storing `status` here if it is per-file is likely to cause desyncs between various FD's, no?

litebox_shim_linux/src/lib.rs
```diff
@@ -72,8 +72,17 @@ pub(crate) fn litebox_page_manager<'a>() -> &'a PageManager<'static, Platform, P
...
+struct DescriptorEntry {
+    desc: Descriptor,
+    status: alloc::sync::Arc<core::sync::atomic::AtomicU32>,
```
jaybosamiya-ms: I'm not sure why this is in an arc. Also, without `SETFL`, we don't even need it to be atomic.

CvvT: This is to support `dup` as `status` is per-file (I didn't implement `dup` yet). Maybe it is better to let each file to maintain its `status`? Then, we also need to change fs interface to allow to set/get status. I'm not sure if it is possible to do it by having a `status` outside and without changes to `fs`.

jaybosamiya-ms: I've opened #74 to track this; as I mentioned IRL, some changes to `fs` will support a more generic version of this at the relevant layers.

To prevent this from being a blocker for this PR, as long as you add a `TODO: Fix after #74 is closed` comment here, I think it is fine.

litebox_shim_linux/src/lib.rs
CvvT: Sorry the comment is not accurate. `status` is per-file-struct not per-file. I removed it for now. It might be better to be maintained by each File.

litebox_common_linux/src/lib.rs
CvvT: I can do it in another PR.

jaybosamiya-ms: I just opened #80; we can do the cleanup in a different PR but at least that makes it cleaner for future.

PR #75, opened 2025-04-10
Fix futex

CvvT:
There might be some potential issues with futex:
1. `futex_requeue`: the document says it returns the number of woken threads, but it seems that it actually returns the number of tasks requeued or woken (same as `futex_cmp_requeue`).
2. use i32::MAX instead u32::MAX for the maximum number of threads to requeue/wake: see https://elixir.bootlin.com/linux/v5.19.17/source/kernel/futex/requeue.c#L374
3. It looks like `wake_many` resets `num_to_wake_up` to zero, 

https://github.com/MSRSSP/litebox/blob/3c2c0bb3ea2c3e4d36b275e236e6ebdeb9e0ace7/litebox_platform_linux_userland/src/lib.rs#L285-L286

and then all woken threads  get stuck at https://github.com/MSRSSP/litebox/blob/3c2c0bb3ea2c3e4d36b275e236e6ebdeb9e0ace7/litebox_platform_linux_userland/src/lib.rs#L204-L206

### DIFF
--- a/litebox/src/platform/mod.rs
+++ b/litebox/src/platform/mod.rs
@@ -178,7 +178,7 @@ pub trait RawMutex: Send + Sync {
     ///
     /// Returns the number of waiters that were woken up.
     fn wake_all(&self) -> usize {
-        self.wake_many(u32::MAX as usize)
+        self.wake_many(i32::MAX as usize)
     }
 
     /// If the underlying value is `val`, block until a wake operation wakes us up.


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -247,13 +247,12 @@ impl litebox::platform::RawMutex for RawMutex {
...
-            /* number to requeue */ u32::MAX,
+            /* number to requeue */ i32::MAX as u32,
             Some(&temp_q),
             /* val3: ignored */ 0,
         ) {
-            0 => {
-                // Since we didn't ask for anyone to wake up, this should just move everyone over,
-                // and tell us no one was woken up.
+            0.. => {
+                // on success, return the number of tasks requeued or woken;
```
jaybosamiya-ms: Nit:
```suggestion
                // On success, returns the number of tasks requeued or woken, which we ignore
```

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -276,21 +275,25 @@ impl litebox::platform::RawMutex for RawMutex {
             &temp_q,
             FutexOperation::Requeue,
```
jaybosamiya-ms, line 276: Don't we end up with the same problem here? `num_woken_up` is incorrect due to what you just discovered, yes? IT also includes the number that were re-queued, so we don't actually get an accurate number?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -276,21 +275,25 @@ impl litebox::platform::RawMutex for RawMutex {
...
-        let final_num_to_wake_up = self.num_to_wake_up.swap(0, SeqCst);
-
-        // Confirm that no one has clobbered the lock bits (which would indicate an implementation
-        // failure somewhere).
-        assert_eq!(
-            final_num_to_wake_up >> 30,
-            0b11,
-            "lock bits should remain unclobbered"
-        );
+        let remain = (0b11 << 30) | (n - u32::try_from(num_woken_up).unwrap());
```
jaybosamiya-ms: I think this should just be this, right?  Once the "number remaining" goes down to 0, we know that everything is done, and we can swap it out, yes?
```suggestion
        let remain = 0b11 << 30;
```

litebox_platform_linux_userland/src/lib.rs
CvvT: Right! I think the actual number could be computed by `min(n, num_woken_up)`.

litebox_platform_linux_userland/src/lib.rs
CvvT: Got it!

litebox_platform_linux_userland/src/lib.rs
CvvT: `n` might be larger than the total number of waiting threads. I tweaked the unit test a little bit, in which it tries to `wake_many(10)` while there is only on thread waiting. The remainning number would be 9.

jaybosamiya-ms: Thanks, that clarifies things and helps!

PR #76, opened 2025-04-12
Add metadata operations for all file systems

### LINKED ISSUE #74: Consider exposing a extended metadata storage for file systems
See https://github.com/MSRSSP/litebox/pull/73#discussion_r2034011765

In short, we want to expose something that allows storing shim-layer metadata into the entries in file systems, without needing storage management higher up.

The core necessity of this is to account for situations where the same file is opened multiple times, and needs to have equivalent metadata _aliased_ across the FDs that point to the same underlying file.

We essentially need to support storing type-parametric metadata at both the file and the FD entry layers, with a consistent interface.

My current thought on design is to have an any-abstraction with support for an `Err(Unsupported)` so that some file systems (namely tar-ro) can return that, so that we can support this at the layered abstraction and in-mem layers.

Some details need to be sketched out further for this interface to be nice, but I believe it is feasible to make it non-Linux-specific at the LiteBox layer, while supporting all the Linux-isms that we'd need at the Linux shim layer.

jaybosamiya-ms:
This PR introduces metadata storage to the file system, allowing storing shim-polymorphic data at either the file OR fd layers, depending on the aliasing needs of the shim.

The main reason to add this came from recognizing that there are situations where we want to add data to the "file struct" (shared across duplicated fds, but not across non-duplicated fds) and to the file itself (shared across all fds on the file), as was noticed in #73. This PR sets up the necessary scaffolding for it.

The way it is shim-polymorphic is by not dictating what types are valid metadata. Instead _any_ type is valid metadata to be attached (although only one valid value can be stored for any particular type: the type itself is the key to the metadata maps). Some things may be supported by some file systems, and others might not. For now, all our file systems will support all possible metadata we want to throw into it.

Minor caveat: layered FS has slight oddities in how we'd need to update metadata if the lower layer is read-only. Currently, it simply throws an `unimplemented` exception and would be handled in a separate PR in the future (I am considering this low priority, unless we actually hit such a use case).

@CvvT: you should test (on a temporary local merge) to sure this would be able to help your #73 use case, before we merge this. Let me know if any design changes are needed.

---
Closes #74.

### DIFF
--- a/litebox/src/fs/errors.rs
+++ b/litebox/src/fs/errors.rs
@@ -128,6 +128,23 @@ pub enum FileStatusError {
     PathError(#[from] PathError),
 }
 
+/// Possible errors from [`FileSystem::with_metadata`] and [`FileSystem::with_metadata_mut`]
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum MetadataError {
+    #[error("no such metadata available")]
+    NoSuchMetadata,
+}
+
+/// Possible errors from  [`FileSystem::set_file_metadata`] and [`FileSystem::set_fd_metadata`]
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum SetMetadataError<T> {
+    #[error("the file resides on a read-only filesystem")]
+    // Note: we return the T just so we are not dropping data
+    ReadOnlyFileSystem(T),
+}
+
 /// Possible errors in any file-system function due to path errors.
 #[derive(Error, Debug)]
 pub enum PathError {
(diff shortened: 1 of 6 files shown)


CvvT:
Regarding #73 , since the shim layer also maintains a file table, do we need to move the per-fd metadata from the shim layer to fs?

jaybosamiya-ms:
Currently (on `main`) there isn't any metadata in the shim-layer file table. #73 does indeed introduce metadata there, and yes, that metadata would need to be moved.

I guess things will get simpler once we move the file table entirely out of the shim layer.  Discussing specific details offline is likely easier since I think for the semantics of some metadata (evidenced by #73), we need three layers of indirection, but that seems suboptimal, and I think we could do better.

PR #77, opened 2025-04-14
Fix SIGSYS signal

CvvT:
1. The seccomp filter does not allow calling `rt_sigaction` to replace our SIGSYS handler. However, if the second argument is null, user may only want to read the current registered handler without changing it, in which case we directly handle it.
2. User program may call `rt_sigmask` to block SIGSYS, resulting in crashing. We intercept this syscall, modify its arguments, and then forward it to the kernel. To allow both blocking and invoking this syscall, we set up another seccomp rule which checks whether the fifth argument is a magic number or not (this is set by our code and unused by the kernel). A user program having this knowledge can intentionally use this to invoke the syscall without interception. However, it should be fine as it only leads to the same behavior as not intercepting it (i.e., the program crashes).

### DIFF
--- a/litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
+++ b/litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
@@ -104,8 +104,8 @@ static SET_FS_BASE: spin::Once<fn(u64)> = spin::Once::new();
 /// Need to change it to per-thread.
 static FS_BASE: core::sync::atomic::AtomicU64 = core::sync::atomic::AtomicU64::new(0);
 
-// Certain syscalls with this magic argument are allowed.
-// This is useful for syscall interception where we need to invoke the original syscall.
+/// Certain syscalls with this magic argument are allowed.
+/// This is useful for syscall interception where we need to invoke the original syscall.
 const SYSCALL_ARG_MAGIC: u64 = u64::from_le_bytes(*b"LITE BOX");
 
 /// Get fs register value via syscall `arch_prctl`.
@@ -450,7 +450,7 @@ fn register_seccomp_filter() {
             ],
         ),
         (
-            // allow rt_sigaction that does not block SIGSYS
+            // allow rt_sigprocmask that does not block SIGSYS
             libc::SYS_rt_sigprocmask,
             vec![
                 SeccompRule::new(vec![


litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
```diff
@@ -274,6 +279,56 @@ unsafe extern "C" fn syscall_dispatcher(syscall_number: i64, args: *const usize)
...
+                    let oldaction = TransparentMutPtr { inner: oldaction };
+                    let mut sigset = core::mem::MaybeUninit::uninit();
+                    let _ = unsafe { libc::sigemptyset(sigset.as_mut_ptr()) };
+                    if unsafe {
+                        oldaction.write_at_offset(
+                            0,
+                            libc::sigaction {
+                                sa_sigaction: sigsys_handler as usize,
+                                sa_flags: libc::SA_SIGINFO,
+                                sa_mask: sigset.assume_init(),
```
jaybosamiya-ms, line 301: Nit: `assume_init` is a fairly dangerous function to invoke, so it'd be nice to include a comment saying why it is safe, just in case things change around later.
```suggestion
                                // SAFETY: Initialized by `libc::sigemptyset`
                                sa_mask: sigset.assume_init(),
```

litebox_common_linux/src/lib.rs
```diff
@@ -336,4 +336,5 @@ pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
         buf: Platform::RawMutPointer<FileStat>,
         flags: AtFlags,
     },
+    Ret(i64),
```
jaybosamiya-ms, line 340: Nit: this feels a little surprising given everything else in this enum, thus would be good to just place an explicit comment here.
```suggestion
    /// A sentinel that is expected to be "handled" by trivially returning its value.
    Ret(i64),
```

CvvT: Got it!

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
CvvT: Fixed, thanks!

PR #79, opened 2025-04-14
Introduce stdio `/dev/` devices to the file system

jaybosamiya-ms:
This PR introduces a new filesystem which is aware of exactly 3 files (`/dev/std{in,out,err}`). It is intended to be used with the layering system to provide these as devices in Linux-y environments.  Non-Linux-y environments are unaffected by simply not including them into the layering tree.

Expected layering:
```mermaid
flowchart TD
    l["layered"] --> im["in_mem"]
    l --> l2["layered"]
    l2 --> d["dev::stdio"]
    l2 --> tr["tar_ro"]
```

---
Stacked PR on top of #78

### DIFF
--- a/litebox/src/fs/devices/stdio.rs
+++ b/litebox/src/fs/devices/stdio.rs
@@ -161,7 +161,12 @@ impl<Platform: crate::platform::StdioProvider> super::super::FileSystem
     }
 
     fn fd_file_status(&self, fd: &FileFd) -> Result<FileStatus, FileStatusError> {
-        unimplemented!()
+        assert!(matches!(fd.x.as_usize(), 0..=2));
+        Ok(FileStatus {
+            file_type: FileType::CharacterDevice,
+            mode: Mode::RUSR | Mode::WUSR | Mode::WGRP,
+            size: 0,
+        })
     }
 
     fn with_metadata<T: core::any::Any, R>(


jaybosamiya-ms:
Quick note: all the `unimplemented`s in `stdio.rs` are because I don't actually expect us to hit those cases in practice, but if we do, it is helpful to get the specific panic so we can actually then take the time to figure out the correct effects there.

litebox/src/fs/devices/stdio.rs
```diff
@@ -0,0 +1,198 @@
...
+    fn rmdir(&self, path: impl Arg) -> Result<(), RmdirError> {
+        unimplemented!()
+    }
+
+    fn file_status(&self, path: impl Arg) -> Result<FileStatus, FileStatusError> {
+        let path = self.absolute_path(path)?;
+        if matches!(path.as_str(), "/dev/stdin" | "/dev/stdout" | "/dev/stderr") {
+            Ok(FileStatus {
+                file_type: FileType::CharacterDevice,
+                mode: Mode::RWXU | Mode::RWXG | Mode::RWXO,
```
CvvT: Not sure if it matters. I checked it on my machine:
```
lrwxrwxrwx 1 root root 15 Apr  9 01:08 /dev/stdin -> /proc/self/fd/0
lrwx------ 1 weiteng weiteng 64 Apr 15 13:52 /proc/self/fd/0 -> /dev/pts/14
crw--w---- 1 weiteng tty 136, 14 Apr 15  2025 /dev/pts/14
```
Should we use mode  `crw--w----` (`Mode::WUSR | Mode::RUSR | Mode::WGRP`)?

litebox/src/fs/devices/stdio.rs
```diff
@@ -0,0 +1,198 @@
...
+            RmdirError, SeekError, UnlinkError, WriteError,
+        },
+    },
+    path::Arg,
+    platform::{StdioOutStream, StdioReadError, StdioWriteError},
+};
+
+/// A backing implementation for [`FileSystem`](super::super::FileSystem).
+///
+/// This provider provides only `/dev/stdin`, `/dev/stdout`, and `/dev/stderr`.
```
CvvT, line 20: For other dev files, are we going to implement one fs for each dev file and chain them together through layered fs? Would that be slow?

litebox/src/fs/devices/stdio.rs
```diff
@@ -0,0 +1,198 @@
...
+                mode: Mode::RWXU | Mode::RWXG | Mode::RWXO,
+                size: 0,
+            })
+        } else {
+            Err(FileStatusError::PathError(PathError::NoSuchFileOrDirectory))
+        }
+    }
+
+    fn fd_file_status(&self, fd: &FileFd) -> Result<FileStatus, FileStatusError> {
+        unimplemented!()
```
CvvT: BTW, this is actually needed when I tested against nodejs.

litebox/src/fs/devices/stdio.rs
jaybosamiya-ms: Ah, you are right, I have checked it with:
```console
$ stat /dev/stdin
  File: /dev/stdin -> /proc/self/fd/0
  Size: 15              Blocks: 0          IO Block: 4096   symbolic link
Device: 0,5     Inode: 161         Links: 1
Access: (0777/lrwxrwxrwx)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-04-15 14:33:38.366812975 -0700
Modify: 2025-04-15 13:11:30.582202316 -0700
Change: 2025-04-15 13:11:30.582202316 -0700
 Birth: -
```
when I should have checked with `-L`:
```console
$ stat -L /dev/stdin
  File: /dev/stdin
  Size: 0               Blocks: 0          IO Block: 1024   character special file
Device: 0,67    Inode: 5           Links: 1     Device type: 136,2
Access: (0620/crw--w----)  Uid: ( 1000/    jayb)   Gid: (    5/     tty)
Access: 2025-04-15 14:35:12.709885706 -0700
Modify: 2025-04-15 14:35:12.709885706 -0700
Change: 2025-04-15 14:33:33.720120709 -0700
 Birth: -
```

I think indeed we should remove the `x` from this. As for the other permissions (user vs group vs other), since we don't _really_ have multiple users around, I don't think it is really going to impact things, but yes, I think matching things closely is a good idea.  Will update it. Good catch!

EDIT: Updated

litebox/src/fs/devices/stdio.rs
jaybosamiya-ms: Important questions to answer I think: whether we actually need other `dev` files (the only ones that come to mind are `random`,`urandom`,`null`,`full`,`zero`---those could be considered one/two file systems, depending on how we want to split it), and what the actual cost of the indirection through the layered system really is.

There is a real cost indeed to things that are "lower" in a deep layered system, but in terms of what the actual impact of that is, I don't think it should be particularly significant for most applications. If it is, there are ways we can flatten the tree, and speed things up (e.g., by having "advertisements" of supported files, which the layered FS can cache and directly forward through with a single hash lookup instead); we could also merge the various device files together if even that is a bottleneck.

But yeah, generally speaking, I am not too worried about the performance for this (at least at the moment), especially given that most "bottleneck" files are likely to end up in in-mem layer, which is guaranteed to be the top layer.  

Your question is 100% valid though, and we should think about what alternative designs give us the nice separation, while also reducing the impact on performance.

litebox/src/fs/devices/stdio.rs
jaybosamiya-ms: Ah, good to know, will implement it shortly, thanks!

EDIT: Updated

litebox/src/fs/devices/stdio.rs
jaybosamiya-ms: Opened #81 to track this

CvvT: I see. Thanks for the clarification. Feel free to merge this PR.

PR #82, opened 2025-04-15
Basic Linux Userland Runner

jaybosamiya-ms:
This PR sets up the basic userland runner. There are some more details to be figured out, the biggest one is how we are supporting the various syscall interception backends, but that would be separate future PRs. As an example, currently, `litebox_platform_linux_userland` hard-commits to the SIGSYS-based syscall interception, but this means that our syscall-rewriter backend (which is what would be used on other platforms) is not actually being tested.

The interface of the runner:
```console
$ cargo run --bin litebox_runner_linux_userland -- --help
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/litebox_runner_linux_userland --help`
Run Linux programs with LiteBox on unmodified Linux

Usage: litebox_runner_linux_userland [OPTIONS] <PROGRAM_AND_ARGUMENTS>...

Arguments:
  <PROGRAM_AND_ARGUMENTS>...
          The program and arguments passed to it (e.g., `python3 --version`)

Options:
      --env <ENVIRONMENT_VARIABLES>
          Environment variables passed to the program (`K=V` pairs; can be invoked multiple times)

      --forward-env
          Forward the existing environment variables

  -Z, --unstable
          Allow using unstable options

  -h, --help
          Print help (see a summary with '-h')

Unstable Options:
      --insert-file <INSERT_FILES>
          Pre-fill files into the initial file system state

      --initial-files <PATH_TO_TAR>
          Pre-fill the files in this tar file into the initial file system state

      --rewrite-syscalls
          Apply syscall-rewriter to the ELF file before running it

          This is meant as a convenience feature; real deployments would likely prefer ahead-of-time rewrite things to amortize costs.
```

### DIFF
--- a/litebox_runner_linux_userland/src/lib.rs
+++ b/litebox_runner_linux_userland/src/lib.rs
@@ -56,8 +56,6 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
         )
     }
 
-    // NOTE: We need to grab this info before initializing the platform, because that installs in
-    // hooks that will start taking control of the syscalls being run.
     let (ancestor_modes, prog_data): (Vec<litebox::fs::Mode>, Vec<u8>) = {
         let prog = PathBuf::from(&cli_args.program_and_arguments[0]);
         let ancestors: Vec<_> = prog.ancestors().collect();
@@ -93,7 +91,6 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
     let platform = Box::leak(Box::new(Platform::new(
         None,
         litebox::platform::trivial_providers::ImpossiblePunchthroughProvider {},
-        litebox_shim_linux::syscall_entry,
     )));
     let initial_file_system = {
         let mut in_mem = litebox::fs::in_mem::FileSystem::new(&*platform);
@@ -139,6 +136,7 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
     };
     litebox_shim_linux::set_fs(initial_file_system);
     litebox_platform_multiplex::set_platform(platform);
+    platform.enable_syscall_interception_with(litebox_shim_linux::syscall_entry);
 
     let argv = cli_args
         .program_and_arguments
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -23,6 +23,7 @@ mod syscall_intercept;
 pub struct LinuxUserland<PunchthroughProvider: litebox::platform::PunchthroughProvider = litebox::platform::trivial_providers::ImpossiblePunchthroughProvider> {
     tun_socket_fd: Option<std::os::fd::OwnedFd>,
     punchthrough_provider: PunchthroughProvider,
+    interception_enabled: std::sync::atomic::AtomicBool,
 }
 
 impl<PunchthroughProvider: litebox::platform::PunchthroughProvider>
@@ -33,22 +34,10 @@ impl<PunchthroughProvider: litebox::platform::PunchthroughProvider>
     /// Takes an optional tun device name (such as `"tun0"` or `"tun99"`) to connect networking (if
     /// not specified, networking is disabled).
     ///
-    /// Registers `syscall_handler` to handle all intercepted syscalls.
-    ///
     /// # Panics
     ///
     /// Panics if the tun device could not be successfully opened.
-    pub fn new(
-        tun_device_name: Option<&str>,
-        punchthrough_provider: PunchthroughProvider,
-        syscall_handler: impl Fn(litebox_common_linux::SyscallRequest<LinuxUserland>) -> i64
-        + Send
-        + Sync
-        + 'static,
-    ) -> Self {
-        // TODO: have better signature and registration of the syscall handler.
-        syscall_intercept::init_sys_intercept(syscall_handler);
-
+    pub fn new(tun_device_name: Option<&str>, punchthrough_provider: PunchthroughProvider) -> Self {
         let tun_socket_fd = tun_device_name.map(|tun_device_name| {
             let tun_fd = nix::fcntl::open(
                 "/dev/net/tun",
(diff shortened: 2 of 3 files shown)


jaybosamiya-ms:
Now that https://github.com/MSRSSP/litebox/pull/79 has been merged, we will need some minor changes to make sure that the stdio layer shows up, but that should not really impact reviewing this PR and should be an uncontroversial change. Will add that to this PR as a new commit on top of existing commits soon.

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+            litebox_syscall_rewriter::hook_syscalls_in_elf(&data, None).unwrap()
+        } else {
+            data
+        };
+        (modes, data)
+    };
+
+    // TODO(jb): Clean up platform initialization once we have https://github.com/MSRSSP/litebox/issues/24
+    //
+    // TODO: We also need to pick the type of syscall interception based on whether we want we want
```
wdcui: Minor: two "we want".

litebox_runner_linux_userland/temp.tar
wdcui, line 1: Should we put this file in this folder?

litebox_shim_linux/src/lib.rs
```diff
@@ -29,26 +29,41 @@ use litebox_platform_multiplex::Platform;
 pub mod loader;
 pub mod syscalls;
 
-static FS: OnceBox<litebox::fs::in_mem::FileSystem<Platform>> = OnceBox::new();
+static FS: OnceBox<
+    litebox::fs::layered::FileSystem<
```
wdcui, line 36: Does this spec have to match with the fs set up in the linux userland runner?

litebox_runner_linux_userland/src/lib.rs
jaybosamiya-ms: Good catch, fixed.

litebox_runner_linux_userland/temp.tar
jaybosamiya-ms: Accidentally left in (as the name might imply), not sure why my local .gitignore did not get triggered; removed.

litebox_shim_linux/src/lib.rs
jaybosamiya-ms: Yes, indeed it must match up, the `set_fs` call is what ensures that both the runner and the shim are in agreement (i.e., the compiler will complain if they disagree).

jaybosamiya-ms: This is one of those things that are part of what I'd like to clean up (so that we don't have a hard-code of equality here), but it is a low-priority item to disentangle those via the component thing I was mentioning (mostly because some more details need to be figured out, and I think it is more of an _internal_ ergonomics issue for us developers, more than actual users of litebox; thus not _particularly_ high priority unless it annoys one of us enough :)).

wdcui: I agree that this cleanup is low priority. One question I have is that how much code change it would require to make to the runner when we replace the Linux kernel with the Nanvix kernel. No need to do anything for it in this PR.

jaybosamiya-ms: That's a good question.

This runner is focused entirely on userland, but the shape of the runner for the Nanvix kernel would look quite similar.  Essentially, we'd introduce a new crate `litebox_runner_nanvix` that has essentially the same style of code, but with updated types focusing on Nanvix instead.  Abstracting to have just one runner _may_ be possible, but I don't think we gain much by doing that (esp because we expect the Linux userland one to be more featureful with lots more debugging support). Instead, having two separate runners is the correct move, imho, especially since the code duplication is going to be minor, and mostly "boring" code that is easier to maintain that the actual core internals of LiteBox itself.

litebox_syscall_rewriter/tests/hello
CvvT, line 1: Why this file is updated?

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+                data = &data[len..];
+            }
+            fs.close(fd).unwrap();
+        });
+        let tar_data = if let Some(tar_file) = cli_args.initial_files.as_ref() {
+            if tar_file.extension().and_then(|x| x.to_str()) != Some("tar") {
+                anyhow::bail!("Expected a .tar file, found {}", tar_file.display());
+            }
+            std::fs::read(tar_file)
+                .map_err(|e| anyhow!("Could not read tar file at {}: {}", tar_file.display(), e))?
```
CvvT: Once platform is initialized (i.e., `syscall_entry` is registered), all syscalls would be handled by litebox so this read to host fs would fail. Either copy the data to memory before setting the platform or add a separate interface to register syscall hooks.

litebox_syscall_rewriter/tests/hello
jaybosamiya-ms: Just a permissions change 644->755. It is meant to be an executable, and the runner uses the mode from it, so using the correct permission bits probably makes sense.

litebox_runner_linux_userland/src/lib.rs
jaybosamiya-ms: Good catch. I think the initialization instantly hooks things is confusing and prone to making mistakes (you'll notice that in the same PR, I've already got the `NOTE: We need to grab this info before initializing the platform ...` portion, but yeah, this requires strict vigilance to make sure we don't accidentally cause these issues.

I agree with having a separate "register hooks" point in time would be better, and easier to get things correct. Phase separation :)  We have a similar thing in WasmBox too (https://github.com/MSRSSP/WasmBox/blob/13b4d8e48105f8617cb1ce220293efab0a89e585/wasmbox_std/src/lib.rs#L221-L229), where restrictions kick in, but it ends up being an explicit call.

I'll introduce that to this PR itself, rather than making a separate PR.

PR #83, opened 2025-04-16
Introduce `copy_from_slice` to make kernel buffer usage easier

### LINKED ISSUE #44: Consider a slightly friendlier interface for `mutate_subslice_with`
> Ah, you are right about `mutate_subslice_with`, and I understand now. I still think using `usize` here is safer. The main reason `mutate_subslice_with` exists with `isize` is that _technically_ you may sometimes have buffers where you want to access negative indexes into it. It is quite rare though, so maybe the interface there needs to change a little bit to improve it (i.e., give a nicer default which uses `usize`s, and a rarely-used-but-more-general `isize` based version). I'll give it some thought on how that interface can be improved.

_Originally posted by @jaybosamiya-ms in https://github.com/MSRSSP/litebox/pull/35#discussion_r2006486489_

### LINKED ISSUE #70: Nicer additional function for `RawMutPointer`
> Maybe the nicer interface that could be provided for `RawMutPointer` is a `copy_from_slice`? It would make such cases easier to see/fix. If that sounds reasonable to you, I am happy to push a PR that does that.

_Originally posted by @jaybosamiya-ms in https://github.com/MSRSSP/litebox/pull/67#discussion_r2021900337_

jaybosamiya-ms:
Essentially, this removes a potential foot-gun of performing some operation that might have a pagefault partway through some operation; additionally, it provides a nicer usize-based interface that is more likely to be the common use case, rather than the more general isize-based interface (which is necessary, but also requires more thought and effort to use properly).

Closes #44 and closes #70.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -187,12 +187,19 @@ pub extern "C" fn close(fd: i32) -> i32 {
     syscalls::file::sys_close(fd).map_or_else(Errno::as_neg, |()| 0)
 }
 
+// This places size limits on maximum read/write sizes that might occur; it exists primarily to
+// prevent OOM due to the user asking for a _massive_ read or such at once. Keeping this too small
+// has the downside of requiring too many syscalls, while having it be too large allows for massive
+// allocations to be triggered by the userland program. For now, this is set to a
+// hopefully-reasonable middle ground.
+const MAX_KERNEL_BUF_SIZE: usize = 64 * 1024;
+
 /// Entry point for the syscall handler
 #[allow(clippy::too_many_lines)]
 pub fn syscall_entry(request: SyscallRequest<Platform>) -> i64 {
     let res: Result<usize, Errno> = match request {
         SyscallRequest::Read { fd, buf, count } => {
-            let mut kernel_buf = vec![0u8; count];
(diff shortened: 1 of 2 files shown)


litebox_shim_linux/src/lib.rs
```diff
@@ -192,15 +192,12 @@ pub extern "C" fn close(fd: i32) -> i32 {
...
-            isize::try_from(count)
-                .map_err(|_| Errno::EINVAL)
-                .and_then(|count| {
-                    buf.mutate_subslice_with(..count, |user_buf| {
-                        // TODO: use kernel buffer to avoid page faults
-                        syscalls::file::sys_read(fd, user_buf, None)
-                    })
-                    .unwrap_or(Err(Errno::EFAULT))
-                })
+            let mut kernel_buf = vec![0u8; count];
```
CvvT: `count` might be arbitrarily large, leading to OOM. I used a fixed-size buffer: https://github.com/MSRSSP/litebox/blob/1ff952fbcf9c3958c394a13fbc06e3bcf4fcf382/litebox_shim_linux/src/syscalls/mm.rs#L66

jaybosamiya-ms: Ah, fair point. I'll update the PR to place upper limits to the kernel buffer size.

PR #84, opened 2025-04-16
Pipe

CvvT:
Add support for syscalls `pipe`/`pipe2`

### DIFF
--- a/litebox_shim_linux/src/channel.rs
+++ b/litebox_shim_linux/src/channel.rs
@@ -244,11 +244,11 @@ mod tests {
     extern crate std;
 
     fn init_platform() {
-        let syscall_entry: Option<
-            fn(litebox_common_linux::SyscallRequest<litebox_platform_multiplex::Platform>) -> i64,
-        > = None;
-        let platform = Platform::new(None, ImpossiblePunchthroughProvider {}, syscall_entry);
-        set_platform(platform);
+        let platform = alloc::boxed::Box::leak(alloc::boxed::Box::new(Platform::new(
+            None,
+            ImpossiblePunchthroughProvider {},
+        )));
+        set_platform(&*platform);
     }
 
     #[test]


litebox_common_linux/src/lib.rs
```diff
@@ -170,7 +170,7 @@ pub struct FileStat {
     pub st_mode: u32,
     pub st_uid: u32,
     pub st_gid: u32,
-    __pad0: core::ffi::c_int,
+    pub pad0: core::ffi::c_int,
```
jaybosamiya-ms: Nit: I think we still want to prefix it with `__` just to clarify that this is not meant to be actually touched.

litebox_common_linux/src/lib.rs
```diff
@@ -181,7 +181,7 @@ pub struct FileStat {
     pub st_mtime_nsec: i64,
     pub st_ctime: i64,
     pub st_ctime_nsec: i64,
-    __unused: [i64; 3],
+    pub unused: [i64; 3],
```
jaybosamiya-ms: Similar to `__pad0`

jaybosamiya-ms: If clippy complains, then an `#[expect(clippy::...)]` specific to this field is perfectly fine.

litebox_shim_linux/Cargo.toml
```diff
@@ -11,6 +11,8 @@ litebox_platform_multiplex = { path = "../litebox_platform_multiplex/", version
 once_cell = { version = "1.20.2", default-features = false, features = ["alloc", "race"] }
 elf_loader = { version = "0.9.2", features = ["fs"] }
 thiserror = { version = "2.0.6", default-features = false }
+ringbuf = { version = "0.4.8", default-features = false, features = ["alloc"] }
+spin = { version = "0.9.8", default-features = false, features = ["spin_mutex"] }
```
jaybosamiya-ms: Rather than introducing `spin`, maybe we should provide `Once` as a primitive within our `Synchronization`?

I am not convinced that the current usage of `Once` is necessary in this PR fwiw. The only usage of it that I can tell is during initialization; we can handle that via an `Arc::get_mut_unchecked` (which is `unsafe` but its safety is maintained by the fact that we are in full control of the _just created_ `Arc`s and their weak pointers, thus it is impossible for anything to dereference them while we fill the values in).

litebox_shim_linux/src/channel.rs
```diff
@@ -0,0 +1,333 @@
...
+    where
+        T: Copy,
+    {
+        if is_nonblocking {
+            self.try_write(buf)
+        } else {
+            // TODO: use poll rather than busy wait
+            loop {
+                match self.try_write(buf) {
+                    Err(Errno::EAGAIN) => {}
```
jaybosamiya-ms, line 118: I know you have a TODO to switch to polling, but at least until then, this should have a `core::hint::spin_loop` call

litebox_shim_linux/src/channel.rs
```diff
@@ -0,0 +1,333 @@
...
+    where
+        T: Copy,
+    {
+        if is_nonblocking {
+            self.try_read(buf)
+        } else {
+            // TODO: use poll rather than busy wait
+            loop {
+                match self.try_read(buf) {
+                    Err(Errno::EAGAIN) => {}
```
jaybosamiya-ms, line 184: Same

litebox_common_linux/src/lib.rs
CvvT: Fixed.

litebox_common_linux/src/lib.rs
CvvT: Fixed.

litebox_shim_linux/Cargo.toml
CvvT: Good point! I didn't think of it before. `Arc::get_mut_unchecked` requires a unstable feature `get_mut_unchecked` though. I figured out a workaround that uses `Arc::get_mut` and `Arc::new_cyclic`.

litebox_shim_linux/src/channel.rs
CvvT: Sure, fixed!

litebox_shim_linux/src/channel.rs
CvvT: Fixed.

CvvT:
> nyways, I would say that we don't need to answer that question in this PR, and a separate PR can migrate if we decide that we want this in `litebox` rather than in the shim layer.

I agree. This should be moved to `litebox` in the future.

litebox_shim_linux/Cargo.toml
jaybosamiya-ms: Ah neat! I had missed that `new_cyclic` exists: that is definitely cleaner (and safer), thanks!

jaybosamiya-ms:
#82 has been merged, thanks for taking care of the rebase. Also thanks for pointing at `Arc::new_cyclic`: I hadn't realized this exists---it is definitely the cleaner way of doing it.

PR #85, opened 2025-04-16
Stat

CvvT:
* add support for syscalls `stat`/`fstat`. Since we don't support symbolic links, `fstat` is identical to `stat` for now.
* fix `sys_newfstatat` to support non-file fd.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -347,7 +347,7 @@ pub fn sys_stat(pathname: impl path::Arg) -> Result<FileStat, Errno> {
 
 /// Handle syscall `lstat`
 ///
-/// lstat() is identical to stat(), except that if pathname is a symbolic link,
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -330,14 +326,41 @@ pub fn sys_fstat(fd: i32) -> Result<FileStat, Errno> {
...
+
+/// Handle syscall `stat`
+pub fn sys_stat(pathname: impl path::Arg) -> Result<FileStat, Errno> {
+    let status = litebox_fs().file_status(pathname)?;
+    Ok(FileStat::from(status))
+}
+
+/// Handle syscall `lstat`
+///
+/// lstat() is identical to stat(), except that if pathname is a symbolic link,
```
jaybosamiya-ms: Nit:
```suggestion
/// `lstat` is identical to `stat`, except that if `pathname` is a symbolic link,
```

PR #86, opened 2025-04-17
Account for spurious or immediate wakeups

CvvT:
Ran into another problem introduced by #75:  

https://github.com/MSRSSP/litebox/blob/1ff952fbcf9c3958c394a13fbc06e3bcf4fcf382/litebox_platform_linux_userland/src/lib.rs#L287-L291
If `n` = 1 and `num_wake_up` = 0, `num_to_wake_up` should remain `(0b11 << 30) | 1`. However, it is possible that there is an immediate wakeup that updates `num_to_wake_up` to `(0b11 << 30)`. To fix it, instead of checking the exact number, we should check if the current value is less than or equal to `remain`.

I don't have a reliable test case for this. This was found when testing some other code that uses mutex.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -286,7 +286,7 @@ impl litebox::platform::RawMutex for RawMutex {
         // Unlock the lock bits, allowing other wakers to run.
         let remain = n - num_woken_up;
         while let Err(v) = self.num_to_wake_up.fetch_update(SeqCst, SeqCst, |v| {
-            // due to spurious or immediate wake-ups (i.e., unexpected wakeups that may decrease `num_to_wake_up`),
+            // Due to spurious or immediate wake-ups (i.e., unexpected wakeups that may decrease `num_to_wake_up`),
             // `num_to_wake_up` might end up being less than expected. Thus, we check `<=` rather than `==`.
             if v & ((1 << 30) - 1) <= remain {
                 Some(0)


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -284,11 +284,15 @@ impl litebox::platform::RawMutex for RawMutex {
...
         // Unlock the lock bits, allowing other wakers to run.
-        let remain = (0b11 << 30) | (n - num_woken_up);
-        while let Err(v) = self
-            .num_to_wake_up
-            .compare_exchange(remain, 0, SeqCst, SeqCst)
-        {
+        let remain = n - num_woken_up;
+        while let Err(v) = self.num_to_wake_up.fetch_update(SeqCst, SeqCst, |v| {
+            // due to spurious or immediate wake-ups, `v` might be less than `remain`
+            if v & ((1 << 30) - 1) <= remain {
```
jaybosamiya-ms, line 291: This current fix has the problem of "not everyone has finished waking up yet", right?  Can `num_woken_up` actually ever be 0? It should be `min(n, num_woken_or_requeued)`, neither of which are 0, yes? Or is this if something gets removed from the queue _before_ the wake-ups begin (because the blocks themselves are supposed to re-wait if it is not explicitly woken up, no?).

Maybe we can sketch out the scenario IRL and make sure to fix the issue, because I think at least with the current update, it might break things (it is possible I am misunderstanding something).

CvvT: Sure. What I found was that `num_woken_or_requeued` can be zero.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -284,11 +284,15 @@ impl litebox::platform::RawMutex for RawMutex {
...
 
         // Unlock the lock bits, allowing other wakers to run.
-        let remain = (0b11 << 30) | (n - num_woken_up);
-        while let Err(v) = self
-            .num_to_wake_up
-            .compare_exchange(remain, 0, SeqCst, SeqCst)
-        {
+        let remain = n - num_woken_up;
+        while let Err(v) = self.num_to_wake_up.fetch_update(SeqCst, SeqCst, |v| {
+            // due to spurious or immediate wake-ups, `v` might be less than `remain`
```
jaybosamiya-ms:
```suggestion
            // Due to spurious/immediate wake-ups, `v` might end up less than `remain`.
            // Thus, we check `<=` rather than `==`.
```

PR #87, opened 2025-04-17
Update nostd CI to fail on cargo build errors

jaybosamiya-ms:
It appears that our `Confirm no_std` CI job was not actually complaining upon usage of `std` (it would show up in the logs, but would not exit with non-zero exit code, and thus would not show up as a :x: in the GitHub interface). Thankfully, we don't have any new _unexpected_ usage of `std` that has shown up yet. I mostly recognized this because the CI did _not_ complain in a situation where new _expected_ `std` usage showed up. There _are_ two crates that it should have complained about (the syscall rewriter, and the runner) that _do_ use `std` (unsurprisingly). These have now been added to the exclusions, as well as fixing up the CI to actually complain.

Essentially, we should see the first commit d2611b61c34e5f880b7cb35964f9bcb15ba39576 (which fixes the CI) have a failure, while the second commit 67c7518ee0b295d052daf6cacf8b5a935bb9d2eb (which adds the two exclusions) should succeed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,13 +74,22 @@ jobs:
           #   more precise check, but as long as we are tracking the underlying
           #   platforms, we are unlikely to hit any significant issues here.
           #
+          # - `litebox_runner_linux_userland` is allowed to have `std` access
+          #   since it needs to actually access the file-system, pull in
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
Will merge without review as soon as I get the CI green, since I expect this to be an uncontroversial improvement to CI.

PR #88, opened 2025-04-17
Eventfd

CvvT:
add support for syscall `eventfd`.

The test case `test_semaphore_eventfd` likely triggers the issue mentioned in #86.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -108,11 +108,11 @@ pub fn sys_read(fd: i32, buf: &mut [u8], offset: Option<usize>) -> Result<usize,
             Descriptor::PipeReader { consumer, .. } => consumer.read(buf),
             Descriptor::PipeWriter { .. } => Err(Errno::EINVAL),
             Descriptor::Eventfd { file, .. } => {
-                if buf.len() < 8 {
+                if buf.len() < size_of::<u64>() {
                     return Err(Errno::EINVAL);
                 }
                 let value = file.read()?;
-                buf[..8].copy_from_slice(&value.to_le_bytes());
+                buf[..size_of::<u64>()].copy_from_slice(&value.to_le_bytes());
                 Ok(size_of::<u64>())
             }
         },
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/syscalls/eventfd.rs
```diff
@@ -0,0 +1,208 @@
...
+                return Ok(8);
+            }
+        }
+
+        Err(Errno::EAGAIN)
+    }
+
+    pub(crate) fn write(&self, value: u64, is_nonblocking: bool) -> Result<usize, Errno> {
+        if is_nonblocking {
+            self.try_write(value)
```
jaybosamiya-ms: Why not just pick up the non-blocking value from the status?

litebox_shim_linux/src/syscalls/eventfd.rs
```diff
@@ -0,0 +1,208 @@
...
+        if on {
+            self.status
+                .fetch_or(flag.bits(), core::sync::atomic::Ordering::Relaxed);
+        } else {
+            self.status.fetch_and(
+                flag.complement().bits(),
+                core::sync::atomic::Ordering::Relaxed,
+            );
+        }
+    }
```
jaybosamiya-ms: Dead code?

jaybosamiya-ms: If so, we don't actually need `status` to be an atomic, and can instead just use an `OFlags` directly (or indeed even a `is_blocking: bool`), which would simplify things.

litebox_shim_linux/src/syscalls/eventfd.rs
CvvT: You're right. It is not necessary. I also fixed a similar issue for `pipe`.

litebox_shim_linux/src/syscalls/eventfd.rs
CvvT: I should've submitted #95 first. Please review #95 first.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -105,10 +105,16 @@ pub fn sys_read(fd: i32, buf: &mut [u8], offset: Option<usize>) -> Result<usize,
...
-            }
+            Descriptor::PipeReader { consumer, .. } => consumer.read(buf),
             Descriptor::PipeWriter { .. } => Err(Errno::EINVAL),
+            Descriptor::Eventfd { file, .. } => {
+                if buf.len() < 8 {
+                    return Err(Errno::EINVAL);
+                }
+                let value = file.read()?;
+                buf[..8].copy_from_slice(&value.to_le_bytes());
+                Ok(size_of::<u64>())
```
jaybosamiya-ms: Nit: It is a little surprising to see both `8` and `size_of::<u64>` show up in close proximity. Just for consistency, we might want to stick to one or the other. I don't think this is actually problematic in any way though.

Issue #68, opened 2025-03-31
File status should return dev/ino/uid/gid fields too


jaybosamiya-ms:
See https://github.com/MSRSSP/litebox/blob/2c0295184e1d45cfbf318aacbe1ddb2ea299c746/litebox_common_linux/src/lib.rs#L198-L205

Unresolved questions:
- How should the layered FS handle inode numbers?

Issue #31, opened 2025-03-08
FD cleanup


jaybosamiya-ms:
> [!NOTE]
> ### Current status of this issue
>
> Most of the bits of this issue have now been handled; latest status of remaining bits is at https://github.com/microsoft/litebox/issues/31#issuecomment-4071407039

Currently, we end up having two different type-distinct namespaces for FileSystem FDs and Socket FDs. This requires the Linux shim to maintain its own storage for FDs that translates between the two. It would likely be better to remove this indirect storage, and instead provide a safe way to maintain the two types of FDs, stored in LiteBox itself, such that the shim doesn't need to deal with the storage.

A fundamental issue with this is that it is unclear whether it is feasible to make a clean type-based separation between the types, and _statically_ prevent FD-confusion bugs (for users who are _not_ a POSIX-like shim, but have better higher-level concepts available), while _also_ giving the simple integer-based handling that the Linux-shim would prefer.

I think with some careful design, this _may_ be feasible, but the details do need to be fleshed out.

jaybosamiya-ms:
Note: this _may_ require #24 to be done first.

jaybosamiya-ms:
Related: https://github.com/MSRSSP/litebox/pull/84#pullrequestreview-2773978790 + https://github.com/MSRSSP/litebox/pull/84#issuecomment-2813713240 (we need to move pipes into litebox, but that kinda depends on FD storage being improved).

jaybosamiya-ms:
Important to account for as part of this: https://github.com/microsoft/litebox/pull/169#issuecomment-2993240055

jaybosamiya-ms:
Explicitly noting here that the natural type for `fd`s specifically within the `litebox_shim_linux` crate is `i32` and not an unsigned type.  See related https://github.com/microsoft/litebox/pull/344#issuecomment-3287212360

jaybosamiya-ms:
Noting remaining bits:
- simplifying redundancy in Linux shim around min/max FD handling to make sure it is maintained more consistently everywhere
- removing the i32->u32->usize repetitive code that is prone to introduce inconsistencies in the long run
- introduce nicer stuff for `run_on_raw_fd(fd, f, f, f, f, ...)`

PR #89, opened 2025-04-18
Stdio

CvvT:
* Pre-open stdin/stdout/stderr
* Add `ioctl` to support some terminal related commands `TCGETS/TCSETS/TIOCGWINSZ/TIOCGPTN` and `FIONBIO`.

Edit:
* Resolve symbolic links for "/proc/self/fd/*"
* `stat/newfstatat` should account for symbolic links

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -337,7 +337,7 @@ pub const TIOCGWINSZ: u32 = 0x5413;
 pub const FIONBIO: u32 = 0x5421;
 pub const TIOCGPTN: u32 = 0x80045430;
 
-/// Commands for use with `fcntl`.
+/// Commands for use with `ioctl`.
 #[non_exhaustive]
 pub enum IoctlArg<Platform: litebox::platform::RawPointerProvider> {
     /// Get the current serial port settings.
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -616,27 +616,18 @@ impl litebox::platform::StdioProvider for LinuxUserland {
         stream: litebox::platform::StdioOutStream,
         buf: &[u8],
     ) -> Result<usize, litebox::platform::StdioWriteError> {
-        let n = match stream {
-            litebox::platform::StdioOutStream::Stdout => unsafe {
-                libc::syscall(
-                    libc::SYS_write,
-                    libc::STDOUT_FILENO,
-                    buf.as_ptr(),
-                    buf.len(),
-                    // Unused by the syscall but would be checked by Seccomp filter if enabled.
-                    syscall_intercept::systrap::SYSCALL_ARG_MAGIC,
-                )
-            },
-            litebox::platform::StdioOutStream::Stderr => unsafe {
-                libc::syscall(
-                    libc::SYS_write,
-                    libc::STDERR_FILENO,
-                    buf.as_ptr(),
-                    buf.len(),
-                    // Unused by the syscall but would be checked by Seccomp filter if enabled.
-                    syscall_intercept::systrap::SYSCALL_ARG_MAGIC,
-                )
-            },
+        let n = unsafe {
+            libc::syscall(
+                libc::SYS_write,
+                match stream {
+                    litebox::platform::StdioOutStream::Stdout => libc::STDOUT_FILENO,
+                    litebox::platform::StdioOutStream::Stderr => libc::STDERR_FILENO,
+                },
+                buf.as_ptr(),
+                buf.len(),
+                // Unused by the syscall but would be checked by Seccomp filter if enabled.
+                syscall_intercept::systrap::SYSCALL_ARG_MAGIC,
+            )
         };
         if n < 0 {
             match latest_errno() {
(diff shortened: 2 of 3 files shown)


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -62,6 +62,10 @@ impl<P: path::Arg> FsPath<P> {
 
 /// Handle syscall `open`
 pub fn sys_open(path: impl path::Arg, flags: OFlags, mode: Mode) -> Result<u32, Errno> {
+    let is_stdio = matches!(
+        path.normalized()?.as_str(),
+        "/dev/stdin" | "/dev/stdout" | "/dev/stderr"
+    );
```
CvvT: I don't have a good idea on how to distinguish between regular file and stdio. This is for supporting `ioctl`. Should we add a new interface `is_terminal` to fs or define a custom type `Stdio` to `FileType`? https://github.com/MSRSSP/litebox/blob/9c9a8515af925357f2ebd0b1015c41512a81dc67/litebox/src/fs/mod.rs#L163-L166

jaybosamiya-ms:
We decided (IRL) to put a pause on this PR for now, will revisit after https://github.com/MSRSSP/litebox/issues/68 (and specifically, once `dev_t st_rdev; /* Device ID (if special file) */` is added)

litebox_shim_linux/src/syscalls/file.rs
jaybosamiya-ms: Making a brief note here that `rdev = 136,X` for a char device implies it is a Linux PTY slave, which we would be able to use to distinguish things after #68 (see https://www.kernel.org/doc/Documentation/admin-guide/devices.txt)

litebox_common_linux/src/lib.rs
```diff
@@ -159,14 +159,15 @@ impl From<litebox::fs::FileType> for InodeType {
...
+            litebox::fs::FileType::CharacterDevice => InodeType::CharDevice,
             _ => unimplemented!(),
         }
     }
 }
 
 /// Linux's `stat` struct
 #[repr(C, packed)]
-#[derive(Clone, Default)]
+#[derive(Clone, Default, PartialEq, Debug)]
```
jaybosamiya-ms, line 170: I'm surprised at the need for `PartialEq` here. It because of a test?  Not really an issue to keep it, just curious.

litebox_common_linux/src/lib.rs
```diff
@@ -306,6 +307,56 @@ bitflags::bitflags! {
...
+
+pub const TCGETS: u32 = 0x5401;
+pub const TCSETS: u32 = 0x5402;
+pub const TIOCGWINSZ: u32 = 0x5413;
+pub const FIONBIO: u32 = 0x5421;
+pub const TIOCGPTN: u32 = 0x80045430;
+
+/// Commands for use with `fcntl`.
+#[non_exhaustive]
+pub enum IoctlArg<Platform: litebox::platform::RawPointerProvider> {
```
jaybosamiya-ms: Typo?

```suggestion
/// Commands for use with `ioctl`.
#[non_exhaustive]
pub enum IoctlArg<Platform: litebox::platform::RawPointerProvider> {
```

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -607,18 +616,35 @@ impl litebox::platform::StdioProvider for LinuxUserland {
...
+                libc::syscall(
+                    libc::SYS_write,
+                    libc::STDERR_FILENO,
+                    buf.as_ptr(),
+                    buf.len(),
+                    // Unused by the syscall but would be checked by Seccomp filter if enabled.
+                    syscall_intercept::systrap::SYSCALL_ARG_MAGIC,
+                )
+            },
+        };
```
jaybosamiya-ms: These could be collapsed using a:
```rs
match stream {
  Stdout => STDOUT_FILENO,
  Stderr => STDERR_FILENO,
}
```
in the second argument.

litebox_shim_linux/src/lib.rs
```diff
@@ -172,6 +207,7 @@ enum Descriptor {
         file: alloc::sync::Arc<syscalls::eventfd::EventFile<Platform>>,
         close_on_exec: core::sync::atomic::AtomicBool,
     },
+    Stdio(stdio::StdioFile),
```
jaybosamiya-ms, line 211: I am not convinced that this should exist at this level. Just added this comment to connect it to #31 (we might wish to add that in the code too, but just this comment should be sufficient).

litebox_common_linux/src/lib.rs
CvvT: Yes, it is for the test:

https://github.com/MSRSSP/litebox/blob/7ca1e083ee3a43e4f58a55bf81d911b9d93fc27c/litebox_shim_linux/src/stdio.rs#L71-L74

litebox_common_linux/src/lib.rs
CvvT: Oops, thanks!

litebox_platform_linux_userland/src/lib.rs
CvvT: Good point! Thanks!

litebox_shim_linux/src/lib.rs
CvvT: Right, this is just a workaround for now. Added a comment.

PR #90, opened 2025-04-21
Add `MappedRwLock` guards for more flexible data handling

jaybosamiya-ms:
This PR adds `map` to the read/write guards for `RwLocks`. I have not yet implemented similar functionality for `Mutex`, since I didn't yet find a need for it.

I ended up needing this functionality on the way to implementing some other functionality, but decided to split this out into a separate PR to make it easier to review.

At a high level, this interface mimics the (currently nightly-only) `mapped_lock_guards` API in Rust std (see https://doc.rust-lang.org/std/sync/struct.RwLockReadGuard.html#method.map for an example), but relevant changes have been made to make sure our lock tracing features still work.

### DIFF
--- a/litebox/src/sync/lock_tracing.rs
+++ b/litebox/src/sync/lock_tracing.rs
@@ -194,6 +194,18 @@ impl<Platform: RawSyncPrimitivesProvider> Drop for LockedWitness<Platform> {
         assert!(self.unlocked, "Someone forgot to call `mark_unlock`");
     }
 }
+impl<Platform: RawSyncPrimitivesProvider> LockedWitness<Platform> {
+    /// This function creates a new witness from a borrowed witness. This is only safe to run if
+    /// `&self` is under a `ManuallyDrop` and thus won't be auto-dropped. This is intended to be
+    /// used only with the mapped guards.
+    pub(crate) unsafe fn reborrow_for_mapped_guard(&mut self) -> Self {
+        Self {
+            idx: self.idx,
+            unlocked: self.unlocked,
+            tracker: self.tracker.clone(),
+        }
+    }
+}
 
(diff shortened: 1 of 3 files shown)


litebox/src/sync/lock_tracing.rs
```diff
@@ -196,6 +196,18 @@ impl<Platform: RawSyncPrimitivesProvider> Drop for LockedWitness<'_, Platform> {
         assert!(self.unlocked, "Someone forgot to call `mark_unlock`");
     }
 }
+impl<Platform: RawSyncPrimitivesProvider> LockedWitness<'_, Platform> {
+    /// This function creates a new witness from a borrowed witness. This is only safe to run if
+    /// `&self` is under a `ManuallyDrop` and thus won't be auto-dropped. This is intended to be
+    /// used only with the mapped guards.
+    pub(crate) unsafe fn reborrow_for_mapped_guard(&mut self) -> Self {
```
CvvT, line 201: Why this requires mutable `self`?

jaybosamiya-ms: It doesn't _require_ mutable self, but as the comment says, it is intended to be used only in _very_ limited circumstances; generally speaking this is only doable if you already have a _consumable_ exclusive `self`; the fact that it needs to be "reborrow" by constructing another `Self` is because of some subtle constraints under `Drop` + moving things out of a `ManuallyDrop`.

Basically, I don't actually want to allow "easy" construction of this object, and ideally I would want this to take `self` rather than `&mut self`/`&self`, but some Rust constraints make that impossible, so I picked the next strictest (i.e., `&mut self`).

CvvT: Got it. Thanks!

PR #92, opened 2025-04-23
LVBS placeholder and skeleton runner

sangho2:
This PR adds some placeholders for LVBS (`litebox_platform_lvbs`, `litebox_runner_lvbs`, and `litebox_shim_optee`). For now, `litebox_platform_lvbs` and `litebox_shim_optee` are clones of `litebox_platform_linux_kernel` and `litebox_shim_linux`, respectively.

`litebox_runner_lvbs` is a bit different because it requires a custom build target for VTL1 with its own linker script (entry point and alignments are fixed). To this end, it requires its own Rust toolchain (`nightly`) and that is why I exclude it from the main Cargo workspace.

### DIFF
--- a/litebox_platform_lvbs/README.md
+++ b/litebox_platform_lvbs/README.md
@@ -1,0 +1,5 @@
+# A LiteBox Platform for running LiteBox in Hyper-V VTL1 kernel space
+
+> [!WARNING]
+> This crate is work in progress. Currently, it is a copy of
+> `litebox_platform_linux_kernel`.
--- a/litebox_runner_lvbs/README.md
+++ b/litebox_runner_lvbs/README.md
@@ -1,0 +1,4 @@
+# A LiteBox Runner for running LiteBox in Hyper-V VTL1 kernel space
+
+> [!WARNING]
+> This crate is work in progress.
--- a/litebox_shim_optee/README.md
+++ b/litebox_shim_optee/README.md
@@ -1,5 +1,5 @@
 # A shim that provides an OP-TEE-compatible ABI via LiteBox
 
-This shim is parametric in the choice of platform, which is initialized by first
-invoking `set_platform`, after which all functionality within the OP-TEE shim is
-able to use that particular platform.
+> [!WARNING]
+> This crate is work in progress. Currently, it is a copy of
+> `litebox_shim_linux`.


.github/workflows/ci.yml
```diff
@@ -78,18 +78,27 @@ jobs:
           #   since it needs to actually access the file-system, pull in
           #   relevant files, and then actually trigger LiteBox itself.
           #
+          # - `litebox_runner_lvbs` has a custom target (no_std), so it does not
```
wdcui: Minor: "does not with"?

.github/workflows/ci.yml
```diff
@@ -78,18 +78,27 @@ jobs:
...
+          #   with the current no_std checker.
+          #
           # - `litebox_shim_linux` (in its default feature set) depends on
           #   `litebox_platform_multiplex`; similarly, ideally we'd do a more
           #   precise check.
           #
+          # - `litebox_shim_optee` is expected to work with platform_lvbs (no_std)
```
wdcui: Minor: litebox_platform_lvbs, litebox_platform_linux_userland?

litebox_runner_lvbs/src/main.rs
```diff
@@ -0,0 +1,26 @@
+#![no_std]
+#![no_main]
+
+use core::{arch::asm, panic::PanicInfo};
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn _start() -> ! {
+    unsafe {
+        asm!("\
```
wdcui: Minor: Does this indentation meet the style check?

.github/workflows/ci.yml
sangho2: fixed typo

.github/workflows/ci.yml
sangho2: fixed style

litebox_runner_lvbs/src/main.rs
sangho2: I'll follow the style of some existing asm blocks.

jaybosamiya-ms:
The github diff seems to also be including the changes from PRs #90 and #91 for some reason, which is making reviewing somewhat difficult. Could you squash things into a single commit in the same PR (I don't think any of the current set of commit messages in this PR need to be maintained separately) so that reviewing is easier, thanks!

jaybosamiya-ms:
Maybe we want to also leave README messages in the respective directories stating that they are clones currently. @sangho2 I'll let you make a decision on that before merging, thanks!

PR #94, opened 2025-04-23
Introduce a top-level `LiteBox` type

jaybosamiya-ms:
This introduces a top-level `LiteBox` type to hold all the litebox-internal "globals" we might want to maintain. This is necessary for things such as the shared FD storage (#31).

The members of `LiteBoxX` will be updated/added to as we proceed, but that should not change any public interfaces. This PR on the other hand _does_ change public interfaces (in particular, all the different subcomponents now require a `&LiteBox` to initialize, rather than a `&Platform`). This helps with #31, but would also help if/when we get the full version of #24 working. For now though, it unblocks a simpler variant of it, that I believe would be sufficient for #31.

### DIFF
--- a/litebox_platform_lvbs/src/mm/tests.rs
+++ b/litebox_platform_lvbs/src/mm/tests.rs
@@ -4,6 +4,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 use arrayvec::ArrayVec;
 use litebox::{
+    LiteBox,
     mm::{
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
Good question; the network code was not depending directly on `Synchronization` and thus the type system didn't complain when I took `Synchronization::new` away. I will update it to have the same sort of signature as the other bits.

PR #95, opened 2025-04-24
support fcntl `SETFL`

CvvT:
* Support one more command `SETFL` for fcntl to set file status.
* Fix `STATUS_FLAGS_MASK` as it should include RW bits.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -461,14 +461,24 @@ pub fn sys_fcntl(fd: i32, arg: FcntlArg) -> Result<u32, Errno> {
                 | OFlags::NDELAY
                 | OFlags::DIRECT
                 | OFlags::NOATIME;
+            macro_rules! toggle_flags {
+                ($t:ident) => {
+                    let diff = $t.get_status() ^ flags;
+                    if diff.intersects(OFlags::APPEND | OFlags::DIRECT | OFlags::NOATIME) {
+                        todo!("unsupported flags");
+                    }
+                    $t.set_status(flags & setfl_mask, true);
+                    $t.set_status(flags.complement() & setfl_mask, false);
+                };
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -444,6 +455,24 @@ pub fn sys_fcntl(fd: i32, arg: FcntlArg) -> Result<u32, Errno> {
...
+            match desc {
+                Descriptor::File(file) => todo!(),
+                Descriptor::Socket(socket) => todo!(),
+                Descriptor::PipeReader { consumer, .. } => {
+                    toggle_flags!(consumer, flags, setfl_mask);
+                }
+                Descriptor::PipeWriter { producer, .. } => {
+                    toggle_flags!(producer, flags, setfl_mask);
+                }
+            }
```
jaybosamiya-ms: Minor: you can move the `macro_rules!` closer to the actual usage, which simplifies the macro (fewer meta variables) and makes it easier to track what is happening.
```suggestion
            macro_rules! toggle_flags {
                ($t:ident) => {
                    let diff = $t.get_status() ^ flags;
                    if diff.intersects(OFlags::APPEND | OFlags::DIRECT | OFlags::NOATIME) {
                        todo!("unsupported flags");
                    }
                    $t.set_status(flags & setfl_mask, true);
                    $t.set_status(flags.complement() & setfl_mask, false);
                };
            }
            match desc {
                Descriptor::File(file) => todo!(),
                Descriptor::Socket(socket) => todo!(),
                Descriptor::PipeReader { consumer, .. } => {
                    toggle_flags!(consumer);
                }
                Descriptor::PipeWriter { producer, .. } => {
                    toggle_flags!(producer);
                }
            }
```

CvvT: Nice, thanks!

PR #96, opened 2025-04-24
Remove some Dependencies with STD

ppenna:
This commit patches the build system to reduce the number of dependencies with STD.

### DIFF
--- a/litebox_syscall_rewriter/Cargo.toml
+++ b/litebox_syscall_rewriter/Cargo.toml
@@ -4,11 +4,13 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
-anyhow = "1.0"
+anyhow = { version = "1.0.97", default-features = false }
 capstone = { version = "0.13.0", default-features = false }
-clap = { version = "4.5.32", features = ["derive"] }
 memmap2 = "0.9"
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
Thanks @ppenna, if you don't mind me pushing changes to this branch, I would like to make some changes to this before I merge it. I don't think for the small set of changes, a back-and-forth with the review system is helpful.

ppenna:
> Thanks @ppenna, if you don't mind me pushing changes to this branch, I would like to make some changes to this before I merge it. I don't think for the small set of changes, a back-and-forth with the review system is helpful.

@jaybosamiya-ms  sure!

jaybosamiya-ms:
As part of briefly looking at this (I like to add checks to CI to make sure they don't back-slide), turns out that capstone itself would not work in no-std mode without changes to `captone-sys` that we'd need to upstream or fork.

On the other hand, as we discussed IRL, we don't actually need the syscall rewriter to be no-std, since it is not actually expected to be run on any real no-std environments. It is only intended as an AOT pass.

Thus, I am closing this particular PR as not-planned. I have a separate PR I'll be making (soon) that gets the 32-bit no-std support needed for the actual crates we want, but that won't be as part of this PR/branch.

PR #100, opened 2025-04-29
LVBS VTL1 kernel

sangho2:
This PR adds essential support for LVBS VTL1 kernel. It includes
- Hyper-V Hypercall functions and data structures
- VTL switch functions and data structures
- Support for GDT, TSS, and IDT
- Support for Multiple AP/VP
- Basic VSM functions
- I/O Port based serial communication

### DIFF
--- a/litebox_platform_lvbs/src/kernel_context.rs
+++ b/litebox_platform_lvbs/src/kernel_context.rs
@@ -6,7 +6,7 @@ use crate::{
 };
 use x86_64::structures::tss::TaskStateSegment;
 
-pub const MAX_CORES: usize = 8; // TODO: MAX_CORES = 96?
+pub const MAX_CORES: usize = 8; // TODO: use cpumask
 pub const INTERRUPT_STACK_SIZE: usize = 2 * PAGE_SIZE;
 pub const KERNEL_STACK_SIZE: usize = 8 * PAGE_SIZE;
 
@@ -51,7 +51,7 @@ impl KernelContext {
     }
 }
 
-// TODO: use heap later
+// TODO: use heap
 static mut PER_CORE_KERNEL_CONTEXT: [KernelContext; MAX_CORES] = [KernelContext {
     hv_vp_assist_page: [0u8; PAGE_SIZE],
     interrupt_stack: [0u8; INTERRUPT_STACK_SIZE],
--- a/litebox_platform_lvbs/src/mshv/vtl_switch.rs
+++ b/litebox_platform_lvbs/src/mshv/vtl_switch.rs
@@ -194,16 +194,18 @@ fn load_vtl1_state() {
     pop_vtl_state(&vtl1_state);
 }
 
-/// VTL switch loop (entry point)
-pub fn vtl_switch_loop() -> ! {
+pub fn vtl_switch_loop_entry() -> ! {
     save_vtl0_state();
-    vtl_switch_loop_internal(0);
+    // This is a dummy call to satisfy load_vtl0_state() with reasonable register values.
+    // We do not save VTL0 registers during VTL1 initialization.
+
+    vtl_switch_loop(0);
 }
 
-/// VTL switch loop (internal)
+/// VTL switch loop
 /// # Panics
 /// Panics if VTL call parameter 0 is greater than u32::MAX
-pub fn vtl_switch_loop_internal(result: u64) -> ! {
+pub fn vtl_switch_loop(result: u64) -> ! {
     loop {
         save_vtl1_state();
         load_vtl0_state();
@@ -225,20 +227,20 @@ pub fn vtl_switch_loop_internal(result: u64) -> ! {
                     serial_println!("unknown function ID = {:#x}", params[0]);
                 } else {
                     let new_result = vsm_dispatch(&params);
-                    vtl_switch_loop_internal(new_result)
+                    vtl_switch_loop(new_result)
                 }
             }
             VtlEntryReason::Interrupt => {
                 serial_println!("handle VTL interrupt");
-                vtl_switch_loop_internal(0)
+                vtl_switch_loop(0)
             }
             VtlEntryReason::Intercept => {
                 serial_println!("handle VTL intercept");
-                vtl_switch_loop_internal(0)
+                vtl_switch_loop(0)
             }
             VtlEntryReason::Unknown => {
                 serial_println!("Unknown VTL entry reason");
-                vtl_switch_loop_internal(0)
+                vtl_switch_loop(0)
             }
         }
         // do not put any code which might corrupt registers
--- a/litebox_runner_lvbs/src/main.rs
+++ b/litebox_runner_lvbs/src/main.rs
@@ -5,7 +5,7 @@ use core::arch::asm;
 use litebox_platform_lvbs::{
     arch::instrs::hlt_loop,
     kernel_context::{get_core_id, get_per_core_kernel_context},
-    mshv::vtl_switch::vtl_switch_loop,
+    mshv::vtl_switch::vtl_switch_loop_entry,
     serial_println,
 };
 
@@ -36,10 +36,9 @@ pub fn kernel_main() -> ! {
         serial_println!("==============================");
     }
 
-    // TODO: BSP init (e.g., heap, ...)
+    // TODO: common init (e.g., heap, ...)
 
     litebox_runner_lvbs::per_core_init();
 
-    // TODO: platfrom.switch_loop()?
-    vtl_switch_loop()
+    vtl_switch_loop_entry()
 }
--- a/litebox_platform_lvbs/src/mshv/hvcall_vp.rs
+++ b/litebox_platform_lvbs/src/mshv/hvcall_vp.rs
@@ -107,7 +107,6 @@ fn hvcall_enable_vp_vtl(
     )
 }
 
-// TODO: better way to get the entry point?
 unsafe extern "C" {
     static _start: u8;
 }
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -23,7 +23,7 @@ pub fn mshv_vsm_enable_aps(_cpu_present_mask: u64) -> u64 {
     #[cfg(debug_assertions)]
     serial_println!("VSM: Enable VTL of APs");
 
-    let num_cores = 6; // TODO: decode cpu_present_mask instead of using hardcoded value
+    let num_cores = 6; // TODO: use cpumask
     if let Err(result) = init_vtl_aps(num_cores) {
         serial_println!("Err: {:?}", result);
         let err: u32 = result.into();
--- a/litebox_platform_lvbs/src/mshv/vtl1_mem_layout.rs
+++ b/litebox_platform_lvbs/src/mshv/vtl1_mem_layout.rs
@@ -29,7 +29,6 @@ unsafe extern "C" {
     static _memory_base: u8;
 }
 
-// TODO: should be removed in the future
 #[inline]
 pub fn get_memory_base_address() -> u64 {
     &raw const _memory_base as u64


litebox_platform_lvbs/Cargo.toml
```diff
@@ -10,18 +10,33 @@ bindgen = "0.71.0"
 
 [dependencies]
 bitflags = "2.9.0"
-buddy_system_allocator = { version = "0.11.0", default-features = false, features = ["use_spin"] }
+buddy_system_allocator = { version = "0.11.0", default-features = false, features = [
```
wdcui: Minor: the two lines are the same?

litebox_platform_lvbs/Cargo.toml
```diff
@@ -10,18 +10,33 @@ bindgen = "0.71.0"
...
-x86_64 = { version = "0.15.2", default-features = false, features = ["instructions"] }
+x86_64 = { version = "0.15.2", default-features = false, features = [
+  "abi_x86_interrupt",
+  "instructions",
+] }
 
-[lints]
-workspace = true
+[lints.clippy]
+pedantic = { level = "warn", priority = -1 }
```
wdcui, line 34: Are these choices consistent with other crates in LiteBox?

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -0,0 +1,152 @@
+//! I/O Port-based serial communication
+
+use core::{arch::asm, fmt};
+use spin::{Mutex, Once};
+
+// LVBS uses COM PORT 2 for printing out debug messages
+const COM_PORT_2: u16 = 0x2F8;
+
+// const DATA_OFFSET: u16 = 0;
```
wdcui: Minor: do we need to keep these constants that are commented off?

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        match byte {
+            0x20..=0x7e => outb(self.port, byte),
+            b'\n' => self.new_line(),
+            _ => outb(self.port, 0xfe),
+        }
+    }
+
+    fn new_line(&mut self) {
```
wdcui: Minor: why do we need this function? Is it okay to just call outb() twice in the match branch directly?

.github/workflows/ci.yml
```diff
@@ -55,6 +55,34 @@ jobs:
...
+      - run: pushd litebox_runner_lvbs; ../.github/tools/github_actions_run_cargo build; popd
+      - run: pushd litebox_platform_lvbs; ../.github/tools/github_actions_run_cargo build; popd
+      - run: |
+          pushd litebox_runner_lvbs; ../.github/tools/github_actions_run_cargo test --doc; popd
+          pushd litebox_platform_lvbs; ../.github/tools/github_actions_run_cargo test --doc; popd
+          # We need to run `cargo test --doc` separately because doc tests
+          # aren't included in nextest at the moment. See relevant discussion at
+          # https://github.com/nextest-rs/nextest/issues/16
+          #
+          # sanghle: I'll to change the directory structure of LVBS later to simplify CI
```
jaybosamiya-ms: `cargo` supports using `--manifest-path`, which allows you to do things iwthout the pushd/popd.

E.g.,:
```
./.github/tools/github_actions_run_cargo build --manifest-path=litebox_runner_lvbs/Cargo.toml
```

.github/workflows/ci.yml
```diff
@@ -55,6 +55,34 @@ jobs:
...
+    env:
+      RUSTFLAGS: -Dwarnings
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v4
+      - run: sudo apt install -y gcc-multilib
+      - name: Set up Rust
+        run: |
+          rustup toolchain install $(awk -F'"' '/channel/{print $2}' litebox_runner_lvbs/rust-toolchain.toml) --profile minimal --no-self-update --component rustfmt,clippy --target i686-unknown-linux-gnu
+          rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
```
jaybosamiya-ms: You can combine these I think:
```suggestion
          rustup toolchain install $(awk -F'"' '/channel/{print $2}' litebox_runner_lvbs/rust-toolchain.toml) --profile minimal --no-self-update --component rustfmt,clippy,rust-src --target i686-unknown-linux-gnu
```

Also, you are using a 64-bit toolchain, but installing a 32-bit target. Is that intentional?

You probably instead want to do this:
```suggestion
          rustup toolchain install $(awk -F'"' '/channel/{print $2}' litebox_runner_lvbs/rust-toolchain.toml) --profile minimal --no-self-update --component rustfmt,clippy,rust-src --target x86_64-unknown-linux-gnu
```

litebox_platform_lvbs/rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "nightly"
```
jaybosamiya-ms: I would _strongly_ suggest using a specifically tagged nightly, to ensure that there isn't a sudden breakage if nightly decides to update some things. For example:
```suggestion
channel = "nightly-2025-04-29"
```

.github/workflows/ci.yml
jaybosamiya-ms: Additionally, if you move away from the 32-bit variant, you can remove the `gcc-multilib` installation, which would help shave off ~30s of CI time.

.github/workflows/ci.yml
```diff
@@ -55,6 +55,34 @@ jobs:
...
+          curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
+      - uses: Swatinem/rust-cache@v2
+      - run: pushd litebox_runner_lvbs; ../.github/tools/github_actions_run_cargo build; popd
+      - run: pushd litebox_platform_lvbs; ../.github/tools/github_actions_run_cargo build; popd
+      - run: |
+          pushd litebox_runner_lvbs; ../.github/tools/github_actions_run_cargo test --doc; popd
+          pushd litebox_platform_lvbs; ../.github/tools/github_actions_run_cargo test --doc; popd
+          # We need to run `cargo test --doc` separately because doc tests
+          # aren't included in nextest at the moment. See relevant discussion at
+          # https://github.com/nextest-rs/nextest/issues/16
```
jaybosamiya-ms: You don't seem to be running nextest at all. If you are explicitly opting out of all non-doc tests, then you don't need to install nextest. Personally, I suggest keeping nextest running too (unless some of the tests are absolutely _unrunnable_ on non-correct target, in which case, there are ways to mark tests as skipped/ignored).

litebox_platform_lvbs/src/mm/tests.rs
```diff
@@ -85,26 +85,26 @@ impl super::MemoryProvider for MockKernel {
...
 
-#[test]
-fn test_buddy() {
-    let ptr = MockKernel::mem_allocate_pages(1);
-    assert!(ptr.is_some_and(|p| p as usize != 0));
-    unsafe {
-        MockKernel::mem_free_pages(ptr.unwrap(), 1);
-    }
-}
+// #[test]
```
jaybosamiya-ms: Do you intend to bring these tests back later or something? Why are they just directly commented out?

Note: you can either add `#[ignore]` which would just ignore a test, or you can remove the `#[test]` attribute which would stop it from being marked as a test.

jaybosamiya-ms:
Added comments mostly regarding CI and testing. I haven't reviewed the rest of the code.

litebox_platform_lvbs/src/host/lvbs_impl.rs
```diff
@@ -0,0 +1,112 @@
...
+    fn block_or_maybe_timeout(
+        _mutex: &core::sync::atomic::AtomicU32,
+        _val: u32,
+        _timeout: Option<core::time::Duration>,
+    ) -> Result<(), Errno> {
+        unimplemented!()
+    }
+
+    fn switch(result: u64) -> VtlCallParam {
+        // save VTL1 registers
```
wdcui: Minor: remove code that's commented out?

litebox_platform_lvbs/src/kernel_context.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub fn hv_vp_assist_page_as_u64(&self) -> u64 {
+        &raw const self.hv_vp_assist_page as u64
+    }
+
+    pub fn hv_hypercall_page_as_u64(&self) -> u64 {
+        get_hypercall_page_address()
+    }
+}
+
+// TODO: use heap later
```
wdcui: Minor: consider if it makes sense to log issues for some of the TODOs in this PR.

litebox_platform_lvbs/src/kernel_context.rs
```diff
@@ -0,0 +1,128 @@
...
+static mut PER_CORE_KERNEL_CONTEXT: [KernelContext; MAX_CORES] = [KernelContext {
+    hv_vp_assist_page: [0u8; PAGE_SIZE],
+    interrupt_stack: [0u8; INTERRUPT_STACK_SIZE],
+    _guard_page_0: [0u8; PAGE_SIZE],
+    kernel_stack: [0u8; KERNEL_STACK_SIZE],
+    _guard_page_1: [0u8; PAGE_SIZE],
+    tss: gdt::AlignedTss(TaskStateSegment::new()),
+    vtl0_state: VtlState {
```
wdcui, line 62: Do we need to worry about other registers?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -21,85 +21,31 @@ extern crate alloc;
 
 pub mod arch;
 pub mod host;
+pub mod kernel_context;
```
wdcui, line 24: Does this make these modules public outside the crate?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -392,6 +294,13 @@ pub trait HostInterface {
 
     /// For Debugging
     fn log(msg: &str);
+
+    /// Switch
+    ///
+    /// Switch enables a context switch from VTL1 kernel to VTL0 kernel while passing a value
+    /// through a CPU register. VTL1 kernel will execute the next instruction of `switch()`
+    /// when VTL0 kernel switches back to VTL1 kernel.
+    fn switch(result: u64) -> VtlCallParam;
```
wdcui: Just want to leave a question/issue for us to track: we will need to improve the definition of the host interface when we merge the existing platform code (if we ever do that) since some of the definitions are host specific such as VtlCallParam.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -0,0 +1,248 @@
...
+        asm!("mov r15, rax", in("rax") state.r15);
+        asm!("mov rdi, rax", in("rax") state.rdi);
+    }
+}
+
+#[expect(clippy::inline_always)]
+#[inline(always)]
+fn drop_vtl_state_from_stack() {
+    unsafe {
+        asm!("add rsp, {}", const mem::size_of::<VtlState>());
```
wdcui: Potential issue: does this size_of VtlState guarantee to match the actual usage on the stack (assuming some random changes are made to VtlState)?

litebox_runner_lvbs/src/main.rs
```diff
@@ -1,26 +1,52 @@
...
-use core::{arch::asm, panic::PanicInfo};
+use core::arch::asm;
+use litebox_platform_lvbs::{
+    arch::instrs::hlt_loop,
+    kernel_context::{get_core_id, get_per_core_kernel_context},
+    mshv::vtl_switch::vtl_switch_loop,
+    serial_println,
+};
 
+// shared? per-core?
```
wdcui: Remove these lines if you don't need them?

litebox_platform_lvbs/Cargo.toml
jaybosamiya-ms: To ensure consistency with lints, there is the `workspace = true` that most crates use. However, Sangho and I had a chat, and decided that if he wants to have a less-aggressive set of lints for LVBS (esp since it is running on nightly, where "pedantic" (enabled by the general workspace) would lead to too much burden), then he is welcome to allow a larger set of lints in the LVBS-related crates.

Updating here is one way to do it, the other way is to place a `#![allow(....)]` within the specific crate to allow more lints. I don't think either of them helps/hurts over the other.

litebox_platform_lvbs/src/kernel_context.rs
jaybosamiya-ms: When logging issues, feel free to also use the following to group related TODOs together using checkboxes if you think it'd become too much to open a whole bunch at once.
```md
- [ ] asd
- [ ] qwe
```
But yeah, feel free to use it as a TODO tracker if you find it easier to keep that around.

litebox_platform_lvbs/Cargo.toml
sangho2: VSCode auto changed this based on its rule. I can revert it.

.github/workflows/ci.yml
sangho2: I'm testing CI in a different branch. I'll update the script once everything works fine.

.github/workflows/ci.yml
sangho2: I'll revive all tests.

litebox_platform_lvbs/Cargo.toml
sangho2: This lint is copied and pasted from the main `Cargo.toml`. As Jay mentioned, LVBS building for now is not well integrated with the main LiteBox due to stable versus nightly issues.

litebox_platform_lvbs/src/arch/x86/ioport.rs
sangho2: We don't need them. I'll clean up all commented-out constants.

litebox_platform_lvbs/src/arch/x86/ioport.rs
sangho2: We don't particularly need that function. I'll simplify this code.

litebox_platform_lvbs/src/kernel_context.rs
sangho2: Our VTL1 kernel for now does not use other registers including SSE and AVX (enforced by `-mmx,-sse,+soft_float`). In the future, we might need to save/restore some other registers as well if we plan to use accelerated cryptographic instructions.

litebox_platform_lvbs/src/lib.rs
sangho2: I'll remove all unnecessary `pub`s.

litebox_platform_lvbs/src/lib.rs
sangho2: Let me revert this change. For now, the integration between LiteBox host/platform and VTL1 kernel is unclear and not yet discussed well.

litebox_platform_lvbs/src/mm/tests.rs
sangho2: Will revert this change.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: VtlState is C struct consisting of u64's. Thus, I think its size is equal to the total size of registers in the stack. To avoid any potential issues, I'll replace it with 8 * NUM_REGS.

wdcui: I understand that it is fine for now because the structure is 8 byte aligned. This assumption is implicit, so it might be better to make it explicit either in the code or the comment.

.github/workflows/ci.yml
sangho2: @jaybosamiya-ms I've made a minor change to `github_actions_run_cargo` to get `cargo nightly` through an environment variable. Let me know whether this change is reasonable.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: Done.

litebox_platform_lvbs/src/lib.rs
sangho2: Removed some pub mods. For now, the LVBS runner needs `kernel_context` to get per-core kernel stack pointer and `mshv` to initialize hypercall MSRs/VPs. Perhaps it's better to separate these init functions.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -0,0 +1,233 @@
...
+#[inline(always)]
+fn load_vtl1_state() {
+    let kernel_context = get_per_core_kernel_context();
+    let vtl1_state: VtlState = kernel_context.vtl1_state;
+    pop_vtl_state(&vtl1_state);
+}
+
+/// VTL switch loop (entry point)
+pub fn vtl_switch_loop() -> ! {
+    save_vtl0_state();
```
CvvT, line 198: This is called by `kernel_main`, which performs some initialization before calling this function. At this pointer, does all the registers still have the state of vtl0?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -0,0 +1,233 @@
...
+
+#[expect(clippy::inline_always)]
+#[inline(always)]
+fn save_vtl0_state() {
+    let vtl0_state = push_vtl_state_to_stack();
+    let kernel_context = get_per_core_kernel_context();
+    kernel_context
+        .vtl0_state
+        .clone_from(unsafe { &*vtl0_state });
+    drop_vtl_state_from_stack();
```
CvvT: It feels a little bit off to separate `push_vtl_state_to_stack` and `drop_vtl_state_from_stack` as the stack pointer may change in between. I guess in this case it does not use more stack in the current frame, so it won't break anything.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: This `save_vtl0_state()` is like a dummy function call to populate `vtl0_state` with reasonable register values for `load_vtl0_state()`. During this initialization, VTL0 doesn't assume that VTL1 would well save its registers (due to unknown complexity), so VTL0 has saved its registers before triggering VTL switch. I'll leave a comment on this.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: true. Let me double check this. Previously, I only dealt with a static variable between push and drop (for a single core case). Multi-core support introduces some uncertainty (and this trick might not work if we use heap for kernel context).

sangho2: added `assert` to confirm this.

.github/workflows/ci.yml
```diff
@@ -55,6 +55,49 @@ jobs:
...
+      RUSTFLAGS: -Dwarnings
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v4
+      - name: Set up Rust
+        run: |
+          RUST_CHANNEL=$(awk -F'"' '/channel/{print $2}' litebox_runner_lvbs/rust-toolchain.toml)
+          rustup toolchain install ${RUST_CHANNEL} --profile minimal --no-self-update --component rustfmt,clippy --target x86_64-unknown-none
+          rustup component add rust-src --toolchain ${RUST_CHANNEL}-x86_64-unknown-linux-gnu
+          echo "CARGO_BIN_ENV=cargo +${RUST_CHANNEL}" >> $GITHUB_ENV
```
jaybosamiya-ms: Rather than editing `.github/tools/github_actions_run_cargo`, the following should be sufficient:
```suggestion
          rustup default ${RUST_CHANNEL}
```

.github/workflows/ci.yml
jaybosamiya-ms: Thanks, but I'm a little bit surprised by needing to explicitly do `cargo +nightly`, even when toolchain files are around. I guess this showed up because you switched to the `--manifest-path` version?  Anyways, I've made a suggestion that should allow you to drop the changes from `github_actions_run_cargo` entirely. If that works, then please use that. If not, we can use the current version.

.github/workflows/ci.yml
sangho2: turns out that we need to run `rustup override set ${RUST_CHANNEL}` to use nightly instead of stable (`rust-toolchain.toml` at the project root overrides the system default). I'll rollback `github_actions_run_cargo`.

PR #103, opened 2025-05-01
Patch rewritten binary to set syscall callback

CvvT:
This PR adds support to run 32bit executable with some caveats. Because the binary is patched to set the syscall callback pointer during loading time, it doesn't work with dependencies loaded by the ld loader. For now, I just assume the binary is statically linked.

Another thing I noticed was that 32bit does not support `call [eip + off]` (cannot use eip relative addressing). The workaround I have is to simulate it with the following sequence (doubling the size though):
```
    push eax
    call next
next:
    pop eax
    call [eax + off]
```

I have written a integrate test for that at: https://github.com/MSRSSP/litebox/blob/9f29004a3163034601bfde8ae70b02150599460b/litebox_shim_linux/tests/exec.rs#L104

To run it with commands, follow the following steps:
```
# copy the C source code of the test from `test_syscall_rewriter` to test.c
gcc test.c -o test -static -nostdlib -m32
# 6363377504867338572 is a magic number to locate the placeholder where we should set syscall callback
cargo run -p litebox_syscall_rewriter -- --trampoline-addr 6363377504867338572 test
cargo run -p litebox_runner_linux_userland --target=i686-unknown-linux-gnu -- test.hooked
```

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -12,9 +12,7 @@ use std::time::Duration;
 use litebox::platform::ImmediatelyWokenUp;
 use litebox::platform::UnblockedOrTimedOut;
 use litebox::platform::page_mgmt::MemoryRegionPermissions;
-use litebox_common_linux::ProtFlags;
 use litebox_common_linux::PunchthroughSyscall;
-use litebox_common_linux::errno::Errno;
 
 mod syscall_intercept;
 
@@ -104,21 +102,6 @@ impl LinuxUserland {
         );
         // TODO: have better signature and registration of the syscall handler.
         syscall_intercept::init_sys_intercept(syscall_handler);
-    }
-
-    /************************ Punchthroughs ******************************/
-
-    pub fn mprotect(&self, addr: usize, len: usize, prot: ProtFlags) -> Result<(), Errno> {
-        let ret = unsafe { libc::syscall(libc::SYS_mprotect, addr, len, prot.bits()) };
-        if ret == 0 {
-            Ok(())
-        } else {
-            let errno = latest_errno();
-            let Ok(errno) = Errno::try_from(errno) else {
-                unreachable!();
-            };
-            Err(errno)
-        }
     }
 }
 
--- a/litebox_shim_linux/src/loader/elf.rs
+++ b/litebox_shim_linux/src/loader/elf.rs
@@ -192,6 +192,7 @@ type Ehdr = elf::file::Elf32_Ehdr;
 type Shdr = elf::section::Elf64_Shdr;
 #[cfg(target_arch = "x86")]
 type Shdr = elf::section::Elf32_Shdr;
+
 /// Get the syscall callback placeholder address from the ELF file if it is modified by our syscall
 /// rewriter. The placeholder initially has a magic number that is going to be replaced by the actual
 /// syscall callback.
@@ -225,6 +226,7 @@ fn get_syscall_callback_placeholder(object: &mut ElfFile) -> Option<NonNull<usiz
         )
         .ok()?;
     let magic_number = u64::from_ne_bytes(placeholder);
+    // TODO: check section name instead of magic number
     if magic_number != super::REWRITER_MAGIC_NUMBER {
         return None;
     }
@@ -273,19 +275,17 @@ impl ElfLoader {
                 .map_err(ElfLoaderError::LoaderError)?;
             if let Some(placeholder) = placeholder {
                 // mprotect the memory to make it writable
-                let protflags = litebox_common_linux::ProtFlags::PROT_READ
-                    | litebox_common_linux::ProtFlags::PROT_WRITE;
-                // `mmap` requires the address to be page-aligned
+                let pm = litebox_page_manager();
+                // `mprotect` requires the address to be page-aligned
                 let start_addr = placeholder.as_ptr() as usize & !(PAGE_SIZE - 1);
-                litebox_platform_multiplex::platform()
-                    .mprotect(start_addr, 0x1000, protflags)
-                    .unwrap();
+                let ptr = unsafe {
+                    core::mem::transmute::<*mut u8, crate::MutPtr<u8>>(start_addr as *mut u8)
+                };
+                unsafe { pm.make_pages_writable(ptr, 0x1000) }
+                    .expect("failed to make pages writable");
                 unsafe { placeholder.write(crate::syscall_callback as usize) };
-                let protflags = litebox_common_linux::ProtFlags::PROT_READ
-                    | litebox_common_linux::ProtFlags::PROT_EXEC;
-                litebox_platform_multiplex::platform()
-                    .mprotect(start_addr, 0x1000, protflags)
-                    .unwrap();
+                unsafe { pm.make_pages_executable(ptr, 0x1000) }
+                    .expect("failed to make pages executable");
             }
             elf
         };
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -300,6 +300,8 @@ fn hook_syscalls_in_section(
             trampoline_data.extend_from_slice(&[0xFF, 0x90]); // CALL [EAX + offset]
             let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() - 3);
             trampoline_data.extend_from_slice(&disp32.to_le_bytes());
+            // Note we skip `POP EAX` here as it is done by the callback `syscall_callback`
+            // from litebox_shim_linux/src/lib.rs, which helps reduce the size of the trampoline.
         }
 
         // Add jmp back to original after syscall
--- a/litebox_shim_linux/tests/common/mod.rs
+++ b/litebox_shim_linux/tests/common/mod.rs
@@ -77,9 +77,9 @@ pub fn init_platform(enable_systrap: bool) {
     install_dir("/lib/x86_64-linux-gnu");
 }
 
-pub fn compile(input: &str, output: &std::path::Path, exec_or_lib: bool, nolibc: bool) {
-    // Compile the hello.c file to an executable
-    let mut args = vec!["-o", output.to_str().unwrap(), input];
+/// Compile C code into an executable
+pub fn compile(input: &str, output: &str, exec_or_lib: bool, nolibc: bool) {
+    let mut args = vec!["-o", output, input];
     if exec_or_lib {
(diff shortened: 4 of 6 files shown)


CvvT:
@jaybosamiya-ms I'd like to discuss how to implement `punchthroughs`. Briefly, I would like to have the shim layer to define an enum for `Punchthroughs` and implement `PunchthroughToken` in the platform.

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -217,9 +264,29 @@ impl ElfLoader {
...
+            let elf = loader
+                .easy_load(object)
+                .map_err(ElfLoaderError::LoaderError)?;
+            if let Some(placeholder) = placeholder {
+                // mprotect the memory to make it writable
+                let protflags = litebox_common_linux::ProtFlags::PROT_READ
+                    | litebox_common_linux::ProtFlags::PROT_WRITE;
+                let start_addr = placeholder.as_ptr() as usize & !(PAGE_SIZE - 1);
+                litebox_platform_multiplex::platform()
+                    .mprotect(start_addr, 0x1000, protflags)
```
wdcui: Minor: add a comment or check start_addr is page aligned?

CvvT:
Weidong pointed out that probably we shouldn't rely on `ld` to load dependencies. The crate we use seems to have the basic functionality to support that. I just to need to search the paths of the dependencies and load them.

CvvT:
Per the discussion with Jay, we need to define the `Punchthroughs` in `litebox_common_linux` and implement `PunchthroughToken` in the platform.

jaybosamiya-ms:
Quickly summarizing IRL discussion: this will be merged _after_ #110 (i.e., the PR for #105).

CvvT:
Update:  rebased on #104 to remove `exit` and `exit_group` punchthroughs. Once #107 is complete, I can also remove that punchthrough.

litebox_shim_linux/src/loader/elf.rs
CvvT: Sure, added comment. Thanks!

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -103,6 +105,21 @@ impl LinuxUserland {
...
+        if ret == 0 {
+            Ok(())
+        } else {
+            let errno = latest_errno();
+            let Ok(errno) = Errno::try_from(errno) else {
+                unreachable!();
+            };
+            Err(errno)
+        }
+    }
```
jaybosamiya-ms: I know you've already mentioned you'll remove this after the other PR is merged, but just adding this comment to more easily track it.

litebox_shim_linux/src/lib.rs
```diff
@@ -429,3 +430,166 @@ pub fn syscall_entry(request: SyscallRequest<Platform>) -> isize {
...
+        ::syscalls::Sysno::write => SyscallRequest::Write {
+            fd: syscall_args[0].reinterpret_as_signed().truncate(),
+            buf: unsafe { core::mem::transmute::<usize, ConstPtr<u8>>(syscall_args[1]) },
+            count: syscall_args[2],
+        },
+        ::syscalls::Sysno::exit | ::syscalls::Sysno::exit_group => {
+            litebox_platform_multiplex::platform()
+                .exit(syscall_args[0].reinterpret_as_signed().truncate())
+        }
+        _ => todo!("syscall {sysno} not implemented"),
```
jaybosamiya-ms, line 591: As I understand it, this particular portion is slowly moving things over from the systrap `syscall_dispatcher` to here, yes? If so, I am glad this is not done in one go in this PR, as that full move should indeed be a separate PR.

This was roughly the "future PRs" thing I was mentioning in #110, interestingly enough :)

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -184,6 +184,53 @@ pub struct ElfLoadInfo {
...
 
+#[cfg(target_arch = "x86_64")]
+type Ehdr = elf::file::Elf64_Ehdr;
+#[cfg(target_arch = "x86")]
+type Ehdr = elf::file::Elf32_Ehdr;
+#[cfg(target_arch = "x86_64")]
+type Shdr = elf::section::Elf64_Shdr;
+#[cfg(target_arch = "x86")]
+type Shdr = elf::section::Elf32_Shdr;
+/// Get the syscall callback placeholder address from the ELF file if it is modified by our syscall
```
jaybosamiya-ms: An extra newline, just to _visually_ clarify that the `get_syscall_callback_placeholder` is not restricted by the cfg.  I know the Rust parser would handle it fine, but it still helps distinguish things visually.
```suggestion
#[cfg(target_arch = "x86")]
type Shdr = elf::section::Elf32_Shdr;

/// Get the syscall callback placeholder address from the ELF file if it is modified by our syscall
```

litebox_shim_linux/Cargo.toml
```diff
@@ -9,13 +9,15 @@ litebox = { path = "../litebox/", version = "0.1.0" }
 litebox_common_linux = { path = "../litebox_common_linux/", version = "0.1.0" }
 litebox_platform_multiplex = { path = "../litebox_platform_multiplex/", version = "0.1.0" }
 once_cell = { version = "1.20.2", default-features = false, features = ["alloc", "race"] }
+elf = { version = "0.7.4", default-features = false }
```
jaybosamiya-ms, line 12: We are already using `object` in a different part of the codebase, and that supports the necessary ELF parsing. Maybe we should use that, rather than pulling in a different crate?

litebox_shim_linux/src/loader/mod.rs
```diff
@@ -11,3 +11,7 @@ pub fn load_program(
 ) -> Result<elf::ElfLoadInfo, elf::ElfLoaderError> {
     elf::ElfLoader::load(path, argv, envp)
 }
+
+/// The magic number used to identify the LiteBox rewriter and where we should
+/// update the syscall callback pointer.
+pub const REWRITER_MAGIC_NUMBER: u64 = u64::from_le_bytes(*b"LITE BOX");
```
jaybosamiya-ms, line 17: Rather than doing this via a _new_ magic number, we should be checking the section name instead:
https://github.com/MSRSSP/litebox/blob/37db22002fbb7870a744e99c3bab3e75efea7e37/litebox_syscall_rewriter/src/lib.rs#L59-L67

Pulling in the `linux_syscall_rewriter` as a dependency is a bit too much (and problematic, due to `std` reasons) just to get this constant, but copying this single constant string into this crate is perfectly fine. I think we can clean it up so that things stay in sync, but that can be a separate PR if need be.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -288,9 +288,19 @@ fn hook_syscalls_in_section(
...
+            let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() + 4);
+            trampoline_data.extend_from_slice(&disp32.to_le_bytes());
+        } else {
+            // For 32-bit, use a different approach to simulate `call [rip + disp32]`
+            trampoline_data.push(0x50); // PUSH EAX
+            trampoline_data.extend_from_slice(&[0xE8, 0x0, 0x0, 0x0, 0x0]); // CALL next instruction
+            trampoline_data.push(0x58); // POP EAX (effectively store IP in EAX)
+            trampoline_data.extend_from_slice(&[0xFF, 0x90]); // CALL [EAX + offset]
+            let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() - 3);
+            trampoline_data.extend_from_slice(&disp32.to_le_bytes());
```
jaybosamiya-ms, line 302: I'm not sure I understand why the initial `push eax` is done. Maybe I need to re-think this a bit more, but just making a note here right now.

jaybosamiya-ms: Chatted with @CvvT IRL, the `push eax` is handled by the first couple of instructions in `syscall_dispatcher`; a comment here (and there) linking the two would help.

That code location: https://github.com/MSRSSP/litebox/blob/6d032120df224891d34ebc17e177cfc58ba1d3a4/litebox_shim_linux/src/lib.rs#L518-L520

litebox_platform_linux_userland/src/lib.rs
CvvT: Removed.

litebox_shim_linux/Cargo.toml
CvvT: The reason I used this crate was because `elf_loader` depends on this one. We can decide later which one we should use.

litebox_shim_linux/src/lib.rs
CvvT: You're right! Eventually I would replace systrap `syscall_dispatcher` with this one.

litebox_shim_linux/src/loader/elf.rs
CvvT: Fixed!

litebox_shim_linux/src/loader/mod.rs
CvvT: Yes, I was aware of this. Using magic number is just an easy hack to make it work. Checking section name requires some more work (e.g., reading symbol table). I added a comment here and will create a separate issue for this.

litebox_syscall_rewriter/src/lib.rs
CvvT: Thanks for pointing it out. I have added some comment here.

CvvT: One sidenote:
```
Opcode | Mnemonic | Description
E8 cw | CALL rel16 | Call near, relative, displacement relative to next instruction
E8 cd | CALL rel32 | Call near, relative, displacement relative to next instruction
```
It seems that `CALL rel16` is for 16bit mode so we have to use the longer one `CALL rel32`.

litebox_shim_linux/Cargo.toml
jaybosamiya-ms: Ah, fair point.

litebox_shim_linux/src/loader/mod.rs
jaybosamiya-ms: Sounds good, thanks for opening a separate issue for tracking this!

litebox_syscall_rewriter/src/lib.rs
jaybosamiya-ms: Makes sense, thanks for checking up on that!

PR #107, opened 2025-05-02
Sys mprotect

CvvT:
Adding support for syscall `mprotect`. Similar to the interface about page allocation, it only allows certain combinations of protection flags.

With this change, we could get rid of the punchthrough I added in #103 

https://github.com/MSRSSP/litebox/blob/9f29004a3163034601bfde8ae70b02150599460b/litebox_shim_linux/src/loader/elf.rs#L275-L287

### DIFF
--- a/litebox/src/mm/mod.rs
+++ b/litebox/src/mm/mod.rs
@@ -194,6 +194,20 @@ where
         unsafe { vmem.remove_mapping(range) }
     }
 
+    /// Internal common function used by `make_pages_*` to change page permissions.
+    fn change_page_permissions(
+        &self,
+        ptr: Platform::RawMutPointer<u8>,
+        len: usize,
+        new_permissions: MemoryRegionPermissions,
+    ) -> Result<(), VmemProtectError> {
+        let mut vmem = self.vmem.write();
+        let start = ptr.as_usize();
+        let range = PageRange::new(start, start + len)
+            .ok_or(VmemProtectError::InvalidRange(start..start + len))?;
+        unsafe { vmem.protect_mapping(range, new_permissions) }
+    }
+
     /// Make pages readable and writable.
     ///
     /// # Safety
@@ -204,16 +218,11 @@ where
         ptr: Platform::RawMutPointer<u8>,
         len: usize,
     ) -> Result<(), VmemProtectError> {
-        let mut vmem = self.vmem.write();
-        let start = ptr.as_usize();
-        let range = PageRange::new(start, start + len)
-            .ok_or(VmemProtectError::InvalidRange(start..start + len))?;
-        unsafe {
-            vmem.protect_mapping(
-                range,
-                MemoryRegionPermissions::READ | MemoryRegionPermissions::WRITE,
-            )
-        }
+        self.change_page_permissions(
+            ptr,
+            len,
+            MemoryRegionPermissions::READ | MemoryRegionPermissions::WRITE,
+        )
     }
 
     /// Make pages readable and executable.
@@ -226,16 +235,11 @@ where
         ptr: Platform::RawMutPointer<u8>,
         len: usize,
     ) -> Result<(), VmemProtectError> {
-        let mut vmem = self.vmem.write();
-        let start = ptr.as_usize();
-        let range = PageRange::new(start, start + len)
-            .ok_or(VmemProtectError::InvalidRange(start..start + len))?;
-        unsafe {
-            vmem.protect_mapping(
-                range,
-                MemoryRegionPermissions::READ | MemoryRegionPermissions::EXEC,
-            )
-        }
+        self.change_page_permissions(
+            ptr,
+            len,
+            MemoryRegionPermissions::READ | MemoryRegionPermissions::EXEC,
+        )
     }
 
     /// Make pages readable only.
@@ -248,11 +252,20 @@ where
         ptr: Platform::RawMutPointer<u8>,
         len: usize,
     ) -> Result<(), VmemProtectError> {
-        let mut vmem = self.vmem.write();
-        let start = ptr.as_usize();
-        let range = PageRange::new(start, start + len)
-            .ok_or(VmemProtectError::InvalidRange(start..start + len))?;
-        unsafe { vmem.protect_mapping(range, MemoryRegionPermissions::READ) }
+        self.change_page_permissions(ptr, len, MemoryRegionPermissions::READ)
+    }
+
+    /// Make pages inaccessible.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure there is no concurrent access to the memory region.
+    pub unsafe fn make_pages_inaccessible(
+        &self,
+        ptr: Platform::RawMutPointer<u8>,
+        len: usize,
+    ) -> Result<(), VmemProtectError> {
+        self.change_page_permissions(ptr, len, MemoryRegionPermissions::empty())
     }
 
     /// Returns all mappings in a vector.
--- a/litebox_shim_linux/src/syscalls/mm.rs
+++ b/litebox_shim_linux/src/syscalls/mm.rs
@@ -186,6 +186,7 @@ pub(crate) fn sys_mprotect(
         ProtFlags::PROT_READ_EXEC => unsafe { pm.make_pages_executable(addr, len) },
         ProtFlags::PROT_READ_WRITE => unsafe { pm.make_pages_writable(addr, len) },
         ProtFlags::PROT_READ => unsafe { pm.make_pages_readable(addr, len) },
+        ProtFlags::PROT_NONE => unsafe { pm.make_pages_inaccessible(addr, len) },
         _ => todo!("Unsupported prot flags {:?}", prot),
     }
     .map_err(Errno::from)


litebox/src/mm/linux.rs
```diff
@@ -475,6 +475,8 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
     /// `op` is a callback for caller to initialize the created pages.
     ///
     /// `before_flags` and `after_flags` are the flags to set before and after the call to `op`.
+    /// Note `before_flags` needs to set `VM_MAY_ACCESS_FLAGS` bits and `after_flags` is only used
+    /// when `before_flags` is different from `after_flags` on those `VM_ACCESS_FLAGS` bits.
```
jaybosamiya-ms: I am confused by these semantics, can you explain what this is saying? Is it saying that `before_flags` must _always_ set all bits of `VM_MAY_ACCESS_FLAGS`? 

Also, "only used" is confusing. Maybe phrasing it as "Additionally, protections are not updated after `op` (i.e., `after_flags` are ignored) if the `VM_ACCESS_FLAGS` bits are identical between `before_flags` and `after_flags`".

jaybosamiya-ms: Is the skipping portion as a performance optimization? If so, you could mention "as a performance optimization, ...".

Also, do we expect that `after_flags` will ever differ from `before_flags` on any bits other than the `VM_ACCESS_FLAGS` bits? If so, we should throw a debug message or something, just so that we don't miss out on some edge case and things suddenly break.

litebox/src/mm/mod.rs
```diff
@@ -194,6 +199,60 @@ where
         unsafe { vmem.remove_mapping(range) }
     }
 
+    /// Make pages readable and writable.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure the memory region is used only for reading and writing
+    /// but not executing if this call succeeds.
```
jaybosamiya-ms: I am not sure this is a helpful safety specification. The caller cannot actually ensure things to the end of time, correct?

I think the main thing that the caller needs to ensure is that the `ptr` (up to len) is valid, and that it is safe to update the permissions (i.e., nothing else is concurrently writing), yes?

litebox/src/mm/mod.rs
```diff
@@ -194,6 +199,60 @@ where
...
+            .ok_or(VmemProtectError::InvalidRange(start..start + len))?;
+        unsafe { vmem.protect_mapping(range, VmFlags::VM_READ | VmFlags::VM_WRITE) }
+    }
+
+    /// Make pages readable and executable.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure the memory region is used only for reading and executing
+    /// but not writing if this call succeeds.
```
jaybosamiya-ms: Similar

litebox/src/mm/mod.rs
```diff
@@ -194,6 +199,60 @@ where
...
+            .ok_or(VmemProtectError::InvalidRange(start..start + len))?;
+        unsafe { vmem.protect_mapping(range, VmFlags::VM_READ | VmFlags::VM_EXEC) }
+    }
+
+    /// Make pages readable only.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure the memory region is used only for reading
+    /// but not writing or executing if this call succeeds.
```
jaybosamiya-ms: Similar

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -583,9 +583,20 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Li
...
+                    range.start,
+                    range.len(),
+                    prot_flags(new_permissions).bits(),
+                    // This is to ensure it won't be intercepted by Seccomp if enabled.
+                    syscall_intercept::systrap::SYSCALL_ARG_MAGIC,
+                )
+            },
+            0,
+            "mprotect failed",
+        );
```
jaybosamiya-ms: I generally am weary of side-effectful code that have a wrapping macro around them like an `assert`. We don't expect to change the meaning of `assert_eq` anytime soon, but it can lead to surprising behavior if (at some point) we decide that assertions could be optimized out in release mode. Thus, I find it best to keep side-effectful code outside of macros, such as like this. Additionally, `assert_eq` will already include the two sides of the equality in the panic message, thus "mprotect failed" does not add much to the error message, so I've dropped it in the version below.
```suggestion
        let mprotect_ret = 
            unsafe {
                libc::syscall(
                    libc::SYS_mprotect,
                    range.start,
                    range.len(),
                    prot_flags(new_permissions).bits(),
                    // This is to ensure it won't be intercepted by Seccomp if enabled.
                    syscall_intercept::systrap::SYSCALL_ARG_MAGIC,
                )
            }
        };
        assert_eq!(mprotect_ret, 0);
```

litebox/src/mm/mod.rs
CvvT: You're right. Fixed the comments. If the `ptr` is invalid. an error code would be returned.

litebox/src/mm/linux.rs
CvvT: I agree it is confusing. `VM_MAY_ACCESS_FLAGS` is the max permission this memory region should have, which would be checked against when calling `protect_mapping`. Probably, I should follow `mmap` in Linux where by default the max permission is read/write/execute so that we could simplify the argument here: define a separate flag only containing READ/WRITE/EXECUTE instead of using `VmFlags`.

litebox/src/mm/mod.rs
CvvT: Fixed

litebox/src/mm/mod.rs
CvvT: Fixed

litebox_platform_linux_userland/src/lib.rs
CvvT: I see. Fixed. Thanks!

litebox/src/mm/linux.rs
CvvT: Edit: Now `before_flags` and `after_flags` uses ~~`ProtFlags`~~ `MemoryRegionPermissions`. ~~same as the protection flags used by `mmap`~~. For now, the max permissions by default are READ, WRITE, and EXECUTE, and hence the caller doesn't need to set it. If we find we need to allow caller to specify it, we could add a separate argument in the future.

jaybosamiya-ms: Thanks, that does help simplify things. I think the current version is good, and indeed, if we do need to handle it in the future, we can update it.

litebox/src/mm/linux.rs
```diff
@@ -605,6 +642,8 @@ pub enum VmemMoveError {
 /// Error for protecting mappings
 #[derive(Error, Debug)]
 pub enum VmemProtectError {
+    #[error("the range {0:?} is not aligned")]
+    MisAligned(Range<usize>),
```
jaybosamiya-ms: Just for consistency with naming that we've used elsewhere, this is probably best called `Unaligned` instead.
```suggestion
    #[error("the range {0:?} is not aligned")]
    Unaligned(Range<usize>),
```

Example other place we've used `Unaligned`: https://github.com/MSRSSP/litebox/blob/37db22002fbb7870a744e99c3bab3e75efea7e37/litebox/src/platform/page_mgmt.rs#L75-L76

CvvT: Sure, I also fixed other places where I used `MisAligned`.

litebox/src/mm/mod.rs
```diff
@@ -190,10 +190,84 @@ where
...
+    ///
+    /// The caller must ensure there is no concurrent access to the memory region.
+    pub unsafe fn make_pages_inaccessible(
+        &self,
+        ptr: Platform::RawMutPointer<u8>,
+        len: usize,
+    ) -> Result<(), VmemProtectError> {
+        self.change_page_permissions(ptr, len, MemoryRegionPermissions::empty())
+    }
+
```
CvvT, line 270: I also made one more change: adding `make_pages_inaccessible` similar to other `make_pages_*`, which is required when I tested against Nodejs.

jaybosamiya-ms: Thanks, seems reasonable to me!

CvvT:
Looks like there are cases where application requires readable, writable, and executable pages:

https://github.com/nodejs/node/blob/9148e965b41dfb322a9e1f38dfaf11ddab0f042c/deps/v8/src/heap/code-range.cc#L242

If we support this, effectively we allow application to change the memory permission arbitrarily, and we only need a single API `change_page_permissions` instead of multiple `make_pages_*`.

jaybosamiya-ms:
I don't think v8 _needs_ RWX pages (specifically, unless I'm mistaken, they have modes that prevent usage of RWX pages, since RWX pages make exploitation easier once a vuln is found). We should try to see if we can trigger it to run in one of those modes.

I do agree that if they don't have this, then we are forced to have the more general `change_page_permissions`, but yeah, we should chat.

PR #110, opened 2025-05-05
De-parameterize Linux punchthroughs

### LINKED ISSUE #105: De-parameterize the LinuxUserland crate
Now that we have a linux-common crate, we don't need the punchthrough to be parametric. This will help simplify some things.

Related: https://github.com/MSRSSP/litebox/pull/103#issuecomment-2845365667

jaybosamiya-ms:
Closes #105

This PR removes the generalized parametricity that we were keeping for punchthrough for Linux-y LiteBox systems. The old reason for keeping these was because we had not yet finalized `litebox_common_linux` or what we actually needed from punchthroughs, or how the shim layer would look like. At this point, we have a relatively good design for each of them, thus we don't need the parametricity, and can instead put the shared elements together in a nicer form, so that there is a clearer indication of what needs to be punched through in order to support a Linux-y north interface.

Future PR(s) are going to do even more cleanup to pull out shared elements across the different Linux-y platforms.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -467,5 +467,40 @@ pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
         flags: litebox::fs::OFlags,
     },
     /// A sentinel that is expected to be "handled" by trivially returning its value.
-    Ret(i64),
+    Ret(isize),
+}
+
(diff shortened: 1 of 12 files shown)


jaybosamiya-ms:
All semver checks notes by the bot here are the expected ones.

jaybosamiya-ms:
Thanks, rebased to fix a merge conflict w/ `main`, will merge into `main` when CI succeeds.

Issue #66, opened 2025-03-31
Support swapping per-thread fs register value


CvvT:
Briefly noting an idea we discussed IRL: since our main usage for storing fs register value is to support a swap, we can maintain a (global, rwlocked) hashmap that tells us the swaps. We would need to set up a state-machine-like thing to leave values "unlinked" until we find the link (during libc initialization), but generally speaking, the idea seems feasible.

@CvvT is also exploring some pthread internals to see if they might be helpful as an alternative, just in case that ends up simpler, but yeah, we have ways to solve this :)

_Originally posted by @jaybosamiya-ms in https://github.com/MSRSSP/litebox/issues/56#issuecomment-2759672534_

PR #111, opened 2025-05-05
Litebox allocator

CvvT:
This PR moves the allocator from `litebox_platform_linux_kernel` to `litebox` because I believe `litebox_platform_linux_userland` also needs to use it.

The primary reason for `litebox_platform_linux_userland` to use it is to address the issue mentioned in #66. I realized that the allocator from libc also relies on `fs` registers to maintain per-thread metadata for better performance. Instead of maintaining two TLS (thread local storage) used by two libc instances within the same address space, I feel it's cleaner to not let other dependencies (i.e., libc in this case) uses `fs` registers at all.

I also tried to intercept `clone` syscall and create a separate `TLS` for litebox to use, however, it's not as easy as I thought. It seems that `libc` allocates `TLS` together with the `Thread` struct. That's why I have seen cases where libc accesses some data using `fs` register with a negative offset. It's unclear to me whether maintaining two `TLS` for two `libc` instances would cause any issues (e.g., how to clean up both when a thread exits).

Our current implement for allocator is slow, but I expect we would optimize this for linux kernel platform anyway. Another benefit is we do not need to swtich fs register anymore.

### DIFF
--- a/litebox/src/mm/allocator.rs
+++ b/litebox/src/mm/allocator.rs
@@ -16,7 +16,9 @@ use spin::mutex::SpinMutex;
 pub trait MemoryProvider {
     /// For page allocation from host.
     ///
-    /// Note this is only called when the buddy allocator is out of memory.
+    /// Note this is only called when the allocator is out of memory.
+    /// To add memory to the allocator at any time (e.g., initialize the allocator with
+    /// pre-allocated fixed-size memory), use [`SafeZoneAllocator::fill_pages`].
     ///
     /// It can return more than requested size. On success, it returns the start address
     /// and the size of the allocated memory.


litebox/Cargo.toml
```diff
@@ -10,13 +10,15 @@ either = { version = "1.13.0", default-features = false }
...
+spin = { version = "0.9.8", default-features = false, features = ["spin_mutex"] }
 tar-no-std = { version = "0.3.3", default-features = false, features = ["alloc"] }
 thiserror = { version = "2.0.6", default-features = false }
 rangemap = { version = "1.5.1", features = ["const_fn"] }
+buddy_system_allocator = { version = "0.11.0", default-features = false, features = ["use_spin"] }
+slabmalloc = { git = "https://github.com/gz/rust-slabmalloc.git", rev = "19480b2e82704210abafe575fb9699184c1be110" }
 
 [features]
-lock_tracing = ["dep:arrayvec", "dep:spin", "spin?/mutex", "spin?/spin_mutex"]
+lock_tracing = ["dep:arrayvec", "spin/mutex", "spin/spin_mutex"]
```
jaybosamiya-ms: With `spin_mutex` being enabled by default above, this doesn't need to explicitly re-depend on it here.
```suggestion
lock_tracing = ["dep:arrayvec", "spin/mutex"]
```

litebox/Cargo.toml
```diff
@@ -10,13 +10,15 @@ either = { version = "1.13.0", default-features = false }
 hashbrown = "0.15.2"
 smallvec = "1.13.2"
 smoltcp = { version = "0.12.0", default-features = false, features = ["log", "proto-ipv4", "medium-ip", "socket", "socket-tcp", "socket-udp", "socket-icmp", "socket-raw", "alloc"] }
-spin = { version = "0.9.8", default-features = false, optional = true }
+spin = { version = "0.9.8", default-features = false, features = ["spin_mutex"] }
 tar-no-std = { version = "0.3.3", default-features = false, features = ["alloc"] }
 thiserror = { version = "2.0.6", default-features = false }
 rangemap = { version = "1.5.1", features = ["const_fn"] }
+buddy_system_allocator = { version = "0.11.0", default-features = false, features = ["use_spin"] }
+slabmalloc = { git = "https://github.com/gz/rust-slabmalloc.git", rev = "19480b2e82704210abafe575fb9699184c1be110" }
```
jaybosamiya-ms, line 19:
```suggestion
# Depend on (currently unreleased) slabmalloc `main`, which contains some fixes on top of `0.11.0`
slabmalloc = { git = "https://github.com/gz/rust-slabmalloc.git", rev = "19480b2e82704210abafe575fb9699184c1be110" }
```

litebox/src/mm/alloc.rs
```diff
@@ -10,7 +10,23 @@ use buddy_system_allocator::LockedHeapWithRescue;
...
+    /// and the size of the allocated memory.
+    fn alloc(layout: &Layout) -> Option<(usize, usize)>;
+
+    /// Returns the memory back to host.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the `addr` is valid and was allocated by [`Self::alloc`].
+    unsafe fn free(addr: usize);
+}
```
jaybosamiya-ms: Some minor recommendations to this trait.  I recommend taking `&self` as argument, even if you don't need it, since it allows for a more general usage of this trait. In particular, the platform can use its own self as storage area if it _really_ wants to, rather than depending on globals.

```suggestion
/// Managing blocks of memory allocated from the underlying platform
pub trait MemoryProvider {
    /// Allocate a block of memory with `layout`.
    ///
    /// If successful, returns the start address and size of allocated memory.
    ///
    /// It may return a size larger than the requested size.
    fn alloc(&self, layout: &Layout) -> Option<(usize, usize)>;

    /// Returns the previously-allocated memory back.
    ///
    /// # Safety
    ///
    /// The caller must ensure that the `addr` is valid and was allocated by [`Self::alloc`].
    unsafe fn free(&self, addr: usize);
}
```

litebox/src/mm/alloc.rs
```diff
@@ -10,7 +10,23 @@ use buddy_system_allocator::LockedHeapWithRescue;
 use slabmalloc::{AllocationError, Allocator, LargeObjectPage, ObjectPage, ZoneAllocator};
 use spin::mutex::SpinMutex;
 
-use super::MemoryProvider;
```
jaybosamiya-ms: Rather than naming this file `alloc.rs`, it might be better named `allocator.rs`, just to prevent the need for `::alloc` and such, which reduces our consistency with things.

litebox/src/mm/alloc.rs
```diff
@@ -10,7 +10,23 @@ use buddy_system_allocator::LockedHeapWithRescue;
...
+    /// Returns the memory back to host.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the `addr` is valid and was allocated by [`Self::alloc`].
+    unsafe fn free(addr: usize);
+}
 
 /// Allocator that uses buddy allocator for pages and slab allocator for small objects.
 pub struct SafeZoneAllocator<'a, const ORDER: usize, M: MemoryProvider> {
```
jaybosamiya-ms: It would help to add a comment about what `ORDER` is.

litebox/src/mm/alloc.rs
```diff
@@ -28,22 +44,45 @@ impl<const ORDER: usize, M: MemoryProvider> Default for SafeZoneAllocator<'_, OR
...
+    ///
+    /// In addition to [`MemoryProvider::alloc`], this is used to add memory
+    /// to the buddy allocator when the system is initialized.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the memory range is valid and not used by any others.
+    pub unsafe fn fill_pages(&self, addr: usize, size: usize) {
+        unsafe { self.buddy_allocator.lock().add_to_heap(addr, addr + size) };
+    }
```
jaybosamiya-ms: I was a little confused by this function. Is the following a valid way to rephrase the comment? If so, I think it should help with clarity.
```suggestion
    /// Adds a range of memory to the buddy allocator.
    ///
    /// This range of memory can be provided from (say) [`MemoryProvider::alloc`] in order to allow it to be
    /// controlled by the buddy allocator. Morally, the buddy allocator takes ownership of this range of memory.
    ///
    /// # Safety
    ///
    /// The caller must ensure that the memory range is valid and not used by any others.
    pub unsafe fn fill_pages(&self, addr: usize, size: usize) {
        unsafe { self.buddy_allocator.lock().add_to_heap(addr, addr + size) };
    }
```

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -68,7 +68,10 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
             .collect();
         let data = std::fs::read(prog).unwrap();
         let data = if cli_args.rewrite_syscalls {
-            litebox_syscall_rewriter::hook_syscalls_in_elf(&data, None).unwrap()
+            // capstone declares a global allocator in conflict with our own.
+            // https://github.com/capstone-rust/capstone-rs/blob/14e855ca58400f454cb7ceb87d2c5e7b635ce498/capstone-rs/src/lib.rs#L16
+            // litebox_syscall_rewriter::hook_syscalls_in_elf(&data, None).unwrap()
```
jaybosamiya-ms, line 73: I think this is just a bug in `capstone-rs`, I just opened https://github.com/capstone-rust/capstone-rs/pull/171 to fix the bug.

litebox/Cargo.toml
CvvT: Thanks!

litebox/Cargo.toml
CvvT: Fixed

litebox/src/mm/alloc.rs
CvvT: yeah, make sense. Fixed.

litebox/src/mm/alloc.rs
CvvT: The problem is `LockedHeapWithRescue` only accepts `fn(&mut Heap<ORDER>, &Layout)` so I cannot pass an instance of `MemoryProvider`

https://github.com/MSRSSP/litebox/blob/95dd160b349f2ff64957e0d5e2ad6b6eee273087/litebox/src/mm/alloc.rs#L55-L65

We can reimplement `LockedHeapWithRescue` if we find it necessary. Also, this is the reason why I added a new interface `fill_pages` so that the platform can add its storage to the allocator.

litebox/src/mm/alloc.rs
CvvT: Sure, basically `ORDER` specifies the maximum size of memory the allocator can allocate -- (1 << `ORDER`) * PAGE_SIZE.

litebox/src/mm/alloc.rs
CvvT: `MemoryProvider::alloc` is called whenever the buddy allocator is out of memory, while `fill_pages` is a public interface that user can call at any point to add memory to the buddy allocator. One use case might be for platform that has fixed pre-allocated memory, it can call `fill_pages` once to let the allocator manage all the memory when it initializes and leave `MemoryProvider::alloc` empty.

litebox_runner_linux_userland/src/lib.rs
CvvT: Thanks! Let's wait a little bit more to see how they respond.

litebox/src/mm/alloc.rs
jaybosamiya-ms: Reasonable, we can keep it without the `&self`. Please do see the possible comment changes if you'd like to update/use some of those for clarity if you prefer. Otherwise, please do not consider this one a blocker.

litebox/src/mm/alloc.rs
jaybosamiya-ms: That makes sense. Thanks for taking the comment I wrote and updating it to make it make sense with the correct version.

litebox_runner_linux_userland/src/lib.rs
jaybosamiya-ms: Yep, we don't need to wait on that to merge. As long as we add a quick comment in the code pointing at this, I think this should not be a blocker to us merging this PR.

PR #112, opened 2025-05-06
Remove dependence on `std` for loader

jaybosamiya-ms:
The only usage of `fs` in `elf_loader` in our shim is for `IOError` part of the enum. The actual choice of the enum doesn't _really_ matter for this error, since it is only used to print out the message.

`elf_loader` itself doesn't _need_ to know if things are being `mmap`d in, or they are being I/O'd in. Thus, by switching the error codes to be an `MmapError`, we are able to remove `fs` from the required features for `elf_loader`. This in turn allows us to remove the (transitive) dependency both `libc` and `syscalls` within the shim (specifically, `elf_loader` requires you to depend on at least one of the two, if you have the `fs` feature around, even if you are not using them).

### DIFF
--- a/litebox_shim_linux/Cargo.toml
+++ b/litebox_shim_linux/Cargo.toml
@@ -13,7 +13,7 @@ once_cell = { version = "1.20.2", default-features = false, features = ["alloc",
 # https://github.com/weizhiao/rust-elfloader/ causes trouble on the latest
 # released version of elf-loader, thus we limit ourselves to the released
 # version right before it.
-elf_loader = { version = "=0.10.0", features = ["fs"] }
+elf_loader = { version = "=0.10.0", default-features = false, features = [] }
 thiserror = { version = "2.0.6", default-features = false }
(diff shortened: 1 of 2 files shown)


jaybosamiya-ms:
This was found as part of the litebox-on-nanvix bring-up that @ppenna and I did yesterday. However, this is a cleaner solution than the temporary version we did yesterday.

Specifically, yesterday I wrote:
> Some miscellaneous issues with `elf_loader` that require attention, that could be a blocker, but seemingly (at least currently) might not be. We need to keep a look out for a `compile_error` that is thrown if we don't give it a `use-syscalls` feature---giving it the `use-syscalls` feature seems to allow things to work, but is a non-default usage that we haven't tested enough, so definitely need to look into it more.

Yesterday, we were using `use-syscalls`, just to get past this blocker, but that had me concerned. Now with this PR, that concern is removed, since we don't need `use-syscalls`.

PR #114, opened 2025-05-07
LVBS VSM functions and setup through VTL0

sangho2:
This PR adds support for some VSM security functions and VTL1 initialization through VTL0 kernel. It includes
- Configure VTL1 partition
- Configure VTL0 security features (e.g., MBEC) from VTL1
- Lock up security-critical registers (e.g., MSR, CR4, ...)
- Protect VTL1 memory addresses
- Obtain CPU and physical memory information from pages populated by VTL0 kernel

### DIFF
--- a/litebox_platform_lvbs/src/host/bootparam.rs
+++ b/litebox_platform_lvbs/src/host/bootparam.rs
@@ -1,13 +1,13 @@
 //! VTL1 kernel boot parameters (compatible with Linux kernel's boot_params structure and command line)
 
-use crate::mshv::vtl1_mem_layout::{
-    VTL1_BOOT_PARAMS_PAGE, VTL1_CMDLINE_PAGE, VtlMemoryError, get_address_of_special_page,
+use crate::{
+    debug_serial_println,
+    mshv::vtl1_mem_layout::{
+        VTL1_BOOT_PARAMS_PAGE, VTL1_CMDLINE_PAGE, VtlMemoryError, get_address_of_special_page,
+    },
 };
 use core::ffi::{CStr, c_char};
 use num_enum::TryFromPrimitive;
-
-#[cfg(debug_assertions)]
-use crate::serial_println;
 
 // This module defines a simplified Linux boot params structure
 // (based on arch/x86/include/uapi/asm/bootparam.h and
@@ -60,7 +60,7 @@ impl BootParams {
             } else {
                 let addr_val = entry.addr;
                 let size_val = entry.size;
-                serial_println!(
+                debug_serial_println!(
                     "addr: {:#x}, size: {:#x}, type: {:?}",
                     addr_val,
                     size_val,
@@ -113,7 +113,7 @@ pub fn dump_cmdline() {
     }
 
     if let Some(cmdline_str) = unsafe { CStr::from_ptr(cmdline).to_str().ok() } {
-        serial_println!("{}", cmdline_str);
+        debug_serial_println!("{}", cmdline_str);
     }
 }
 
--- a/litebox_platform_lvbs/src/mshv/hvcall.rs
+++ b/litebox_platform_lvbs/src/mshv/hvcall.rs
@@ -105,6 +105,7 @@ pub fn init() -> Result<(), HypervError> {
 
     debug_serial_println!("HV_REGISTER_VP_INDEX: {:#x}", rdmsr(HV_REGISTER_VP_INDEX));
 
+    #[cfg(debug_assertions)]
     if get_core_id() == 0 {
         debug_serial_println!(
             "HV_X64_MSR_VP_ASSIST_PAGE: {:#x}",


litebox_platform_lvbs/src/kernel_context.rs
```diff
@@ -49,6 +50,11 @@ impl KernelContext {
     pub fn hv_hypercall_page_as_u64(&self) -> u64 {
         get_hypercall_page_address()
     }
+
+    pub fn hv_hypercall_input_page_as_mut_ptr(&mut self) -> *mut [u8; PAGE_SIZE] {
+        self.hvcall_input = [0u8; PAGE_SIZE];
```
wdcui: I don't understand this code. self.hvcall_input is already pointed to a page. Why do we assign another slice to it here?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -40,11 +76,91 @@ pub fn mshv_vsm_boot_aps(_cpu_online_mask_pfn: u64, _boot_signal_pfn: u64) -> u6
...
+    0
+}
+
+pub fn mshv_vsm_configure_partition() -> u64 {
+    #[cfg(debug_assertions)]
+    serial_println!("VSM: Configure partition");
+
+    let mut config = HvRegisterVsmPartitionConfig::new();
+    config.set_default_vtl_protection_mask(HvPageProtFlags::HV_PAGE_FULL_ACCESS.bits().into());
+    config.set_enable_vtl_protection();
```
wdcui, line 101: Minor: the name of set_enable_vtl_protection is a little confusing. Is it just "enable_vtl_protection"?

litebox_platform_lvbs/src/kernel_context.rs
sangho2: Thanks. I've fixed it by using `fill(0)` at the caller (better not to have a dual-purpose function).

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Hyper-V VSM calls this field `EnableVtlProtection` and our convention here is having `set_<field_name>` for each field. I agree with you that `set_enable_...` is confusing.

litebox_platform_lvbs/src/host/bootparam.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub typ: u32,
+}
+
+#[derive(Clone, Copy)]
+#[repr(C, packed)]
+struct BootParams {
+    _unused0: [u8; 720], // Elements in this area are not used.
+    pub e820_table: [BootE820Entry; E820_MAX_ENTRIES],
+    _unused1: [u8; 816], // Elements in this area are not used.
+}
```
CvvT: The struct itself does not have pub but its fields have?

litebox_platform_lvbs/src/host/bootparam.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+}
+
+/// Funtion to get the guest physical start address and size of VTL1 memory
+pub fn get_vtl1_memory_info() -> Result<(u64, u64), VtlMemoryError> {
+    let boot_params = get_address_of_special_page(VTL1_BOOT_PARAMS_PAGE) as *const BootParams;
+    unsafe { (*boot_params).memory_info() }
+}
+
+/// Funtion to get the number of possible cpus from the command line (Linux kerne's num_possible_cpus())
```
CvvT: "Linux kerne's" --> "Linux kernel's"

litebox_platform_lvbs/src/mshv/hvcall_vp.rs
```diff
@@ -5,52 +5,64 @@ use crate::{
...
 pub fn hvcall_set_vp_registers(
     reg_name: u32,
     value: u64,
     input_vtl: u8,
 ) -> Result<u64, HypervCallError> {
-    let mut hvin = HvSetVpRegistersInput::new();
+    let kernel_context = get_per_core_kernel_context();
+    let hvin = unsafe {
+        let ptr = kernel_context.hv_hypercall_input_page_as_mut_ptr();
+        (*ptr).fill(0);
```
CvvT: Why you need to call `fill(0)` when it will be overwritten by the following assignment `*hvin = ... `? to prevent info leak?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -40,11 +76,91 @@ pub fn mshv_vsm_boot_aps(_cpu_online_mask_pfn: u64, _boot_signal_pfn: u64) -> u6
...
+        let err: u32 = result.into();
+        return err.into();
+    }
+
+    0
+}
+
+pub fn mshv_vsm_configure_partition() -> u64 {
+    #[cfg(debug_assertions)]
+    serial_println!("VSM: Configure partition");
```
CvvT: You may want to have a debug version `serial_debug_println` with the config check inside the macro so that you can save one line.

litebox_platform_lvbs/src/host/bootparam.rs
sangho2: No other code needs to access this struct, so I made it private. `e820_table` also doesn't have to be pub, so I fixed it.

litebox_platform_lvbs/src/host/bootparam.rs
sangho2: Fixed

litebox_platform_lvbs/src/mshv/hvcall_vp.rs
sangho2: This is conservative memory cleaning. Hyper-V hypercall accepts variable-length input and `HvSetVpRegisterInput` is smaller then the input buffer (page) size. Let me re-think whether this `fill(0)` is really needed.

PR #115, opened 2025-05-09
Avoid calling syscalls using libc

CvvT:
Similar to the PR #111 that addresses the issue mentioned in #66, calling syscalls using libc would access `__errno_location` which is also a per-thread data. This PR replaces all syscalls with the crate `syscalls`.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -587,10 +587,16 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Li
             });
         let ptr = unsafe {
             syscalls::syscall6(
-                #[cfg(target_arch = "x86_64")]
-                syscalls::Sysno::mmap,
-                #[cfg(target_arch = "x86")]
-                syscalls::Sysno::mmap2,
+                {
+                    #[cfg(target_arch = "x86_64")]
+                    {
+                        syscalls::Sysno::mmap
+                    }
+                    #[cfg(target_arch = "x86")]
+                    {
+                        syscalls::Sysno::mmap2
+                    }
+                },
                 range.start,
                 range.len(),
                 prot_flags(initial_permissions)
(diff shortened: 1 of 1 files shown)


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -549,43 +577,50 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Li
...
-                    | (if can_grow_down {
-                        nix::sys::mman::MapFlags::MAP_GROWSDOWN
-                    } else {
-                        nix::sys::mman::MapFlags::empty()
-                    }),
+            syscalls::syscall6(
+                #[cfg(target_arch = "x86_64")]
+                syscalls::Sysno::mmap,
+                #[cfg(target_arch = "x86")]
+                syscalls::Sysno::mmap2,
```
jaybosamiya-ms: While this technically works, it makes it _seem_ like the 2nd argument is `mmap2`. To make it clearer that it is a single argument, we can use something like:
```suggestion
                if cfg!(target_arch = "x86_64") {
                  syscalls::Sysno::mmap
                } else if cfg!(target_arch = "x86") {
                  syscalls::Sysno::mmap2,
                } else {
                  unreachable!()
                },
```

CvvT: Thanks, I just tried it, but there is an issue. According to the doc, "`cfg!`, unlike `#[cfg]`, does not remove any code and only evaluates to true or false. For example, all blocks in an if/else expression need to be valid when cfg! is used for the condition, regardless of what cfg! is evaluating." However, syscalls::Sysno::mmap2 only exist under x86 arch.

I added more curly bracket as follows, hopefully it would make it clearer.

```rust
syscalls::syscall6(
    {
        {
            syscalls::Sysno::mmap
        }
        #[cfg(target_arch = "x86")]
        {
            syscalls::Sysno::mmap2
        }
    },
    0,
    size,
    ProtFlags::PROT_READ_WRITE.bits().reinterpret_as_unsigned() as usize,
    (MapFlags::MAP_PRIVATE | MapFlags::MAP_ANON)
        .bits()
        .reinterpret_as_unsigned() as usize,
    usize::MAX,
    0,
)
```

jaybosamiya-ms: Ah, indeed `cfg!` only evaluates to true/false, so your version is what we need to go with. Thanks for checking and updating it!

jaybosamiya-ms: Please feel free to merge, thanks!

PR #116, opened 2025-05-12
Network shim layer

CvvT:
Add socket related syscalls to the linux shim layer, including socket, connect, bind, listen, accept, sendto, and recvfrom. Currently only IPv4 is supported.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -205,6 +205,9 @@ impl Descriptors {
 
 enum Descriptor {
     File(litebox::fd::FileFd),
+    // Note we are using `Arc` here so that we can hold a reference to the socket
+    // without holding a lock on the file descriptor (see `sys_accept` for an example).
+    // TODO: this could be addressed by #120.
     Socket(alloc::sync::Arc<crate::syscalls::net::Socket>),
     PipeReader {
         consumer: alloc::sync::Arc<crate::channel::Consumer<u8>>,
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -51,6 +51,7 @@ pub(super) enum SocketAddress {
     Inet(SocketAddr),
 }
 
+// TODO: move `status` and `close_on_exec` to litebox once #119 is completed
 pub(crate) struct Socket {
     pub(crate) fd: Option<SocketFd>,
     /// File status flags (see [`litebox::fs::OFlags::STATUS_FLAGS_MASK`])


litebox/src/net/mod.rs
```diff
@@ -870,8 +871,8 @@ where
                 .get_mut::<tcp::Socket>(socket_handle.handle)
                 .recv_slice(buf)
                 .map_err(|e| match e {
-                    tcp::RecvError::InvalidState => ReceiveError::OperationFinished,
-                    tcp::RecvError::Finished => ReceiveError::SocketInInvalidState,
+                    tcp::RecvError::InvalidState => ReceiveError::SocketInInvalidState,
+                    tcp::RecvError::Finished => ReceiveError::OperationFinished,
```
jaybosamiya-ms, line 875: Nice catch!

litebox_common_linux/src/lib.rs
```diff
@@ -368,6 +369,74 @@ bitflags::bitflags! {
...
+impl From<u32> for SockType {
+    fn from(value: u32) -> Self {
+        match value {
+            1 => Self::Stream,
+            2 => Self::Datagram,
+            3 => Self::Raw,
+            _ => unimplemented!(),
+        }
+    }
+}
```
jaybosamiya-ms: Why does one of them have constants, while the other has only values directly set in the enum?

I recommend the enum approach, fwiw, but yeah, we should make things consistent.

Additionally, rather than `From`, it should probably be `TryFrom`.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -351,21 +354,21 @@ impl litebox::platform::IPInterfaceProvider for LinuxUserland {
         match unsafe {
             syscalls::syscall4(
                 syscalls::Sysno::write,
-                usize::try_from(tun_socket_fd.as_raw_fd()).unwrap(),
+                tun_socket_fd.as_raw_fd().reinterpret_as_unsigned() as usize,
```
jaybosamiya-ms: Why was this change needed? Is there actually a negative value that needs to be reinterpreted?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -381,19 +384,16 @@ impl litebox::platform::IPInterfaceProvider for LinuxUserland {
         unsafe {
             syscalls::syscall4(
                 syscalls::Sysno::read,
-                usize::try_from(tun_socket_fd.as_raw_fd()).unwrap(),
+                tun_socket_fd.as_raw_fd().reinterpret_as_unsigned() as usize,
```

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -381,19 +384,16 @@ impl litebox::platform::IPInterfaceProvider for LinuxUserland {
...
         }
-        .map_err(|e| {
-            if e == syscalls::Errno::EWOULDBLOCK || e == syscalls::Errno::EAGAIN {
-                litebox::platform::ReceiveError::WouldBlock
-            } else {
-                unimplemented!("unexpected error {}", e)
-            }
+        .map_err(|errno| match errno {
+            syscalls::Errno::EWOULDBLOCK => litebox::platform::ReceiveError::WouldBlock,
+            _ => unimplemented!("unexpected error {errno}"),
```
jaybosamiya-ms: Why is `EAGAIN` not being sent back as wouldblock?  The `_ => unimplemented!()` will catch it and throw an error, so I guess it is ok, but I am surprised by this change.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -0,0 +1,594 @@
...
+        std::fs::write(src_path.as_path(), EXTERNAL_CLIENT_C_CODE).unwrap();
+        let output_path = std::path::Path::new(dir_path.as_str()).join("external_client");
+        crate::syscalls::tests::compile(
+            src_path.to_str().unwrap(),
+            output_path.to_str().unwrap(),
+            true,
+            false,
+        );
+
+        #[allow(clippy::zombie_processes)]
```
jaybosamiya-ms: I think you could eliminate this by _after_ the `test_tcp_socket`, you added a `child.kill`/`child.wait`/etc.

litebox_shim_linux/src/lib.rs
```diff
@@ -195,7 +205,7 @@ impl Descriptors {
 
 enum Descriptor {
     File(litebox::fd::FileFd),
-    Socket(litebox::fd::SocketFd),
+    Socket(alloc::sync::Arc<crate::syscalls::net::Socket>),
```
jaybosamiya-ms: I'm not sure I understand why this one needs the `arc<socket>` thing showing up.  We do _not_ need `close_on_exec` semantics right now.  And as for status flags, we could add something similar to #76, yes?

I'm happy to add something similar to the metadata such as with #76, if you open up an issue to track this, and add a comment to the `Socket` structure in this PR as "update when #<newissuenumber> is resolved".  That way, it won't be a blocker, but it is tracked properly.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -183,21 +183,25 @@ pub fn sys_pwrite64(fd: i32, buf: &[u8], offset: usize) -> Result<usize, Errno>
     sys_write(fd, buf, Some(offset))
 }
 
+fn do_close(file: Descriptor) -> Result<(), Errno> {
+    match file {
```
jaybosamiya-ms: Nit: just for consistency/readability:
```suggestion
fn do_close(desc: Descriptor) -> Result<(), Errno> {
    match desc {
```

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -183,21 +183,25 @@ pub fn sys_pwrite64(fd: i32, buf: &[u8], offset: usize) -> Result<usize, Errno>
...
+        Descriptor::File(file) | Descriptor::Stdio(crate::stdio::StdioFile { file, .. }) => {
+            litebox_fs().close(file).map_err(Errno::from)
+        }
+        Descriptor::Socket(socket) => Ok(()),
```
jaybosamiya-ms: Sockets also need to be closed.  I know that the dropped `Socket` is what is doing the close, but that should at least have a comment here.  I don't think we should silently mark as `Ok(())` here.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -183,21 +183,25 @@ pub fn sys_pwrite64(fd: i32, buf: &[u8], offset: usize) -> Result<usize, Errno>
...
-        Some(Descriptor::File(file) | Descriptor::Stdio(crate::stdio::StdioFile { file, .. })) => {
-            litebox_fs().close(file).map_err(Errno::from)
-        }
-        Some(Descriptor::Socket(socket)) => todo!(),
-        Some(
-            Descriptor::PipeReader { .. }
-            | Descriptor::PipeWriter { .. }
-            | Descriptor::Eventfd { .. },
-        ) => Ok(()),
+        Some(desc) => do_close(desc),
```
jaybosamiya-ms: Fwiw, I don't quite see how refactoring this into a separate function is helpful, other than add indirection.

litebox_common_linux/src/lib.rs
CvvT: Good point. I found a simple crate https://crates.io/crates/int-enum to do it.

litebox_platform_linux_userland/src/lib.rs
CvvT: You're right, I changed it back.

litebox_platform_linux_userland/src/lib.rs

litebox_platform_linux_userland/src/lib.rs
CvvT: It reports a warning because EWOULDBLOCK equals EAGAIN and thus there is a unreachable branch. I guess I can just suppress this warning.

litebox_shim_linux/src/lib.rs
CvvT: `Arc` is for supporting blocking operations on the socket. 

https://github.com/MSRSSP/litebox/blob/c912a339276d624b409bea993f45a528c20b4a91/litebox_shim_linux/src/syscalls/net.rs#L260-L270

litebox_shim_linux/src/syscalls/file.rs

litebox_shim_linux/src/syscalls/file.rs
CvvT: Oh this is needed for another PR. I changed it back.

litebox_shim_linux/src/syscalls/file.rs
CvvT: Sure, added some comment.

litebox_shim_linux/src/syscalls/net.rs
CvvT: Fixed! Thanks!

litebox_common_linux/src/lib.rs
jaybosamiya-ms: Thanks, that's a nice crate for it.  

(Tangentially) I think an even simpler crate could be written for this (`int-enum` pulls in `syn`, despite not needing the _vast_ majority of `syn`'s features), but this is not really an issue for us. Using `int-enum` is the right move for us.

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: Ah, I missed that they are literally equal, yes suppressed warning here makes it clearer.  I am surprised that the prior version had not triggered the warning (I guess that is one place rustc/clippy are not checking for redundancy?).

litebox_shim_linux/src/lib.rs
jaybosamiya-ms: Ah, thanks! That explains the `Arc` part, but I still think there is a better move possible by improving the interfaces a little bit.

I've opened #119 and #120 to track these.

CvvT: Thanks! Another reason I'd like to use `Arc` is to support `dup`. I just submitted a PR to support `dup` for stdio: #121

jaybosamiya-ms: Thanks, seems reasonable, I just did a quick review of it.  I think we should still include a comment in the codebase around here to link to #119 and #120 for tracking reasons, but yeah, feel free to merge once that is done, thanks!

CvvT: Sure, just added some comments. Thanks!

PR #121, opened 2025-05-14
Support syscall dup/dup2/dup3

CvvT:
This PR adds support for syscall dup/dup2/dup3. Currently, it only supports `dup` for stdio.

#74/#76 provide storage for per-file and per-fd metadata. `close_on_exec` is per-fd metadata while file `status` is per-file-struct metadata (which is different from per-file metadata). Basically, `status` is shared among `dup`ed fds, while `open`ed fds that point to same file do not share the same `status`. I added a [test](https://github.com/MSRSSP/litebox/blob/5d30eb0a8fc407bb59359a7cddbf73cf70d53eb8/litebox_shim_linux/src/stdio.rs#L136) that might be clearer.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -486,15 +486,8 @@ pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
     },
     Dup {
         oldfd: i32,
-    },
-    Dup2 {
-        oldfd: i32,
-        newfd: i32,
-    },
-    Dup3 {
-        oldfd: i32,
-        newfd: i32,
-        flags: litebox::fs::OFlags,
+        newfd: Option<i32>,
+        flags: Option<litebox::fs::OFlags>,
     },
     Socket {
         domain: AddressFamily,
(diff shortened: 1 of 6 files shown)


litebox_common_linux/src/lib.rs
```diff
@@ -484,6 +484,18 @@ pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
...
+    },
+    Dup2 {
+        oldfd: i32,
+        newfd: i32,
+    },
+    Dup3 {
+        oldfd: i32,
+        newfd: i32,
+        flags: litebox::fs::OFlags,
+    },
```
jaybosamiya-ms: dup/dup2/dup3 exist as separate things because C does not nicely support optional. We can collapse them here by using the following:
```suggestion
    Dup {
        oldfd: i32,
        newfd: Option<i32>,
        flags: Option<litebox::fs::OFlags>,
    },
```

Given that `SyscallRequest` is already a stronger-typed notion of syscalls, we can afford to collapse things where we find the original syscalls to be deficient, imho.

CvvT: Good point! Fixed now.

PR #122, opened 2025-05-14
Support syscall setsockopt

CvvT:
This PR adds support for syscall `setsockop`. Currently, it only handles a few options (with some TODOs left), including `IP_TOS, SO_RCVTIMEO, SO_SNDTIMEO, SO_REUSEADDR, SO_BROADCAST, SO_KEEPALIVE, SO_RCVBUF, SO_SNDBUF, SO_TYPE, SO_PEERCRED, TCP_NODELAY, and TCP_CORK`.

### DIFF
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -977,7 +977,7 @@ pub enum TcpOptionName {
 
 /// Data for TCP options
 ///
-/// Note it should be paired with the correct `TcpOptionName` variant.
+/// Note it should be paired with the correct [`TcpOptionName`] variant.
 /// For example, `TcpOptionName::NODELAY` should be paired with `TcpOptionData::NODELAY(true)`.
 #[non_exhaustive]
 pub enum TcpOptionData {
--- a/litebox_shim_linux/Cargo.toml
+++ b/litebox_shim_linux/Cargo.toml
@@ -18,7 +18,6 @@ elf_loader = { version = "=0.10.0", default-features = false, features = [] }
 thiserror = { version = "2.0.6", default-features = false }
 ringbuf = { version = "0.4.8", default-features = false, features = ["alloc"] }
 syscalls = { version = "0.6", default-features = false }
-smoltcp = { version = "0.12.0", default-features = false }
 
 [target.'cfg(target_arch = "x86")'.dependencies]
 # ELF loader requires us to enable the `rel` dependency on 32-bit to work.


litebox/src/net/mod.rs
```diff
@@ -881,6 +881,21 @@ where
         self.automated_platform_interaction(PollDirection::Ingress);
         ret
     }
+
+    /// Invoke the given function with the given socket if it is a valid tcp socket.
+    pub fn call_on_tcp_socket_mut<F, R>(&mut self, fd: &SocketFd, f: F) -> Option<R>
+    where
+        F: FnOnce(&mut smoltcp::socket::tcp::Socket) -> R,
```
jaybosamiya-ms: I'm not comfortable with this type signature in the public API of LiteBox, since it leaks our internal implementation details (specifically, what happens if we decide to move away from smoltcp and have our own implementation?).  I know we are not planning to change the internal implementation, but it still feels unclean to do this.

I think this needs more thinking as to how we can expose a nice enough interface for supporting this syscall without completely breaking all the abstractions we've set up.

Let us chat more IRL on how to handle this.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -51,29 +54,159 @@ pub(super) enum SocketAddress {
...
+                                tcp.set_nagle_enabled(val == 0);
+                            })
+                            .ok_or(Errno::EOPNOTSUPP)?;
+                        Ok(())
+                    }
+                    TcpOption::CORK => {
+                        // Some applications use Nagle’s Algorithm (via the TCP_NODELAY option) for a similar effect.
+                        // However, TCP_CORK offers more fine-grained control, as it’s designed for applications that
+                        // send variable-length chunks of data that don’t necessarily fit nicely into a full TCP segment.
+                        // Because smoltcp does not support TCP_CORK, we emulate it by enabling/disabling Nagle’s Algorithm.
```
jaybosamiya-ms: Minor: I was surprised to see `’` rather than `'` in this comment; was this comment (partly at least) generated by an LLM?

litebox_common_linux/src/lib.rs
```diff
@@ -409,6 +408,99 @@ pub enum Protocol {
...
+    if #[cfg(all(target_arch = "x86"))] {
+        pub type time_t = u32;
+        pub type suseconds_t = u64;
+    } else if #[cfg(all(target_arch = "x86_64"))] {
+        pub type time_t = u64;
+        pub type suseconds_t = u64;
+    } else {
+        compile_error!("Unsupported architecture");
+    }
+}
```
jaybosamiya-ms: Isn't this equivalent to `time_t = usize`?  I'm not sure why this is cfg-gated.

Also `suseconds_t` is supposed to be a _signed_ integer type (https://www.man7.org/linux/man-pages/man3/time_t.3type.html) correct?

Additionally, you've set it to u64 for both 32 and 64-bit, but I think that is incorrect:
```console
$ cat x.c
#include <stdio.h>
#include <sys/types.h>
int main() {
  printf("%d\n", sizeof(suseconds_t));
}

$ gcc -m32 -o x x.c && ./x
4

$ gcc -Wno-format -m64 -o x x.c && ./x
8
```

litebox/src/net/mod.rs
CvvT: I see. We can make it less generic by adding new interface for each sepcific sock option that requires accessing internal API like `set_keep_alive ` and `set_nagle_enabled` if you believe that's better.

litebox_common_linux/src/lib.rs
CvvT: Right, it's a typo and should be i32. Also changed others unsigned types to signed types. Thanks!

For some other arch, `time_t` is not `usize`:
https://github.com/rust-lang/libc/blob/151c3a971e423c76e7acb54aa2d21a6e2706c4e6/src/unix/linux_like/linux/gnu/b32/mod.rs#L22

litebox_shim_linux/src/syscalls/net.rs
CvvT: I'm surprised you could notice it. This comment was copied from VSBox, I forgot if I used copilot or not. It is fixed now.

litebox/src/net/mod.rs
jaybosamiya-ms: I think adding new interfaces for every _group_ of functionality would be good. For example, we can choose to collect many related ones under a single function that takes an enum as an argument. If we are not fully exhaustive in options (i.e., we expect that the enum might be expanded in the future), we can also mark the enum as `#[non_exhaustive]`.

A simple grouping might even just be to go with `set_tcp_option`, so you could say things like `set_tcp_option(Nagle, true)`.  If you can think of a more precise grouping (I haven't spent much time thinking of a better grouping fwiw), that might be better, but even a simpler grouping is fine.

litebox_shim_linux/src/syscalls/net.rs
jaybosamiya-ms: Thanks. Fwiw, the last line of the comment still has the `’`

litebox_common_linux/src/lib.rs
jaybosamiya-ms: Huh, interesting. That is a neat finding and might be a good idea to place as a comment around there.

litebox/src/net/mod.rs
jaybosamiya-ms: Quick note to add: you may need to allow for either supporting separate `set_option_bool`/`set_option_i32`/.. style things, or you may need to support two different enums for `set_option` and `get_option` (specificaly, allowing something like `SockOptName::Nagle` vs `SockOptData::Nagle(bool)` or similar, where `set_option` takes a `SockOptData` while `get_option` takes a `SockOptName` and returns a `SockOptData`).

Technically we want dependent types there, but I think that is overkill for this application (also not _directly_ supported by Rust, so we'd need to spend some time hacking around it).

There is also the option of saying that the data is _always_ an `i32` (or similar), and returning a `Err(SetSockOptError::InvalidValue)` or similar if we expected it to be a bool.

I'll let you pick whichever of the three options seems most reasonable to you based on the specifics you are noticing.

CvvT: Thanks for all the suggestions! I go with the second option, which separate TcpOptionData from TcpOptionName.

I don't have a better idea on how to group options. `set_tcp_option` sounds reasonable to me.

litebox_shim_linux/src/syscalls/net.rs
CvvT: Good catch! Fixed now.

litebox_common_linux/src/lib.rs
CvvT: Sure, I added some comments. Thanks!

litebox/src/net/mod.rs
```diff
@@ -879,6 +879,40 @@ where
...
+        match socket_handle.protocol() {
+            Protocol::Tcp => {
+                let tcp_socket = self.socket_set.get_mut::<tcp::Socket>(socket_handle.handle);
+                match (option, data) {
+                    (TcpOptionName::NODELAY, TcpOptionData::NODELAY(nodelay)) => {
+                        tcp_socket.set_nagle_enabled(!nodelay);
+                    }
+                    (TcpOptionName::KEEPALIVE, TcpOptionData::KEEPALIVE(keepalive)) => {
+                        tcp_socket.set_keep_alive(keepalive.map(smoltcp::time::Duration::from));
+                    }
```
jaybosamiya-ms: Since the `data` is always in sync with the `option`, it is essentially redundant right?  If so, do we need to keep it as an extra argument (in `set_` function)?  I think we can remove the `option` argument and simplify these to just
```suggestion
                    TcpOptionData::NODELAY(nodelay) => {
                        tcp_socket.set_nagle_enabled(!nodelay);
                    }
                    TcpOptionData::KEEPALIVE(keepalive) => {
                        tcp_socket.set_keep_alive(keepalive.map(smoltcp::time::Duration::from));
                    }
```

litebox/src/net/mod.rs
```diff
@@ -935,3 +969,24 @@ bitflags! {
...
+    /// segments are always sent as soon as possible, even if there
+    /// is only a small amount of data.
+    NODELAY,
+    /// Enable sending of keep-alive messages.
+    KEEPALIVE,
+}
+
+/// Data for TCP options
+///
+/// Note it should be paired with the correct `TcpOptionName` variant.
```
jaybosamiya-ms: Minor update (to automatically get the documentation hyperlink)
```suggestion
/// Note it should be paired with the correct [`TcpOptionName`] variant.
```

litebox_shim_linux/Cargo.toml
```diff
@@ -18,6 +18,7 @@ elf_loader = { version = "=0.10.0", default-features = false, features = [] }
 thiserror = { version = "2.0.6", default-features = false }
 ringbuf = { version = "0.4.8", default-features = false, features = ["alloc"] }
 syscalls = { version = "0.6", default-features = false }
+smoltcp = { version = "0.12.0", default-features = false }
```
jaybosamiya-ms: This should no longer be needed, yes?

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -51,29 +54,171 @@ pub(super) enum SocketAddress {
...
+                            litebox::net::TcpOptionData::KEEPALIVE(Some(
+                                core::time::Duration::from_secs(2 * 60 * 60),
+                            )),
+                        ) {
+                            match err {
+                                litebox::net::errors::SetTcpOptionError::InvalidFd => {
+                                    return Err(Errno::EBADF);
+                                }
+                                // This option probably has no effect on other protocols
+                                litebox::net::errors::SetTcpOptionError::NotTcpSocket => {}
```
jaybosamiya-ms: In the case of `probably`, I think I prefer that there is an explicit panic instead, just so we will know when we hit it, rather than chasing it down during debugging
```suggestion
                                litebox::net::errors::SetTcpOptionError::NotTcpSocket => {
                                  unimplemented!("KEEPALIVE probably has no effect on other protocols")
                                }
```

litebox/src/net/mod.rs
CvvT: Good point! I removed this argument.

litebox/src/net/mod.rs
CvvT: `TcpOptionName` is not used in this PR, thus I removed it completely.

litebox_shim_linux/Cargo.toml
CvvT: Good catch! I forgot this.

litebox_shim_linux/src/syscalls/net.rs
CvvT: Sure, fixed.

litebox/src/net/mod.rs
jaybosamiya-ms: The enum is still around fwiw.  Also, you raise a valid point, maybe we should include:
```
    pub fn get_tcp_option(
        &self,
        fd: &SocketFd,
        data: TcpOptionName,
    ) -> Result<TcpOptionData, errors::GetTcpOptionError> {
```
?

We don't need to expose the shim layer interfaces for it just yet, but for consistency, we might wish to include it in.  Not a blocker to the PR though, there is a clean way to do it later too.

Either way, I think given that `TcpOptionName` _is_ a valid enum (I don't think there is a good reason to remove it, since we'll add it back in later too), we should add the square brackets.

litebox_shim_linux/Cargo.toml
jaybosamiya-ms: I think the fix 6d48c440ef28f37921ef6c482f573de2f26c1d29 commit might have accidentally skipped removing this?

litebox/src/net/mod.rs
CvvT: I see. I will keep it then.

litebox_shim_linux/Cargo.toml
CvvT: Thanks! I may forget to commit the change.

PR #123, opened 2025-05-15
Adding `litebox_platform_lvbs` to `litebox_platform_multiplex`

sangho2:
This small PR adds `litebox_platform_lvbs` to `litebox_platform_multiplex`. First, it uses `cfg-if` to specify the `Platform` type (to work around clippy `--all-features` issue). Second, it selectively enables the interrupt service routines (which rely on an unstable Rust feature `abi_x86_interrupt`) of `litebox_platform_lvbs`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,19 +77,29 @@ jobs:
           rustup default ${RUST_CHANNEL}
           rustup override set ${RUST_CHANNEL}
           rustup show
+      - name: Set up Nextest
+        run: |
+          curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
       - uses: Swatinem/rust-cache@v2
       - name: cargo fmt
         run: |
           .github/tools/github_actions_run_cargo fmt --manifest-path=litebox_runner_lvbs/Cargo.toml
+          .github/tools/github_actions_run_cargo fmt --manifest-path=litebox_platform_lvbs/Cargo.toml
       - name: cargo clippy
         run: |
           .github/tools/github_actions_run_cargo clippy --lib --bins --examples --no-deps --all-features -Z build-std-features=compiler-builtins-mem -Z build-std=core,alloc --manifest-path=litebox_runner_lvbs/Cargo.toml --target litebox_runner_lvbs/x86_64_vtl1.json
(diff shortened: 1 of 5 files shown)


jaybosamiya-ms:
This is an interesting way to do it, and might be neat for some scenarios, but I think as a workaround around the `--all-features` issue, we just need to place a prioritization order (which is what your `build.rs` approach is doing anyways). The standard way to do this is to use the [`cfg-if`](https://lib.rs/crates/cfg-if) crate, which allows for this:
```rs
cfg_if::cfg_if! {
    if #[cfg(feature = "platform_linux_userland")] {
        pub type Platform = litebox_platform_linux_userland::LinuxUserland;
    } else if #[cfg(feature = "platform_lvbs")] {
        pub type Platform = litebox_platform_lvbs::host::LvbsLinuxKernel;
    } else {
        compile_error!(
            r##"No platforms specified.  Please enable the feature for the platform you want."##
        );
    }
}
```

`cfg_if!` does the correct chaining of placing the necessary `all(x, not(any(y, z))`-style conditions, that we would have to do manually otherwise.

jaybosamiya-ms:
For the current CI build failures on f09fbae, merging in the changes in `main` should fix it. See #124.

sangho2:
`cfg-if` works. Thanks, @jaybosamiya-ms !

litebox_platform_lvbs_isr/README.md
```diff
@@ -0,0 +1,4 @@
+# Interrupt service routines for LiteBox LVBS Platform
```
jaybosamiya-ms: I'm a little surprised by why `isr` was made into a separate crate (especially as part of this particular PR, based on title and description).  Any particular reason for that? Is that in plans for other upcoming changes?  The code changes themselves seem reasonable, it is just that I am not seeing why we need a separate crate for it.  As it _currently_ stands in this PR, this might fit better just as a module inside of `litebox_runner_lvbs` (as it previously was), no?

Note: I am _not_ against splitting it into a crate, it is just that I am not seeing justification at the moment for it, and imho, every additional crate _does_ carry some additional maintenance cost that modules don't (because each crate is a full compilation unit of its own, there are ordered edge constraints that start getting added by using crates, rather than modules).

sangho2: I also don't want to separate `isr` from `litebox_platform_lvbs`. It's still all about `clippy --all-features` and `stable` versus `nightly`. Let me check a few other workarounds.

sangho2:
Let me open this PR later.

PR #125, opened 2025-05-15
Fix partial read

CvvT:
Looks like when `nodejs` imports modules, it assumes no partial reads (`read` should reads as much as possible as long as the provided buffer is sufficient).

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -273,30 +273,22 @@ pub extern "C" fn close(fd: i32) -> i32 {
 // has the downside of requiring too many syscalls, while having it be too large allows for massive
 // allocations to be triggered by the userland program. For now, this is set to a
 // hopefully-reasonable middle ground.
-const MAX_KERNEL_BUF_SIZE: usize = 64 * 1024;
+const MAX_KERNEL_BUF_SIZE: usize = 0x80_000;
 
 /// Entry point for the syscall handler
 #[allow(clippy::too_many_lines)]
 pub fn syscall_entry(request: SyscallRequest<Platform>) -> isize {
     let res: Result<usize, Errno> = match request {
         SyscallRequest::Read { fd, buf, count } => {
+            // Note some applications (e.g., `node`) seem to assume that getting fewer bytes than
+            // requested indicates EOF.
+            debug_assert!(count <= MAX_KERNEL_BUF_SIZE);
             let mut kernel_buf = vec![0u8; count.min(MAX_KERNEL_BUF_SIZE)];
-            let mut total_read = 0;
-            loop {
-                match syscalls::file::sys_read(fd, &mut kernel_buf, None).and_then(|size| {
-                    buf.copy_from_slice(total_read, &kernel_buf[..size])
-                        .map(|()| size)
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
How large is the read it is attempting/checking in one go?! Also, it seems like this behavior is incorrect on `node`'s end, no?

Knowing what the read size it is actually attempting, we can just bump up `MAX_KERNEL_BUF_SIZE` to be at least that in size.  

By placing the loop here, we _technically_ make it feasible for `read` to do a partial overwrite of the user buffer and bump the cursor forwards, before returning an error (in which case, things _may_ be in a surprising state that is much harder to debug).

Can you dump the read size it is attempting, and then we can make a decision of whether we want to include the complexity of rewinding (which sometimes may not even be possible on some file system backends) or if we want to handle this just by bumping up our kernel buffer size.

CvvT:
> How large is the read it is attempting/checking in one go?! Also, it seems like this behavior is incorrect on `node`'s end, no?
> 
> Knowing what the read size it is actually attempting, we can just bump up `MAX_KERNEL_BUF_SIZE` to be at least that in size.
> 
> By placing the loop here, we _technically_ make it feasible for `read` to do a partial overwrite of the user buffer and bump the cursor forwards, before returning an error (in which case, things _may_ be in a surprising state that is much harder to debug).
> 
> Can you dump the read size it is attempting, and then we can make a decision of whether we want to include the complexity of rewinding (which sometimes may not even be possible on some file system backends) or if we want to handle this just by bumping up our kernel buffer size.

The requested size is 0x80000. I didn't find the source code where `node` does the read, but the symptom was that `node` threw an exception:

```
file:///.netlify/dist/run/handlers/server.js:1082
              return t4 && "function" == typeof Symbol && t
                                                           

SyntaxError: Unexpected end of input
    at compileSourceTextModule (node:internal/modules/esm/utils:338:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:106:18)
    at #translate (node:internal/modules/esm/loader:470:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:517:27)
```

The actual size of this file is 141108.

I guessed it was caused by partial read and fixed it. We can bump up the size to 0x80000? and add an assertion to check if the requested size is bigger.

jaybosamiya-ms:
I think that is a reasonable move. 0x80000 == 512KiB, which seems perfectly reasonable for a buffer size for us to hold on to (it also only gets triggered if someone requests such large buffers; if the user requests a smaller size, the kernel buffer size is smaller anyways).  So yeah, we should bump it up and add a `debug_assert!`. But if anyone is _actually_ hitting that assert and expecting a _guaranteed_ full read, then I would consider that program to be suspicious.

I _personally_ believe that `node` is doing a non-compliant thing because it is supposed to attempt an additional `read` to see if it gets an EOF, rather than assuming that it has reached EOF.  Specifically, the manpage for `read` literally says:
> (zero indicates end of file) [...] It is not an error if this number is smaller than the number of bytes requested; this may happen for example [...] `read()` was interrupted by a signal

Anyways, we need to handle it the way it is currently implemented, so yeah, bumping up the internal max size is reasonable.

PR #126, opened 2025-05-16
Add mremap and mmap

CvvT:
* Add support for `mremap`
* Previously only file-backed mmap requests were intercepted, now all `mmap` are intercepted and handled by Litebox first. For Linux userland, it sill forwards the request to the kernel, but it allows page manager to keep track of all allocated memory.

To leave a backdoor for invoking the original syscalls, seccomp filters are set to check the additional argument. For example, a syscall with `N` arguments will not be intercepted if the `N+1` argument is set to a specific value. This only works if the syscall has less than 6 arguments. For `mmap` with 6 arguments, it checks the 4th argument `flags` to see if a specific bit (1 << 31) is set. It seems that Linux ignores unknown flags.

### DIFF
--- a/litebox_platform_lvbs/src/mm/tests.rs
+++ b/litebox_platform_lvbs/src/mm/tests.rs
@@ -227,9 +227,13 @@ fn test_vmm_page_fault() {
     let mut vmm = PageManager::<_, PAGE_SIZE>::new(&litebox);
     unsafe {
         assert_eq!(
-            vmm.create_writable_pages(PageRange::new(start_addr, start_addr + 4 * PAGE_SIZE).unwrap(), true, |_: UserMutPtr<u8>| Ok(0),)
-                .unwrap()
-                .as_usize(),
+            vmm.create_writable_pages(
+                PageRange::new(start_addr, start_addr + 4 * PAGE_SIZE).unwrap(),
+                true,
+                |_: UserMutPtr<u8>| Ok(0),
+            )
+            .unwrap()
+            .as_usize(),
             start_addr
         );
     }
@@ -261,9 +265,12 @@ fn test_vmm_page_fault() {
     let stack_addr: usize = 0x1000_0000;
     unsafe {
         assert_eq!(
-            vmm.create_stack_pages(PageRange::new(stack_addr, stack_addr + 4 * PAGE_SIZE).unwrap(), true,)
-                .unwrap()
-                .as_usize(),
+            vmm.create_stack_pages(
+                PageRange::new(stack_addr, stack_addr + 4 * PAGE_SIZE).unwrap(),
+                true,
+            )
+            .unwrap()
+            .as_usize(),
             stack_addr
         );
     }


litebox/src/mm/linux.rs
```diff
@@ -385,13 +385,13 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
     ///
     /// # Panics
     ///
-    /// Panics if `new_size` is smaller than the current size of the range.
+    /// Panics if the size of `suggested_new_range` is smaller than the size of `old_range`.
     /// Panics if range is not within exact one mapping.
```
jaybosamiya-ms: Thanks for updating this comment!  Minor note: the last line (currently unchanged in the diff) can also benefit from grammar fix

litebox/src/mm/mod.rs
```diff
@@ -40,6 +43,35 @@ where
...
+            vmem.create_pages(
+                suggested_range,
+                fixed_addr,
+                is_stack,
+                before_perms,
+                after_perms,
+                op,
+            )
+        }
+    }
```
jaybosamiya-ms: This function almost seems to be "laundering" `unsafe`.  I understand that it is to help de-duplicate some code, but nonetheless, it would help making this function `unsafe` so that callers actually take a look inside to know what is happening.  Otherwise, there is no guarantee that the safety constraints are upheld.

jaybosamiya-ms: Sidenote: is this indirection actually helping us save much? It is wrapping away just two basic operations---one of which I even suggest could be bubbled "up" (i.e., we could make functions like `create_executable_pages` take a `PageRange` instead of the two `usize`s).  `PageRange` already is public, so having this be part of the public interface doesn't seem too bad, no?

litebox/src/mm/tests.rs
```diff
@@ -44,8 +44,10 @@ impl crate::platform::PageManagementProvider<PAGE_SIZE> for DummyVmemBackend {
         &self,
         old_range: Range<usize>,
         new_range: Range<usize>,
-    ) -> Result<(), crate::platform::page_mgmt::RemapError> {
-        Ok(())
+    ) -> Result<Self::RawMutPointer<u8>, crate::platform::page_mgmt::RemapError> {
+        Ok(unsafe {
+            core::mem::transmute::<*mut u8, TransparentMutPtr<u8>>(new_range.start as *mut u8)
+        })
```
jaybosamiya-ms: Please don't unnecessarily transmute. You can `TransparentMutPtr { inner: new_range.start as *mut u8 }`, which is entirely safe.

As per Rust docs:
> `transmute` should be the absolute last resort.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -16,6 +16,11 @@ fn align_up(addr: usize, align: usize) -> usize {
     (addr + align - 1) & !(align - 1)
 }
 
+#[inline]
+fn align_down(addr: usize, align: usize) -> usize {
+    addr & !(align - 1)
```
jaybosamiya-ms: Minor (just to prevent accidental issues in future, if this gets used somewhere else; I don't actually expect this to ever be triggered, but if it _does_ get triggered, then it would be _very_ surprising behavior that is hard to debug)
```suggestion
fn align_down(addr: usize, align: usize) -> usize {
    debug_assert!(align.is_power_of_two());
    addr & !(align - 1)
```

Funnily, Rust has a [`next_multiple_of`](https://doc.rust-lang.org/std/primitive.usize.html#method.next_multiple_of)  but not a `prev_multiple_of`, which is what we need here :)

litebox/src/mm/linux.rs
CvvT: Got it. It's fixed now.

litebox/src/mm/mod.rs
CvvT: Good point! I think it's better to take one `PageRange` instead of two usize. Thanks!

litebox/src/mm/tests.rs
CvvT: Good catch! Thanks!

litebox_shim_linux/src/syscalls/mm.rs
CvvT: Fair, I also added the assertion to `align_up`:

https://github.com/MSRSSP/litebox/blob/c9ff8b7836f1ea4065599cc50f12ac3852bfce66/litebox_shim_linux/src/syscalls/mm.rs#L15-L18

Do you think we should use `next_multiple_of` for `align_up`? `next_multiple_of` uses modulo operation but I guess it would be optimized by the compiler.

jaybosamiya-ms: Indeed it will be optimized, I don't think it matters much one way or the other. For now, leaving it as-is for consistency seems reasonable to me.

litebox/src/mm/mod.rs
```diff
@@ -42,8 +45,10 @@ where
 
     /// Create readable and executable pages.
     ///
-    /// `suggested_addr` is the hint address where to create the pages. Provide `0` to let the kernel
-    /// choose an available memory region.
+    /// The start address of `suggested_range` is the hint address for where to create the pages.
+    /// Provide `0` to let the kernel choose an available memory region.
```
jaybosamiya-ms, line 49: `0` to let kernel pick seems like what a C-like API would do. A more Rust-y way to do it would be to be an `Option`.  I don't think it matters too much in this case, since it is _already_ a fairly low-level API, but generally speaking, rather than sentinel values, it is preferred to use `Option` so that the type-system is tracking such situations.

For this case, I'll let you make a decision on whether to update it to be `Option<..>` or not.  I _personally_ suggest updating it if it is not too much hassle, but if it is, then leaving it as-is is fine too, and not a blocker to merging.

jaybosamiya-ms: If updating, for consistency, obviously the other functions in this with similar comment should be updated too.

CvvT: Good point! I will address it in a separate PR. `PageRange` also provides `len` so it cannot be simply replaced with `Option<PageRange>`. I can replace it with two new types: `Option<AlignedAddr<ALIGN>>` and `AlignedLen<ALIGN>`

litebox_shim_linux/src/syscalls/mm.rs
CvvT: Got it, thanks!

PR #127, opened 2025-05-20
Support MAP_POPULATE

CvvT:
@sangho2 mentioned LVBS does not forward page fault exceptions (a bug maybe?), so we need to set up the mapping immediately to avoid page faults. Also, `mmap` requires supporting `MAP_POPULATE`.

### DIFF
--- a/litebox/src/mm/linux.rs
+++ b/litebox/src/mm/linux.rs
@@ -77,16 +77,16 @@ bitflags::bitflags! {
     pub(super) struct CreatePagesFlags: u8 {
         const FIXED_ADDR     = 1 << 0;
         const IS_STACK       = 1 << 1;
-        const POPULATE_PAGES = 1 << 2;
+        const POPULATE_PAGES_IMMEDIATELY = 1 << 2;
     }
 }
 
 impl CreatePagesFlags {
-    pub(super) fn new(fixed_addr: bool, is_stack: bool, populate_pages: bool) -> Self {
+    pub(super) fn new(fixed_addr: bool, is_stack: bool, populate_pages_immediately: bool) -> Self {
         let mut flags = Self::empty();
         flags.set(Self::FIXED_ADDR, fixed_addr);
         flags.set(Self::IS_STACK, is_stack);
-        flags.set(Self::POPULATE_PAGES, populate_pages);
+        flags.set(Self::POPULATE_PAGES_IMMEDIATELY, populate_pages_immediately);
         flags
     }
 }
@@ -268,7 +268,7 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
         &mut self,
         range: PageRange<ALIGN>,
         vma: VmArea,
-        populate_pages: bool,
+        populate_pages_immediately: bool,
     ) -> Option<Platform::RawMutPointer<u8>> {
         let (start, end) = (range.start, range.end);
         if start < Self::TASK_ADDR_MIN || end > Self::TASK_ADDR_MAX {
@@ -296,7 +296,7 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
                 range.into(),
                 MemoryRegionPermissions::from_bits(permissions).unwrap(),
                 vma.flags.contains(VmFlags::VM_GROWSDOWN),
-                populate_pages,
+                populate_pages_immediately,
             )
         }
         .ok()?;
@@ -311,10 +311,11 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
     /// The length of `suggested_range` is the size of the pages to be created.
     ///
     /// Set `fixed_addr` to `true` to force the mapping to be created at the given address, resulting in any
-    /// existing overlapping mappings being removed.
+    /// existing overlapping mappings being removed. Otherwise, the kernel will choose an available memory region
+    /// if the suggested address is not available.
     ///
     /// By default, the pages are not populated until they are accessed.
-    /// Set `populate_pages` to `true` to populate the pages immediately.
+    /// Set `populate_pages_immediately` to `true` to populate the pages immediately.
     ///
     /// Return `Some(new_addr)` if the mapping is created successfully.
     /// The returned address is `ALIGN`-aligned.
@@ -329,12 +330,12 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
         suggested_range: PageRange<ALIGN>,
         vma: VmArea,
         fixed_addr: bool,
-        populate_pages: bool,
+        populate_pages_immediately: bool,
     ) -> Option<Platform::RawMutPointer<u8>> {
         let new_addr = self.get_unmmaped_area(suggested_range, fixed_addr)?;
         // new_addr must be ALIGN aligned
         let new_range = PageRange::new(new_addr, new_addr + suggested_range.len()).unwrap();
-        unsafe { self.insert_mapping(new_range, vma, populate_pages) }
+        unsafe { self.insert_mapping(new_range, vma, populate_pages_immediately) }
     }
 
     /// Resize a range in the virtual address space.
@@ -559,7 +560,7 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
                         },
                 ),
                 flags.contains(CreatePagesFlags::FIXED_ADDR),
-                flags.contains(CreatePagesFlags::POPULATE_PAGES),
+                flags.contains(CreatePagesFlags::POPULATE_PAGES_IMMEDIATELY),
             )
         }
         .ok_or(MappingError::OutOfMemory)?;
--- a/litebox/src/mm/mod.rs
+++ b/litebox/src/mm/mod.rs
@@ -52,7 +52,7 @@ where
     /// Set `fixed_addr` to `true` to force the mapping to be created at the given address, resulting in any
     /// existing overlapping mappings being removed.
     ///
-    /// Set `populate_pages` to `true` to populate the pages immediately.
+    /// Set `populate_pages_immediately` to `true` to populate the pages immediately.
     ///
     /// `op` is a callback for caller to initialize the created pages.
     ///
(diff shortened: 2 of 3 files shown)


litebox/src/mm/linux.rs
```diff
@@ -72,6 +72,24 @@ impl From<MemoryRegionPermissions> for VmFlags {
...
+    /// Options for page creation.
+    pub(super) struct CreatePagesFlags: u8 {
+        const FIXED_ADDR     = 1 << 0;
+        const IS_STACK       = 1 << 1;
+        const POPULATE_PAGES = 1 << 2;
+    }
+}
+
+impl CreatePagesFlags {
+    pub(super) fn new(fixed_addr: bool, populate_pages: bool) -> Self {
```
sangho2: Out of curiosity. Any reason to exclude `is_stack`?

litebox/src/mm/linux.rs
```diff
@@ -284,8 +304,16 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
     }
 
     /// Create a new mapping in the virtual address space.
-    /// The mapping will be created at the suggested address. If the suggested start address is zero,
-    /// some available range will be choosen by the kernel.
+    ///
+    /// `suggested_range` is the range of pages to create. If the start address is not given (i.e., zero), some
```
sangho2: a bit unclear. Is the start address `PageRange::start`?

litebox/src/mm/linux.rs
```diff
@@ -363,7 +392,7 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
                 return Err(VmemResizeError::RangeOccupied(r));
             }
             let range = PageRange::new(range.end, new_end).unwrap();
-            unsafe { self.insert_mapping(range, *cur_vma) };
+            unsafe { self.insert_mapping(range, *cur_vma, false) };
```
sangho2, line 397: Does this imply that `mremap` doesn't support `MAP_POPULATE` (it seems so)?

litebox/src/mm/linux.rs
CvvT: Simply because there is only one place where I need to set that bit, but now I feel it's better to include it as well. Adding one extra line of code shouldn't introduce too much overhead and probably compiler is smart enough to optimize it out.

litebox/src/mm/linux.rs
CvvT: Yes, the start address of `suggested_range` is the hint address for where to create the pages. I fixed the comment.

litebox/src/mm/linux.rs
CvvT: You're right. I checked Linux's implementation, if `mremap` expands the old memory region, the extra part does not get populated. 

For LVBS, if it requires to populate all pages, I guess we can add a config check (once we merge two platform implements) in the page table implementation and handle all unmapped pages:

https://github.com/MSRSSP/litebox/blob/056a99c8d80840d7b0f2367084201260f0b72c42/litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs#L144-L148

litebox/src/platform/page_mgmt.rs
```diff
@@ -29,11 +29,14 @@ pub trait PageManagementProvider<const ALIGN: usize>: RawPointerProvider {
...
     /// `can_grow_down` specifies if the region is allowed to grow "downward" (i.e., towards zero),
     /// upon a page fault.
+    ///
+    /// `populate_pages` specifies if the pages should be populated immediately or lazily.
     fn allocate_pages(
         &self,
         range: Range<usize>,
         initial_permissions: MemoryRegionPermissions,
         can_grow_down: bool,
+        populate_pages: bool,
```
jaybosamiya-ms: Nit: a slightly more informative name that helps clarify whether true = immediate or true = lazy, because otherwise it is technically ambiguous (in both cases, pages _are_ populated, it is a question of _when_).
```suggestion
        populate_pages_immediately: bool,
```

litebox/src/mm/linux.rs
```diff
@@ -284,8 +305,16 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
...
     /// Create a new mapping in the virtual address space.
-    /// The mapping will be created at the suggested address. If the suggested start address is zero,
-    /// some available range will be choosen by the kernel.
+    ///
+    /// The start address of `suggested_range` is the hint address for where to create the pages.
+    /// Provide `0` to let the kernel choose an available memory region.
+    /// The length of `suggested_range` is the size of the pages to be created.
+    ///
+    /// Set `fixed_addr` to `true` to force the mapping to be created at the given address, resulting in any
+    /// existing overlapping mappings being removed.
```
jaybosamiya-ms: What happens if `fixed_addr` is false?  Adding something like "otherwise, page creation is allowed to fail, returning `None`" or similar would help with this comment.

litebox/src/mm/linux.rs
```diff
@@ -300,11 +329,12 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
         suggested_range: PageRange<ALIGN>,
         vma: VmArea,
         fixed_addr: bool,
+        populate_pages: bool,
```
jaybosamiya-ms: Similar to other place
```suggestion
        populate_pages_immediately: bool,
```

litebox/src/mm/mod.rs
```diff
@@ -64,6 +65,7 @@ where
         &self,
         suggested_range: PageRange<ALIGN>,
         fixed_addr: bool,
+        populate_pages: bool,
```
jaybosamiya-ms: Similar

litebox/src/mm/mod.rs
```diff
@@ -104,14 +107,23 @@ where
...
```
jaybosamiya-ms: Similar

litebox/src/mm/mod.rs
```diff
@@ -207,10 +228,19 @@ where
...
```
jaybosamiya-ms: Similar

litebox/src/mm/mod.rs
```diff
@@ -198,6 +214,11 @@ where
     /// Provide `0` to let the kernel choose an available memory region.
     /// The length of `suggested_range` is the size of the pages to be created.
     ///
+    /// Set `fixed_addr` to `true` to force the mapping to be created at the given address, resulting in any
+    /// existing overlapping mappings being removed.
```
jaybosamiya-ms, line 218: Similar

litebox/src/mm/tests.rs
```diff
@@ -27,9 +27,10 @@ impl crate::platform::PageManagementProvider<PAGE_SIZE> for DummyVmemBackend {
         range: Range<usize>,
         initial_permissions: crate::platform::page_mgmt::MemoryRegionPermissions,
         can_grow_down: bool,
+        populate_pages: bool,
     ) -> Result<Self::RawMutPointer<u8>, crate::platform::page_mgmt::AllocationError> {
-        Ok(unsafe {
-            core::mem::transmute::<*mut u8, TransparentMutPtr<u8>>(range.start as *mut u8)
+        Ok(TransparentMutPtr {
+            inner: range.start as *mut u8,
```
jaybosamiya-ms, line 33: Good catch!

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -81,9 +81,23 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
...
+        flags: VmFlags,
+        populate_pages: bool,
+    ) -> UserMutPtr<u8> {
+        if populate_pages {
+            let flags = vmflags_to_pteflags(flags);
+            for page in range {
+                let page =
+                    Page::<Size4KiB>::from_start_address(VirtAddr::new(page as u64)).unwrap();
+                unsafe {
+                    PageTableImpl::handle_page_fault(self, page, flags, PageFaultErrorCode::empty())
```
jaybosamiya-ms, line 96: Why is this not `self.handle_page_fault(...)`?

litebox_platform_linux_kernel/src/arch/x86/mm/paging.rs
```diff
@@ -80,9 +80,23 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
...
+        flags: VmFlags,
+        populate_pages: bool,
+    ) -> UserMutPtr<u8> {
+        if populate_pages {
+            let flags = vmflags_to_pteflags(flags);
+            for page in range {
+                let page =
+                    Page::<Size4KiB>::from_start_address(VirtAddr::new(page as u64)).unwrap();
+                unsafe {
+                    PageTableImpl::handle_page_fault(self, page, flags, PageFaultErrorCode::empty())
```
jaybosamiya-ms, line 95: Similar

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
CvvT: Because `X64PageTable` implements two traits, both of which have `handle_page_fault`.

litebox/src/platform/page_mgmt.rs
CvvT: Fixed

litebox/src/mm/linux.rs
CvvT: Fixed

litebox/src/mm/linux.rs
CvvT: Fixed

litebox/src/mm/mod.rs
CvvT: Fixed

litebox/src/mm/mod.rs
CvvT: Fixed

litebox/src/mm/mod.rs
CvvT: Fixed

litebox/src/mm/mod.rs
CvvT: Fixed

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
jaybosamiya-ms: Oh, I see, seems reasonable, thanks

PR #128, opened 2025-05-21
LVBS page table and inter-VTL shared memory interaction

sangho2:
This PR adds support for page table manipulation in LVBS and inter-VTL shared memory interaction. It includes
- Enable LiteBox's page management and allocator for `litebox_platform_lvbs`
- Add functions to map/unmap arbitrary (VTL0) physical frames to VTL1 kernel page tables
- Copy data from/to VTL0 through shared memory mapping
- Add some data structures like `CpuMask` and `HekiPage` for VTL0-VTL1 interaction
- Enable and capture Hyper-V synthetic interrupts

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -60,9 +60,9 @@ impl<Host: HostInterface> LinuxKernel<Host> {
     /// This function initializes the VTL1 kernel platform (mostly the kernel page table).
     /// `init_page_table_addr` specifies the physical address of the initial page table prepared by the VTL0 kernel.
     /// `phys_start` and `phys_end` specify the entire range of physical memory that is reserved for the VTL1 kernel.
-    /// Since the VTL0 kernel does not fully maps this physical address range to the initial page table, this function
+    /// Since the VTL0 kernel does not fully map this physical address range to the initial page table, this function
     /// creates and maintains a kernel page table covering the entire VTL1 physical memory range. The caller must
-    /// ensure that the heap is enough space for this page table creation.
+    /// ensure that the heap has enough space for this page table creation.
     ///
     /// # Panics
     ///


litebox_platform_lvbs/src/lib.rs
```diff
@@ -51,19 +57,164 @@ impl<Host: HostInterface> RawPointerProvider for LinuxKernel<Host> {
...
+    /// # Safety
+    ///
+    /// The caller must ensure that the `phys_addr` is a valid VTL0 physical address
+    /// # Panics
+    ///
+    /// Panics if `phys_addr` is invalid
+    pub unsafe fn copy_from_vtl0_phys<T: Copy>(
+        &self,
+        phys_addr: x86_64::PhysAddr,
+    ) -> Option<alloc::boxed::Box<T>> {
```
CvvT, line 147: Why you need to wrap it in a `Box`?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -168,16 +238,60 @@ pub fn mshv_vsm_end_of_boot() -> u64 {
...
+        unsafe { crate::platform_low().copy_from_vtl0_phys::<HekiPage>(x86_64::PhysAddr::new(pa)) }
+    {
+        // TODO: handle multi-paged input
+        for i in 0..core::cmp::min(usize::try_from(nranges).unwrap(), HEKI_MAX_RANGES) {
+            let va = heki_page.ranges[i].va;
+            let pa = heki_page.ranges[i].pa;
+            let epa = heki_page.ranges[i].epa;
+            let attr = heki_page.ranges[i].attributes;
+            let attr: MemAttr = MemAttr::from_bits(attr).unwrap_or(MemAttr::empty());
+            // TODO: load kernel data
```
CvvT: Did you forget to leave `todo!(...)`?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -51,19 +57,164 @@ impl<Host: HostInterface> RawPointerProvider for LinuxKernel<Host> {
...
 
 impl<Host: HostInterface> LinuxKernel<Host> {
-    pub fn new(init_page_table_addr: x86_64::PhysAddr) -> &'static Self {
+    /// # Panics
+    ///
+    /// Panics if `phys_start` or `phys_end` is invalid
+    pub fn new(
+        init_page_table_addr: x86_64::PhysAddr,
+        phys_start: x86_64::PhysAddr,
+        phys_end: x86_64::PhysAddr,
```
CvvT, line 74: What is the meaning of these two?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -1,21 +1,86 @@
...
+
+    if get_core_id() == 0 {
+        if let Ok((start, size)) = get_vtl1_memory_info() {
+            let vtl1_start = x86_64::PhysAddr::new(start);
+            let vtl1_end = x86_64::PhysAddr::new(start + size);
+
+            // Add a small range of mapped memory to the global allocator for populating the kernel page table.
+            let mem_fill_start = usize::try_from(Platform::pa_to_va(vtl1_start).as_u64()).unwrap()
+                + VTL1_INIT_HEAP_START_PAGE * PAGE_SIZE;
+            let mem_fill_size = VTL1_INIT_HEAP_SIZE;
```
CvvT: Is this range within the range [`vtl1_start`, `vtl1_end`) or possibly overlapping with the heap?

litebox_platform_lvbs/src/host/lvbs_impl.rs
```diff
@@ -11,33 +11,47 @@ pub type LvbsLinuxKernel = crate::LinuxKernel<HostLvbsInterface>;
...
+
+        unsafe fn free(addr: usize) {
+            unsafe { super::HostLvbsInterface::free(addr) }
+        }
+    }
 
     impl crate::mm::MemoryProvider for super::LvbsLinuxKernel {
         const GVA_OFFSET: x86_64::VirtAddr = x86_64::VirtAddr::new(0);
         const PRIVATE_PTE_MASK: u64 = 0;
+        const VTL0_GVA_OFFSET: x86_64::VirtAddr = x86_64::VirtAddr::new(VTL0_SHARED_PAGE_OFFSET);
```
CvvT: Why you need to put it at a different offset (different from GVA_OFFSET)?

litebox_platform_lvbs/src/lib.rs
sangho2: these are the start and end of VTL1's physical address range. I'll change the variable names to clarify this.

litebox_platform_lvbs/src/lib.rs
sangho2: To let Rust automatically deallocate it later.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: For now, this function is no-op (it simply prints out VTL0's request) and I want to avoid halting a VM due to the invocation of this function (it's on the critical path).

litebox_runner_lvbs/src/lib.rs
sangho2: This range is within [`vtl1_start`, `vtl1_end`] and belongs to the data section. This range is mapped to the initial page table (by the host) and is not used by any code/data.

litebox_platform_lvbs/src/host/lvbs_impl.rs
sangho2: To distinguish VTL0 addresses from VTL1 addresses (without checking page tables). This is technically no-op for now (no other existing code relies on this separation).

CvvT: I see. Is it because VTL1 can map some pages to VTL1 and later on when you free the pages, you need to know whether they belong to VTL1 or VTL0? Currently, it maps some pages but unmaps them immediately after copying the data, so we don't need it yet.

litebox_platform_lvbs/src/lib.rs
CvvT: My understanding is that they present the memory range of the kernel binary itself, not including the heap.

litebox_platform_lvbs/src/lib.rs
CvvT: Then you don't need `Box`: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#examples
Unless you don't want to use stack.

litebox_platform_lvbs/src/mshv/vsm.rs
CvvT: I see. It would be clearer if you mention it's no-op for now in the comment.

litebox_runner_lvbs/src/lib.rs
CvvT: https://github.com/MSRSSP/litebox/blob/6b5f47c52deb3c54239d1906ecbed952a35b0421/litebox_runner_lvbs/x86_64_vtl1.ld#L28-L30

The address and size are hardcoded, but you didn't specify them in the link file. How does it work?

litebox_platform_lvbs/src/lib.rs
sangho2: The size of data could be huge (e.g., up to several megabytes). So, I think stack doesn't work for this. Any other suggestion is welcomed.

litebox_platform_lvbs/src/lib.rs
sangho2: I'll document the LVBS memory layout for better understanding later (https://github.com/MSRSSP/litebox/blob/6b5f47c52deb3c54239d1906ecbed952a35b0421/litebox_platform_lvbs/src/mshv/vtl1_mem_layout.rs). Long story short, [`phys_start`, `phys_end`] is a range of physical memory addresses (i.e., 128 MiB) reserved for VTL1. VTL0 prepares an initial page table for this range but it does not fully map it (i.e., it only maps 16 MiB out of 128 MiB). Due to this, we should gradually allocate and populate virtual addresses for the heap.

litebox_runner_lvbs/src/lib.rs
sangho2: Oh, that's my mistake. In fact, `VTL1_INIT_HEAP_START` is located within [`_memory_base`, `_text_start`]. This 2 MiB area is prepared by the VTL0 kernel for various stuffs like the initial page table (https://github.com/MSRSSP/litebox/blob/6b5f47c52deb3c54239d1906ecbed952a35b0421/litebox_platform_lvbs/src/mshv/vtl1_mem_layout.rs).

litebox_platform_lvbs/src/host/lvbs_impl.rs
sangho2: True. This is like future proofing because some LVBS's VSM functions want to have identical virtual addresses between VTL0 and VTL1 kernels (to ease their implementation). I'll drop this for now because this is not yet decided.

litebox_platform_lvbs/src/lib.rs
CvvT: Sounds good! Thanks!

Cargo.toml
```diff
@@ -8,7 +8,7 @@ members = [
     "litebox_platform_multiplex",
     "litebox_runner_linux_userland",
     "litebox_shim_linux",
-    "litebox_syscall_rewriter",
+    "litebox_syscall_rewriter"
```
wdcui: Minor: is this change necessary?

litebox_platform_lvbs/src/arch/x86/interrupts.rs
```diff
@@ -30,27 +33,29 @@ pub fn init_idt() {
     idt().load();
 }
 
+// TODO: carefully handle excpetions/interrupts. If an exception or interrupt is due to userspace code,
```
wdcui: typo: exceptions

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -243,13 +250,70 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
...
+        frame_range: PhysFrameRange<Size4KiB>,
+        flags: PageTableFlags,
+        is_vtl1: bool,
+    ) -> Result<*mut u8, MapToError<Size4KiB>> {
+        let mut allocator = PageTableAllocator::<M>::new();
+
+        let mut inner = self.inner.lock();
+        for target_frame in frame_range {
+            let page: Page<Size4KiB> = if is_vtl1 {
+                Page::containing_address(M::pa_to_va(target_frame.start_address()))
```
wdcui: Minor: for naming consistency, call this vtl1_pa_to_va?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -168,16 +238,60 @@ pub fn mshv_vsm_end_of_boot() -> u64 {
...
-pub fn mshv_vsm_protect_memory(_pa: u64, _nranges: u64) -> u64 {
-    debug_serial_println!("VSM: Protect memory");
-    // TODO: protect memory using hv_modify_protection_mask()
+pub fn mshv_vsm_protect_memory(pa: u64, nranges: u64) -> u64 {
+    if let Some(heki_page) =
+        unsafe { crate::platform_low().copy_from_vtl0_phys::<HekiPage>(x86_64::PhysAddr::new(pa)) }
+    {
+        // TODO: handle multi-paged input
+        for i in 0..core::cmp::min(usize::try_from(nranges).unwrap(), HEKI_MAX_RANGES) {
+            let va = heki_page.ranges[i].va;
```
wdcui, line 247: Do we need to validate the input?

Cargo.toml
sangho2: Not needed. I'll revert that change.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: We'll drop the virtual address separation of vtl0 and vtl1 at least for now, so I think this naming consistency problem will disappear.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: LVBS doesn't trust VTL0, so it does validate all inputs from VTL0. The existing LVBS implementation has a lot more input validation/sanitization which we have not implemented yet.

litebox_platform_lvbs/src/lib.rs
```diff
@@ -51,19 +57,170 @@ impl<Host: HostInterface> RawPointerProvider for LinuxKernel<Host> {
 }
 
 impl<Host: HostInterface> LinuxKernel<Host> {
-    pub fn new(init_page_table_addr: x86_64::PhysAddr) -> &'static Self {
+    /// This function initializes the VTL1 kernel platform (mostly the kernel page table).
+    /// `init_page_table_addr` specifies the physical address of the initial page table prepared by the VTL0 kernel.
+    /// `phys_start` and `phys_end` specify the entire range of physical memory that is reserved for the VTL1 kernel.
+    /// Since the VTL0 kernel does not fully maps this physical address range to the initial page table, this function
+    /// creates and maintains a kernel page table covering the entire VTL1 physical memory range. The caller must
+    /// ensure that the heap is enough space for this page table creation.
```
sangho2: typos fixed: fully maps -> fully map, the heap is -> the heap has

PR #131, opened 2025-05-30
LVBS intercept handler and control register checking

sangho2:
This PR adds support for VTL intercept handler and control register checker. In includes
- Parse Hyper-V intercept messages
- Raise VTL0 GP fault or advance its RIP based on whether VTL0 misbehaves or not
- Store control registers and compare them against intercepted write attempts


litebox_platform_lvbs/src/mshv/hvcall_vp.rs
```diff
@@ -51,6 +51,58 @@ pub fn hvcall_set_vp_registers(
...
+        (&raw const *hvin).cast::<core::ffi::c_void>(),
+        (&raw mut *hvout).cast::<core::ffi::c_void>(),
+    )?;
+
+    Ok(hvout.as64().0)
+}
+
+/// Hyper-V Hypercall to get virtual processor (VP) registers
+#[expect(dead_code)]
+pub fn hvcall_get_vp_registers(reg_name: u32) -> Result<u64, HypervCallError> {
```
wdcui, line 104: What's the difference between 0 and TargetVtlMask::VTL0.bits() for input_vtl?

litebox_platform_lvbs/src/mshv/mod.rs
```diff
@@ -285,6 +317,66 @@ impl HvSetVpRegistersInput {
...
+}
+
+#[derive(Default, Clone, Copy)]
+#[repr(C, packed)]
+pub struct HvGetVpRegistersInputElement {
+    pub name0: u32,
+    pub name1: u32,
+}
+
+pub(crate) const HV_GET_VP_MAX_REGISTERS: usize = 1;
```
wdcui, line 349: Why set it to 1?

litebox_platform_lvbs/src/mshv/vsm_intercept.rs
```diff
@@ -0,0 +1,217 @@
...
+    HvX64RegisterTr = HV_X64_REGISTER_TR,
+    Unknown = 0xffff_ffff,
+}
+
+#[allow(clippy::too_many_lines)]
+pub fn vsm_handle_intercept() -> u64 {
+    let kernel_context = get_per_core_kernel_context();
+    let simp_page = kernel_context.hv_simp_page_as_mut_ptr();
+
+    let msg_type = unsafe { (*simp_page).sint_message[0].header.message_type };
```
wdcui, line 53: Do these two operations need to be atomic?

litebox_platform_lvbs/src/mshv/vsm_intercept.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    Ok(())
+}
+
+#[inline]
+fn check_and_write_vtl0_register(reg_name: u32, value: u64, mask: u64) -> bool {
+    let kernel_context = get_per_core_kernel_context();
+    if let Some(allowed_value) = kernel_context.vtl0_locked_regs.get(reg_name) {
+        if value & mask == allowed_value && write_vtl0_register(reg_name, value).is_ok() {
```
wdcui, line 211: Why do we need to call write_vtl0_register if the value is not changed?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -239,8 +240,7 @@ pub fn vtl_switch_loop(result: u64) -> ! {
                 vtl_switch_loop(new_result)
             }
             VtlEntryReason::Intercept => {
-                serial_println!("handle VTL intercept");
-                vtl_switch_loop(0)
+                todo!("handle VTL intercept");
```
wdcui: Did you mean to call vsm_handle_intercept?

litebox_platform_lvbs/src/mshv/hvcall_vp.rs
sangho2: Let me re-check this `input_vtl` stuff. document/code about this is unclear. LVBS code uses `input_vtl = 1 << 4` to set/get VTL0 registers.

litebox_platform_lvbs/src/mshv/mod.rs
sangho2: This hypercall allows us to access several registers at once but we only use this to get a single register for simplicity (and representing this variable-length array in Rust is a bit tricky).

litebox_platform_lvbs/src/mshv/vsm_intercept.rs
sangho2: These are per-core variables.

litebox_platform_lvbs/src/mshv/vsm_intercept.rs
sangho2: Some bits of the value can be changed, and this depends on `mask`.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: Let me remove it. We don't need to worry about `VtlEntryReason = 3`.

litebox_platform_lvbs/src/mshv/hvcall_vp.rs
sangho2: Clarified and revised `HvInputVtl` usages. The current Hyper-V TLFS and Linux kernel header file have a bit incorrect information.

litebox_platform_lvbs/src/mshv/mod.rs
```diff
@@ -883,3 +989,70 @@ pub struct HvMsrInterceptMessage {
...
+    // union of
+    // event_pending: 1;
+    // event_type: 3;
+    // reserved_0: 4;
+    // deliver_error_code: 1;
+    // reserved_1: 7;
+    // vector: 16;
+    // error_code: 32;
+}
+
```
CvvT, line 1005: I recently found two crates that could help implement bitfield-like struct: [bitfield](https://crates.io/crates/bitfield) and [modular-bitfield](https://crates.io/crates/modular-bitfield). You can decide whether it's reasonable to use it (in a separate PR maybe?).

sangho2: Another crate I considered is [bitfield-struct](https://crates.io/crates/bitfield-struct). Using one of these results in multiple (small) changes, so it might be better to consider this later.

jaybosamiya-ms: I think `bitfield` is the most popular crate amongst the three.  I think its interface is quite nicely explicit in design.

Issue #130, opened 2025-05-28
Syscall rewriter: insufficient bytes before syscall/int80 instruction


CvvT:
I just found that our syscall rewriter failed on the 32-bit hello world test we have: 

https://github.com/MSRSSP/litebox/blob/f1ecd6a5fc508ea1e13ae2332ef13c2539f291df/litebox_syscall_rewriter/tests/snapshots/snapshot_tests__hello-32-diff.snap#L104-L122

The original function `_dl_sysinfo_int80` looks like the following:
```
805ca10 <_dl_sysinfo_int80>:
805ca10:	cd 80         int    $0x80          <=== first instruction of this function!
805ca12:	c3              ret
```

The problem is the rewriter disassembles all instructions without knowing the function boundaries, and hence it overwrites the previous function (which causes crash) with no warnings. More generally, if the previous or current instruction is a jump target, we need to make sure the program does not jump to the middle of an instruction that we add.

Also, even if we know the function boundaries, we still need a different strategy to patch the code as there is no space before the int80 instruction.

Related issues: #46, #47, and #57

CvvT:
Potential solutions:

- https://www.usenix.org/conference/atc23/presentation/yasukata
- https://github.com/GJDuck/e9syscall

CvvT:
Had a brief conversation with @jaybosamiya-ms earlier today. This might be a rare case that only occurs in libc. To get around this, we could add some special handling to hook the function instead of the `int 0x80` instruction:

```
08073530 <__libc_do_syscall>:
 8073530:       f3 0f 1e fb             endbr32
 8073534:       53                      push   %ebx
 8073535:       57                      push   %edi
 8073536:       55                      push   %ebp
 8073537:       8b 1f                   mov    (%edi),%ebx
 8073539:       8b 6f 08                mov    0x8(%edi),%ebp
 807353c:       8b 7f 04                mov    0x4(%edi),%edi
 807353f:       65 ff 15 10 00 00 00    call   *%gs:0x10
 8073546:       5d                      pop    %ebp
 8073547:       5f                      pop    %edi
 8073548:       5b                      pop    %ebx
 8073549:       c3                      ret
```

Note the `call   *%gs:0x10` instruction would call `_dl_sysinfo_int80`:

```
080738f0 <_dl_sysinfo_int80>:
 80738f0:       cd 80                   int    $0x80
 80738f2:       c3                      ret
```

PR #132, opened 2025-05-31
Handle 32bit libc: int80 is at the function start of dl_sysinfo_int80

wdcui:
This PR addresses issue #130.

### DIFF
--- a/litebox_syscall_rewriter/tests/snapshots/snapshot_tests__hello-32-diff.snap
+++ b/litebox_syscall_rewriter/tests/snapshots/snapshot_tests__hello-32-diff.snap
@@ -4,28 +4,39 @@ expression: diff
 ---
 --- original
 +++ rewritten
-@@ -1279,9 +1279,10 @@
+@@ -104,7 +104,9 @@
+  804913d:	be 08 00 00 00       	mov    $0x8,%esi
+  8049142:	c7 85 d0 39 00 00 01 	movl   $0x1,0x39d0(%ebp)
+  8049149:	00 00 00
+- 804914c:	65 ff 15 10 00 00 00 	call   *%gs:0x10
++ 804914c:	e9 bb de 09 00       	jmp    80e700c <_end+0x3c4>
++ 8049151:	90                   	nop
++ 8049152:	90                   	nop
+  8049153:	8b 85 d0 39 00 00    	mov    0x39d0(%ebp),%eax
+  8049159:	83 f8 01             	cmp    $0x1,%eax
+  804915c:	0f 85 8b 00 00 00    	jne    80491ed <abort+0x136>
+@@ -1279,9 +1281,10 @@
   8049df3:	74 11                	je     8049e06 <__libc_start_call_main+0x96>
   8049df5:	ba 01 00 00 00       	mov    $0x1,%edx
   8049dfa:	31 db                	xor    %ebx,%ebx
 - 8049dfc:	8d 74 26 00          	lea    0x0(%esi,%eiz,1),%esi
 - 8049e00:	89 d0                	mov    %edx,%eax
 - 8049e02:	cd 80                	int    $0x80
-+ 8049dfc:	e9 0b d2 09 00       	jmp    80e700c <_end+0x3c4>
++ 8049dfc:	e9 1d d2 09 00       	jmp    80e701e <_end+0x3d6>
 + 8049e01:	90                   	nop
 + 8049e02:	90                   	nop
 + 8049e03:	90                   	nop
   8049e04:	eb fa                	jmp    8049e00 <__libc_start_call_main+0x90>
   8049e06:	31 c0                	xor    %eax,%eax
   8049e08:	eb d0                	jmp    8049dda <__libc_start_call_main+0x6a>
-@@ -3543,20 +3544,25 @@
+@@ -3543,20 +3546,25 @@
   804bf83:	ff
   804bf84:	c7 44 24 48 ff ff 0f 	movl   $0xfffff,0x48(%esp)
   804bf8b:	00
 - 804bf8c:	c7 44 24 4c 51 00 00 	movl   $0x51,0x4c(%esp)
 - 804bf93:	00
 - 804bf94:	cd 80                	int    $0x80
-+ 804bf8c:	e9 93 b0 09 00       	jmp    80e7024 <_end+0x3dc>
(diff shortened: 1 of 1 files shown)


litebox_syscall_rewriter/src/lib.rs
```diff
@@ -272,20 +295,26 @@ fn hook_syscalls_in_section(
             .address()
             .checked_add(inst.bytes().len().try_into().unwrap())
             .unwrap();
-        let replace_start = (0..=i)
+        let replace_start = if inst.bytes().len() >= 5 {
```
wdcui: Is this check redundant?

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -113,6 +113,13 @@ pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<usize>) -> R
         Arch::X86_32
     };
 
+    // Get symbols
+    let (libc_do_syscall, dl_sysinfo_int80) = if arch == Arch::X86_32 {
+        get_symbols(&builder)
+    } else {
+        (0, 0)
+    };
```
jaybosamiya-ms: Tracking it as `Option`s is more Rust-y than using sentinels like `0`:
```suggestion
    let (libc_do_syscall, dl_sysinfo_int80) = if arch == Arch::X86_32 {
        get_symbols(&builder)
    } else {
        (None, None)
    };
```

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -327,3 +356,21 @@ fn hook_syscalls_in_section(
 
     Ok(())
 }
+
+fn get_symbols(
+    builder: &object::build::elf::Builder<'_>,
+) -> (u64, u64) {
```
jaybosamiya-ms: This name and function signature could likely be improved:
```suggestion
fn get_symbols(
    builder: &object::build::elf::Builder<'_>,
) -> (Option<u64>, Option<u64>) {
```

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -249,13 +260,25 @@ fn hook_syscalls_in_section(
...
         match arch {
             Arch::X86_32 => {
-                // `int 0x80` (the 32-bit Linux "syscall" instruction).
-                if inst.bytes() != [0xcd, 0x80] {
+                if inst.address() == libc_do_syscall {
+                    in_libc_do_syscall = true;
+                    continue;
+                }
+                if in_libc_do_syscall {
+                    if inst.bytes() == [0x65, 0xff, 0x15, 0x10, 0x00, 0x00, 0x00] {
```
jaybosamiya-ms: These magic bytes probably need a comment for what they are
```suggestion
                    // `call DWORD PTR gs:0x10`
                    if inst.bytes() == [0x65, 0xff, 0x15, 0x10, 0x00, 0x00, 0x00] {
```

jaybosamiya-ms: Sidenote: this actually assumes that we are _guaranteed_ to get a `call DWORD PTR gs:0x10` inside the libc-do-syscall. It would be nice to actually check for `ret` as well, and make sure that if a `ret` is hit _without_ hitting the `call DWORD PTR gs:0x10` then an assumption has gone wrong, and that we should panic.

litebox_syscall_rewriter/src/lib.rs
jaybosamiya-ms: Yeah, I think it is redundant; the existing (unchanged) version of these lines _already_ looks at `i` first (notice the equal sign in `(..=i).rev()....`), thus this is already accounted for.

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs

CvvT:
Instead of replacing the `call   *%gs:0x10` instruction in `__libc_do_syscall`, we should replace `_dl_sysinfo_int80` because there are multiple places where you can find `call   *%gs:0x10`.

wdcui:
> Instead of replacing the `call   *%gs:0x10` instruction in `__libc_do_syscall`, we should replace `_dl_sysinfo_int80` because there are multiple places where you can find `call   *%gs:0x10`.

I thought we don't have enough space in _dl_sysinfo_int80? Maybe we should replace all the call sites of call gs:0x10?

CvvT:
> _dl_sysinfo_int80

Here is the disassembled code from `objdump`:

```
080738f0 <_dl_sysinfo_int80>:
 80738f0:       cd 80                   int    $0x80
 80738f2:       c3                      ret
 80738f3:       8d b4 26 00 00 00 00    lea    0x0(%esi,%eiz,1),%esi
 80738fa:       8d b6 00 00 00 00       lea    0x0(%esi),%esi

08073900 <_dl_aux_init>:
 8073900:       f3 0f 1e fb             endbr32
```

I assume the last two instructions are just padding bytes for address alignment, which we can probably overwrite.

jaybosamiya-ms:
Both of those instructions are indeed nop (specifically, they are standard [multi-byte nops](https://stackoverflow.com/questions/25545470/long-multi-byte-nops-commonly-understood-macros-or-other-notation)), likely used for padding/alignment. Safe to overwrite.

If we decide to overwrite these, I strongly recommend some assertions in place there to make sure they are indeed the bytes we expect.

I don't have strong preference on whether we should replace all call-gs ones, or just replace out all the bytes in dl-sysinfo-int80; whichever is easier is fine.

wdcui:
> Both of those instructions are indeed nop (specifically, they are standard [multi-byte nops](https://stackoverflow.com/questions/25545470/long-multi-byte-nops-commonly-understood-macros-or-other-notation)), likely used for padding/alignment. Safe to overwrite.
> 
> If we decide to overwrite these, I strongly recommend some assertions in place there to make sure they are indeed the bytes we expect.
> 
> I don't have strong preference on whether we should replace all call-gs ones, or just replace out all the bytes in dl-sysinfo-int80; whichever is easier is fine.

I think a good approach is to patch all `call gs[%10]` instructions. This way we don't need to worry about runtime patching of the Thread Control Block (which could be tricky because we need to update it at the right moment). I pushed new changes for this approach.

jaybosamiya-ms:
Just needs a snapshot update (`cargo insta review`) before merging, to make sure CI is happy. Needs to be done on 32-bit.

Issue #47, opened 2025-03-21
Handle the case of not enough space in syscall rewriter


wdcui:
We don't always have enough space to store trampolines in the code segment; need a better solution for it.

jaybosamiya-ms:
Just noting what I had said earlier: I think if we are clever, we may be able to perform some compression (e.g., via lookup tables) due to the repetitive nature of prior instructions, allowing for merging trampolines (possibly at the cost of having to do some decompression), where we can key off of the original address it came from (due to the `call` into the trampoline).

I don't know how good such compression would work, but this _still_ would only be a stop-gap measure and not handle the _general_ case (it will just increase the number of binaries we can handle). For the general case, we need to do more serious rewriting of the ELF itself.

jaybosamiya-ms:
Discussed in a meeting: if we use non-ELF as output of rewriter, we have our own loader and can leverage that.

PR #133, opened 2025-05-31
Append trampoline code to the end of the elf file

wdcui:
This PR addresses issue #47 .

### DIFF
--- a/litebox_syscall_rewriter/tests/snapshots/snapshot_tests__hello-32-diff.snap
+++ b/litebox_syscall_rewriter/tests/snapshots/snapshot_tests__hello-32-diff.snap
@@ -1,7 +1,6 @@
 ---
 source: litebox_syscall_rewriter/tests/snapshot_tests.rs
 expression: diff
-snapshot_kind: text
 ---
 --- original
 +++ rewritten
@@ -159,7 +158,7 @@ snapshot_kind: text
  
  08098e40 <__libc_sigaction>:
   8098e40:	55                   	push   %ebp
-@@ -124321,4 +124331,19 @@
+@@ -124321,4 +124331,21 @@
   80ae605:	81 c3 ef 49 03 00    	add    $0x349ef,%ebx
   80ae60b:	83 c4 08             	add    $0x8,%esp
   80ae60e:	5b                   	pop    %ebx
@@ -180,5 +179,7 @@ snapshot_kind: text
 + 80ae61b:	08 00                	or     %al,(%eax)
 + 80ae61d:	00 00                	add    %al,(%eax)
 + 80ae61f:	00 4c 01 00          	add    %cl,0x0(%ecx,%eax,1)
++ 80ae623:	00 00                	add    %al,(%eax)
++ 80ae625:	00 00                	add    %al,(%eax)
 +	...
 \ No newline at end of file


litebox_syscall_rewriter/src/lib.rs
```diff
@@ -327,3 +331,21 @@ fn hook_syscalls_in_section(
...
+    for seg in &builder.segments {
+        if seg.p_type == object::elf::PT_LOAD {
+            let end_addr = seg.p_vaddr + seg.p_memsz;
+            if end_addr > max_virtual_addr {
+                max_virtual_addr = end_addr;
+            }
+        }
+    }
+    // Round up to the nearest page (assume 0x1000 page size)
+    max_virtual_addr.div_ceil(0x1000) * 0x1000
```
jaybosamiya-ms: Rust has https://doc.rust-lang.org/std/primitive.u64.html#method.next_multiple_of for this
```suggestion
    max_virtual_addr.next_multiple_of(0x1000)
```

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -327,3 +331,21 @@ fn hook_syscalls_in_section(
...
+
+    // Find the highest virtual address among all sections in executable segments
+    for seg in &builder.segments {
+        if seg.p_type == object::elf::PT_LOAD {
+            let end_addr = seg.p_vaddr + seg.p_memsz;
+            if end_addr > max_virtual_addr {
+                max_virtual_addr = end_addr;
+            }
+        }
+    }
```
jaybosamiya-ms: Fwiw, this loop can be written in a more Rust-y way using:
```suggestion
    let max_virtual_addr = builder.segments.iter()
      .filter(|seg| seg.p_type == object::elf::PT_LOAD)
      .map(|seg| seg.p_vaddr + seg.p_memsz)
      .max().unwrap();
```

It makes reading the code almost like a natural sentence: "the max virutal address is computed by taking the `PT_LOAD` segments, computing end addresses, and taking the max".  This also automatically makes sure there is a panic if there is no `PT_LOAD` segment (we don't actually expect this to ever happen, but nonetheless, this one makes it clearer that we are expecting at least one `PT_LOAD` segment to exist in the binary).

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -181,6 +188,10 @@ pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<usize>) -> R
     builder
         .write(&mut out)
         .map_err(|e| Error::GenerateObjFileError(e.to_string()))?;
+    // ensure the start address of the trampoline code is page-aligned
```
wdcui, line 195: @CvvT I don't think we need to align the trampoline data in the elf file.

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -138,9 +138,13 @@ pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<usize>) -> R
         .segments
         .get_mut(executable_segment)
         .append_section(builder.sections.get_mut(trampoline_section));
-    let trampoline_base_addr = builder.sections.get(trampoline_section).sh_addr;
+
+    let trampoline_base_addr = find_addr_for_trampoline_code(&builder);
 
     let mut trampoline_data = vec![];
+    // The magic prefix for the trampoline section
+    trampoline_data.extend_from_slice("LITEBOX0".as_bytes());
```
jaybosamiya-ms: Ideally, we would expose a `pub const` for the `LITEBOX0` u64, such that the `litebox_shim_linux` could pick it up, but I don't think we want to do that because `std` reasons.  So for now, at least a comment saying where things need to get updated would be helpful.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -168,10 +172,13 @@ pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<usize>) -> R
         return Err(Error::NoSyscallInstructionsFound);
     }
 
-    builder.sections.get_mut(trampoline_section).sh_size =
-        trampoline_data.len().try_into().unwrap();
+    let mut trampoline_vec = Vec::new();
+    trampoline_vec.extend_from_slice("LITE BOX".as_bytes());
```
jaybosamiya-ms, line 180: Similar

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -168,10 +172,13 @@ pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<usize>) -> R
...
+    trampoline_vec.extend_from_slice(&trampoline_base_addr.to_le_bytes());
+    trampoline_vec.extend_from_slice(&trampoline_data.len().to_le_bytes());
```
jaybosamiya-ms: I think this makes it dependent on the rewriter host size (since it is a `usize`).  I think we want to force this to always be a `u64` to prevent possible confusion in future.

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -193,10 +193,24 @@ type Shdr = elf::section::Elf64_Shdr;
...
 
-/// Get the syscall callback placeholder address from the ELF file if it is modified by our syscall
-/// rewriter. The placeholder initially has a magic number that is going to be replaced by the actual
-/// syscall callback.
-fn get_syscall_callback_placeholder(object: &mut ElfFile) -> Option<NonNull<usize>> {
+#[repr(C, packed)]
+struct TrampolineSection {
+    magic_number: u64,
+    trampoline_addr: u64,
+    trampoline_size: usize,
```
jaybosamiya-ms: When updating things at the rewriter side, this should be updated too

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs
wdcui: Fixed.

Issue #46, opened 2025-03-21
Handle RIP-dependent instructions in syscall rewriter


wdcui:
There could be an instruction that's RIP dependent before a syscall. When we move it, we need to adjust it based on the new RIP.

PR #134, opened 2025-06-01
Check branch instructions before syscalls

wdcui:
This PR addresses issue #46 .

### DIFF
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -335,7 +335,13 @@ fn hook_syscalls_in_section(
             if replace_end - prev_inst.address() >= 5 {
                 replace_start = Some(prev_inst.address());
                 break;
-            } else if libc_start_call_main.is_none_or(|(addr, size)| {
+            }
+            // We ignore this check inside the __libc_start_call_main function
+            // because it has a redundant backward jmp to repeat the exit syscall
+            // which is not supposed to be returned from the kernel.
+            // Our current patching approach cannot handle this case yet.
+            // A potential solution is to patch both before and after the syscall instruction.
+            else if libc_start_call_main.is_none_or(|(addr, size)| {
                 prev_inst.address() < addr || prev_inst.address() >= addr + size
             }) && control_transfer_targets.contains(&prev_inst.address())
             {
@@ -554,7 +560,7 @@ fn hook_syscall_and_after(
     // Add call [rip + offset_to_shared_target]
     if arch == Arch::X86_64 {
         trampoline_data.extend_from_slice(&[0xFF, 0x15]);
-        let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() + 4);
+        let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() - 4);
         trampoline_data.extend_from_slice(&disp32.to_le_bytes());
     } else {
         // For 32-bit, use a different approach to simulate `call [rip + disp32]`
@@ -562,7 +568,7 @@ fn hook_syscall_and_after(
         trampoline_data.extend_from_slice(&[0xE8, 0x0, 0x0, 0x0, 0x0]); // CALL next instruction
         trampoline_data.push(0x58); // POP EAX (effectively store IP in EAX)
         trampoline_data.extend_from_slice(&[0xFF, 0x90]); // CALL [EAX + offset]
-        let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() - 3);
+        let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() - 11);
         trampoline_data.extend_from_slice(&disp32.to_le_bytes());
         // Note we skip `POP EAX` here as it is done by the callback `syscall_callback`
         // from litebox_shim_linux/src/lib.rs, which helps reduce the size of the trampoline.


litebox_syscall_rewriter/src/lib.rs
```diff
@@ -272,11 +274,27 @@ fn hook_syscalls_in_section(
...
+            if is_control_transfer {
+                // If it's a control transfer, we don't want to cross it
+                return Err(Error::InsufficientBytesBefore(inst.address()));
+            }
+            if replace_end - prev_inst.address() >= 5 {
+                replace_start = prev_inst.address();
+                break;
+            }
+        }
+
```
jaybosamiya-ms, line 352: With the current implementation, it is _technically_ feasible for a situation where `replace_start = 0` after the end of the `for` loop, which can be problematic and hard to debug.

If you want to use the for loop implementation, you can change it to `let mut replace_start = None;`, assign it with a `replace_start = Some(...)`, and then place a `let replace_start = replace_start.ok_or_else(|| Error::InsufficientBytesBefore(inst.address());` at the end here.  With that in place, you can switch out the `if is_control_transfer` body with just a `break;`, that way both cases are accounted for correctly.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -272,11 +274,42 @@ fn hook_syscalls_in_section(
...
+        let mut replace_start = None;
+        for inst_id in (0..=i).rev() {
+            let prev_inst = &instructions[inst_id];
+            let prev_inst_detail = cs.insn_detail(prev_inst).unwrap();
+            // Check if the instruction is an instruction-relative control transfer
+            let is_control_transfer = prev_inst_detail.groups().iter().any(|&grp| {
+                grp.0 == u8::try_from(X86_GRP_JUMP).unwrap()
+                    || grp.0 == u8::try_from(X86_GRP_CALL).unwrap()
+                    || grp.0 == u8::try_from(X86_GRP_RET).unwrap()
+            });
```
jaybosamiya-ms: This comment is technically incorrect, yes?  We might want to update it:
```suggestion
            // Check if the instruction does control transfer
            // TODO: Check if the instruction is an instruction-relative control transfer
            let is_control_transfer = prev_inst_detail.groups().iter().any(|&grp| {
                grp.0 == u8::try_from(X86_GRP_JUMP).unwrap()
                    || grp.0 == u8::try_from(X86_GRP_CALL).unwrap()
                    || grp.0 == u8::try_from(X86_GRP_RET).unwrap()
            });
```

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -327,3 +360,98 @@ fn hook_syscalls_in_section(
...
+) -> Result<()> {
+    let syscall_inst = &instructions[inst_index];
+
+    let replace_start = syscall_inst.address();
+
+    let mut replace_end = None;
+
+    for next_inst in instructions.iter().skip(inst_index) {
+        let next_inst_detail = cs.insn_detail(next_inst).unwrap();
+        // Check if the instruction is an instruction-relative control transfer
```
jaybosamiya-ms: Similar

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -327,3 +360,98 @@ fn hook_syscalls_in_section(
...
+            .unwrap();
+
+        if next_end - syscall_inst.address() >= 5 {
+            replace_end = Some(next_end);
+            break;
+        }
+    }
+
+    let replace_end =
+        replace_end.ok_or_else(|| Error::InsufficientBytesBefore(syscall_inst.address()))?;
```
jaybosamiya-ms: We may wish to extend the `Error` enum to `InsufficientBytesBeforeOrAfter`

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -327,3 +360,98 @@ fn hook_syscalls_in_section(
 
     Ok(())
 }
+
+#[allow(clippy::too_many_arguments)]
+fn hook_syscall_and_after(
```
jaybosamiya-ms, line 503: I think this function has a lot of repetition with the `hook_syscalls_in_section` function, and we might be able to refactor things better.  I don't think we should block merging on that though, just making a note here.

wdcui: I agree with you. The code in these two functions is similar but also has subtle differences. I also thought about moving some code from `hook_syscalls_in_section` to a new function `hook_syscall_and_before`. Then I decided to leave it as a low priority for now.

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub enum Error {
     #[error("failed to disassemble: {0}")]
     DisassemblyFailure(String),
     #[error("insufficient bytes before syscall at {0:#x}")]
-    InsufficientBytesBefore(u64),
+    InsufficientBytesBeforeOrAfter(u64),
```
jaybosamiya-ms: Minor nit:
```suggestion
    #[error("insufficient bytes before or after syscall at {0:#x}")]
    InsufficientBytesBeforeOrAfter(u64),
```

CvvT:
I found a failure case:

```
00000000004468e0 <_exit>:
  4468e0:       f3 0f 1e fa             endbr64
  4468e4:       49 c7 c0 b8 ff ff ff    mov    $0xffffffffffffffb8,%r8
  4468eb:       be e7 00 00 00          mov    $0xe7,%esi
  4468f0:       ba 3c 00 00 00          mov    $0x3c,%edx
  4468f5:       eb 16                   jmp    44690d <_exit+0x2d>
  4468f7:       66 0f 1f 84 00 00 00    nopw   0x0(%rax,%rax,1)
  4468fe:       00 00
  446900:       89 d0                   mov    %edx,%eax
  446902:       0f 05                   syscall
  446904:       48 3d 00 f0 ff ff       cmp    $0xfffffffffffff000,%rax
  44690a:       77 1c                   ja     446928 <_exit+0x48>
  44690c:       f4                      hlt
  44690d:       89 f0                   mov    %esi,%eax
  44690f:       0f 05                   syscall
  446911:       48 3d 00 f0 ff ff       cmp    $0xfffffffffffff000,%rax
  446917:       76 e7                   jbe    446900 <_exit+0x20>
  446919:       f7 d8                   neg    %eax
  44691b:       64 41 89 00             mov    %eax,%fs:(%r8)
  44691f:       eb df                   jmp    446900 <_exit+0x20>
  446921:       0f 1f 80 00 00 00 00    nopl   0x0(%rax)
  446928:       f7 d8                   neg    %eax
  44692a:       64 41 89 00             mov    %eax,%fs:(%r8)
  44692e:       eb dc                   jmp    44690c <_exit+0x2c>
```

and here is the code after the rewriting:

```
00000000004468e0 <_exit>:
  4468e0:       f3 0f 1e fa             endbr64
  4468e4:       49 c7 c0 b8 ff ff ff    mov    $0xffffffffffffffb8,%r8
  4468eb:       be e7 00 00 00          mov    $0xe7,%esi
  4468f0:       ba 3c 00 00 00          mov    $0x3c,%edx
  4468f5:       eb 16                   jmp    44690d <_exit+0x2d>
  4468f7:       e9 ec 68 08 00          jmp    4cd1e8 <_end+0x5a8>
  4468fc:       90                      nop
  4468fd:       90                      nop
  4468fe:       90                      nop
  4468ff:       90                      nop
  446900:       90                      nop
  446901:       90                      nop
  446902:       90                      nop
  446903:       90                      nop
  446904:       48 3d 00 f0 ff ff       cmp    $0xfffffffffffff000,%rax
  44690a:       77 1c                   ja     446928 <_exit+0x48>
  44690c:       e9 ed 68 08 00          jmp    4cd1fe <_end+0x5be>
  446911:       48 3d 00 f0 ff ff       cmp    $0xfffffffffffff000,%rax
  446917:       76 e7                   jbe    446900 <_exit+0x20>
  446919:       f7 d8                   neg    %eax
  44691b:       64 41 89 00             mov    %eax,%fs:(%r8)
  44691f:       eb df                   jmp    446900 <_exit+0x20>
  446921:       0f 1f 80 00 00 00 00    nopl   0x0(%rax)
  446928:       f7 d8                   neg    %eax
  44692a:       64 41 89 00             mov    %eax,%fs:(%r8)
  44692e:       eb dc                   jmp    44690c <_exit+0x2c>
```

Note that there is `jmp    44690d` at 0x4468f5, which would jump to the middle of a patched instruction.

See the second point of #140

jaybosamiya-ms:
Hmm, we might want to calculate a list of _direct_ jump targets, and similar to the check of "there is control flow here", we might wish to check for "there is control flow _to_ here" too.  Would be even more conservative, but I think at least in this case, it'd force it to "after" case which is safe.

litebox_syscall_rewriter/src/lib.rs
wdcui: Good catch! Fixed.

wdcui:
@CvvT @jaybosamiya-ms I pushed new changes to check if there exists a direct control transfer into the middle of a patched range. Please take a look.

I also need to handle the special case in 32-bit libc code. There is a super tight loop for `exit(1)`.  I had to special case it.

 8049df5:       ba 01 00 00 00          mov    $0x1,%edx
 8049dfa:       31 db                   xor    %ebx,%ebx
 8049dfc:       8d 74 26 00             lea    0x0(%esi,%eiz,1),%esi
 8049e00:       89 d0                   mov    %edx,%eax
 8049e02:       cd 80                   int    $0x80
 8049e04:       eb fa                   jmp    8049e00 <__libc_start_call_main+0x90>

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -302,11 +315,51 @@ fn hook_syscalls_in_section(
...
-            .map(|idx| instructions[idx].address())
-            .find(|addr| replace_end - addr >= 5)
-            .ok_or_else(|| Error::InsufficientBytesBefore(inst.address()))?;
+
+        let mut replace_start = None;
+        for inst_id in (0..=i).rev() {
+            let prev_inst = &instructions[inst_id];
+            let prev_inst_detail = cs.insn_detail(prev_inst).unwrap();
+            // Check if the instruction does control transfer
+            // TODO: Check if the instruction is an instruction-relative control transfer
```
CvvT, line 324: Forgot to remove `TODO`?

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -302,11 +315,51 @@ fn hook_syscalls_in_section(
...
+            if is_control_transfer {
+                // If it's a control transfer, we don't want to cross it
+                break;
+            }
+            if replace_end - prev_inst.address() >= 5 {
+                replace_start = Some(prev_inst.address());
+                break;
+            } else if libc_start_call_main.is_none_or(|(addr, size)| {
+                prev_inst.address() < addr || prev_inst.address() >= addr + size
+            }) && control_transfer_targets.contains(&prev_inst.address())
```
CvvT: Is it saying we don't check control transfer inside `libc_start_call_main` function? If so, adding some reasoning would be helpful.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -374,15 +427,178 @@ fn find_addr_for_trampoline_code(builder: &object::build::elf::Builder<'_>) -> u
...
+
+    let replace_end =
+        replace_end.ok_or_else(|| Error::InsufficientBytesBeforeOrAfter(syscall_inst.address()))?;
+
+    let target_addr = trampoline_base_addr + trampoline_data.len() as u64;
+
+    // Add call [rip + offset_to_shared_target]
+    if arch == Arch::X86_64 {
+        trampoline_data.extend_from_slice(&[0xFF, 0x15]);
+        let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() + 4);
```
CvvT: Copied from the old file? It's `- 4` in `hook_syscalls_in_section`.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -374,15 +427,178 @@ fn find_addr_for_trampoline_code(builder: &object::build::elf::Builder<'_>) -> u
...
+        trampoline_data.extend_from_slice(&disp32.to_le_bytes());
+    } else {
+        // For 32-bit, use a different approach to simulate `call [rip + disp32]`
+        trampoline_data.push(0x50); // PUSH EAX
+        trampoline_data.extend_from_slice(&[0xE8, 0x0, 0x0, 0x0, 0x0]); // CALL next instruction
+        trampoline_data.push(0x58); // POP EAX (effectively store IP in EAX)
+        trampoline_data.extend_from_slice(&[0xFF, 0x90]); // CALL [EAX + offset]
+        let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() - 3);
```
CvvT: Similarly, it's `-11`.

litebox_syscall_rewriter/src/lib.rs
wdcui: This TODO is correct. Right now the code only checks control transfers and doesn't check if it's instruction-relative.

litebox_syscall_rewriter/src/lib.rs
wdcui: Yes, we don't check if a relocated instruction is a control transfer target inside this function. I will add a comment to explain it.

litebox_syscall_rewriter/src/lib.rs
wdcui: This is the result of having 3 branches working the same code simultaneously. I will correct it.

litebox_syscall_rewriter/src/lib.rs
wdcui: Yes, it needs to be updated.

PR #135, opened 2025-06-02
Refine CI trigger to only main branch for push events

jaybosamiya-ms:
This PR reduces our CI load to not automatically trigger the CI on all push events; instead, it will only trigger on PRs (or on pushes to the `main` branch).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
This PR was triggered by noticing that our GitHub actions usage has almost doubled in the past month or so as we added new CI runs (32-bit, LVBS, no_std), and near the end of the month, it started causing CI failures (e.g., see #133) since we are still on the GitHub free plan.

jaybosamiya-ms:
@wdcui @CvvT @sangho2 : pinging you folks since this change will impact your branches going forward.  I will merge this PR in a day or so.

PR #137, opened 2025-06-03
Add support for arch_prctl & remove fs swapping

CvvT:
As mentioned in #66, the current implement does not support maintaining per-thread fs register values. The reason why we need to do it was because both the target program we run and litebox use libc, but only one TLS is created when a new thread is spawned. #111  and #115 address this issue by removing the use of libc for allocator and syscalls (both use thread-local variables). Currently, litebox only use libc for constants and struct definitions. Thus, we no longer need to swap the fs register value when switching between litebox and the target program. Eventually, we may still need to swap fs or gs register at the syscall entry if we need to support TLS (some syscalls require maintaining some per-thread metadata).

This PR also adds support for syscall `arch_prctl` for setting and getting fs register values.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -521,7 +521,7 @@ pub enum ArchPrctlCode {
     #[cfg(target_arch = "x86_64")]
     GetFs = 0x1003,
 
-    /* CET (Control-flow Enforcement Technology) ralated operations */
+    /* CET (Control-flow Enforcement Technology) ralated operations; each of these simply will return EINVAL */
     CETStatus = 0x3001,
     CETDisable = 0x3002,
     CETLock = 0x3003,
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -427,7 +427,7 @@ const HWCAP2_FSGSBASE: u64 = 1 << 1;
 
 /// Get the current fs base register value.
 ///
-/// Depending on whether `fsgsbase` instructions are enabled, we can choose
+/// Depending on whether `fsgsbase` instructions are enabled, we choose
 /// between `arch_prctl` or `rdfsbase` to get the fs base.
 #[cfg(target_arch = "x86_64")]
 fn get_fs_base() -> Result<usize, litebox_common_linux::errno::Errno> {
@@ -445,7 +445,7 @@ fn get_fs_base() -> Result<usize, litebox_common_linux::errno::Errno> {
 
 /// Set the fs base register value.
 ///
-/// Depending on whether `fsgsbase` instructions are enabled, we can choose
+/// Depending on whether `fsgsbase` instructions are enabled, we choose
 /// between `arch_prctl` or `wrfsbase` to set the fs base.
 #[cfg(target_arch = "x86_64")]
 fn set_fs_base(fs_base: usize) -> Result<usize, litebox_common_linux::errno::Errno> {


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -421,6 +421,88 @@ impl litebox::platform::Instant for Instant {
...
 }
 
+// from asm/hwcap2.h
+#[cfg(target_arch = "x86_64")]
+const HWCAP2_FSGSBASE: u64 = 1 << 1;
+
+/// Get the current fs base register value.
+///
+/// Depending on whether `fsgsbase` instructions are enabled, we can choose
+/// between `arch_prctl` or `rdfsbase` to get the fs base.
```
jaybosamiya-ms: Nit:
```suggestion
/// Depending on whether `fsgsbase` instructions are enabled, we choose
/// between `arch_prctl` or `rdfsbase` to get the fs base.
```

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -421,6 +421,88 @@ impl litebox::platform::Instant for Instant {
...
+#[cfg(target_arch = "x86_64")]
+fn get_fs_base() -> Result<usize, litebox_common_linux::errno::Errno> {
+    /// Function pointer to get the current fs base.
+    static GET_FS_BASE: spin::Once<fn() -> Result<usize, litebox_common_linux::errno::Errno>> =
+        spin::Once::new();
+    GET_FS_BASE.call_once(|| {
+        if unsafe { libc::getauxval(libc::AT_HWCAP2) } & HWCAP2_FSGSBASE != 0 {
```
jaybosamiya-ms, line 438: I think `unsafe { core::arch::x86::__cpuid(7).ebx } & 1` (assuming the constants [here](https://en.wikipedia.org/wiki/CPUID#EAX=7,_ECX=0:_Extended_Features) are correct) should be able to get us this info without needing to depend on libc, correct?

jaybosamiya-ms: Otherwise I think `getauxval` reads from the aux vector, which I think we need to confirm that the loader is setting up properly (I haven't checked, but adding this comment here as something one of us should confirm)

litebox_common_linux/src/lib.rs
```diff
@@ -413,6 +413,79 @@ pub enum Protocol {
...
+    #[cfg(target_arch = "x86_64")]
+    SetFs = 0x1002,
+    /// Return the 64-bit base value for the FS register of the calling thread
+    #[cfg(target_arch = "x86_64")]
+    GetFs = 0x1003,
+
+    /* CET (Control-flow Enforcement Technology) ralated operations */
+    CETStatus = 0x3001,
+    CETDisable = 0x3002,
+    CETLock = 0x3003,
```
jaybosamiya-ms: It appears that we simply return `EINVAL` for these, in which case including a quick comment here might be helpful, e.g.,
```suggestion
    /* CET (Control-flow Enforcement Technology) ralated operations; each of these simply will return EINVAL */
    CETStatus = 0x3001,
    CETDisable = 0x3002,
    CETLock = 0x3003,
```

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -421,6 +421,88 @@ impl litebox::platform::Instant for Instant {
...
+            || Ok(unsafe { litebox_common_linux::rdfsbase() })
+        } else {
+            get_fs_base_arch_prctl
+        }
+    })()
+}
+
+/// Set the fs base register value.
+///
+/// Depending on whether `fsgsbase` instructions are enabled, we can choose
```
```suggestion
/// Depending on whether `fsgsbase` instructions are enabled, we choose
```

litebox_common_linux/src/lib.rs

litebox_platform_linux_userland/src/lib.rs

litebox_platform_linux_userland/src/lib.rs

litebox_platform_linux_userland/src/lib.rs
CvvT: I believe `cpuid` only tell you whether this is instruction is available or not. Kernel still needs to enable it for userspace by setting `CR4` register properly (see https://docs.kernel.org/arch/x86/x86_64/fsgs.html#fsgsbase-instructions-enablement). 

You can see the aux vector by running `LD_SHOW_AUXV=1 /bin/true`. Note the aux vector litebox reads here is different from the one our elf loader sets.

We do need to be careful about `getauxval` when using it to get other values. I just did a quick check on the implementation: https://elixir.bootlin.com/glibc/glibc-2.41.9000/source/misc/getauxval.c#L62. If we use it to get a non-existing entry, it would fail and set the errno (which requires access to TLS). I will create an issue to track it. Reimplementing this is feasible, just requires more work.

jaybosamiya-ms: Fair point, although I will also note that Linux enables it by default since ~5 years ago: https://github.com/torvalds/linux/commit/b745cfba44c152c34363eea9e052367b6b1d652b

For now, `getauxval` seems reasonable to me, as long as we open a tracking issue, thanks!

CvvT: #142  Good point! We don't even need to reimplement it.

PR #138, opened 2025-06-03
Add exit/exit_group

CvvT:
We need to differentiate between `exit` and `exit_group` as `exit` only terminates the current calling thread while `exit_group` terminates the whole process.

### DIFF
--- a/litebox/src/litebox.rs
+++ b/litebox/src/litebox.rs
@@ -33,7 +33,7 @@ impl<Platform: RawSyncPrimitivesProvider + ExitProvider> LiteBox<Platform> {
         // TODO(jayb): After #24, #31, we will be able to pass along clean-up operations to
         // subcomponents, to request a clean-up. For now, there is no clean-up necessary, we can
         // just exit.
-        self.x.platform.exit(exit_code)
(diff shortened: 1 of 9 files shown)


jaybosamiya-ms:
Just making a quick note that we should discuss this IRL a little bit before I review it.

jaybosamiya-ms:
Discussed IRL with @CvvT. We decided that it makes more sense to have thread-related functionality in its own "Provider" interface. Specifically, adding this to `ExitProvider` when we don't have the ability to spawn threads makes it feel a little asymmetric. Instead, a new `ThreadProvider` interface (with `thread_spawn`/`thread_die` functions, similar to `clone3`/`exit`) should work better. We also thought through briefly whether such an interface would work ok, or be problematic if/when we attempt a Windows platform, and concluded that as long as we have a (subset of) `clone3`, and are willing to write a trampoline function on Windows, such a `ThreadProvider` interface should work well.

---

TL;DR: This particular PR should **not** be merged, instead a new PR will be made for `ThreadProvider` described above.

jaybosamiya-ms:
Closing this as discussed above. Separate PR should be opened for `ThreadProvider`

PR #139, opened 2025-06-04
Add sys_brk

CvvT:
[brk](https://www.man7.org/linux/man-pages/man2/brk.2.html) change the location of the program break, which defines the end of the process's data segment (i.e., the program break is the first location after the end of the uninitialized data segment).  Increasing the program break has the effect of allocating memory to the process; decreasing the break deallocates memory.

A typical Linux process memory layout:
```
+--------------------------+ 0xC0000000 or 0x7ffffffff000
|     Kernel Space         | (not accessible from userland)
+--------------------------+
|     Stack (grows down)   |
+--------------------------+
|     Mmap area (down)     |
|--------------------------| ← mmap base (e.g., 0x7fff0000)
|     Shared libraries     |
|     Heap (brk, grows up) |
|--------------------------| ← brk start
|     BSS / Data / Text    |
|--------------------------| ← program load addr (e.g., 0x400000)
```

To avoid memory overlapping, the `brk` heap starts after the end of the binary's .data section and goes upward, while the mmap allocations typically start just below the stack and go downward. This PR also fixes `get_unmmaped_area` so that it returns memory from high addresses to low addresses, conforming to the Linux kernel's memory layout.

Also add a new API `reserved_pages` to make certain regions as allocated.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -149,7 +149,7 @@ impl LinuxUserland {
             }
             total_read += n;
         }
-        assert!(total_read != buf.len(), "buffer too small");
+        assert!(total_read < buf.len(), "buffer too small");
 
         let mut reserved_pages = alloc::vec::Vec::new();
         let s = core::str::from_utf8(&buf[..total_read]).expect("invalid UTF-8");


litebox/src/platform/page_mgmt.rs
```diff
@@ -73,6 +73,11 @@ pub trait PageManagementProvider<const ALIGN: usize>: RawPointerProvider {
         range: Range<usize>,
         new_permissions: MemoryRegionPermissions,
     ) -> Result<(), PermissionUpdateError>;
+
+    /// Return reserved pages that are not available for allocation.
+    fn reserved_pages(&self) -> impl Iterator<Item = crate::mm::linux::PageRange<ALIGN>> {
+        core::iter::empty()
+    }
```
jaybosamiya-ms: It would be preferable for this not to have a default body, since we actually expect to override it in some platforms at least.  By convention, default bodies for Rust trait functions contain things that you do not expect to override at all.  I know that technically it reduces code duplication, but using defaults for sometimes-overridden purposes makes it harder for the compiler to tell you when you need to/want to override the function.  In this case, an actual decision _must_ be made by the implementor of the trait, thus it should not be a default.

jaybosamiya-ms: Also, I am surprised to see a `crate::mm::linux::...` show up here, which feels like it breaks the layering that we have.  Given that the other functions in this trait all use `Range<usize>` to represent areas in memory, this should also just be that.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -781,6 +785,60 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Li
         .expect("mprotect failed");
         Ok(())
     }
+
+    fn reserved_pages(&self) -> impl Iterator<Item = PageRange<ALIGN>> {
+        // Note this function should be called before we switch to Litebox's FS as it does not
+        // support reading `/proc/self/maps` yet.
```
jaybosamiya-ms: We can guarantee this by running it upon `LinuxUserland::new` and storing the data into the `LinuxUserland` struct; then the `reserved_pages` would only need to be a tiny function that returns that `self.reserved_pages.iter()`.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -781,6 +785,60 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Li
...
+            .expect("read failed");
+            if n == 0 {
+                break;
+            }
+            total_read += n;
+        }
+        assert!(total_read != buf.len(), "buffer too small");
+
+        let mut reserved_pages = alloc::vec::Vec::<PageRange<ALIGN>>::new();
+        let s = core::str::from_utf8(&buf[..total_read]).expect("invalid UTF-8");
```
jaybosamiya-ms: This will read the first 4096 bytes only. If there are more bytes, it will silently truncate, if there are fewer, it will complain. I don't think this code is doing what you intend it to do.

Fwiw, on my local zsh, just to get a quick idea of what size `/proc/self/maps` tends to be, and what kinds of bits can get silently truncated:
```console
$ cat /proc/self/maps | wc -c
4194

$ cat /proc/self/maps | cut -zb 4096-
            [vvar]
7ffcc60ec000-7ffcc60ee000 r-xp 00000000 00:00 0                          [vdso]
```

litebox/src/mm/mod.rs
```diff
@@ -243,6 +243,55 @@ where
...
+            // to the end of the data segment.
+            vmem.brk = brk;
+            return Ok(brk);
+        } else if brk == 0 {
+            // Calling `brk` with 0 can be used to find the current location of the program break.
+            return Ok(vmem.brk);
+        } else if old_brk >= brk {
+            // If the old brk is already greater than or equal to the new brk, we can just return
+            // TODO: shrink the memory region
+            return Ok(brk);
```
jaybosamiya-ms: Due to this `TODO`, do we want to change this to a `todo!()` so we can catch this when it happens?

litebox/src/mm/mod.rs
```diff
@@ -243,6 +243,55 @@ where
...
+    ///
+    /// Increasing the program break has the effect of allocating memory to the process;
+    /// decreasing the break deallocates memory.
+    ///
+    /// Note the initial program break is set to zero and the first call to `brk` would set it
+    /// to the given address, which is usually the end of the data segment.
+    ///
+    /// # Safety
+    ///
+    /// If shrinking the program break, the caller must ensure that the released memory region is no longer used.
```
jaybosamiya-ms: Might want to add something like "A `brk` of 0 leaves the break unchanged." and "Upon success, returns the new program break."

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -95,6 +95,8 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
     // `litebox_platform_linux_userland` does not provide a way to pick between the two.
     let platform = Platform::new(None);
     let litebox = LiteBox::new(platform);
+    let pm = litebox::mm::PageManager::new(&litebox);
+    litebox_shim_linux::set_page_manager(pm);
```
jaybosamiya-ms: Is this needed due to the ordering of operations (for reading `/proc/self/maps` issue)?  If so, fixing up that issue would allow us to remove this change, and keep the runner simpler.

litebox_shim_linux/src/lib.rs
```diff
@@ -77,9 +77,20 @@ pub fn litebox_fs<'a>() -> &'a impl litebox::fs::FileSystem {
...
+    VMEM.set(alloc::boxed::Box::new(page_manager))
+        .map_err(|_| {})
+        .expect("page manager is already set");
+}
+
 pub(crate) fn litebox_page_manager<'a>() -> &'a PageManager<Platform, PAGE_SIZE> {
-    static VMEM: OnceBox<PageManager<Platform, PAGE_SIZE>> = OnceBox::new();
-    VMEM.get_or_init(|| alloc::boxed::Box::new(PageManager::new(litebox())))
+    VMEM.get().expect("page manager has not yet been set")
 }
```
jaybosamiya-ms: Same comment as `litebox_runner_linux_userland/src/lib.rs`:
> Is this needed due to the ordering of operations (for reading `/proc/self/maps` issue)? If so, fixing up that issue would allow us to remove this change, and keep the runner simpler.

litebox/src/mm/mod.rs
CvvT: Added

litebox/src/mm/mod.rs
CvvT: I have seen use cases of this. It just seems okay not to unmap the pages. One minor concern would be memory info leak. Since it only takes a few lines of code to unmap the unused pages, I just added it.

litebox/src/platform/page_mgmt.rs
CvvT: I see. Good to learn that. I just changed PageRange to Range.

litebox_platform_linux_userland/src/lib.rs
CvvT: It's less than 4k on my machine. I doubled the size just to be safer.

I assume the buffer size is large enough. If truncation happens (i.e., we read exactly the same number of bytes as the buffer size), the assertion will fail. I didn't account for the rare case where the size of the file happens to be the same as the buffer size. As the TODO I left says, I still need to fix `mmap` or global allocator. I will create an issue to remind me to improve it.

litebox_platform_linux_userland/src/lib.rs
CvvT: Good point! Fixed!

litebox_runner_linux_userland/src/lib.rs
CvvT: Fixed

litebox_shim_linux/src/lib.rs

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -107,6 +113,58 @@ impl LinuxUserland {
...
+                    buf.len() - total_read,
+                )
+            }
+            .expect("read failed");
+            if n == 0 {
+                break;
+            }
+            total_read += n;
+        }
+        assert!(total_read != buf.len(), "buffer too small");
```
jaybosamiya-ms: Won't this still fail if actual file size is (for example) 6000?

Specifically, the value of `total_read` starts at 0, the `read` syscall will set `total_read` to 6000, and then another `read` is done, where it will return `0`, so the `break` is executed, at which point it is outside the loop, and it reaches the assertion, at which since `buf.len() == 8192` while `total_read = 6000`, the assertion will fail with "buffer too small" even though buffer is not too small.

Am I missing something here?

CvvT: I thought it's the opposite. if `total_read = 6000` != `buf.len() == 8192`, the assertion is true then. I tried to reduce the buffer size to e.g., `256` then I got the assertion failure when running the following unit test:

https://github.com/MSRSSP/litebox/blob/ed77035c4d3795c1ab96cd4431efe7aabd88ea5c/litebox_platform_linux_userland/src/lib.rs#L1048-L1056

jaybosamiya-ms: Sorry about the confusion, you are completely correct.

What I was missing: I was completely misreading the `!=`.  You are correct, and this assertion is correct.

Please merge when you are ready to merge, thanks!

CvvT: Changed to `<` as you suggested. Thanks!

PR #141, opened 2025-06-04
VSM functions for managing VTL0 kernel modules

sangho2:
This PR adds and improves VSM functions for managing VTL0 kernel modules.
It includes
- `mshv_vsm_validate_guest_module`: Apply protection to the memory ranges of a kernel module based on their memory types (assume that module validation is successful)
- `mshv_vsm_free_guest_module_init`: Change the protection applied to module memory ranges once the module is initialized (to let VTL0 free some `__init` code and data pages or to write-protect `__ro_after_init` pages).
- `mshv_vsm_unload_guest_module`: Change the protection applied to module memory ranges to let VTL0 free them.
- Fix VTL return value passing
- Add/improve sanity checks

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -423,7 +423,9 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
     Ok(token)
 }
 
-/// VSM function for supporting the initialization of a guest kernel module.
+/// VSM function for supporting the initialization of a guest kernel module including
+/// freeing the memory ranges that were used only for initialization and
+/// write-protecting the memory ranges that should be read-only after initialization.
 /// `token` is the unique identifier for the module.
 pub fn mshv_vsm_free_guest_module_init(token: i64) -> Result<i64, Errno> {
     debug_serial_println!("VSM: Free kernel module's init (token: {})", token);
@@ -693,6 +695,8 @@ impl ModuleMemoryMap {
         }
     }
 
+    /// Generate a unique key for representing each loaded kernel module.
+    /// It assumes a 64-bit atomic counter is sufficient and there is no run out of keys.
     fn gen_unique_key(&self) -> i64 {
         self.key_gen.fetch_add(1, Ordering::Relaxed)
     }


litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -206,123 +232,292 @@ pub fn mshv_vsm_lock_regs() -> u64 {
...
-            // for now, this function is a no-op and just prints the memory range we should protect.
-            debug_serial_println!(
-                "VSM: Protect memory: va {:#x} pa {:#x} epa {:#x} {:?} (size: {})",
-                va,
-                pa,
-                epa,
-                attr,
-                epa - pa
-            );
+    if !pa.is_multiple_of(u64::try_from(PAGE_SIZE).unwrap_or(4096)) || nranges == 0 {
```
wdcui: Is it okay to replace 4096 with PAGE_SIZE?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -206,123 +232,292 @@ pub fn mshv_vsm_lock_regs() -> u64 {
...
-    debug_serial_println!("VSM: Free kernel module init");
-    // TODO: free kernel module
+/// VSM function for supporting the initialization of a guest kernel module.
+/// `token` is the unique identifier for the module.
+pub fn mshv_vsm_free_guest_module_init(token: u64) -> u64 {
+    debug_serial_println!("VSM: Free kernel module's init (token: {})", token);
+
+    if !crate::platform_low().vtl0_module_memory.contains_key(token) {
+        serial_println!("VSM: invalid module token");
+        return EINVAL;
```
wdcui: I see that you always return EINVAL. In the current C-based LVSM implementation, does it return more information about the error?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -462,3 +654,166 @@ fn save_vtl0_locked_regs() -> Result<u64, HypervCallError> {
...
     Ok(0)
 }
+
+/// Data structure for maintaining the memory ranges of each VTL0 kernel module and their types
+pub struct ModuleMemoryMap {
+    inner: spin::mutex::SpinMutex<BTreeMap<u64, ModuleMemoryMapInner>>,
+    key_gen: AtomicU64,
+}
+
+pub struct ModuleMemoryMapInner {
```
wdcui: Do we need to make ModuleMemoryMapInner public?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -462,3 +654,166 @@ fn save_vtl0_locked_regs() -> Result<u64, HypervCallError> {
...
+
+impl ModuleMemoryMap {
+    pub fn new() -> Self {
+        Self {
+            inner: spin::mutex::SpinMutex::new(BTreeMap::new()),
+            key_gen: AtomicU64::new(0),
+        }
+    }
+
+    pub fn gen_unique_key(&self) -> u64 {
```
wdcui: Is this how the current linux-based LVBS generates a token?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -462,3 +654,166 @@ fn save_vtl0_locked_regs() -> Result<u64, HypervCallError> {
...
+    }
+
+    pub fn insert_memory_range(&self, key: u64, mem_range: ModuleMemoryRange) -> bool {
+        if self.key_gen.load(Ordering::Relaxed) < key {
+            debug_serial_println!("VSM: Invalid key {:#x} for inserting memory range", key);
+            return false;
+        }
+
+        let mut map = self.inner.lock();
+        map.entry(key)
```
wdcui: What if the key was removed?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -462,3 +654,166 @@ fn save_vtl0_locked_regs() -> Result<u64, HypervCallError> {
...
+}
+
+impl<'a> ModuleMemoryIters<'a> {
+    pub fn iter_mem_ranges(&'a self) -> impl Iterator<Item = &'a ModuleMemoryRange> {
+        self.guard.get(&self.key).unwrap().ranges.iter()
+    }
+}
+
+impl ModuleMemoryMap {
+    pub fn iter_entry(&self, key: u64) -> Option<ModuleMemoryIters> {
```
wdcui: Why do we split the implementation of ModuleMemoryMap to two parts?

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Let me remove replace those `unwrap_or` with `unwrap`. Anyway, there will be no panic.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Let me check it again. They do barely use others like `EFAULT` and `ENOMEM`.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: I don't think so. Let me make it private and check whether this works.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: yes. A token is just a global counter. LVBS VTL1 kernel does `++token` with a lock.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: When a module is loaded into the memory (e.g., by `modprobe`), the VTL0 kernel invokes `mshv_vsm_validate_guest_module` for the module which generates a new key. That is, a key is associated with an in-memory module instance. This key will never be reused once the module is unloaded.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: We don't have to do this. Let me merge these two `impl ModuleMemoryMap`.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: As discussed, let me explicitly create an empty `vector`s when we create a new key.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -33,15 +33,23 @@ use crate::{
...
 use core::sync::atomic::{AtomicU64, Ordering};
-use num_enum::TryFromPrimitive;
+use num_enum::{IntoPrimitive, TryFromPrimitive};
 use x86_64::structures::paging::{PhysFrame, Size4KiB, frame::PhysFrameRange};
 
 /// VTL call parameters (param[0]: function ID, param[1-3]: parameters)
 pub const NUM_VTLCALL_PARAMS: usize = 4;
 
-const EINVAL: u64 = u64::MAX - 22 + 1;
+/// Error for VSM functions. This is just negative Linux error code but we use `u64`` as we return `u64` values to VTL0.
```
wdcui: Just want to bring up this option for discussion; I don't have a strong preference here. An alternative approach is to define the error code as i64 directly (then we will have -12, -14, etc. instead of u64::MAX - 12 + 1), then we convert i64 to u64 when we return it to vtl0.

sangho2: That will be better than the current ugly one. Let me do that.

We might need to return a pointer or something similar in the future (e.g., for RPC), so this return logic should be revised later anyhow.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -206,150 +231,314 @@ pub fn mshv_vsm_lock_regs() -> u64 {
...
-                "VSM: Load kernel data: va {:#x} pa {:#x} epa {:#x} {:?} (size: {})",
-                va,
-                pa,
-                epa,
-                attr,
-                epa - pa
-            );
+/// VSM function for loading kernel data (e.g., certificates, blocklist, kernel symbols) into VTL1.
+/// `pa` and `nranges` specify memory areas containing the information about the memory ranges to load.
+pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> i64 {
```
CvvT: It might be better to return `Result` instead of just i64. This also applies to other functions that returns `i64`

sangho2: yes using `Result` is probably fine. Let me test it. I didn't do that because I wanted to avoid any stuff affecting low-level behaviors, but we do have `vsm_dispatch` which would consume any leftovers.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -206,150 +231,314 @@ pub fn mshv_vsm_lock_regs() -> u64 {
...
+    for mod_mem_range in &module_memory {
+        if protect_physical_memory_range(
+            mod_mem_range.phys_frame_range,
+            mod_mem_type_to_mem_attr(mod_mem_range.mod_mem_type),
+        )
+        .is_err()
+        {
+            return Errno::EFAULT.as_neg().into();
+        }
+    }
```
CvvT: Just curious: is the memory range controllable by VTL0 and does the validation ensure that the given module does not overlap with existing modules?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -206,150 +231,314 @@ pub fn mshv_vsm_lock_regs() -> u64 {
...
+        .register_module_memory(module_memory)
 }
 
-/// VSM function for initializing guest kernel module
-pub fn mshv_vsm_free_guest_module_init(_token: u64) -> u64 {
-    debug_serial_println!("VSM: Free kernel module init");
-    // TODO: free kernel module
+/// VSM function for supporting the initialization of a guest kernel module.
+/// `token` is the unique identifier for the module.
+pub fn mshv_vsm_free_guest_module_init(token: i64) -> i64 {
```
CvvT: The name is confusing.  Maybe add the comment you had to this function:

`mshv_vsm_free_guest_module_init: Change the protection applied to module memory ranges once the module is initialized (to let VTL0 free some __init code and data pages or to write-protect __ro_after_init pages).`

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -462,3 +652,183 @@ fn save_vtl0_locked_regs() -> Result<u64, HypervCallError> {
...
+    pub fn new() -> Self {
+        Self {
+            inner: spin::mutex::SpinMutex::new(BTreeMap::new()),
+            key_gen: AtomicI64::new(0),
+        }
+    }
+
+    fn gen_unique_key(&self) -> i64 {
+        self.key_gen.fetch_add(1, Ordering::Relaxed)
+    }
```
CvvT, line 702: I guess we assume we never run out of keys so we don't need to recycle it.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: interesting observation. This might result in ambiguous SLAT permission. Whether this is exploitable or not is a bit unclear for now.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: yes. We assume that a 64-bit counter is enough to cover all module loading attempts.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: looks like this one's security implication is weak and there might be benign cases (i.e., KSM/deduplication). Let's not too much worry about this for now (an issue is there).

PR #143, opened 2025-06-06
Add syscall set_thread_area

CvvT:
Add syscall [set_thread_area](https://www.man7.org/linux/man-pages/man2/set_thread_area.2.html) as a punchthrough.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -772,6 +772,8 @@ pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
         pipefd: Platform::RawMutPointer<u32>,
         flags: litebox::fs::OFlags,
     },
+    /// Manipulate thread-local storage information.
+    /// Returns `ENOSYS` on 64-bit.
     SetThreadArea {
         user_desc: Platform::RawMutPointer<UserDesc>,
     },


CvvT:
> At least the manpage seems to imply (I am not sure though) that at least in some cases ENOSYS can occur (e.g., on 64-bit), which means that higher-level code/libraries would probably have a fallback, no?

On  x86,  Linux  dedicates three global descriptor table (GDT) entries for thread-local storage: https://elixir.bootlin.com/linux/v5.19.17/source/arch/x86/kernel/tls.c#L145

which userspace program cannot modify.

I agree for x86_64, we can just return `ENOSYS`.

CvvT:
> Code looks fine to me, but a quick question (I'll let you make a decision on) before we merge this: every additional punchthrough causes more difficulty with supporting on more platforms, so an important question to ask is if it is good enough for us to just return `ENOSYS` for this syscall (or just generally before we add any new punchthroughs). At least the manpage seems to imply (I am not sure though) that at least in some cases ENOSYS can occur (e.g., on 64-bit), which means that higher-level code/libraries would probably have a fallback, no?

on x86_64, it is preferred to use fs/gs base via `arch_prctl`, which x86 doesn't have.

jaybosamiya-ms:
The question is more of why we need this syscall in the first place.   Basically, is it the case that a specific LiteBox guest program is invoking `set_thread_area`, and _thus_ we need the punchthrough, or is it that you anticipate LiteBox itself needing it? Based on the implementation I am seeing thus far, I suspect it is the former, and not the latter.  If it is the former, then I think we can do it without a punchthrough.

CvvT:
> The question is more of why we need this syscall in the first place. Basically, is it the case that a specific LiteBox guest program is invoking `set_thread_area`, and _thus_ we need the punchthrough, or is it that you anticipate LiteBox itself needing it? Based on the implementation I am seeing thus far, I suspect it is the former, and not the latter. If it is the former, then I think we can do it without a punchthrough.

It's used to support thread local storage (TLS). This syscall would update GDT and gs selector. Then the program can use `mov %gs:offset reg` to access TLS. How to reimplement this within Litebox?

litebox_common_linux/src/lib.rs
```diff
@@ -763,6 +772,9 @@ pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
         pipefd: Platform::RawMutPointer<u32>,
         flags: litebox::fs::OFlags,
     },
+    SetThreadArea {
+        user_desc: Platform::RawMutPointer<UserDesc>,
+    },
```
jaybosamiya-ms, line 779:
```suggestion
    /// Manipulate thread-local storage information.
    /// Returns `ENOSYS` on 64-bit.
    SetThreadArea {
        user_desc: Platform::RawMutPointer<UserDesc>,
    },
```

PR #144, opened 2025-06-06
Add rt_sigprocmask

CvvT:
Add `rt_sigprocmask` as a punchthrough.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -703,7 +703,7 @@ unsafe extern "C" fn syscall_handler(syscall_number: usize, args: *const usize) 
     let dispatcher = match sysno {
         ::syscalls::Sysno::write => SyscallRequest::Write {
             fd: syscall_args[0].reinterpret_as_signed().truncate(),
-            buf: unsafe { core::mem::transmute::<usize, ConstPtr<u8>>(syscall_args[1]) },
+            buf: unsafe { transmute_ptr(syscall_args[1] as *const u8) },
             count: syscall_args[2],
         },
         ::syscalls::Sysno::rt_sigprocmask => {
@@ -757,9 +757,7 @@ unsafe extern "C" fn syscall_handler(syscall_number: usize, args: *const usize) 
         }
         ::syscalls::Sysno::set_thread_area => SyscallRequest::SetThreadArea {
             user_desc: unsafe {
-                core::mem::transmute::<usize, MutPtr<litebox_common_linux::UserDesc>>(
-                    syscall_args[0],
-                )
+                transmute_ptr_mut(syscall_args[0] as *mut litebox_common_linux::UserDesc)
             },
         },
         _ => todo!("syscall {sysno} not implemented"),


litebox_common_linux/src/lib.rs
```diff
@@ -509,6 +509,79 @@ impl TryFrom<TimeVal> for core::time::Duration {
...
+    SIGWINCH = 28,
+    SIGIO = 29,
+    // SIGPOLL = 29, // Alias for SIGIO
+    SIGPWR = 30,
+    SIGSYS = 31,
+}
+
+pub const SIGIOT: Signal = Signal::SIGABRT;
+pub const SIGPOLL: Signal = Signal::SIGIO;
+pub const SIGUNUSED: Signal = Signal::SIGSYS;
```
jaybosamiya-ms: I don't think these should be `pub const` on their own, they are probably better suited as `pub const` inside an `impl Signal`.

litebox_common_linux/src/lib.rs
```diff
@@ -783,15 +862,13 @@ pub enum PunchthroughSyscall<Platform: litebox::platform::RawPointerProvider> {
...
         /// If `set` is NULL, then the signal mask is unchanged (i.e., `how` is ignored), but the
         /// current value of the signal mask is nevertheless returned in `oldset` (if it is not
         /// NULL).
-        set: Platform::RawConstPointer<u8>,
+        set: Option<Platform::RawConstPointer<SigSet>>,
         /// If `oldset` is non-NULL, the previous value of the signal mask is stored in `oldset`.
-        old_set: Platform::RawMutPointer<u8>,
-        /// Specifies the size in bytes of the signal sets in `set` and `oldset`.
-        sigsetsize: usize,
+        oldset: Option<Platform::RawMutPointer<SigSet>>,
```
jaybosamiya-ms: These comments likely require updates

litebox_shim_linux/src/lib.rs
```diff
@@ -652,6 +664,20 @@ unsafe extern "C" fn syscall_handler_32(args: *const usize) -> isize {
...
+{
+    unsafe { core::mem::transmute::<*const T, ConstPtr<T>>(ptr) }
+}
+
+unsafe fn transmute_ptr_mut<T>(ptr: *mut T) -> MutPtr<T>
+where
+    T: Clone,
+{
+    unsafe { core::mem::transmute::<*mut T, MutPtr<T>>(ptr) }
+}
```
jaybosamiya-ms: Weird, how did clippy not complain about `unsafe` functions without a `SAFETY` doc comment?

Also, these particular functions are technically unsound in general, since the transmutation is not guaranteed to be always safe.  We need to overhaul our `RawPointerProvider` to reduce the number of transmutations, but I'll open a separate issue to track that.

jaybosamiya-ms: https://github.com/MSRSSP/litebox/issues/145

litebox_common_linux/src/lib.rs
CvvT: Good point! Fixed.

litebox_common_linux/src/lib.rs

litebox_shim_linux/src/lib.rs
CvvT: Looks like clippy only checks public functions. These two functions should only be used by `syscall_handler` to convert raw pointer to ConstPtr/MutPtr. I added some comments.

jaybosamiya-ms: Thanks!  Interesting to know it is only on public functions.  The comments help.

PR #146, opened 2025-06-10
Add sys_rt_sigaction

CvvT:
Add [rt_sigaction](https://www.man7.org/linux/man-pages/man2/sigaction.2.html) as a punchthrough.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -918,7 +918,7 @@ pub enum PunchthroughSyscall<Platform: litebox::platform::RawPointerProvider> {
     RtSigaction {
         /// The signal number to change the action for.
         signum: Signal,
-        /// If `act` is None, the new action for signal `signum` is installed from `act`.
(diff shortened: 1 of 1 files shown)


litebox_common_linux/src/lib.rs
```diff
@@ -871,6 +900,15 @@ pub enum PunchthroughSyscall<Platform: litebox::platform::RawPointerProvider> {
         /// If `oldset` is not None, the previous value of the signal mask is stored in `oldset`.
         oldset: Option<Platform::RawMutPointer<SigSet>>,
     },
+    /// Change the action taken by a process on receipt of a specific signal
+    RtSigaction {
+        /// The signal number to change the action for.
+        signum: Signal,
+        /// If `act` is None, the new action for signal `signum` is installed from `act`.
+        act: Option<Platform::RawConstPointer<SigAction>>,
```
jaybosamiya-ms: Possible missed word?
```suggestion
        /// If `act` is not None, the new action for signal `signum` is installed from `act`.
        act: Option<Platform::RawConstPointer<SigAction>>,
```

PR #147, opened 2025-06-10
Fix syscall_handler

CvvT:
Previously we had two syscall handlers, one for seccomp and one for rewriter. This PR merges them into one syscall handler.

Also, fix a subtle difference in `pread64` between x86 and x64: https://elixir.bootlin.com/linux/v5.19.17/source/arch/x86/kernel/sys_ia32.c#L68


litebox_common_linux/src/lib.rs
```diff
@@ -680,6 +680,8 @@ pub unsafe fn wrfsbase(fs_base: usize) {
     }
 }
 
+pub type loff_t = i64;
+
```
jaybosamiya-ms: Do we really need to name a new type here?  It is used only for 2 places (for the same syscall)?  Couldn't we just change it to a `i64` at each of those two places?

If you think we need a new type here, please add a doc string for it.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -89,10 +89,7 @@ impl LinuxUserland {
     /// Panics if this function has already been invoked on the platform earlier.
     pub fn enable_syscall_interception_with(
         &self,
-        syscall_handler: impl Fn(litebox_common_linux::SyscallRequest<LinuxUserland>) -> isize
-        + Send
-        + Sync
-        + 'static,
+        syscall_handler: unsafe extern "C" fn() -> isize,
```
jaybosamiya-ms: This change completely removes any type-level protections we might have for this.  We would need strong justification if we want to switch to this.

Any reason we can't use a Rust `Fn`?

CvvT: I just realized that we don't need to pass any functions.

Now the syscall entry function is moved to linux userland as it is platform specific (e.g., Linux kernel does not need to save and restore rip/rsp, which is done by the hardware). The entry function would call 
`pub unsafe extern "C" fn syscall_handler(syscall_number: usize, args: *const usize) -> isize` defined in `litebox_shim_linux`. To break the circular dependency, linux userland could just declare an extern function as follows:

```
unsafe extern "C" {
    pub fn syscall_handler(syscall_number: usize, args: *const usize) -> isize;
}
```

litebox_common_linux/src/lib.rs
CvvT: I agree it is unnecessary. Removed.

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: I think rather than an extern function, we should be doing it via a registration.  Extern functions are a very C-like notion, but they leave things too implicit, and any issues only get caught only at link time (or worse). Providing something a "here's a function you can use now" is much more Rust-y, and allows the type system to be able to catch issues better.

The shim indeed should have a `syscall_handler` (or similar) function. The runner should perform the final connection (i.e., invoke the platform's registration with the shim's exposed function), that way there is no circular dependency involved.

jaybosamiya-ms: If the approach I am suggesting is not very clear, I'm happy to sketch it out and push to a PR, do let me know if you prefer that.

CvvT: Thanks for the suggestion. If you already have a solution in your mind, please take over. Thanks!

My understanding is that `extern "C"` is necessary because `syscall_handler` is called by the assembly code `syscall_callback`. I'm not sure how function registration would help it.

jaybosamiya-ms: Update: I have a version of this partially completed.  For review purposes, it is probably easiest to look at it in chunks though, so will do those in separate PRs.  For now, I have split away just the first part into #154, which should fairly independent to review.  I'll push other bits as time permits.

To keep things simple, should we merge this current PR (after resolving conflicts with `main` introduced due to #139), and then just use a new PR for the Rust-y solution I am mentioning?  That way, it is not blocking other PRs, and conflict management is simpler.

PR #148, opened 2025-06-10
Add rseq

CvvT:
Do not support syscall `rseq` for now.


jaybosamiya-ms:
Semver checks results like above are confusing, so I opened #149.

PR #149, opened 2025-06-10
semver-checks: add check to ensure PR base is `main`

jaybosamiya-ms:
Turns out semver-checks is quite confusing if the PR base is not `main`. This PR adds a check to ensure that the PR base is `main`.  It also updates the "breaking changes" text to have ⚠️ symbols to make it visually stand out.

I expect this PR to initially fail (as a test) and then to succeed once I force-push and then update base (currently I've chosen a random other branch as base, just as a test).

EDIT: Tested to fail in situations (where base is not set to `main`) as expected. I will merge it upon a CI green.

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -11,6 +11,11 @@ jobs:
     name: Check SemVer Correctness
     runs-on: ubuntu-latest
     steps:
+      - name: Fail if PR base is not main
+        if: github.event.pull_request.base.ref != 'main'
+        run: |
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
@CvvT @sangho2 @wdcui : heads up on this upcoming change to CI behavior.  If the failure state turns out too annoying (it should not be triggered very often, so hopefully is ok) then we can switch to a less noisy version, but for now, I would prefer it to be slightly noisy that it is *not* on a `main` branch base.

PR #150, opened 2025-06-10
ci: build and check documentation in main CI job using cargo doc

### LINKED ISSUE #101: Add documentation building to CI
Just to make sure we don't hit a scenario where there are bad links/warnings in documentation, it is probably a good idea to add documentation building to the CI.

jaybosamiya-ms:
Closes #101 

> [!NOTE]
> PR made (mostly) with "Copilot Agent Mode"

### My notes on the experience:

Copilot Agent Mode appears to be different to the [copilot coding agent](https://github.blog/news-insights/product-news/github-copilot-meet-the-new-coding-agent/).  I thought it was the latter one, and thought it might be able to quickly figure out and solve this long-open but relatively simple issue fully autonomously.  Seems like it was not fully able to do it automatically, and there is some manual stuff I need to walk it through.  For example, I needed to manually push it into knowing that the initial code it suggested was wrong (see the CI results on the first commit).  

There are still more failures that show up, I had to prompt Copilot to fix those up, it fixed a few, but then ignored others. I had to then prompt it again to fix more. I also had to do a couple of manual edits near the end to clear things up because it had added superfluous stuff along the way.

Overall, the process seemed suboptimal for what I think should have been an easy job for Copilot to blast through. I am still curious to attempt to _full_ agent mode (which is supposed to automatically make a draft PR and start updating and doing things there) but it appears I (or at least this GitHub org) don't have access to that yet.  But for the more manual agent mode, I am left largely unimpressed.
 
This PR _does_ fix up the issue #101 though, so overall not a complete loss of time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,6 @@ jobs:
     runs-on: ubuntu-latest
     env:
       RUSTFLAGS: -Dwarnings
-      RUSTDOCFLAGS: -Dwarnings
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
@@ -75,7 +74,6 @@ jobs:
     runs-on: ubuntu-latest
     env:
       RUSTFLAGS: -Dwarnings
-      RUSTDOCFLAGS: -Dwarnings
     steps:
       - name: Check out repo
         uses: actions/checkout@v4


jaybosamiya-ms:
@CvvT added you as a reviewer mostly as a quick sanity check, I checked the PR myself (and after sufficient guiding and manual fixing, see above) it seems reasonable overall.  But yeah, I am curious to know if you've tried the agent mode (this is within VSCode---GitHub just gave me an in-browser VSCode workspace for this) and if so, does the above matches your experience, or if I was just holding the tool wrong?

CvvT:
Thanks for sharing your experience. I only tried copilot agent mode within VSCode to see how it works with MCP. It's pretty easy to set up/register a new agent, but my experience with copilot was similar to yours. I don't expect it to work out of box. My strategy is asking for small and specific change, which works pretty well but requires more interaction/prompting, or directly taking the output and fixing the details (which is faster than asking LLM to fix it in most of times). It's possible that a better model with reasoning capability could make your task done without much human efforts in the loop.

jaybosamiya-ms:
Thanks for sharing your experience too.  I assumed that this PR would be small and self-contained enough (I didn't expect it to also automatically go fix up all the doc comments, but at least the ci.yml update I would've expected it to get right on the first go).

Agreed that possibly better model might help.

PR #152, opened 2025-06-13
Add uname

CvvT:
Add syscall [uname](https://linux.die.net/man/2/uname)

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -690,6 +690,17 @@ pub struct UserDesc {
     pub flags: i32,
 }
 
+#[repr(C)]
+#[derive(Clone)]
+pub struct Utsname {
+    pub sysname: [u8; 65],
+    pub nodename: [u8; 65],
+    pub release: [u8; 65],
+    pub version: [u8; 65],
+    pub machine: [u8; 65],
+    pub domainname: [u8; 65],
+}
+
 /// Request to syscall handler
 #[non_exhaustive]
 pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
@@ -849,6 +860,9 @@ pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
         optname: SocketOptionName,
         optval: Platform::RawConstPointer<u8>,
         optlen: usize,
+    },
+    Uname {
+        buf: Platform::RawMutPointer<Utsname>,
     },
     Fcntl {
         fd: i32,
@@ -1195,6 +1209,9 @@ impl<Platform: litebox::platform::RawPointerProvider> SyscallRequest<Platform> {
             },
             Sysno::exit_group => SyscallRequest::ExitGroup {
                 status: args[0].reinterpret_as_signed().truncate(),
+            },
+            ::syscalls::Sysno::uname => SyscallRequest::Uname {
+                buf: Platform::RawMutPointer::from_usize(args[0]),
             },
             Sysno::fcntl => SyscallRequest::Fcntl {
                 fd: args[0].reinterpret_as_signed().truncate(),
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -278,6 +278,36 @@ pub extern "C" fn close(fd: i32) -> i32 {
 // hopefully-reasonable middle ground.
 const MAX_KERNEL_BUF_SIZE: usize = 0x80_000;
 
+/// A const function to convert a str to a fixed-size array of bytes
+///
+/// Note the fixed-size array is terminated with a null byte, so the string must be
+/// at most `N - 1` bytes long.
+const fn to_fixed_size_array<const N: usize>(s: &str) -> [u8; N] {
+    assert!(
+        s.len() < N,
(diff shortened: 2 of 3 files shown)


litebox_shim_linux/src/lib.rs
```diff
@@ -280,6 +280,28 @@ pub extern "C" fn close(fd: i32) -> i32 {
 // hopefully-reasonable middle ground.
 const MAX_KERNEL_BUF_SIZE: usize = 0x80_000;
 
+const fn make_field(s: &str) -> [u8; 65] {
```
wdcui: This function's name `make_field` doesn't look like a good match with its functionality which converts a string to an array?

litebox_shim_linux/src/lib.rs
```diff
@@ -914,6 +939,11 @@ pub unsafe extern "C" fn syscall_handler(syscall_number: usize, args: *const usi
             litebox_platform_multiplex::platform()
                 .exit(syscall_args[0].reinterpret_as_signed().truncate())
         }
+        ::syscalls::Sysno::uname => SyscallRequest::Uname {
+            buf: unsafe {
```
wdcui: How do we know if the buffer is big enough to store Utsname?

CvvT: It's checked when writing to this ptr:

https://github.com/MSRSSP/litebox/blob/c0f28ce653f988778c9119cf11ffc607787553a3/litebox_shim_linux/src/lib.rs#L451-L453

litebox_shim_linux/src/lib.rs
CvvT: This supposes to be only used to construct `SYS_INFO`. I made it more general and changed its name.

litebox_shim_linux/src/lib.rs
wdcui: I'm still confused. Here we create a `SyscallRequest::Uname` by assuming the input buffer is big enough. Then in the place where we write to this ptr, it assumes it's a `SyscallRequest::Uname`. How does it check it?

litebox_shim_linux/src/lib.rs
```diff
@@ -280,6 +280,29 @@ pub extern "C" fn close(fd: i32) -> i32 {
 // hopefully-reasonable middle ground.
 const MAX_KERNEL_BUF_SIZE: usize = 0x80_000;
 
+/// A const function to convert a str to a fixed-size array of bytes
+const fn to_fixed_size_array<const N: usize>(s: &str) -> [u8; N] {
```
jaybosamiya-ms: Minor comment: might want to add a `if s.len() >= N { panic!() }` just to catch any future issues related to this.

Also, it might be good to explicitly mention that this will be a C-terminated byte array (i.e., terminated by `\0`), or if we are explicitly deciding that we don't want that (e.g., changing the `if` condition I mentioned above to only `>`), then we should document that.

litebox_shim_linux/src/lib.rs
```diff
@@ -918,6 +944,11 @@ pub unsafe extern "C" fn syscall_handler(syscall_number: usize, args: *const usi
             litebox_platform_multiplex::platform()
                 .exit(syscall_args[0].reinterpret_as_signed().truncate())
         }
+        ::syscalls::Sysno::uname => SyscallRequest::Uname {
+            buf: unsafe {
+                transmute_ptr_mut(syscall_args[0] as *mut litebox_common_linux::Utsname)
```
jaybosamiya-ms: Due to this, please wait until https://github.com/MSRSSP/litebox/pull/154 is merged, thank you.

litebox_shim_linux/src/lib.rs
CvvT: Oh, I see. If it runs in Linux userland, yes, we don't have any checks (the user program can modify arbitrary memory anyway).

If it runs in kernel model, similar to Linux, it relies on page fault to catch invalid memory access and return an error (i.e., EFAULT), which is not implemented yet.

litebox_shim_linux/src/lib.rs
CvvT: Thanks for the suggestion! I added some comments and assertion.

litebox_shim_linux/src/lib.rs
CvvT: Fixed! Thanks!

litebox_common_linux/src/lib.rs
```diff
@@ -689,6 +689,17 @@ pub struct UserDesc {
...
 
+#[repr(C)]
+#[derive(Clone)]
+pub struct Utsname {
+    pub sysname: [u8; 65usize],
+    pub nodename: [u8; 65usize],
+    pub release: [u8; 65usize],
+    pub version: [u8; 65usize],
+    pub machine: [u8; 65usize],
+    pub domainname: [u8; 65usize],
```
jaybosamiya-ms: Nit: we might not actually need to explicitly say `usize` here, in `[u8; N]`, Rust already expects `N` to be a usize.
```suggestion
    pub sysname: [u8; 65],
    pub nodename: [u8; 65],
    pub release: [u8; 65],
    pub version: [u8; 65],
    pub machine: [u8; 65],
    pub domainname: [u8; 65],
```

Issue #145, opened 2025-06-09
Remove all the `transmute`s


jaybosamiya-ms:
There are some transmutes, and more being added (e.g., https://github.com/MSRSSP/litebox/pull/144), which seem to imply that the `RawPointerProvider` interface needs some improvement, since as it currently stands, such transmutes can be unsound (they are being added to the shim, which (as an example) is unaware of the _actual_ size of the `Platform::Raw*Pointer`, and needs to stay parametric over it).

If we added some more constraints, this would possibly be fine, but adding such constraints could also instead be done by adding some transitions to/from actual raw Rust pointers with a safer interface.

Not a high-priority, but just wanted to open a tracking issue as a reminder.

PR #154, opened 2025-06-14
Provenance-tracked integer-to-pointer conversion

jaybosamiya-ms:
While this doesn't _fully_ solve #145, it clears out a whole bunch of `transmute`s by adding a safer interface for converting integers to pointers (`from_usize`) using the Rust ["Exposed Provenance"](https://doc.rust-lang.org/std/ptr/index.html#exposed-provenance) model.

This PR directly deletes ~54 lines of unsafe code, replacing it all with safer versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,14 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
+      - name: "Ratchet: ensure number of 'transmute's in codebase is not increasing"
+        run: |
+          if [ $(find . -name \*.rs -print0 | xargs -0 grep 'transmute' | wc -l) -eq 5 ]; then \
+            echo 'Ratchet maintained'; \
+          else \
+            echo 'Ratchet failed'; \
+            false; \
+          fi
       - name: Set up Rust
         run: |
           rustup toolchain install $(awk -F'"' '/channel/{print $2}' rust-toolchain.toml) --profile minimal --no-self-update --component rustfmt,clippy
@@ -97,7 +105,7 @@ jobs:
         run: |
           .github/tools/github_actions_run_cargo clippy --lib --bins --examples --no-deps --all-features -Z build-std-features=compiler-builtins-mem -Z build-std=core,alloc --manifest-path=litebox_runner_lvbs/Cargo.toml --target litebox_runner_lvbs/x86_64_vtl1.json
           .github/tools/github_actions_run_cargo clippy --lib --bins --examples --no-deps --all-features -Z build-std-features=compiler-builtins-mem -Z build-std=core,alloc --manifest-path=litebox_platform_lvbs/Cargo.toml --target litebox_runner_lvbs/x86_64_vtl1.json
-      - name: cargo build 
+      - name: cargo build
         run: |
           .github/tools/github_actions_run_cargo build -Z build-std-features=compiler-builtins-mem -Z build-std=core,alloc --manifest-path=litebox_runner_lvbs/Cargo.toml --target litebox_runner_lvbs/x86_64_vtl1.json
           .github/tools/github_actions_run_cargo build -Z build-std-features=compiler-builtins-mem -Z build-std=core,alloc --manifest-path=litebox_platform_lvbs/Cargo.toml --target litebox_runner_lvbs/x86_64_vtl1.json


wdcui:
@jaybosamiya-ms I'm a little confused about this comment in the "Exposed Provenance" doc: "This list is purely conceptual, it exists for the purpose of specifying Rust but is not materialized in actual executions, except in tools like [Miri](https://github.com/rust-lang/miri)". Does the compiler actually maintain a list or not?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -101,30 +101,37 @@ mod tests {
...
         init_platform(None);
 
         // Save old FS base
         let mut old_fs_base = MaybeUninit::<usize>::uninit();
-        let ptr: MutPtr<usize> = unsafe { core::mem::transmute(old_fs_base.as_mut_ptr()) };
+        let ptr = MutPtr {
+            inner: old_fs_base.as_mut_ptr(),
+        };
         sys_arch_prctl(ArchPrctlArg::GetFs(ptr)).expect("Failed to get FS base");
         let old_fs_base = unsafe { old_fs_base.assume_init() };
```
wdcui, line 115: Does this have the same effect as calling `with_exposed_provenance`?

jaybosamiya-ms:
Thanks, the run-time code does **not** have any concretized list.  The compiler is free to keep or not-keep such a list (purely at compile-time) but it is needed for it to uphold the guarantee that a `with_exposed_provenance(ptr.expose_provenance()` does **not** cause UB.  This is roughly speaking the guarantee we would normally expect with `ptr as usize as *const T`, except that the (in)formal specification of `as` is a _lot_ more complicated and subtle.  The Exposed Provenance work is trying to make the specifications simpler, so that it is easier to reason about _when_ things become UB and when not.

TL;DR: For our use cases, this as a guaranteed 0-runtime-cost change.

litebox_shim_linux/src/syscalls/process.rs
jaybosamiya-ms: There is no need to expose provenance here, the `.as_mut_ptr()` is simply converting a `&mut` to a `*mut`; there is no integer conversion involved, thus no need to (explicitly) reason about provenance. Placing it into the `MutPtr` struct is just a normal data-movement operation.

I've mentioned in the past (e.g., https://github.com/MSRSSP/litebox/pull/126#discussion_r2094296113) that we should just not use transmute to make the user-pointers (`MutPtr`/`ConstPtr`/...) whenever this struct-movement approach is possible.  This particular change is simply cleaning up yet another case where this existed.

wdcui:
Thank you for the offline discussion, @jaybosamiya-ms . Please wait for Weiteng's approval.

jaybosamiya-ms:
@CvvT I've finished fixing conflicts and rebased this onto `main`.

jaybosamiya-ms:
I'm also just about to push a new commit to this PR that sets up a [ratchet](https://qntm.org/ratchet) on the CI for the number of `transmute`s that show up in the code, so that future instances of `transmute` are immediately highlighted.

EDIT: Pushed, tested, works as intended on CI.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -101,30 +101,37 @@ mod tests {
...
         use litebox_common_linux::ArchPrctlArg;
 
         init_platform(None);
 
         // Save old FS base
         let mut old_fs_base = MaybeUninit::<usize>::uninit();
-        let ptr: MutPtr<usize> = unsafe { core::mem::transmute(old_fs_base.as_mut_ptr()) };
+        let ptr = MutPtr {
+            inner: old_fs_base.as_mut_ptr(),
+        };
```
CvvT, line 113: If we switch to a different platform, `MutPtr` may not be able to be constructed in this way, but this is just test and probably we'll stick to `LinuxUserland`, it should be fine. The previous code does not assume the concretized type of `MutPtr`.

jaybosamiya-ms: Thanks, yes, previous one did not assume concretized type _but_ it implicitly assumed that the shape is correct, which cannot be caught by the compiler. This one at least _if_ it fails on a new platform, then we get a compiler error, in which case we can deal with it at that time.

PR #155, opened 2025-06-14
Add an integration test for runner

CvvT:
For dynamically linked executable, we need to prepare a rootfs tar file containing all the dependencies it needs. This PR adds an integration test for the runner that compiles a simple C program, creating a rootfs, and runs it w/ seccomp (w/o rewriter).

Currently, there are still some syscalls we don't support but rewrite would intercept all syscalls and route them to litebox. With seccomp, we ony selectively intercept syscall we supported.

The test also outputs the command so we can run it manually (for testing purposes).

### DIFF
--- a/litebox_runner_linux_userland/src/lib.rs
+++ b/litebox_runner_linux_userland/src/lib.rs
@@ -76,14 +76,20 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
         )
     }
 
-    let (ancestor_modes, prog_data): (Vec<litebox::fs::Mode>, Vec<u8>) = {
+    let (ancestor_modes_and_users, prog_data): (Vec<(litebox::fs::Mode, u32)>, Vec<u8>) = {
         let prog = PathBuf::from(&cli_args.program_and_arguments[0]);
         let ancestors: Vec<_> = prog.ancestors().collect();
         let modes: Vec<_> = ancestors
             .into_iter()
             .rev()
             .skip(1)
-            .map(|path| litebox::fs::Mode::from_bits(path.metadata().unwrap().st_mode()).unwrap())
+            .map(|path| {
+                let metadata = path.metadata().unwrap();
+                (
+                    litebox::fs::Mode::from_bits(metadata.st_mode()).unwrap(),
+                    metadata.st_uid(),
+                )
+            })
             .collect();
         let data = std::fs::read(prog).unwrap();
         let data = if cli_args.rewrite_syscalls {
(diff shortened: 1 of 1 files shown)


litebox_runner_linux_userland/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    // Just to make sure cargo sets up the `OUT_DIR` environment variable
```
wdcui, line 2: Why would this empty function trigger cargo to set up the `OUT_DIR`?

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -95,23 +95,23 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
...
-            for (path, &mode) in ancestors
-                .into_iter()
-                .skip(1)
-                .rev()
-                .skip(1)
-                .zip(&ancestor_modes)
-            {
-                fs.mkdir(path.to_str().unwrap(), mode).unwrap();
-            }
+            // let ancestors: Vec<_> = prog.ancestors().collect();
```
wdcui: Why can we ignore these lines?

litebox_runner_linux_userland/build.rs
CvvT: See cargo doc: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

`OUT_DIR — If the package has a build script, this is set to the folder where the build script should place its output. See below for more information. (Only set during compilation.)`

litebox_runner_linux_userland/src/lib.rs
CvvT: The test works when running locally without this change. However, it causes EPERM error when running on CI. I suspect it was because the permission of some dir along the path to the executable: `/home/weiteng/workplace/litebox/target/debug/build/litebox_runner_linux_userland-9f859b4272cf86f9/out/hello_dylib`.

To get around it, instead of copying the executable to the same path in our in-memory fs, I copy it to the root folder, e.g., `/hello_dylib`.

wdcui: I see. Let's see if @jaybosamiya-ms has any comment about this workaround.

wdcui:
Please wait for @jaybosamiya-ms 's approval before you merge it.

litebox_runner_linux_userland/src/lib.rs
jaybosamiya-ms: Huh, that EPERM is definitely weird.  Are you talking about https://github.com/MSRSSP/litebox/actions/runs/15655785004/job/44106688910#annotation:10:127 failure?  I see no reason it should be different on CI server or locally due to this.

We should definitely try to debug the issue, because removing this bit to hide the issue is not ideal (essentially, it will cause disagreements in terms of paths of executables between the sandboxed world and the "real" world, which is suboptimal).

Let me know if you need help with debugging it, I could take a look sometime later this week.

CvvT: Thanks Jay! This is just to allow Weidong or intern to run and test litebox in a convenient way. It is not on my priority list. Feel free to take over and debug this. I can also spend some time later this week if you don't have time.

jaybosamiya-ms: Let's debug/merge this later this week, in that case, thanks!

CvvT: I figured out the root cause.

The code creates dirs as root user:

https://github.com/MSRSSP/litebox/blob/a40e90e6c9a8b1f684a381c32490098312dbd299/litebox_runner_linux_userland/src/lib.rs#L119-L130

The mode for `/home/runner` is `Mode(RWXU | RGRP | XGRP | 0x4000)`. On my local machine, it was `Mode(RWXU | RGRP | XGRP | ROTH | XOTH | 0x4000)`. When opening a file as non-root user, it would check `XOTH`.  The root cause I feel is that every dir it creates has to be owned by root. There is no interface to change the owner of a folder.

jaybosamiya-ms: Ah, good catch!  And fair point, we have added `chmod` but have not yet added `chown` (despite already tracking users and such).  I'll add that interface later this week.

jaybosamiya-ms: Chown has been implemented in #164, once we merge that and then fix this PR, I think we'll be ready to merge

CvvT:
> The changes _overall_ look good to me, but I think we should fix up the directory creation thing, especially since #164 now exists, before we merge this.

Thanks Jay! It's fixed.

PR #156, opened 2025-06-17
Support TLS

CvvT:
This PR tries to address #66. Similar to how libc uses fs on x86_64 (or gs on x86) for thread-local storage, we could use the other one (i.e., gs on x86_64 and fs on x86) for litebox's TLS.

Note that on x86_64, we can directly modify fs/gs base, while on x86, we have to call syscall `set_thread_area` to add an entry to GDT and update `fs` selector to point to the new entry.

### DIFF
--- a/litebox/src/platform/mod.rs
+++ b/litebox/src/platform/mod.rs
@@ -556,7 +556,7 @@ pub trait ThreadLocalStorageProvider {
     /// # Panics
     ///
     /// Panics if TLS is not set yet.
-    /// Panics if TLS is borrowed already (e.g., recursive call to [`Self::release_thread_local_storage`]).
+    /// Panics if TLS is borrowed already (e.g., recursive call).
     fn with_thread_local_storage_mut<F, R>(&self, f: F) -> R
     where
         F: FnOnce(&mut Self::ThreadLocalStorage) -> R;


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1155,12 +1160,131 @@ impl litebox::platform::SystemInfoProvider for LinuxUserland {
...
+            base_addr: Box::into_raw(tls) as u32,
+            limit: u32::try_from(core::mem::size_of::<Self::ThreadLocalStorage>()).unwrap() - 1,
+            flags,
+        };
+        let user_desc_ptr = litebox::platform::trivial_providers::TransparentMutPtr {
+            inner: &raw mut user_desc,
+        };
+        set_thread_area(user_desc_ptr).expect("Failed to set thread area for TLS");
+
+        let new_gs_selector = ((user_desc.entry_number & 0xfff) << 3) | 0x3; // user mode
```
wdcui: We use fs on x86. Why is this called new_gs_selector?

wdcui: set_thread_area doesn't update `gs`. Why do we need to save and restore `gs`?

CvvT: You're right, I got confused by this line: https://elixir.bootlin.com/linux/v5.19.17/source/arch/x86/kernel/tls.c#L171

CvvT: > We use fs on x86. Why is this called new_gs_selector?

Changed to `new_fs_selector`. Thanks!

litebox/src/platform/mod.rs
```diff
@@ -539,3 +539,18 @@ pub trait SystemInfoProvider {
...
+    /// Get a thread-local storage value for the current thread.
+    fn get_thread_local(&self) -> *mut Self::ThreadLocalStorage;
+
+    /// Set a thread-local storage value for the current thread.
+    fn set_thread_local(&self, value: Self::ThreadLocalStorage);
+
+    /// Release the thread-local storage value for the current thread
+    /// and invoke the provided callback function on it.
+    fn release_thread_local(&self, f: fn(&mut Self::ThreadLocalStorage));
+}
```
jaybosamiya-ms: I think that this interface can be improved.

Here's an alternative design which is a little more Rust-y, be safer to use, and I think should cover all uses.
```rs
pub trait ThreadLocalStorageProvider {
  type ThreadLocalStorage;
  #[must_use]
  fn set_thread_local_storage(&self, value: Self::ThreadLocalStorage) -> Option<Self::ThreadLocalStorage>;
  fn release_thread_local_storage(&self) -> Self::ThreadLocalStorage;
  fn with_thread_local_storage<R>(&self, impl FnOnce(&Self::ThreadLocalStorage) -> R) -> R;
  fn with_thread_local_storage_mut<R>(&self, impl FnOnce(&mut Self::ThreadLocalStorage) -> R) -> R;
}
```

Notes:
- For `set_...`, we can skip the return here if you think it is unnecessary. If so, I think we might want to explicitly document a panic if something has already been set. _Or_ we can document that any existing one is overwritten.
- For both the `with_...`, we should document that they panic if there is no thread local storage has been set. We should also document that they can panic if it has already been borrowed (otherwise, a recursive caller can lead to breaking Rust's memory safety contract with this signature).
- The (currently implemented) `release_thread_local` has the recursive caller UB problem, the signature allows users to do the equivalent of `platform.release_thread_local(|x| platform.release_thread_local(|y| /* At this point, you have violated Rust's memory safety */));`.  My suggested version documents the panic, which means that an existing borrow can/should lead to a panic, which means that the inner `y` is never created, and thus no safety has been violated.
- It might be difficult to easily implement a version that supports both `with_thread_local_storage`/`with_thread_local_storage_mut` _and_ correctly panics in all relevant circumstances. In that case, I recommend dropping `with_thread_local_storage`, which should make the panic support easier.

litebox_common_linux/src/lib.rs
```diff
@@ -681,13 +683,84 @@ pub unsafe fn wrfsbase(fs_base: usize) {
...
+    pub limit_in_pages, set_limit_in_pages: 4;
+    /// Segment not present
+    pub seg_not_present, set_seg_not_present: 5;
+    /// Usable by userland
+    pub useable, set_useable: 6;
+    /// 1 if the segment is 64-bit (x86_64 only)
+    pub lm, set_lm: 7;
+}
+
+pub struct ThreadLocalStorage<Platform: litebox::platform::RawPointerProvider> {
```
jaybosamiya-ms, line 755: Public item needs a doc comment, thanks!

litebox/src/platform/mod.rs
CvvT: Thanks, Jay for the suggestions! I made the following changes:

1. `set_thread_local_storage` now panics if TLS is set already.
2. `release_thread_local_storage` panics if TLS is not set yet or it is being used, and returns `Self::ThreadLocalStorage` directly.
3. `with_thread_local_storage_mut` also panics if TLS is not set yet or it is being used.

To check if TLS is borrowed multiple times, I added a bool flag to it.

litebox_common_linux/src/lib.rs
CvvT: Fixed. Thanks!

litebox/src/platform/mod.rs
```diff
@@ -539,3 +539,33 @@ pub trait SystemInfoProvider {
...
+    ///
+    /// Panics if TLS is set already.
+    fn set_thread_local_storage(&self, value: Self::ThreadLocalStorage);
+
+    /// Invokes the provided callback function with the thread-local storage value for the current thread.
+    ///
+    /// # Panics
+    ///
+    /// Panics if TLS is not set yet.
+    /// Panics if TLS is borrowed already (e.g., recursive call to [`Self::release_thread_local_storage`]).
```
jaybosamiya-ms: Nit: I think mentioning release here as part of recursive might be slightly misleading/confusing.  The biggest issue is not actually calling `release_thread_local_storage`, but a recursive call to _itself_.
```suggestion
    /// Panics if TLS is borrowed already (e.g., recursive call).
```

PR #158, opened 2025-06-17
Support clone3/set_tid_address/gettid

CvvT:
This PR adds a new thread provider for spawning new thread and terminate the calling thread.

Since the new thread does not share the same stack as the calling thread, we need to divert its execution to some trampoline code that could restore its registers to resume the execution. Instead of just passing the syscall arguments, now it passes the execution context (i.e., `pt_regs` used in Linux kernel).

To make the interface general enough, certain actions that supposed to be taken by Linux kernel now is done in the shim layer. In particular, 

```
CLONE_CHILD_CLEARTID: Clear  (zero)  the  child  thread ID at the location pointed to by child_tid (clone()) or cl_args.child_tid (clone3()) in child memory when the child exits, and do a wakeup on the futex at that address.  The address involved may be changed by the set_tid_address(2) system call.  This is used by threading libraries.
```
This is now done in `sys_exit` in the shim layer. One minor concern is that if the thread is forcefully killed, we don't have the chance to do the cleanup. In contrast, kernel always does it.

Since `set_tid_address` can change it, I also add the support for this syscall, which in turn requires the support for `gettid`.

```
CLONE_CHILD_SETTID (since Linux 2.5.49)
              Store the child thread ID at the location pointed to by child_tid (clone()) or cl_args.child_tid (clone3()) in the child's memory.  The store operation completes before the clone call returns  control  to  user space in the child process.  (Note that the store operation may not have completed before the clone call returns in the parent process, which will be relevant if the CLONE_VM flag is also employed.)

CLONE_PARENT_SETTID (since Linux 2.5.49)
              Store the child thread ID at the location pointed to by parent_tid (clone()) or cl_args.parent_tid (clone3()) in the parent's memory.  (In Linux 2.5.32-2.5.48 there was a flag CLONE_SETTID that did
              this.)  The store operation completes before the clone call returns control to user space.
```
These two can be trivially done in the shim layer after the new thread is spawned.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -293,9 +293,7 @@ extern "C" fn thread_start(
         );
     }
 
-    loop {
-        unsafe { core::arch::asm!("hlt") }
-    }
+    unreachable!();
 }
 
 impl litebox::platform::ThreadProvider for LinuxUserland {
@@ -424,10 +422,7 @@ impl litebox::platform::ThreadProvider for LinuxUserland {
         }
         .expect("Failed to exit");
 
-        // should not reach here
-        loop {
-            unsafe { core::arch::asm!("hlt") };
-        }
+        unreachable!("exit should not return");
     }
 }
 


litebox_common_linux/src/lib.rs
```diff
@@ -632,7 +633,7 @@ pub enum ArchPrctlCode {
 #[non_exhaustive]
 pub enum ArchPrctlArg<Platform: litebox::platform::RawPointerProvider> {
     #[cfg(target_arch = "x86_64")]
-    SetFs(Platform::RawConstPointer<u8>),
+    SetFs(Platform::RawMutPointer<u8>),
```
wdcui: @CvvT and I discussed. Should this be a `usize`? @jaybosamiya-ms

litebox_common_linux/src/lib.rs
```diff
@@ -751,16 +752,111 @@ bitfield::bitfield! {
...
-    pub __phantom: core::marker::PhantomData<Platform>,
+    pub current_task: alloc::boxed::Box<Task<Platform>>,
 }
 
-pub struct Task {
-    pub tid: u32,
+pub struct Task<Platform: litebox::platform::RawPointerProvider> {
+    /// Thread ID
+    pub tid: i32,
+    /// When a thread whose `clear_child_tid` is not None terminates, then, if the thread
```
wdcui: Minor: polish the comment here? It's hard to follow this long comment with multiple ifs and thens.

litebox_common_linux/src/lib.rs
```diff
@@ -979,16 +1076,24 @@ pub enum SyscallRequest<Platform: litebox::platform::RawPointerProvider> {
         pipefd: Platform::RawMutPointer<u32>,
         flags: litebox::fs::OFlags,
     },
+    Clone {
+        args: Platform::RawConstPointer<CloneArgs>,
+        ctx: &'a PtRegs,
```
wdcui, line 1124: Why do we need this lifetime thing for this argument?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -197,7 +216,226 @@ impl litebox::platform::ExitProvider for LinuxUserland {
...
+            tls: 0,
+            set_tid: 0,
+            set_tid_size: 0,
+            cgroup: 0,
+        };
+        let mut ret: usize;
+        #[cfg(target_arch = "x86_64")]
+        unsafe {
+            core::arch::asm!(
+                "syscall",
```
wdcui, line 344: Where is the syscall number specified?

litebox_common_linux/src/lib.rs
CvvT: All references require a lifetime. In this case, it specifies that `ctx` should outlives `SyscallRequest`.

litebox_common_linux/src/lib.rs
CvvT: I copied the comment from the man page without changes. Asked copilot to polish it.

litebox_platform_linux_userland/src/lib.rs
CvvT: At this line:

https://github.com/MSRSSP/litebox/blob/a32346aa42815c1cb2427cfa4ed1d3fb7913c9f7/litebox_platform_linux_userland/src/lib.rs#L367

litebox_common_linux/src/lib.rs
CvvT: Weidong suggested it should be either
```
SetFs(usize)
GetFs(Platform::RawMutPointer<usize>)
```

or 

```
SetFs(Platform::RawMutPointer<u8>),
GetFs(Platform::RawMutPointer<Platform::RawMutPointer<u8>>)
```

litebox/src/platform/mod.rs
```diff
@@ -50,6 +50,46 @@ pub trait ExitProvider: Sized {
     fn exit(&self, code: Self::ExitCode) -> !;
 }
 
+/// Thread management provider.
+pub trait ThreadProvider: RawPointerProvider {
+    /// Exit code for [`ThreadProvider::terminate_thread`].
+    type ExitCode;
```
jaybosamiya-ms: We can require the fact that any platform that has support for threads _also_ has support for being able to do full exits.  That allows us to re-use the `ExitCode` type, rather than repeat ourselves.
```suggestion
/// Thread management provider.
pub trait ThreadProvider: RawPointerProvider + ExitProvider {
```

litebox/src/platform/mod.rs
```diff
@@ -50,6 +50,46 @@ pub trait ExitProvider: Sized {
...
+    /// Execution context for the current thread of the guest program.
+    type ExecutionContext;
+    /// Arguments to the callback function for the new thread.
+    type ThreadArgs;
+    /// Error type for [`ThreadProvider::spawn_thread`].
+    type ThreadSpawnError: core::error::Error;
+    /// Thread identifier type for the spawned thread.
+    type ThreadId: Send + 'static;
```
jaybosamiya-ms: I think `ThreadId` can safely be _required_ to be `Sync` (nb. `T` is Sync if and only if `&T` is Send)
```suggestion
    type ThreadId: Send + Sync + 'static;
```

We also possibly want to ensure that `ThreadId` is `Clone`able.

litebox/src/platform/mod.rs
```diff
@@ -50,6 +50,46 @@ pub trait ExitProvider: Sized {
...
+
+    /// Spawn a new thread with the given entry point.
+    ///
+    /// `stack` is an optional pointer to the stack for the new thread, and `stack_size` is the
+    /// size of the stack.
```
jaybosamiya-ms: We might want to document what is expected to happen if `None` is passed?

litebox/src/platform/mod.rs
```diff
@@ -50,6 +50,46 @@ pub trait ExitProvider: Sized {
...
+    ///
+    /// The `entry_point` must be a valid entry point for the new thread.
+    unsafe fn spawn_thread(
+        &self,
+        ctx: &Self::ExecutionContext,
+        stack: Option<<Self as RawPointerProvider>::RawMutPointer<u8>>,
+        stack_size: usize,
+        entry_point: usize,
+        thread_args: alloc::boxed::Box<Self::ThreadArgs>,
+        new_thread_callback: extern "C" fn(&Self::ThreadArgs),
```
jaybosamiya-ms: I'm not quite sure why this is set up this way.  Might be easiest to discuss in person, but roughly speaking, I think a single `FnOnce` is sufficient, rather than doing it this way.

litebox/src/utils/mod.rs
```diff
@@ -56,6 +56,19 @@ impl_truncate! { i32, i16 }
...
+            #[inline(always)]
+            fn truncate(self) -> $ty {
+                self
+            }
+        }
+    };
+}
+impl_truncate_same! { u32 }
+impl_truncate_same! { i32 }
+
```
jaybosamiya-ms: I don't think this is necessary. We can reuse the existing macro.

```suggestion
impl_truncate! { u32, u32 }
impl_truncate! { i32, i32 }
```

litebox_common_linux/src/lib.rs
jaybosamiya-ms: I agree with the first of the options, i.e.
```
SetFs(usize),
GetFs(Platform::RawMutPointer<usize>)
```

This better matches the C interfaces too, which use an `unsigned long` and an `unsigned long*`.

litebox/src/platform/mod.rs
```diff
@@ -50,6 +50,46 @@ pub trait ExitProvider: Sized {
...
+
+    /// Spawn a new thread with the given entry point.
+    ///
+    /// `stack` is an optional pointer to the stack for the new thread, and `stack_size` is the
+    /// size of the stack.
+    ///
+    /// `entry_point` is the address of the first instruction that will be executed in the new thread.
+    ///
+    /// `new_thread_callback` would be invoked with the given `thread_args` in the new thread after
+    /// it is spawned.
```
jaybosamiya-ms: This is confusing: it says that `entry_point` is the first instruction that will be executed in the new thread. If that is true, when is the `new_thread_callback` executed?

We probably want to clear up the ordering here.  Also, I think it _may_ be feasible to eliminate one of the two, but we should chat IRL about this.

litebox_common_linux/src/lib.rs
```diff
@@ -751,16 +752,116 @@ bitfield::bitfield! {
...
+    pub tls: u64,
+    pub set_tid: u64,
+    pub set_tid_size: u64,
+    pub cgroup: u64,
+}
+
+#[cfg(target_arch = "x86_64")]
+pub type ThreadLocalDescriptor = u8;
+#[cfg(target_arch = "x86")]
+pub type ThreadLocalDescriptor = UserDesc;
```
jaybosamiya-ms: This is surprising, `UserDesc` is a fairly large struct, while `u8` is tiny.  At minimum, there should be a comment here as to how this is so different, and why we need separate types for the two.

litebox_common_linux/src/lib.rs
```diff
@@ -1411,3 +1561,128 @@ impl<Platform: litebox::platform::RawPointerProvider> litebox::platform::Punchth
...
+    pub eflags: u64,
+    pub rsp: u64,
+    pub ss: u64,
+    /* top of stack page */
+}
+
+#[cfg(target_arch = "x86")]
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct PtRegs {
```
jaybosamiya-ms, line 1665: This definitely needs a doc string to it, otherwise, the doc strings become dependent on whether you are compiling on 32-bit or 64-bit.

litebox_common_linux/src/lib.rs
```diff
@@ -999,118 +1109,128 @@ impl<Platform: litebox::platform::RawPointerProvider> SyscallRequest<Platform> {
     /// is allowed to panic upon receiving a syscall number (or arguments) that it does not know how
     /// to handle.
     #[expect(clippy::too_many_lines)]
-    pub fn try_from_raw(syscall_number: usize, args: &[usize; 6]) -> Result<Self, errno::Errno> {
+    pub fn try_from_raw(syscall_number: usize, ctx: &'a PtRegs) -> Result<Self, errno::Errno> {
```
jaybosamiya-ms, line 1155: I think this could be cleaned up a little bit, as it currently stands, I think this takes a hard dependency on the specific shape of `PtRegs` even though all it needs is the `&[usize; 6]`.  I don't think this is a blocker though, although I would possibly have named `get_arg` as `syscall_arg` to prevent confusion with the non-syscall calling convention.

litebox_common_linux/src/lib.rs
```diff
@@ -1411,3 +1561,128 @@ impl<Platform: litebox::platform::RawPointerProvider> litebox::platform::Punchth
...
+    pub orig_eax: u32,
+    pub eip: u32,
+    pub xcs: i32,
+    pub eflags: u32,
+    pub esp: u32,
+    pub xss: i32,
+}
+
+impl PtRegs {
+    /// Get the i-th syscall argument.
```
jaybosamiya-ms: Improve consistency
```suggestion
    /// Get the `idx`th syscall argument.
```

litebox_common_linux/src/lib.rs
```diff
@@ -1411,3 +1561,128 @@ impl<Platform: litebox::platform::RawPointerProvider> litebox::platform::Punchth
...
+    pub r13: u64,
+    pub r12: u64,
+    pub rbp: u64,
+    pub rbx: u64,
+    /* These regs are callee-clobbered. Always saved on kernel entry. */
+    pub r11: u64,
+    pub r10: u64,
+    pub r9: u64,
+    pub r8: u64,
+    pub rax: u64,
```
jaybosamiya-ms: If you use `usize`s here, then you don't need the conversions in the methods on this struct.  Things get simpler for the `get_ip` function too.

litebox/src/platform/mod.rs
CvvT: Good point! Thanks!

litebox/src/platform/mod.rs
CvvT: Added both `Clone` and `Sync`

litebox/src/platform/mod.rs
CvvT: Actually I misunderstood the document

    In the case where the CLONE_VM flag (see below) is specified, a stack must be explicitly allocated and specified.

Stack is not optional for spawning a new thread.

litebox/src/platform/mod.rs
CvvT:
```
/// `entry_point` is the address of the first instruction in the guest program that will be executed
/// in the new thread.
///
/// `new_thread_callback` would be invoked with the given `thread_args` in the new thread after
/// it is spawned, allowing the upper shim layers to perform any necessary initialization
```
I updated the comment, hope it's clear now.

litebox/src/utils/mod.rs
CvvT: You're right. Fixed!

litebox_common_linux/src/lib.rs
CvvT: I agree. Fixed!

litebox_common_linux/src/lib.rs
CvvT: This might be confusing. In x86, the provided TLS is a pointer to `UserDesc` because it invokes `set_thread_area`, while in x64, it is a pointer to an arbitrary-size memory region. Any suggestions?

litebox_common_linux/src/lib.rs
CvvT: only `clone` (and probably `rt_sigreturn`) needs more than `&[usize; 6]`; I could have a dedicate function for `clone`, but don't feel it is necessary (at least for now).

CvvT: Changed to `syscall_arg`.

litebox_common_linux/src/lib.rs
CvvT: Good point! Fixed!

litebox_common_linux/src/lib.rs
CvvT: Good catch! Added.

litebox_common_linux/src/lib.rs
CvvT: Thanks! Fixed!

litebox/src/platform/mod.rs
CvvT: Update: I removed `new_thread_callback` from the interface, which is part of `thread_args` for linux userland.

litebox_common_linux/src/lib.rs
jaybosamiya-ms: Thanks for the clarification.  I think it'd help to write a comment here saying that this is set to `u8` since it is a pointer to an arbitrary-sized memory region.

litebox_common_linux/src/lib.rs
```diff
@@ -751,34 +752,131 @@ bitfield::bitfield! {
...
+    /// Thread ID
+    pub tid: i32,
+    /// When a thread whose `clear_child_tid` is not `None` terminates, and it shares memory with other threads,
+    /// the kernel writes 0 to the address specified by `clear_child_tid` and then executes:
+    ///
+    /// futex(clear_child_tid, FUTEX_WAKE, 1, NULL, NULL, 0);
+    ///
+    /// This operation wakes a single thread waiting on the specified memory location via futex.
+    /// Any errors from the futex wake operation are
+    pub clear_child_tid: Option<Platform::RawMutPointer<i32>>,
```
jaybosamiya-ms: I think this comment might have accidentally been cut short?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -197,7 +212,225 @@ impl litebox::platform::ExitProvider for LinuxUserland {
...
+                // Unused by the syscall but would be checked by Seccomp filter if enabled.
+                syscall_intercept::systrap::SYSCALL_ARG_MAGIC,
+            )
+        }
+        .expect("Failed to exit group");
+
+        // should not reach here
+        loop {
+            unsafe { core::arch::asm!("hlt") };
+        }
```
jaybosamiya-ms: Nit: the Rust-y way to do this is simply a `unreachable!()`.  Not a big deal though, esp since it _is_ quite common to just see a `f4 eb fd` or an `eb fe` tight loop to indicate "unreachable" in quite a bit of systems-y code.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1183,18 +1436,20 @@ unsafe extern "C" {
...
-    let syscall_args: &[usize; 6] = unsafe { core::slice::from_raw_parts(args, 6) }
-        .try_into()
-        .unwrap();
-    match litebox_common_linux::SyscallRequest::try_from_raw(syscall_number, syscall_args) {
+#[unsafe(no_mangle)]
+unsafe extern "C" fn syscall_handler(
+    syscall_number: usize,
+    ctx: *mut litebox_common_linux::PtRegs,
+) -> isize {
+    // SAFETY: By the requirements of this function, it's to dereference a valid pointer to `PtRegs`.
```
jaybosamiya-ms: Nit: missing word
```suggestion
    // SAFETY: By the requirements of this function, it's safe to dereference a valid pointer to `PtRegs`.
```

litebox_platform_linux_userland/src/lib.rs
CvvT: Good point! I totally forgot this. Thanks!

litebox_common_linux/src/lib.rs
CvvT: Good catch! Thanks!

PR #159, opened 2025-06-18
Add support for extracting/managing the symbols of VTL0 kernel and modules

sangho2:
This PR adds support for extracting/managing the symbols of VTL0 kernel and modules. That is, it maintains their names and addresses using maps. These kernel symbol maps are needed to support kernel module ELF relocation/validation (which will be another PR). Specifically, this PR includes
- Parse Linux kernel's `ksymtab` and `ksymtab_gpl` which are arrays consisting of `KernelSymbol` data structures. Typically (depending on Linux kernel build configuration), `KernelSymbol` is a compact representation of value, name, and namespace offsets.
- Parse Linux kernel module ELF files (in memory) to get their export symbols (and get their `modinfo`).
- Add a new data structure to sort (with virtual addresses) and analyze memory pages copied from VTL0.


sangho2:
Discussed with @wdcui . I'll simplify this PR (will remove kernel/module symbol resolving and check whether we can use the `elf` crate).

PR #161, opened 2025-06-18
fix allocator

### LINKED ISSUE #160: LVBS memory size and contiguous memory chunk issues
LVBS VTL1 kernel does have limited memory which prevents us from copying a large amount of data from VTL0 to VTL1 (e.g., to validate large kernel modules whose sizes are several tens or hundreds of megabytes). Further, our buddy allocator prevents us from allocating a large contiguous memory chunk (e.g., `vec` for storing the whole ELF file) which is often required by `no_std` ELF parsers. We might need to either increase the VTL1 memory size or reserve a portion of VTL0 memory for module validation (with temporal SLAT protection). Also, selective ELF parsing is desired (or use `mmap`).
 
https://github.com/MSRSSP/litebox/blob/fce6a4d8f6bcba6481107e9696eb5db824ee7a4b/litebox_platform_lvbs/src/mshv/vsm.rs#L492-L496

Panics for reference. It seems that, somehow, this ended up with `rescue.alloc()`. According to `stats_alloc_actual()` and `stats_total_bytes()`, heap did have larger than 80 MB empty space.

```
// ORDER = 23
panicked at /workspace/litebox/litebox/src/mm/allocator.rs:65:21:
not implemented: requested size 16777216 is too large
```

```
// ORDER = 24
dynamic memory allocation is not supported (layout = Layout { size: 16777216, align: 16777216 (1 << 24) })
```

CvvT:
Fix #160

### DIFF
--- a/litebox/src/mm/allocator.rs
+++ b/litebox/src/mm/allocator.rs
@@ -35,7 +35,7 @@ pub trait MemoryProvider {
 /// Allocator that uses buddy allocator for pages and slab allocator for small objects.
 ///
 /// `ORDER` is the maximum order of the buddy allocator, specifying the maximum size of the
-/// allocation that can be done using the buddy allocator -- i.e., 2^ORDER * PAGE_SIZE.
+/// allocation that can be done using the buddy allocator -- i.e., 1 << (ORDER - 1).
 pub struct SafeZoneAllocator<'a, const ORDER: usize, M: MemoryProvider> {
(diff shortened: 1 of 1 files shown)


sangho2:
Cool. `page_aligned_size.trailing_zeros() as usize >= ORDER` seems to work!

CvvT:
This is a straightforward bug fix that only touches two lines (including one line for comment). I don't think an extra review is necessary and will merge this in a few hours if no objection.

PR #162, opened 2025-06-19
Support epoll related syscalls

CvvT:
This PR adds support for epoll-related syscalls like `epoll_wait`, `epoll_pwait`, `epoll_ctl`, `epoll_create`, and `epoll_create1`.

Currently, only pipes and eventfds are fully supported for epoll. I added some support for sockets so that nodejs can run, but I don't think it's complete. Litebox has [`EventManager`](https://github.com/MSRSSP/litebox/blob/3b9235c272e2b7bfafa13f955a4831d793f79b28/litebox/src/event.rs#L59), I will try to see how to connect it with epoll (in a separate PR maybe).

I also fixed some minor issues: 1. use polling instead of spin loop; and 2) fix blocking read/write while holding the lock of the file table.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1413,8 +1413,6 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
         sigsetsize: usize,
     },
     EpollCreate {
-        // the `size` argument is ignored, but must be greater than zero;
-        size: i32,
         flags: EpollCreateFlags,
     },
     ArchPrctl {
@@ -1947,12 +1945,18 @@ impl<'a, Platform: litebox::platform::RawPointerProvider> SyscallRequest<'a, Pla
                 },
                 sigsetsize: ctx.syscall_arg(5),
             },
-            Sysno::epoll_create => SyscallRequest::EpollCreate {
-                size: ctx.syscall_arg(0).reinterpret_as_signed().truncate(),
-                flags: EpollCreateFlags::empty(),
-            },
+            Sysno::epoll_create => {
+                // the `size` argument is ignored, but must be greater than zero;
+                let size: i32 = ctx.syscall_arg(0).reinterpret_as_signed().truncate();
+                if size > 0 {
+                    SyscallRequest::EpollCreate {
+                        flags: EpollCreateFlags::empty(),
+                    }
+                } else {
+                    SyscallRequest::Ret(errno::Errno::EINVAL)
+                }
+            }
             Sysno::epoll_create1 => SyscallRequest::EpollCreate {
-                size: 1,
                 flags: EpollCreateFlags::from_bits_truncate(ctx.syscall_arg(0).truncate()),
             },
             Sysno::arch_prctl => {
--- a/litebox_shim_linux/src/event.rs
+++ b/litebox_shim_linux/src/event.rs
@@ -57,7 +57,8 @@ impl<E> ObserverKey<E> {
 /// A Subject notifies interesting events to registered observers.
 struct Subject<E, F: EventsFilter<E>> {
     /// A table that maintains all interesting observers.
-    observers: spin::Mutex<BTreeMap<ObserverKey<E>, F>>,
+    observers:
+        litebox::sync::Mutex<litebox_platform_multiplex::Platform, BTreeMap<ObserverKey<E>, F>>,
     /// Number of observers.
     nums: core::sync::atomic::AtomicUsize,
 }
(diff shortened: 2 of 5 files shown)


litebox_shim_linux/src/event.rs
```diff
@@ -0,0 +1,269 @@
...
+        self.observer.upgrade()
+    }
+}
+
+/// A Subject notifies interesting events to registered observers.
+struct Subject<E, F: EventsFilter<E>> {
+    /// A table that maintains all interesting observers.
+    observers: spin::Mutex<BTreeMap<ObserverKey<E>, F>>,
+    /// Number of observers.
+    nums: core::sync::atomic::AtomicUsize,
```
wdcui, line 63: Can we get the number of observers from the BTreeMap?

litebox_shim_linux/src/syscalls/epoll.rs
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    pub(crate) fn wait(
+        &self,
+        maxevents: usize,
+        timeout: Option<Duration>,
+    ) -> Result<Vec<EpollEvent>, Errno> {
+        let mut events = Vec::new();
+        match self.ready.pollee.wait_or_timeout(
+            Events::IN,
```
wdcui, line 113: Why is it okay to just use `Events::IN` here?

litebox_shim_linux/src/syscalls/epoll.rs
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    fn add_interest(&self, fd: u32, file: &Descriptor, event: EpollEvent) -> Result<(), Errno> {
+        let mut interests = self.interests.lock();
+        let key = EpollEntryKey::new(fd, file);
+        if let Some(entry) = interests.get(&key) {
+            if entry.desc.upgrade().is_some() {
+                return Err(Errno::EEXIST);
+            }
+            // find stale entry because we don't remove it immediately after the file is closed
```
wdcui: This comment is for the code above or below?

litebox_shim_linux/src/event.rs
CvvT: Yes, just a minor optimization that does not require lock operation.

litebox_shim_linux/src/syscalls/epoll.rs
CvvT: If any event occurs, it would trigger the callback first:

https://github.com/microsoft/litebox/blob/db2123e0667aed8350d8e9b90b17071315919eaf/litebox_shim_linux/src/syscalls/epoll.rs#L309-L311

which in turn notifies all observers with `Event::IN`:

https://github.com/microsoft/litebox/blob/db2123e0667aed8350d8e9b90b17071315919eaf/litebox_shim_linux/src/syscalls/epoll.rs#L345

litebox_shim_linux/src/syscalls/epoll.rs
CvvT: The comment is for the else branch. It should fall through and replace the stale entry. I updated the comment to make it clear.

litebox_common_linux/src/lib.rs
```diff
@@ -1310,6 +1334,25 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
...
+        epfd: i32,
+        events: Platform::RawMutPointer<EpollEvent>,
+        maxevents: u32,
+        timeout: i32,
+        sigmask: Option<Platform::RawConstPointer<SigSet>>,
+        sigsetsize: usize,
+    },
+    EpollCreate {
+        // the `size` argument is ignored, but must be greater than zero;
+        size: i32,
```
jaybosamiya-ms: If it is ignored but must be greater than zero, why not just assume that it is always (say) 1, and not even have this be in the request?

If we keep the argument, we should document what happens if it is not greater than zero; for example "if not, will return EINVAL (see HISTORY in `man 2 epoll_create`)" or similar.

litebox_shim_linux/src/event.rs
```diff
@@ -0,0 +1,269 @@
...
+impl<E> ObserverKey<E> {
+    fn upgrade(&self) -> Option<alloc::sync::Arc<dyn Observer<E>>> {
+        self.observer.upgrade()
+    }
+}
+
+/// A Subject notifies interesting events to registered observers.
+struct Subject<E, F: EventsFilter<E>> {
+    /// A table that maintains all interesting observers.
+    observers: spin::Mutex<BTreeMap<ObserverKey<E>, F>>,
```
jaybosamiya-ms: Any reason this is a `spin::Mutex` rather than our `Mutex`?

litebox_common_linux/src/lib.rs
CvvT: Right. Removing `size` feels better.

litebox_shim_linux/src/event.rs
CvvT: I forgot why I used spin. Switched to our `Mutex`. Thanks!

PR #166, opened 2025-06-20
LVBS kernel module validation

sangho2:
This PR adds initial support for validating VTL0’s Linux kernel modules. It ensures that "non-relocatable bytes" in certain sections (e.g., `.text` and `.init.text`) are loaded into memory pages without modification before those pages become executable and/or write-protected. It examines the original, signed module ELF file to identify relocatable and non-relocatable bytes, considering both direct (e.g., `.rela.text`) and indirect relocations (e.g., `rela.altinstructions`). Then, it selectively compares the relocated section within VT0-provided memory pages against the original section from the signed module ELF. Note that this feature is mainly for defense-in-depth. Additionally, this PR introduces a new data structure to sort and analyze physical pages copied from VTL0 based on their virtual addresses.

### DIFF
--- a/litebox_platform_lvbs/rust-toolchain.toml
+++ b/litebox_platform_lvbs/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly-2025-03-31"
+channel = "nightly-2025-05-09"
--- a/litebox_runner_lvbs/rust-toolchain.toml
+++ b/litebox_runner_lvbs/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly-2025-03-31"
+channel = "nightly-2025-05-09"


litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -406,10 +436,26 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
...
+    let mut elf_buf = vec![0u8; elf_size];
+    memory_elf
+        .read_bytes(memory_elf.start().unwrap(), &mut elf_buf)
+        .map_err(|_| Errno::EINVAL)?;
+    memory_elf.clear();
+
+    parse_modinfo(&elf_buf).map_err(|_| Errno::EINVAL)?;
+    if !validate_kernel_module_elf(&elf_buf, &module_memory_content).map_err(|_| Errno::EINVAL)? {
+        serial_println!("VSM: Found unexpected relocations in the loaded module");
+        // return Err(Errno::EINVAL);
```
sangho2: will fix this mistake.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -0,0 +1,312 @@
...
+    },
+    endian::AnyEndian,
+    parse::ParsingTable,
+    section::SectionHeader,
+    string_table::StringTable,
+    symbol::Symbol,
+};
+use rangemap::set::RangeSet;
+
+/// This function validates the code sections (`.text` and `.init.text`) of a kernel module ELF.
```
wdcui: Do we need to check readonly data sections?

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -0,0 +1,312 @@
...
+        let mut to_validate = vec![0u8; section_buf.len()];
+        section_memory
+            .read_bytes(section_memory.start().unwrap(), &mut to_validate)
+            .map_err(|_| KernelElfError::SectionReadFailed)?;
+
+        // check whether non-relocatable bytes are modified
+        let mut diffs = Vec::new();
+        for non_reloc in reloc_ranges.gaps(&(0..section_buf.len())) {
+            for i in non_reloc {
+                if section_buf[i] != to_validate[i] {
```
wdcui: TODO: do we need a more efficient comparison than byte-wise?

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -0,0 +1,312 @@
...
+                .shdr_strtab
+                .get(usize::try_from(s.sh_name).unwrap())
+                .is_ok_and(|n| n == [".rela", target_section].join(""))
+    }) {
+        let relas = elf_params
+            .elf
+            .section_data_as_relas(&rela_shdr)
+            .map_err(|_| KernelElfError::ElfParseFailed)?;
+        for rela in relas {
+            if let Ok(sym) = elf_params.symtab.get(usize::try_from(rela.r_sym).unwrap()) {
```
wdcui: All three levels of `if` can be merged into one?

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -0,0 +1,312 @@
...
+            }
+
+            let reloc_size: u64 = match rela.r_type {
+                R_X86_64_64 => 8,
+                R_X86_64_32 | R_X86_64_32S | R_X86_64_PLT32 | R_X86_64_PC32 => 4,
+                _ => {
+                    todo!("Unsupported relocation type {:?}", rela.r_type);
+                }
+            };
+            if usize::try_from(rela.r_offset + reloc_size).unwrap() <= section_buf.len() {
```
wdcui: Is the `r_offset` section or elf relative?

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -0,0 +1,312 @@
...
+            {
+                let reloc_size: u64 = match rela.r_type {
+                    R_X86_64_64 => 8,
+                    R_X86_64_32 | R_X86_64_32S | R_X86_64_PLT32 | R_X86_64_PC32 => 4,
+                    _ => {
+                        todo!("Unsupported relocation type {:?}", rela.r_type);
+                    }
+                };
+
+                // section-based relocation relies on `r_addend` to specify the relocation offset
```
wdcui: Double check this? I thought it also uses `r_offset`?

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -0,0 +1,312 @@
...
+
+    if let Some(shdr) = shdrs.iter().find(|s| {
+        s.sh_flags & u64::from(SHF_ALLOC) != 0
+            && s.sh_size > 0
+            && shdr_strtab
+                .get(usize::try_from(s.sh_name).unwrap())
+                .is_ok_and(|n| n == ".modinfo")
+    }) {
+        let start = usize::try_from(shdr.sh_offset).unwrap();
+        let end = start + usize::try_from(shdr.sh_size).unwrap();
```
wdcui: Should we use a safe add that checks overflow?

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -0,0 +1,312 @@
...
+                        );
+                    }
+                }
+            }
+        }
+    }
+    Ok(())
+}
+
+/// This function parses the `.modinfo` section of a kernel module ELF
```
wdcui, line 316: This function only ensures .modinfo is valid? It doesn't look like it has any output.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -364,8 +373,12 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
         nranges,
     );
 
-    // collect and maintain the memory ranges of a module locally until the module is validated and registered in the global map
-    let mut module_memory = ModuleMemory::new();
+    let mut module_memory_metadata = ModuleMemoryMetadata::new();
+
+    // collect and maintain the memory ranges of a module locally until the module is validated and its metadata is registered in the global map
+    // we don't maintain this content in the global map due to memory overhead. Instead, we could add its hash value to the global map to check the integrity.
+    let mut module_memory_content = ModuleMemoryContent::new();
```
wdcui: These two names `ModuleMemoryContent` and `MemoryContent` are not clear about what each one is for.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -406,10 +436,26 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
...
+    );
+
+    let mut elf_buf = vec![0u8; elf_size];
+    memory_elf
+        .read_bytes(memory_elf.start().unwrap(), &mut elf_buf)
+        .map_err(|_| Errno::EINVAL)?;
+    memory_elf.clear();
+
+    parse_modinfo(&elf_buf).map_err(|_| Errno::EINVAL)?;
+    if !validate_kernel_module_elf(&elf_buf, &module_memory_content).map_err(|_| Errno::EINVAL)? {
```
wdcui: I don't understand the difference between the sources of `elf_buf` and `module_memory_content` if both were provided by vtl0.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -795,3 +891,239 @@ fn protect_physical_memory_range(
...
+
+    pub fn find_section_by_name(&self, name: &str) -> Option<&MemoryContent> {
+        match name {
+            ".text" => Some(&self.text),
+            ".init.text" => Some(&self.init_text),
+            _ => None,
+        }
+    }
+
+    pub fn write_vtl0_phys_bytes_by_type(
```
wdcui: Should these functions be either private or (crate) pub?

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: I think checking `.rodata` is doable. Let me take a look at this.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: I think so. The current one is mostly for debugging purposes (we don't need to know the exact indexes in practice). To improve the performance, we can use our old approach: patch `section_buf` first and compare it against `to_validate` buffer-wise.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: Let me see whether I can simplify this. `let chain` in Rust is still an unstable feature, so I might need to use `and_then()` or `filter()`.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: `r_offset` is section relative.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: `section_relocation` is misleading. Let me change it (`binary_patch`?).

This function aims to reproduce the Linux kernel module loader's binary patching behaviors. For example, when the kernel module loader processes a module like below

```Bash
$ readelf -a evdev.ko
...
Relocation section '.rela.altinstructions' at offset 0x7758 contains 6 entries:
  Offset          Info           Type           Sym. Value    Sym. Name + Addend
000000000000  000100000002 R_X86_64_PC32     0000000000000000 .text + 17f3
000000000004  000400000002 R_X86_64_PC32     0000000000000000 .altinstr_replacement + 0
00000000000e  000100000002 R_X86_64_PC32     0000000000000000 .text + 17fb
000000000012  000400000002 R_X86_64_PC32     0000000000000000 .altinstr_replacement + 3
...
```
the module loader can selectively patch `.text + 0x17f3` and `.text + 0x17fb` using the bytes stored at `.altinstr_replacement + 0x0` and `.altinstr_replacement + 0x3` according to kernel and hardware configurations. In practice, the module loader patches the `.altinstructions` section using `.rela.altinstructions` first and then patches `.text` using the patched `.altinstructions`. We skip this and focus on the final result. Other sections like `.rela.call_sites` are with different patching algorithms but they also selectively patch `<section> + r_addend` in the end.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: True. This function is mainly for debugging purposes. Let me use `#[cfg(debug_assertions)]`.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: Let me use `checked_add()` and `checked_sub()`.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Agree with you. Let me rename them (`ModuleMemory` and `MemoryContainer`?). Basically, I wanted to maintain the sections of a kernel module separately to ease analysis. `MemoryContent` (or `MemoryContainer`) is a generic data structure.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: `elf_buf` contains the original module ELF binary which is cryptographically signed by the kernel build pipeline. VTL0 does not have access to the private portion of this signing key. My next [PR](https://github.com/microsoft/litebox/commit/9ea8c1108702ae04b243a213d9af7ee03edfac1f) will verify this signature using the system certificate (also provided by VTL0 but secured by UEFI Secure Boot).

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Let me change the scope of these member functions.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -15,13 +18,13 @@ use elf::{
 };
 use rangemap::set::RangeSet;
 
-/// This function validates the code sections (`.text` and `.init.text`) of a kernel module ELF.
-/// In particular, this function checks the integrity of "non-relocatable bytes" in ELF. We check the
-/// signature of the entire ELF (including relocation information) before calling this function,
-/// so what the adversary can do with relocatable bytes is limited.
+/// This function validates certain code (`.text`, `.init.text`) and data (`.init.rodata`, `.data..ro_after_init`)
```
wdcui: Please update this comment based on our offline discussion.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -61,44 +65,76 @@ pub fn validate_kernel_module_elf(
...
         // figure out relocatable ranges
         let mut reloc_ranges = RangeSet::<usize>::new();
         get_symbol_relocations(elf_params, target_section, &section_buf, &mut reloc_ranges)?;
-        get_section_relocations(elf_params, target_section, &section_buf, &mut reloc_ranges)?;
+        get_binary_patches(elf_params, target_section, &section_buf, &mut reloc_ranges)?;
 
         let mut to_validate = vec![0u8; section_buf.len()];
         section_memory
-            .read_bytes(section_memory.start().unwrap(), &mut to_validate)
+            .read_bytes(
```
wdcui, line 100: The function definition of `read_bytes` is a little unusual. It takes a `self` but never uses it?

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -61,44 +65,76 @@ pub fn validate_kernel_module_elf(
...
-        let mut diffs = Vec::new();
-        for non_reloc in reloc_ranges.gaps(&(0..section_buf.len())) {
-            for i in non_reloc {
-                if section_buf[i] != to_validate[i] {
-                    diffs.push(i);
-                }
+        #[cfg(not(debug_assertions))]
+        {
+            for reloc in reloc_ranges {
+                section_buf[reloc.clone()].copy_from_slice(&to_validate[reloc.clone()]);
```
wdcui: Minor: it might be helpful for future readers of this code to name variables in a way that it's clear that they are from memory or elf. For instance, `section_memory` and `to_validate` are from memory, but `section_buf` is from the elf.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -182,9 +210,9 @@ fn is_allowed_rela_section(name: &str) -> bool {
     )
 }
 
-// This function handles `.rela.*` sections which can relocate `.text` or `.init.text` (section-based relocations).
-// A rela section can relocate `.text` or `init.text` sections if it has unnamed symbols which belong to these sections.
-fn get_section_relocations(
+// This function handles `.rela.*` sections which can patch our target sections. For example,
+// a rela section can relocate `.text` if it has unnamed symbols which belong to `.text`.
+fn get_binary_patches(
```
wdcui: How is `get_binary_patches` different from `get_symbol_relocations`?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -377,8 +379,8 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
 
     // collect and maintain the memory ranges of a module locally until the module is validated and its metadata is registered in the global map
     // we don't maintain this content in the global map due to memory overhead. Instead, we could add its hash value to the global map to check the integrity.
-    let mut module_memory_content = ModuleMemoryContent::new();
-    let mut memory_elf = MemoryContent::new();
+    let mut module_memory = ModuleMemory::new();
+    let mut memory_elf = MemoryContainer::new();
```
wdcui: Minor: would `elf_memory` be slightly better? Of course, a more tedious one would `module_from_memory` and `module_from_elf`. :-)

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: I've updated the comments.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: it does uses `self.pages.range()`

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: Yes. I agree with you that variable and function names were unclear. I renamed them like `original_elf_data`,  `section_from_elf`, `section_in_memory`, `module_in_memory`, `module_as_elf`, ...

litebox_platform_lvbs/src/mshv/kernel_elf.rs
sangho2: I renamed these two functions like `identify_direct_relocations` and `identify_indirect_relocations`. Technically, the former analyzes `.rela.text` whereas the latter analyzes all other `rela*` affecting `.text`.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Renamed them: `module_in_memory`, `module_as_elf`. kind of difficult to name them because both are in memory anyhow. lol.

PR #173, opened 2025-06-23
Use ld audit to patch loaded shared objects

wdcui:
This PR adds the support to use ld audit to patch loaded shared objects.

The ld provides an audit interface that allows a shared object to monitor loading (and symbol lookups) of other shared objects. We create our own audit.so to patch other shared objects for syscall hooking. It works in the following way:

1. Litebox's loader sets the syscall entry into the environment variable LITEBOX_SYSCALL_ENTRY.
2. Audit.so picks up the syscall entry from the environment variable.
3. Audit.so uses this syscall entry to make its own syscalls such as open/close/mmap/mprotect.
4. Audit.so patches shared objects by updating its syscall placeholder and loading the trampoline code into memory.

Note that it's okay to have the audit.so to use libc as long as it doesn't use libc to make syscalls.

### DIFF
--- a/litebox_shim_linux/src/loader/elf.rs
+++ b/litebox_shim_linux/src/loader/elf.rs
@@ -28,7 +28,7 @@ use arrayvec::ArrayString;
 use core::fmt::Write;
 
 fn number_to_hex(n: u64) -> ArrayString<16> {
-    let mut buf = ArrayString::<16>::new(); // <-- constant generic
+    let mut buf = ArrayString::<16>::new();
     write!(&mut buf, "{n:x}").unwrap();
     buf
 }


wdcui:
This is the build command I use to compile the audit.so:

`cc -Wall -fPIC -shared -o litebox_rtld_audit.so rtld_audit.c`

Can someone help put this into the build.rs?

CvvT:
One question: if the audit lib requires `getenv` from libc but also needs to patch libc before it is loaded, how the circular dependency is resolved?

CvvT:
It seems that libc is loaded and runs `__libc_early_init` before our audit lib runs. and `__libc_early_init` would call syscalls, e.g., `getrlimit`.

wdcui:
`getenv` works possibly because it was after the early init. `getenv` itself doesn't involve any syscalls. It's a great observation that `__libc_early_init` is called before the audit library is loaded. I think a potential solution is to patch libc to hook `__libc_early_init` to set up things that are necessary.

CvvT:
I created a new PR #200 for it.

PR #174, opened 2025-06-23
Add prlimit64

CvvT:
Add syscall handlers for `getrlimit`, `setrlimit`, and `prlimit64`. For now, setting new resource limits is not supported, only getting current limit is allowed, and hence we only need to return constant/default values for current limit.

Only `RLIMIT_STACK` and `RLIMIT_NOFILE` are supported now, but they are not enforced yet.

Also, getting/setting resource limit for other process is not supported.


litebox_common_linux/src/lib.rs
```diff
@@ -929,6 +929,92 @@ pub struct EpollEvent {
...
+        },
+    }
+}
+
+pub fn rlimit64_to_rlimit(rlim: Rlimit64) -> Rlimit {
+    Rlimit {
+        rlim_cur: if rlim.rlim_cur == u64::MAX {
+            rlim_t::MAX
+        } else {
+            rlim.rlim_cur as rlim_t
```
wdcui: `rlim_cur` may be greater than a `u32::MAX` when `usize` is `u32`?

litebox_common_linux/src/lib.rs
```diff
@@ -1163,6 +1249,25 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
...
+    },
+    Setrlimit {
+        resource: RlimitResource,
+        rlim: Platform::RawConstPointer<Rlimit>,
+    },
+    Prlimit {
+        pid: Option<i32>,
+        /// The resource for which the limit is being queried.
+        resource: RlimitResource,
+        /// If the new_limit argument is a not None, then the rlimit structure to which it points
```
wdcui: Minor (typo): is a not None => is not a None?

litebox_common_linux/src/lib.rs
```diff
@@ -1163,6 +1249,25 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
...
+        /// is used to set new values for the soft and hard limits for resource.
+        new_limit: Option<Platform::RawConstPointer<Rlimit64>>,
+        /// If the old_limit argument is a not None, then a successful call to prlimit() places the
```
wdcui: "is a not None"?

litebox_common_linux/src/lib.rs
```diff
@@ -1515,6 +1620,65 @@ impl<'a, Platform: litebox::platform::RawPointerProvider> SyscallRequest<'a, Pla
                 buf: Platform::RawMutPointer::from_usize(ctx.syscall_arg(2)),
                 bufsiz: ctx.syscall_arg(3),
             },
+            #[cfg(target_arch = "x86_64")]
+            Sysno::getrlimit => {
+                let resource: u32 = ctx.syscall_arg(0).truncate();
```
wdcui: The function definition in Linux has the first argument as `int`?

litebox_common_linux/src/lib.rs
CvvT: Good catch! Changed to `>= rlim_t::MAX`

litebox_common_linux/src/lib.rs

litebox_common_linux/src/lib.rs

litebox_common_linux/src/lib.rs

PR #176, opened 2025-06-24
Support set_robust_list/get_robust_list

CvvT:
This PR adds syscall handler for `set_robust_list` and `get_robust_list`. However, the most critical function [`handle_futex_death`](https://elixir.bootlin.com/linux/v6.16-rc3/source/kernel/futex/core.c#L1014) is not implemented yet because most applications call `set_robust_list` with an empty list and do not need it.


litebox_common_linux/src/lib.rs
```diff
@@ -1015,6 +1020,53 @@ pub enum RlimitResource {
...
+     * _might_ have taken. We check the owner TID in any case,
+     * so only truly owned locks will be handled.
+     */
+    pub list_op_pending: Platform::RawConstPointer<RobustList<Platform>>,
+}
+
+impl<Platform: litebox::platform::RawPointerProvider> Clone for RobustListHead<Platform> {
+    fn clone(&self) -> Self {
+        Self {
+            list: self.list.clone(),
```
wdcui, line 1041: Is this clone() a deep copy?

litebox_common_linux/src/lib.rs
```diff
@@ -1272,6 +1324,14 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
         tidptr: Platform::RawMutPointer<i32>,
     },
     Gettid,
+    SetRobustList {
+        head: Platform::RawConstPointer<RobustListHead<Platform>>,
+    },
+    GetRobustList {
+        pid: Option<i32>,
+        head: Platform::RawMutPointer<usize>,
```
wdcui, line 1291: Why is `head` in `GetRobustList` a `usize`?

litebox_common_linux/src/lib.rs
CvvT: No, this just clones the pointer.

litebox_common_linux/src/lib.rs
CvvT: This is similar to what we have discussed for `setfs/getfs`. I changed `SetRobustList` to:

```
SetRobustList {
        head: usize,
},
```

Issue #142, opened 2025-06-06
DO NOT use getauxval


CvvT:
Fair point, although I will also note that Linux enables it by default since ~5 years ago: https://github.com/torvalds/linux/commit/b745cfba44c152c34363eea9e052367b6b1d652b

For now, `getauxval` seems reasonable to me, as long as we open a tracking issue, thanks!

_Originally posted by @jaybosamiya-ms in https://github.com/MSRSSP/litebox/pull/137#discussion_r2130739150_

CvvT:
Now we can use libc/std freely in the platform.

PR #178, opened 2025-06-24
Support uid/gid/euid/egid

CvvT:
Support syscalls `getuid`/`geteuid`/`getgid`/`getegid`. Currently, it reads these values from the auxiliary vector via `libc::getauxval` which we should avoid (see #142).

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -948,10 +948,10 @@ pub struct ThreadLocalStorage<Platform: litebox::platform::RawPointerProvider> {
 /// Credentials of a process
 #[derive(Clone)]
 pub struct Credentials {
-    pub uid: u32,
-    pub euid: u32,
-    pub gid: u32,
-    pub egid: u32,
+    pub uid: usize,
+    pub euid: usize,
+    pub gid: usize,
+    pub egid: usize,
 }
 
 impl<Platform: litebox::platform::RawPointerProvider> ThreadLocalStorage<Platform> {
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -184,18 +184,12 @@ impl LinuxUserland {
     fn get_user_info() -> litebox_common_linux::Credentials {
         litebox_common_linux::Credentials {
             // Alternatively, we could read those from `/proc/self/aux`
-            uid: unsafe { syscalls::syscall0(syscalls::Sysno::getuid) }
-                .expect("failed to get UID")
-                .truncate(),
+            uid: unsafe { syscalls::syscall0(syscalls::Sysno::getuid) }.expect("failed to get UID"),
             euid: unsafe { syscalls::syscall0(syscalls::Sysno::geteuid) }
-                .expect("failed to get EUID")
-                .truncate(),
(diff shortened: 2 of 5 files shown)


litebox_common_linux/src/lib.rs
```diff
@@ -964,6 +973,8 @@ pub struct Task<Platform: litebox::platform::RawPointerProvider> {
     /// of the futex has died. This notification consists of two pieces: the FUTEX_OWNER_DIED bit is set in the futex word,
     /// and the kernel performs a futex(2) FUTEX_WAKE operation on one of the threads waiting on the futex.
     pub robust_list: Option<Platform::RawConstPointer<RobustListHead<Platform>>>,
+    /// The thread's credentials
```
wdcui: uid/gid are associated with a process, not a thread?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -181,6 +181,16 @@ impl LinuxUserland {
         reserved_pages
     }
 
+    fn get_user_info() -> litebox_common_linux::Credentials {
```
wdcui, line 184: Why do we have two `litebox_common_linux::Credentials` here?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -398,6 +398,13 @@ impl ElfLoader {
             elf.entry()
         };
 
+        let user_info = litebox_platform_multiplex::platform()
+            .with_thread_local_storage_mut(|tls| (*tls.current_task.credentials).clone());
```
wdcui, line 449: What code originally stores the information into `tls.current_task.credentials`?

litebox_platform_linux_userland/src/lib.rs
CvvT: You meant the first one for the type of function return and the second one in the following line for constructing the return value?

litebox_shim_linux/src/loader/elf.rs
CvvT: Set by the first thread when initializing the platform:

https://github.com/microsoft/litebox/blob/086e6cbf70608b414326ed7cd85205e88e9636bc/litebox_platform_linux_userland/src/lib.rs#L202

All new spawn threads would have a copy.

litebox_common_linux/src/lib.rs
CvvT: It's shared across threads. That's why I use `Arc`. Updated the comment accordingly.

CvvT:
> Support syscalls `getuid`/`geteuid`/`getgid`/`getegid`. Currently, it reads these values from the auxiliary vector via `libc::getauxval` which we should avoid (see #142).

Update: changed to use raw syscalls (e.g., getuid) to get that info. Note that those syscalls are invoked during the initialization of `platform` and before seccomp filter is enabled. They only need to be called once.

CvvT:
> LGTM thanks! Minor comment: it seems like we are doing a lot of conversion back and forth to/from `u32` and `usize`. I wonder if this might be cleaned up a little bit by just having the `Credentials` store `usize`s rather than `u32`s. Not crucial though, feel free to merge without any further changes too.

Good point! Using usize is better.

PR #179, opened 2025-06-24
Allow making pages readable, writable, and executable

CvvT:
As mentioned in https://github.com/microsoft/litebox/pull/107#issuecomment-2856711807, nodejs requires RWX pages because of JIT. It's possible to run node in jitless mode with the flag `--jitless`. However, running wasm with jitless node is not supported yet: https://github.com/nodejs/node/issues/28800#issuecomment-514027732, which causes problem when running the netlify example:

node:internal/deps/undici/undici:6340
        mod = await WebAssembly.compile(llhttpWasmData || require_llhttp_wasm());
        ^

ReferenceError: WebAssembly is not defined
    at lazyllhttp (node:internal/deps/undici/undici:6340:9)
    at lib/dispatcher/client-h1.js (node:internal/deps/undici/undici:6429:25)
    at __require (node:internal/deps/undici/undici:6:50)
    at lib/dispatcher/client.js (node:internal/deps/undici/undici:8123:21)
    at __require (node:internal/deps/undici/undici:6:50)
    at lib/dispatcher/pool.js (node:internal/deps/undici/undici:8576:18)
    at __require (node:internal/deps/undici/undici:6:50)
    at lib/dispatcher/agent.js (node:internal/deps/undici/undici:8677:16)
    at __require (node:internal/deps/undici/undici:6:50)
    at lib/global.js (node:internal/deps/undici/undici:8791:17)

Node.js v24.2.0

### DIFF
--- a/litebox/src/mm/mod.rs
+++ b/litebox/src/mm/mod.rs
@@ -475,7 +475,7 @@ where
     /// It is highly recommended to minimize the use of this function and to prefer safer alternatives
     /// whenever possible. If this function must be used, ensure that the memory region is locked down
     /// and access is strictly controlled.
-    pub unsafe fn make_pages_rw_exec(
+    pub unsafe fn make_pages_rwx(
         &self,
         ptr: Platform::RawMutPointer<u8>,
         len: usize,
--- a/litebox_shim_linux/src/syscalls/mm.rs
+++ b/litebox_shim_linux/src/syscalls/mm.rs
@@ -197,7 +197,7 @@ pub(crate) fn sys_mprotect(
         ProtFlags::PROT_READ_WRITE => unsafe { pm.make_pages_writable(addr, len) },
         ProtFlags::PROT_READ => unsafe { pm.make_pages_readable(addr, len) },
         ProtFlags::PROT_NONE => unsafe { pm.make_pages_inaccessible(addr, len) },
-        ProtFlags::PROT_READ_WRITE_EXEC => unsafe { pm.make_pages_rw_exec(addr, len) },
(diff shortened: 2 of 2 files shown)


litebox/src/mm/mod.rs
```diff
@@ -459,6 +459,36 @@ where
...
+    ///
+    /// The caller must ensure the following:
+    /// 1. The memory region is only used for legitimate purposes, such as JIT compilation,
+    ///    where writable and executable permissions are strictly necessary.
+    /// 2. The memory region is properly sanitized and does not contain malicious or unintended code.
+    ///
+    /// It is highly recommended to minimize the use of this function and to prefer safer alternatives
+    /// whenever possible. If this function must be used, ensure that the memory region is locked down
+    /// and access is strictly controlled.
+    pub unsafe fn make_pages_rw_exec(
```
jaybosamiya-ms: Nit: `rw_exec` is completely understandable, but it feels like `r`/`w` are placed differently from `x` due to how it is written.  Maybe we might simply prefer `rwx` directly?  Not terribly crucial, since I think this is a rare function to invoke anyways.

jaybosamiya-ms: Discussion point: do we want to feature-gate this function, so that it is easier to immediately tell whether an application has `rwx` or not?  I don't think we want to do this as part of this PR, but writing this here since the general idea of feature gates has been on my mind for a while.  There are upsides and downsides to feature-gates, so we want to be careful with when/how we use them, which is why I am _not_ recommending it here, but only using it as a starting point to jump-start a discussion (I'll push this to a new issue shortly).

CvvT: I agree. Changed to `rwx`. Thanks!

PR #180, opened 2025-06-25
Use a shared FD storage within LiteBox's subsystems

jaybosamiya-ms:
This is part 1 of multiple PRs towards migrating to a shared FD storage (see #31). I am splitting this into multiple parts in order to not have the final PR be _too_ massive to review (or indeed even just manage). However, unfortunately, the overall migration to this is a fairly deep change, thus it is inevitable that there is going to be non-trivial amounts to review.

This specific PR focuses on setting up the actual shared storage itself (i.e., `Descriptors`) and migrating the various `litebox` crate subsystems into using it.  The cleanup for the shims (to remove their extra storage) will be in a future PR.

Essentially the `Descriptors` storage supports storing descriptor entries, and managing FDs for any `FdEnabledSubsystem` (a new trait introduced in this PR).  A subsystem can become an `FdEnabledSubsystem` simply by invoking a macro, simplifying the addition/update of subsystems, or indeed even if we need to add additional functionality to the shared FD storage.

Each `FdEnabledSubsystem` can define the type of a descriptor entry, and the shared descriptor storage is able to handle these varied types automatically into a shared storage.  Additionally, the file descriptors provided by `Descriptors` storage are all strongly typed (i.e., you cannot accidentally use a socket FD for a file-system; or even use an in-mem-fs FD for a device).  This strong typing discipline not only allows us to get strong guarantees from ownership, but also good ergonomics for getting descriptor entries from within a subsystem.  In particular, despite the shared storage, when accessing with a strongly-typed FD, when a subsystem asks for an entry, it gets back an entry of the correct type, and it cannot _accidentally_ access a different type of descriptor entry than what it is aware of.

This PR also sets up the basics for removing the extra storage that exists in the Linux shim.  This is the "raw integer" interfaces that I've set up within the `Descriptors`.  The actual usage of the raw integer interfaces will show up in a future PR.

---

### Expected API breakages

- Some of the types no longer implement the `*UnwindSafe` marker traits: this is not an issue, since we don't expect to `catch_unwind` them anyways.  Mostly just a technicality in terms of breakage.
- We no longer have `FileFd`/`SocketFd` as two types, instead `TypedFd` is the parametric type to be used. This actually increases our precision with strong-typing (in the past, you could use a `FileFd` produced by one file system in another file system; this would have had weird effects; we now actually statically ensure this cannot happen). 
- `WaitableBuilder` no longer needs two interfaces.

These breakages are all accounted for in the shims above them; I am writing the above notes mostly for documentation, and to note that the semver-checks _will_ complain about these.

### DIFF
--- a/litebox/src/litebox.rs
+++ b/litebox/src/litebox.rs
@@ -24,6 +24,10 @@ impl<Platform: RawSyncPrimitivesProvider + ExitProvider> LiteBox<Platform> {
     /// Create a new (empty) [`LiteBox`] instance for the given `platform`.
     pub fn new(platform: &'static Platform) -> Self {
         let sync = Synchronization::new_from_platform(platform);
+        // We set `descriptors` to `None` and replace it out with a `Some` after creation due to a
+        // circular dependency between the two types for their initialization. The public interfaces
+        // here do not need to deal with any of this though; see the post-creation invariant
+        // guarantee written on `LiteBoxX::descriptors`.
         let descriptors = sync.new_rwlock(None);
         let ret = Self {
             x: Arc::new(LiteBoxX {


litebox/src/fd/mod.rs
```diff
@@ -0,0 +1,486 @@
...
+
+    /// Insert `entry` into the descriptor table, returning an `OwnedFd` to this entry.
+    pub(crate) fn insert<Subsystem: FdEnabledSubsystem>(
+        &mut self,
+        entry: impl Into<Subsystem::Entry>,
+    ) -> TypedFd<Subsystem> {
+        let entry = DescriptorEntry {
+            entry: alloc::boxed::Box::new(entry.into()),
+        };
+        let idx = self
```
wdcui, line 55: This code pattern is new to me. Why do we do `push` then `replace`?

litebox/src/fs/in_mem.rs
```diff
@@ -272,15 +271,20 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
+        &self,
+        fd: &FileFd<Platform>,
+        offset: isize,
+        whence: SeekWhence,
+    ) -> Result<usize, SeekError> {
+        let descriptor_table = self.litebox.descriptor_table();
         let Descriptor::File {
             file,
             read_allowed: _,
             write_allowed,
```
wdcui: Why do we have `write_allowed` for `seek`?

litebox/src/litebox.rs
```diff
@@ -23,9 +24,18 @@ impl<Platform: RawSyncPrimitivesProvider + ExitProvider> LiteBox<Platform> {
     /// Create a new (empty) [`LiteBox`] instance for the given `platform`.
     pub fn new(platform: &'static Platform) -> Self {
         let sync = Synchronization::new_from_platform(platform);
-        Self {
-            x: Arc::new(LiteBoxX { platform, sync }),
-        }
+        let descriptors = sync.new_rwlock(None);
```
wdcui, line 31: Just for my own understanding (kind of similar to my question earlier): why do we `new` then `replace`?

litebox/src/fd/mod.rs
jaybosamiya-ms: The `push` is done only if no empty (i.e., `None`) location is found.  This way this statement simply is guaranteeing that `idx` is pointing to a `None` location.  We could alternatively write it via a `if None is found { do replace } else { do push }`, but idk, in this case I felt it was simpler just to have each step be on its own, rather than repeating the new generated arc at each point.

Codegen wise, I don't think it would make any noticeable impact.

litebox/src/fs/in_mem.rs
jaybosamiya-ms: Good point, I believe this is just an unused variable, and it should have been `write_allowed: _`.  Weird that Rust didn't complain about it.  I should check up settings again to make sure we are enabling things.  Thanks for pointing this out, we are probably just accidentally suppressing the warning somewhere.

litebox/src/litebox.rs
jaybosamiya-ms: Ah, this one is a slightly different scenario than the other one.  In this case, to create a `Descriptors` object, we need to finish making a `LiteBox` object, but to make a `LiteBox` object, we need a `Descriptors` object.  To break this dependency loop, I instead make it an `Option<Descriptors>` so that I can do deferred initialization of the `Descriptors` object.  This way, a `LiteBox` object can be finished being made _without_ a `Descriptors` object yet, and then I can make the `Descriptors` object using this `LiteBox` object, and then connect the two of them up.  Thus you see `new(None)` followed by `replace` (which makes it a `Some`).

litebox/src/fs/in_mem.rs
jaybosamiya-ms: Turns out it is a fairly old suppression from many months ago:
https://github.com/microsoft/litebox/blob/db371be74220899f950cc33378ffd81af4b77c7c/litebox/src/lib.rs#L13-L15

I'll remove the suppression and fix up anything that gets triggered by this soon.

jaybosamiya-ms: Opened #183 to fix this

litebox/src/litebox.rs
wdcui: I see. I didn't realize the circular dependency. If you didn't have a comment in the code, it might be helpful to point out the workaround for the circular dependency in a comment.

litebox/src/fs/devices/stdio.rs
```diff
@@ -65,30 +68,43 @@ impl<Platform: crate::platform::StdioProvider> FileSystem<Platform> {
...
+    ) -> Result<FileFd<Platform>, OpenError> {
         let path = self.absolute_path(path)?;
         match path.as_str() {
             "/dev/stdin" => {
                 if flags == OFlags::RDONLY && mode.is_empty() {
-                    Ok(FileFd { x: OwnedFd::new(0) })
+                    Ok(self
+                        .litebox
+                        .descriptor_table_mut()
+                        .insert(StdioStream::Stdin))
```
CvvT, line 87: How can we ensure their corresponding fd integer are 0, 1, and 2?

litebox/src/fd/mod.rs
```diff
@@ -0,0 +1,486 @@
...
+
+    /// Obtain the typed FD for the raw integer value of the `fd`.
+    ///
+    /// This operation will "consume" the raw integer (thus future [`Self::fd_from_raw_integer`]
+    /// might not refer to this file descriptor unless it is returned back via
+    /// [`Self::fd_into_raw_integer`]).
+    ///
+    /// You almost definitely want [`Self::fd_from_raw_integer`] instead, and should only use this
+    /// if you really know you want to consume the descriptor.
+    pub fn fd_consume_raw_integer<Subsystem: FdEnabledSubsystem>(
```
CvvT, line 294: Why not just consume it in `remove`?

litebox/src/fs/devices/stdio.rs
```diff
@@ -181,33 +219,40 @@ impl<Platform: crate::platform::StdioProvider> super::super::FileSystem for File
...
-        fd: &FileFd,
+        fd: &FileFd<Platform>,
         metadata: T,
     ) -> Result<Option<T>, crate::fs::errors::SetMetadataError<T>> {
         unimplemented!()
     }
 }
+
+crate::fd::enable_fds_for_subsystem! {
+    @ Platform: { crate::sync::RawSyncPrimitivesProvider + crate::platform::StdioProvider };
```
CvvT, line 264: What's the meaning of this `@`?

litebox/src/fs/devices/stdio.rs
jaybosamiya-ms: This is something that the shim needs to do during setup.  Will be part of the next PR.  There is no reason for the `litebox` crate itself to be aware of 0, 1, 2 assignment.

litebox/src/fs/devices/stdio.rs
jaybosamiya-ms: This `@` is just an arbitrary symbol I picked when writing the `enable_fds_for_subsystem` macro to resolve an inherent ambiguity that showed up when attempting to support optional-parametricity for the subsystems.  To have more "natural" looking syntax, we would need a **lot** more complexity in the macro (essentially needing us to switch to introducing a new crate to write a procedural macro), for practically no gain.  Thus, we just need to have this (mildly ugly) `@` sign show up before the parametricity constraints.

litebox/src/fd/mod.rs
jaybosamiya-ms: `remove` consumes a `TypedFd`, `fd_consume_raw_integer` consumes the raw integer value, and gives you the `TypedFd` that can be consumed.

We need the raw integer values for some shims (such as Linux), but don't need it for other shims (such as Wasm).  Raw integers are a form of "weakening" of the guarantees, which is why I don't want to force everything to use raw integers only.  This is why we end up with a two-stage process for creation/removal of raw integer FDs.

PR #182, opened 2025-06-26
LVBS kernel module signature verification

sangho2:
This PR adds support for verifying Linux kernel module signature. If module signing is enabled, a Linux kernel build pipeline signs each kernel module and appends the signature to it. We extract and verify the signature of each kernel module before validating its relocations and changing its page permissions.
This signature verification is based on the system certificate which contains the public portion of the build pipeline key. It is provided by the VTL0 kernel and locked down before the `end_of_boot` signal. Its authenticity depends on UEFI Secure Boot which ensures only trusted software is loaded during the boot process. Also, both the VTL0 kernel and VTL1 do not have access to the private portion of the key.

### DIFF
--- a/litebox_platform_lvbs/Cargo.toml
+++ b/litebox_platform_lvbs/Cargo.toml
@@ -27,6 +27,7 @@ cms = { version = "0.2.3", default-features = false, features = ["alloc"] }
 rsa = { version = "0.9.8", default-features = false }
 sha2 = { version = "0.10.9", default-features = false, features = ["force-soft", "oid"] }
 x509-cert = { version = "0.2.5", default-features = false }
+const-oid = { version = "0.9.6", default-features = false, features = ["db"] }
 
 [target.'cfg(target_arch = "x86_64")'.dependencies]
 x86_64 = { version = "0.15.2", default-features = false, features = ["instructions"] }
--- a/litebox_platform_lvbs/src/mshv/mem_integrity.rs
+++ b/litebox_platform_lvbs/src/mshv/mem_integrity.rs
@@ -1,4 +1,4 @@
-//! Functions for checking the integrity of VTL0 kernel modules
+//! Functions for checking the memory integrity of VTL0 kernel image and modules
 
 #[cfg(debug_assertions)]
 use alloc::vec::Vec;
@@ -8,6 +8,7 @@ use crate::{
 };
 use alloc::vec;
 use cms::{content_info::ContentInfo, signed_data::SignedData};
+use const_oid::db::rfc5912::{ID_SHA_256, ID_SHA_512, RSA_ENCRYPTION};
 use elf::{
     ElfBytes,
     abi::{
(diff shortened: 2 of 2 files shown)


litebox_platform_lvbs/Cargo.toml
```diff
@@ -22,6 +22,10 @@ num_enum = { version = "0.7.3", default-features = false }
 once_cell = { version = "1.20.2", default-features = false, features = ["alloc", "race"] }
 hashbrown = "0.15.2"
 elf = { version = "0.7.4", default-features = false }
+cms = { version = "0.2.3", default-features = false, features = ["alloc"] }
+rsa = { version = "0.9.8", default-features = false }
```
wdcui, line 27: It might make sense to add an issue to track a potential update to verified crypto code.

litebox_platform_lvbs/src/mshv/kernel_elf.rs
```diff
@@ -352,10 +364,123 @@ pub fn parse_modinfo(original_elf_data: &[u8]) -> Result<(), KernelElfError> {
     Ok(())
 }
 
+/// This function verifies the signature of a Linux kernel module.
+/// When module signing is configured, the Linux kernel build pipeline signs each kernel module and appends
+/// the signature to it. This function extracts the signature and verifies it using the system certificate which
+/// contains the pubic portion of the build pipeline key. VTL0 does not have access to the private portion.
+///
+/// Currently, this function is slow because it uses the `sha2` crate with the `force-soft` feature.
+/// We should consider using HW-accelerated SHA-512 in the future (need to save/restore vector registers).
```
wdcui, line 371: Create an issue to track this?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -325,32 +326,60 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
...
+        serial_println!("VSM: No system certificate found");
         Err(Errno::EINVAL)
+    } else {
+        let mut cert_buf = vec![0u8; system_certs_mem.len()];
+        system_certs_mem
+            .read_bytes(system_certs_mem.start().unwrap(), &mut cert_buf)
+            .map_err(|_| Errno::EINVAL)?;
+
+        // The system certificate is loaded into VTL1 before `end_of_boot` is signaled. That is, its authenticity
+        // depends on UEFI Secure Boot. Since our VTL1 kernel does not have root certificates, it cannot
```
wdcui: I don't understand this comment. Why do we need a root certificate to verify the system certificate?

sangho2: Let me remove this misleading comment. I meant we might be able to use a root certificate to check whether the system certificate is valid. Currently, the system certificate itself is not signed. Instead, the Linux kernel image (PE) embedding the system certificate is signed by companies like Microsoft and Canonical.

wdcui: Please also update the comment for this PR. BTW, how does LVBS verify the system certificate now?

sangho2: I'll update the comment. LVBS does not verify the system certificate. It relies on UEFI Secure Boot and assumes that the kernel behaves as expected until compromised. It assumes all basic setups like VTL1 memory protection and system certificate transfer can be completed before any compromise.

wdcui: I see, so you are trying to be more secure than what LVBS does today by checking if the system certificate is from a root certificate that's acceptable, right?

sangho2: exactly. Anyway, I think we don't need to worry about this now. Maintaining a root certificate in a secure manner is difficult.

PR #183, opened 2025-06-26
Reactivate deadcode/unused lints on `litebox`

jaybosamiya-ms:
@wdcui noticed [here](https://github.com/microsoft/litebox/pull/180#discussion_r2167906575) an anomaly where it appeared like we were depending on a value when we weren't.

Indeed, I had sometimes been surprised that Rust was not complaining about things even when I said `must_use` but I hadn't had a chance to take a look to find out why.  The lack of an actual unused variable complaint from Rust actually pushed me into taking a look, and indeed turns out this was due to a fairly old suppression from many months ago, when initially setting up LiteBox's codebase:
https://github.com/microsoft/litebox/blob/db371be74220899f950cc33378ffd81af4b77c7c/litebox/src/lib.rs#L13-L15

This is clearly not needed anymore.  However, since we have lived with these lints suppressed for a while, I had to make updates across `litebox/` wherever things were suppressed.  Many of these places, I did the actual fixes (this actually reduces out memory footprint by a small amount in some places!), but in some others, I placed more precise suppressions that should be more locally triggered (e.g., for `unimplemented!()` code that we expect to implement).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
+      - run: sudo apt install -y musl-tools
       - name: "Ratchet: ensure number of 'transmute's in codebase is not increasing"
         run: |
           if [ $(find . -name \*.rs -print0 | xargs -0 grep 'transmute' | wc -l) -eq 5 ]; then \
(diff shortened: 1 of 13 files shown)


litebox/src/net/mod.rs
```diff
@@ -521,18 +524,18 @@ where
...
-                    smoltcp::storage::PacketBuffer::new(
-                        vec![smoltcp::storage::PacketMetadata::EMPTY; MAX_PACKET_COUNT],
-                        vec![0u8; SOCKET_BUFFER_SIZE],
-                    ),
-                    smoltcp::storage::PacketBuffer::new(
-                        vec![smoltcp::storage::PacketMetadata::EMPTY; MAX_PACKET_COUNT],
-                        vec![0u8; SOCKET_BUFFER_SIZE],
-                    ),
-                ))
+                // self.socket_set.add(raw::Socket::new(
```
wdcui: Why are these lines commented off?

jaybosamiya-ms: The `return Err` before it causes these lines to be dead code.  We expect to bring them back when we introduce raw-protocol support (and answer the TODO that is right above it).  We could instead place an `#[expect(dead_code)]` here too; practically identical, and won't impact things too much.

wdcui: Since we put `#[expect(dead_code)]` in other places, to be consistent, it may be better to do it here (instead of just commenting them out).

PR #185, opened 2025-06-28
No libc

CvvT:
1. Remove the dependency on nix/libc: mostly just copy/paste macro and struct definitions
2. As mentioned in #142, assume `rdfsbase`/`wrfsbase` are available so we don't need to use `libc::getauxvec`.
3. `seccompiler` and registering signal handlers are still requiring non-trivial work to get rid of libc. For now, use conditional compilation to disable the code as it is not needed for integrating with python.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -27,45 +27,47 @@ pub const FUTEX_WAIT: i32 = 0;
 pub const FUTEX_WAKE: i32 = 1;
 pub const FUTEX_REQUEUE: i32 = 3;
 
-/* constants for encoding ioctl commands (see the macro `ioc`) */
-/// The number of bits allocated for the ioctl command number field.
-pub const NRBITS: u32 = 8;
-/// The number of bits allocated for the ioctl command type field.
-pub const TYPEBITS: u32 = 8;
-/// The number of bits allocated for the ioctl command size field.
-pub const SIZEBITS: u32 = 14;
-/// The bit offset for the ioctl command number field.
-pub const NRSHIFT: u32 = 0;
-/// The bit offset for the ioctl command type field.
-pub const TYPESHIFT: u32 = NRSHIFT + NRBITS;
-/// The bit offset for the ioctl command size field.
-pub const SIZESHIFT: u32 = TYPESHIFT + TYPEBITS;
-/// The bit offset for the ioctl command direction field.
-pub const DIRSHIFT: u32 = SIZESHIFT + SIZEBITS;
-/// Represents no data transfer direction for the ioctl command.
-pub const NONE: u32 = 0;
-/// Represents the write data transfer direction for the ioctl command.
-pub const WRITE: u32 = 1;
-/// Represents the read data transfer direction for the ioctl command.
-pub const READ: u32 = 2;
+/// Encoding for ioctl commands.
+pub mod ioctl {
+    /// The number of bits allocated for the ioctl command number field.
+    pub const NRBITS: u32 = 8;
+    /// The number of bits allocated for the ioctl command type field.
+    pub const TYPEBITS: u32 = 8;
+    /// The number of bits allocated for the ioctl command size field.
+    pub const SIZEBITS: u32 = 14;
+    /// The bit offset for the ioctl command number field.
+    pub const NRSHIFT: u32 = 0;
+    /// The bit offset for the ioctl command type field.
+    pub const TYPESHIFT: u32 = NRSHIFT + NRBITS;
+    /// The bit offset for the ioctl command size field.
+    pub const SIZESHIFT: u32 = TYPESHIFT + TYPEBITS;
+    /// The bit offset for the ioctl command direction field.
+    pub const DIRSHIFT: u32 = SIZESHIFT + SIZEBITS;
+    /// Represents no data transfer direction for the ioctl command.
+    pub const NONE: u32 = 0;
+    /// Represents the write data transfer direction for the ioctl command.
+    pub const WRITE: u32 = 1;
+    /// Represents the read data transfer direction for the ioctl command.
+    pub const READ: u32 = 2;
 
-/// Encode an ioctl command.
-#[macro_export]
-macro_rules! ioc {
-    ($direction:expr, $type:expr, $number:expr, $size:expr) => {
-        (($direction as u32) << $crate::DIRSHIFT)
-            | (($type as u32) << $crate::TYPESHIFT)
-            | (($number as u32) << $crate::NRSHIFT)
-            | (($size as u32) << $crate::SIZESHIFT)
-    };
-}
+    /// Encode an ioctl command.
+    #[macro_export]
+    macro_rules! ioc {
+        ($direction:expr, $type:expr, $number:expr, $size:expr) => {
+            (($direction as u32) << $crate::ioctl::DIRSHIFT)
+                | (($type as u32) << $crate::ioctl::TYPESHIFT)
+                | (($number as u32) << $crate::ioctl::NRSHIFT)
+                | (($size as u32) << $crate::ioctl::SIZESHIFT)
+        };
+    }
 
-/// Encode an ioctl command that writes.
-#[macro_export]
-macro_rules! iow {
-    ($ty:expr, $nr:expr, $sz:expr) => {
-        $crate::ioc!($crate::WRITE, $ty, $nr, $sz)
-    };
+    /// Encode an ioctl command that writes.
+    #[macro_export]
+    macro_rules! iow {
+        ($ty:expr, $nr:expr, $sz:expr) => {
+            $crate::ioc!($crate::ioctl::WRITE, $ty, $nr, $sz)
+        };
+    }
 }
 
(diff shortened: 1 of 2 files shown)


litebox_common_linux/src/lib.rs
```diff
@@ -18,6 +18,57 @@ extern crate alloc;
...
+pub const STDIN_FILENO: i32 = 0;
+pub const STDOUT_FILENO: i32 = 1;
+pub const STDERR_FILENO: i32 = 2;
+
+// linux/futex.h
+pub const FUTEX_WAIT: i32 = 0;
+pub const FUTEX_WAKE: i32 = 1;
+pub const FUTEX_REQUEUE: i32 = 3;
+
+/// The number of bits used for the number field.
```
wdcui: Minor: "for the number field in ioctl"?

litebox_common_linux/src/lib.rs
```diff
@@ -489,6 +540,18 @@ cfg_if::cfg_if! {
     }
 }
 
+// From libc's `timespec` definition.
+//
+// linux x32 compatibility
+// See https://sourceware.org/bugzilla/show_bug.cgi?id=16437
+pub struct timespec {
+    pub tv_sec: time_t,
+    #[cfg(all(target_arch = "x86_64", target_pointer_width = "32"))]
```
wdcui, line 633: Why would we have `target_pointer_width=32` for x86_64?

litebox_platform_linux_userland/Cargo.toml
```diff
@@ -4,13 +4,15 @@ version = "0.1.0"
...
 litebox = { path = "../litebox/", version = "0.1.0" }
 litebox_common_linux = { path = "../litebox_common_linux", version = "0.1.0" }
-nix = { version = "0.29.0", default-features = false, features = ["fs", "ioctl", "mman", "signal"] }
-seccompiler = "0.5.0"
+seccompiler = { version = "0.5.0", optional = true }
 spin = "0.9.8"
 syscalls = { version = "0.6", default-features = false }
 
+[features]
+seccomp = ["seccompiler", "dep:libc"]
```
wdcui: Minor: `dep:libc` can be replaced by `libc`?

litebox_shim_linux/Cargo.toml
```diff
@@ -29,11 +29,12 @@ elf_loader = { version = "=0.10.0", default-features = false, features = ["rel"]
...
 unstable-testing = []
+seccomp = ["litebox_platform_multiplex/seccomp"]
 
 [lints]
 workspace = true
 
 [dev-dependencies]
 arrayvec = { version = "0.7.6", default-features = false }
-litebox_shim_linux = { path = ".", features = ["unstable-testing"] }
+litebox_shim_linux = { path = ".", features = ["unstable-testing", "seccomp"] }
```
wdcui: Why do we have a dev dependency on the crate itself?

litebox_platform_linux_userland/Cargo.toml
jaybosamiya-ms: I think I might suggest in the other direction, replace `seccompiler` with `dep:seccompiler` to make it more obvious that it is a "enable this dependency".

Either way, good to be consistent with both of them.

litebox_common_linux/src/lib.rs
jaybosamiya-ms: Presumably this struct is copied from the libc crate; the comment above it already points at the compatibility thing (x32 == 32-bit on x86-64).  x32 is weird in _many_ ways, I don't know if we want to explicitly support it.  Otoh, copying this over from the libc crate does not hurt, so why not.

Sidenote: wherever stuff is copied from the libc crate, we should say "From libc crate's ..." rather than "From libc's" because otherwise it is confusing.

litebox_shim_linux/Cargo.toml
jaybosamiya-ms: I believe this is to turn it on for tests.

litebox_platform_multiplex/Cargo.toml
```diff
@@ -5,12 +5,13 @@ edition = "2024"
...
 [dependencies]
 litebox = { path = "../litebox/", version = "0.1.0" }
-litebox_platform_linux_userland = { path = "../litebox_platform_linux_userland/", version = "0.1.0", optional = true }
+litebox_platform_linux_userland = { path = "../litebox_platform_linux_userland/", version = "0.1.0", default-features = false, optional = true }
 once_cell = { version = "1.20.2", default-features = false, features = ["alloc", "race"] }
 
 [features]
 default = ["platform_linux_userland"]
 platform_linux_userland = ["dep:litebox_platform_linux_userland"]
+seccomp = ["litebox_platform_linux_userland/seccomp"]
```
jaybosamiya-ms: Within `platform_multiplex/src/lib.rs`, we may wish to have something like:
```
#[cfg(all(feature = "seccomp", not(feature = "platform_linux_userland")))]
compile_error!("The `seccomp` feature requires `platform_linux_userland`")
```

litebox_platform_linux_userland/Cargo.toml
jaybosamiya-ms: Sidenote: we may wish to name this feature `support_systrap_backend` or `systrap_backend_available` rather than `seccomp`.  Currently, it implies that enabling this conditional compilation _forces_ seccomp, which is a slightly different thing than we intend.  We also wish to (in the future) for non-systrap backend too support seccomp (but in that case, have the stronger "disable everything else").  Thus, making this feature explicitly about the systrap backend makes sense.

litebox_common_linux/src/lib.rs
CvvT: Right, fixed!

litebox_common_linux/src/lib.rs
CvvT: It was directly copied from libc crate. Fixed the comment as suggested.

litebox_platform_linux_userland/Cargo.toml
CvvT: There is a subtle difference between `dep:feature` and `feature`. See https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

Changed the name of  the feature and use `dep` as suggested.

litebox_platform_multiplex/Cargo.toml
CvvT: I made `systrap_backend` depends on `platform_linux_userland` in the Cargo.toml:

`systrap_backend = ["platform_linux_userland", "litebox_platform_linux_userland/systrap_backend"]`

So, the check for detecting enabling multiple platforms should catch this.

jaybosamiya-ms: Ah, that makes sense, thanks!

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -50,17 +96,29 @@ impl LinuxUserland {
...
             .map(|tun_device_name| {
-                let tun_fd = nix::fcntl::open(
-                    "/dev/net/tun",
-                    nix::fcntl::OFlag::O_RDWR
-                        | nix::fcntl::OFlag::O_CLOEXEC
-                        | nix::fcntl::OFlag::O_NONBLOCK,
-                    nix::sys::stat::Mode::empty(),
-                )
-                .unwrap();
+                let tun_path = "/dev/net/tun\0";
```
jaybosamiya-ms: Nit: it is probably slightly more preferable to use `b"/dev/net/tun\0"` here.

litebox_common_linux/src/lib.rs
```diff
@@ -18,6 +18,56 @@ extern crate alloc;
...
+/// The bit offset for the ioctl command size field.
+pub const SIZESHIFT: u32 = TYPESHIFT + TYPEBITS;
+/// The bit offset for the ioctl command direction field.
+pub const DIRSHIFT: u32 = SIZESHIFT + SIZEBITS;
+/// Represents no data transfer direction for the ioctl command.
+pub const NONE: u32 = 0;
+/// Represents the write data transfer direction for the ioctl command.
+pub const WRITE: u32 = 1;
+/// Represents the read data transfer direction for the ioctl command.
+pub const READ: u32 = 2;
```
jaybosamiya-ms: I am mildly concerned by the number of constants being added.  The particularly worrisome ones are things like `NONE`/`READ`/`WRITE`.

It might be better to either use enums where it makes sense, but if we want to keep them as constants, we can just wrap them in simple `pub mod whatever { ... }` wrappers (so for example, all the ioctl constants can stay within a `pub mod ioctl { pub const .... ; ....... }`).

Note: we don't need to each of these out to a new file, this would just be a couple of lines around each group of constants above to place it into a mod.

litebox_platform_linux_userland/src/lib.rs
CvvT: Got it, fixed!

litebox_common_linux/src/lib.rs
CvvT: Good point! Fixed.

PR #187, opened 2025-07-01
Chuqiz/newlib/newlib shim static

Icegrave0391:
This pull request is the skeleton for supporting `newlib` in a static library scheme. It contains a new crate named `litebox_runner_newlib` and the final output will be `liblitebox_newlibrunner.a` library.

Also, `litebox_platform_multiplex` is slightly modified (removed default platform backend).

To build `litebox_runner_newlib`:

```bash
cd /path/to/litebox_runner_newlib
cargo build --features "platform_mock_nostd" --target x86_64-unknown-none
```

### DIFF
--- a/litebox_runner_newlib/src/lib.rs
+++ b/litebox_runner_newlib/src/lib.rs
@@ -83,7 +83,7 @@ pub unsafe extern "C" fn _start(argc: ffi::c_int, argv: *const *const ffi::c_cha
     let envp = argv.add((argc + 1) as usize) as *const *const ffi::c_char;
     
     // Initialize the LiteBox environment
-    __init_litebox_runtime();
+    unsafe {__init_litebox_runtime(); }
     
     // // Call newlib's initialization functions
     // __libc_init_array();
@@ -268,7 +268,7 @@ pub unsafe extern "C" fn getpid() -> i32 {
 pub unsafe extern "C" fn kill(pid: i32, sig: i32) -> i32 {
     if pid == 1 && sig == 9 {
         // SIGKILL to our own process - exit
-        _exit(1);
+        unsafe { _exit(1); }
     }
     // For other cases, just return success
     0


Cargo.toml
```diff
@@ -14,7 +15,8 @@ members = [
 # exclude binaries which need different toolchain and build target.
 exclude = [
     "litebox_platform_lvbs",
-    "litebox_runner_lvbs"
+    "litebox_runner_lvbs",
+    "litebox_runner_newlib",
```
jaybosamiya-ms, line 19: Any reason this is added to the exclusions?  Even a static library should not require you to place this in the exclusions list, no?

litebox/Cargo.toml
```diff
@@ -17,6 +17,8 @@ rangemap = { version = "1.5.1", features = ["const_fn"] }
 buddy_system_allocator = { version = "0.11.0", default-features = false, features = ["use_spin"] }
 # Depend on (currently unreleased) slabmalloc `main`, which contains some fixes on top of `0.11.0`
 slabmalloc = { git = "https://github.com/gz/rust-slabmalloc.git", rev = "19480b2e82704210abafe575fb9699184c1be110" }
+# todo(chuqi): support my mock-nostd platform's debuglog
+syscalls = { version = "0.6", default-features = false }
```
jaybosamiya-ms, line 21: I don't see why this is needed

litebox_platform_mock_nostd/README.md
```diff
@@ -0,0 +1,31 @@
+# litebox_platform_mock_nostd
+
+A `#![no_std]` mock platform implementation for LiteBox, suitable for testing and development on embedded or bare-metal targets.
```
jaybosamiya-ms, line 3: I don't think we are planning to merge the `mock_nostd` in (since it was meant to be a placeholder until we were able to make `linux_userland` fully no-std + no-libc).  Just making a note here.

litebox_platform_mock_nostd/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+        //     current_time: AtomicU64::new(0),
+        // };
+        // &INSTANCE
+        Box::leak(Box::new(Self {
+            current_time: AtomicU64::new(0),
+            reserved_pages: Self::read_proc_self_maps(),
+        }))
+    }
+
+    fn read_proc_self_maps() -> alloc::vec::Vec<core::ops::Range<usize>> {
```
jaybosamiya-ms, line 45: It is important to document where you are copying things in from if you are copying things in, because otherwise, things might get out of sync, or something does not get fixed.  Again, we aren't planning on merging this mock, so it is fine, but good to make a habit of it.

litebox_platform_mock_nostd/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+    current_time: AtomicU64,
+    reserved_pages: Vec<core::ops::Range<usize>>,
+}
+
+impl MockNoStdPlatform {
+    pub fn new(_name: Option<&'static str>) -> &'static Self {
+        // static INSTANCE: MockNoStdPlatform = MockNoStdPlatform {
+        //     current_time: AtomicU64::new(0),
+        // };
+        // &INSTANCE
```
jaybosamiya-ms, line 38: Why is this commented-out thing here?

litebox_platform_mock_nostd/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+    const EXIT_SUCCESS: Self::ExitCode = 0;
+    const EXIT_FAILURE: Self::ExitCode = 1;
+    
+    fn exit(&self, _code: Self::ExitCode) -> ! {
+        // In a no_std environment, we can't actually exit cleanly
+        // This would typically be implemented by the platform-specific exit mechanism
+        // For now, just loop indefinitely
+        loop {
+           core::hint::spin_loop();
+        }
```
jaybosamiya-ms, line 111: I prefer an explicit panic rather than a spin-loop.

litebox_platform_mock_nostd/README.md
```diff
@@ -0,0 +1,31 @@
...
+let platform = MockNoStdPlatform::new(None);
+// Use with LiteBox
+```
+
+## Limitations
+
+- No actual memory management (mock implementations)
+- No network functionality 
+- Limited I/O capabilities
+- No real synchronization primitives (uses spin loops)
```
jaybosamiya-ms, line 29: The "uses spin loops" is incorrect; you are not using spin loops to implement the `RawMutexProvider`, you are simply returning directly. 

Out of curiosity, was this README generated by an LLM attempting to summarize things?

litebox_platform_mock_nostd/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+    }
+
+    fn block_or_timeout(
+        &self,
+        _val: u32,
+        _time: core::time::Duration,
+    ) -> Result<UnblockedOrTimedOut, ImmediatelyWokenUp> {
+        // No-op implementation for mock - always immediately wake up
+        Err(ImmediatelyWokenUp)
+    }
```
jaybosamiya-ms, line 141: Keeping these as `unimplemented!` is probably a better move unless you have a good reason to implement them.  Most of the locks inside LiteBox don't _actually_ stop the thread, but _internally_ have their own spin-loops that "bottom out" to these.  See how our actual mock uses just `unimplemented!()`s here and still everything works fine.

litebox_platform_mock_nostd/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+}
+
+impl DebugLogProvider for MockNoStdPlatform {
+    fn debug_log_print(&self, msg: &str) {
+        let msg_ptr = msg.as_ptr() as usize;
+        let msg_len = msg.len();
+        
+        unsafe {
+            let _ = syscall!(Sysno::write, 1, msg_ptr, msg_len);
+        }
```
jaybosamiya-ms, line 394: Use stderr for logging, rather than stdout
```suggestion
        unsafe {
            let _ = syscall!(Sysno::write, 2, msg_ptr, msg_len);
        }
```

litebox_platform_mock_nostd/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+use alloc::{boxed::Box, vec::Vec};
+use syscalls::{syscall, Sysno};
+
+use litebox::fs::OFlags;
+use litebox::platform::*;
+use litebox::platform::page_mgmt::MemoryRegionPermissions;
+use litebox::utils::ReinterpretUnsignedExt as _;
+use litebox_common_linux::{MRemapFlags, MapFlags, ProtFlags, PunchthroughSyscall};
+
+mod syscall_intercept;
```
jaybosamiya-ms, line 15: The entire `syscall_intercept` module seems unrelated and irrelevant to a mock platform

litebox_runner_newlib/.cargo/config.toml
```diff
@@ -0,0 +1,6 @@
+[build.cfg]
+target = { value = "x86_64-unknown-none", condition = "cfg(feature = \"platform_mock_nostd\")" }
+
+# [unstable]
+# build-std = ["core", "compiler_builtins", "alloc"]
+# build-std-features = ["compiler-builtins-mem"]
```
jaybosamiya-ms, line 6: Why is there commented-out code here?

litebox_platform_multiplex/Cargo.toml
```diff
@@ -6,11 +6,15 @@ edition = "2024"
...
 litebox = { path = "../litebox/", version = "0.1.0" }
 litebox_platform_linux_userland = { path = "../litebox_platform_linux_userland/", version = "0.1.0", optional = true }
+litebox_platform_mock_nostd = { path = "../litebox_platform_mock_nostd/", version = "0.1.0", optional = true }
 once_cell = { version = "1.20.2", default-features = false, features = ["alloc", "race"] }
 
 [features]
-default = ["platform_linux_userland"]
+# default = ["platform_linux_userland"]
+# chuqi: necessary to support multiplexing
+default = []
```
jaybosamiya-ms, line 15: You shouldn't need to change the defaults at all.  `default-features = false` on any of the caller crates of it should automatically take care of disabling `platform_linux_userland`.  Please revert the changing of the default.

litebox_runner_newlib/README.md
```diff
@@ -0,0 +1,18 @@
...
+It also provides backend implementation for POSIX stubs.
+
+## Usage
+
+This crate is compiled as a static library (`liblitebox_newlibrunner.a`) and is intended to be linked with applications that use Newlib's libc.
+
+## Target Architecture
+
+The target architecture (ideally) should be named as x86_64-litebox (or x86_64-unknown-none).
+For now, I would simply use native gnu-linux target.
```
jaybosamiya-ms, line 18: I think this is mildly confusing, best dropped from this README.

litebox_runner_newlib/build.rs
```diff
@@ -0,0 +1,44 @@
+pub fn main() {
+    // // Tell Cargo to re-run this build script if the target specification file changes
+    // println!("cargo:rerun-if-changed=x86_64-litebox.json");
```
jaybosamiya-ms, line 3: Please check through and remove all commented-out code in this file too

litebox_runner_newlib/build.rs
```diff
@@ -0,0 +1,44 @@
...
+        // Exclude dynamic linking related objects
+        println!("cargo:rustc-link-arg=-Wl,--exclude-libs=libdl.a");
+        println!("cargo:rustc-link-arg=-Wl,--exclude-libs=ld-linux-x86-64.so.2");
+
+        // Remove dynamic sections entirely
+        println!("cargo:rustc-link-arg=-Wl,--discard-all");
+        println!("cargo:rustc-link-arg=-Wl,-s");  // Strip symbols
+        
+        // Set relocation model to static
+        println!("cargo:rustc-codegen-opt=relocation-model=static");
```
jaybosamiya-ms, line 41: There is non-trivial redundancy in these, and I am not sure most of these options are even necessary.  Are all of these actually crucial?  I expect Rust to simply pick things up since you've correctly since you've defined it as a `crate-type = ["staticlib"]`.

Why are all these details necessary?

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+// #![feature(linkage)]
+
+extern crate alloc;
+use core::arch::asm;
+use core::ffi;
+use syscalls::{syscall, Sysno};
+
+#[expect(clippy::inline_always)]
+#[inline(always)]
+pub fn hlt_loop() -> ! {
```
jaybosamiya-ms, line 22: Why is this as `pub fn` rather than simply an `fn`.  Does the linking process with newlib need this functionality?

Also, why is this `inline(always)`?  There is a clippy lint for it for a reason, have you profiled this code and found it to be a bottleneck that requires inlining?  If so, please add a reasoning to the `expect(..., reason = "")`.  In this particular case, I see no reason that this is actually going to be a bottleneck.

Also, if the meaning is "inline everywhere it is used in this crate" then that is **not** what `inline(always)` means.  Crate-internal inlining is a decision the compiler will do _anyways_, writing `inline` attributes a _lot_ more when you are doing cross-crate optimization, which cannot happen in this particular crate anyways, since its public API is exposed as a _binary_ interface.

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+    loop {
+        unsafe {
+            asm!("hlt");
+        }
+    }
+}
+
+/// Print formatted arguments to console output by writing to stdout
+#[doc(hidden)]
+pub fn print(args: core::fmt::Arguments) {
```
jaybosamiya-ms, line 32: Why is this `pub`? Also, why is it `doc(hidden)`?

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+macro_rules! newlib_println {
+    () => ($crate::newlib_print!("\n"));
+    ($($arg:tt)*) => ($crate::newlib_print!("{}\n", format_args!($($arg)*)));
+}
+
+
+#[cfg(feature = "platform_mock_nostd")]
+#[panic_handler]
+fn panic(info: &core::panic::PanicInfo) -> ! {
+    newlib_println!("newlib-runtime panic: {}", info);
```
jaybosamiya-ms, line 55: Any panic, if it happens, is not _inside_ newlib, but inside the whole litebox-newlib combination, so you cannot say it is a newlib-specific panic by defining such a global panic handler.

```suggestion
    newlib_println!("litebox-newlib-runtime panic: {}", info);
```

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+pub unsafe extern "C" fn _start(argc: ffi::c_int, argv: *const *const ffi::c_char) -> ! {
+    // Calculate envp pointer (argv + (argc + 1))
+    // envp starts after argv array + null terminator
+    let envp = argv.add((argc + 1) as usize) as *const *const ffi::c_char;
+    
+    // Initialize the LiteBox environment
+    unsafe {__init_litebox_runtime(); }
+    
+    // // Call newlib's initialization functions
+    // __libc_init_array();
```
jaybosamiya-ms, line 89: Why is this commented out?

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+    );
+    
+    // Set up filesystem and platform
+    litebox_shim_linux::set_fs(initial_file_system);
+    litebox_platform_multiplex::set_platform(platform);
+    
+    // Register syscall handler
+    // chuqi: I don't think this is necessary for newlib, as we don't hook real syscall
+    //        instructions; instead, we provide our own implementations. 
+    // platform.register_syscall_handler(litebox_shim_linux::handle_syscall_request);
```
jaybosamiya-ms, line 151: Yes, we don't need it for newlib runner, can be safely removed.

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+        unsafe extern "C" {
+            static _end: u8;
+        }
+        
+        // Set initial brk to the end of data sections
+        let heap_start_raw = &_end as *const u8 as *mut u8;
+        
+        // Align heap start to page boundary (4KB) for better memory management
+        const PAGE_SIZE: usize = 4096;
+        let heap_start_aligned = (heap_start_raw as usize + PAGE_SIZE - 1) & !(PAGE_SIZE - 1);
```
jaybosamiya-ms, line 167: Use [`usize::next_multiple_of`](https://doc.rust-lang.org/std/primitive.usize.html#method.next_multiple_of) instead

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+        let heap_start = heap_start_aligned as *mut u8;
+
+        current_brk = match _do_syscall_shim(12, &[heap_start as usize]) {
```
jaybosamiya-ms, line 170: Please don't use magic numbers, especially without any comments.

jaybosamiya-ms: Here, you have a very specific syscall in mind.  In that case, it might even be a good idea to do it as a `litebox_shim_linux::handle_syscall_request` which gives you a nicer interface

jaybosamiya-ms, line 170: You are casting a `usize` to a `*mut u8` to immediately cast it back to `usize`.  Any particular reason for this indirection?

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+            neg_value if neg_value < 0 => {
+                panic!("Initial brk syscall failed with error {}", neg_value);
+            }
+            brk => {
+                newlib_println!("[_start] Initial brk set to {:p}", brk as *mut u8);
```
jaybosamiya-ms, line 175: Is this meant to be debug logging? If so, do not do it this way, use the actual Platform `debug_log_print`

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+#[macro_export]
+macro_rules! newlib_print {
+    ($($arg:tt)*) => ($crate::print(format_args!($($arg)*)));
+}
+
+#[macro_export]
+macro_rules! newlib_println {
+    () => ($crate::newlib_print!("\n"));
+    ($($arg:tt)*) => ($crate::newlib_print!("{}\n", format_args!($($arg)*)));
+}
```
jaybosamiya-ms, line 49: I think these seem to be set up for debug logging, in that case, you should not write your own raw syscalls here for `print` defined here but instead should do it via Platform's `debug_log_print`.

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+        
+        // Store the initial brk
+        CURRENT_BRK.store(current_brk, core::sync::atomic::Ordering::Relaxed);
+    }
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn __prog_exit() {
+    // do nothing for now
+}
```
jaybosamiya-ms, line 188: "for now": what is intended to be done in the future?

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+
+
+/// Handles the common logic of creating a SyscallRequest and calling the handler.
+fn _do_syscall_shim(sys_id: usize, args: &[usize]) -> isize {
+    // todo(chuqi): lazy but a bad way of using an additional memory-copy
+    //        we should use a fixed-size array instead to avoid this costly copy.
+    let mut padded_args = [0usize; 6];
+    let copy_len = args.len().min(6);
+    padded_args[..copy_len].copy_from_slice(&args[..copy_len]);
```
jaybosamiya-ms, line 197: The compiler can optimize some things, but even if it does not I don't expect this to be a bottleneck.  It is a copy of `6 * 8 = 48` bytes to/from warm locations, so _incredibly_ unlikely to be a bottleneck.

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+
+/// Read from a file descriptor
+///
+/// This function implements the POSIX read() syscall for newlib.
+/// It translates the call to litebox_shim_linux's handle_syscall_request function.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn read(fd: i32, buf: *mut u8, count: usize) -> isize {
+    if buf.is_null() {
+        return Errno::EFAULT.as_neg() as isize;
+    }
```
jaybosamiya-ms, line 229: This (and similar) seem to be a workaround for something that should be fixed at `SyscallRequest::try_from_raw`.  These specific shims should essentially just be a direct forwarding.

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+/// 
+/// This is called by newlib's exit() function and should never return.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn _exit(status: ffi::c_int) -> ! {
+    // Perform any cleanup
+    __prog_exit();
+    
+    // In a real implementation, this would terminate the process
+    // For now, we'll halt the CPU
+    hlt_loop()
```
jaybosamiya-ms, line 257: Platform has an `ExitProvider`, please use that

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+    1  // Always return PID 1 for our simple environment
+}
+
+/// Kill a process (stub implementation)
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn kill(pid: i32, sig: i32) -> i32 {
+    if pid == 1 && sig == 9 {
+        // SIGKILL to our own process - exit
+        unsafe { _exit(1); }
+    }
```
jaybosamiya-ms, line 272: This seems incorrect, because `_exit` (at least how you've implemented) says "perform any cleanup" while SIGKILL would not even allow doing cleanup.  This should just die via `ExitProvider`.

litebox_runner_newlib/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn isatty(fd: i32) -> i32 {
+    // Check if file descriptor refers to a terminal
+    // For our LiteBox environment, we'll consider stdin/stdout/stderr as TTYs
+    // and everything else as not a TTY
+    match fd {
+        0 | 1 | 2 => 1,  // stdin, stdout, stderr are TTYs
+        _ => 0,          // everything else is not a TTY
+    }
```
jaybosamiya-ms, line 403: We have providers for these, do not invent new results when there is a provider giving you a ground truth for it.

litebox_runner_newlib/x86_64-litebox.json
```diff
@@ -0,0 +1,22 @@
+{
```
jaybosamiya-ms, line 1: This file doesn't seem to actually be used, please remove it, or use it as intended.

PR #188, opened 2025-07-01
Support kexec validation and protection

sangho2:
This PR adds support for kexec validation and protection (`mshv_vsm_kexec_validate`). To validate the authenticity of a kernel image (or blob) loaded through kexec, it verifies the [Authenticode signature](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format) embedded in the kernel blob's PE header (kernel blob has a PE header to be loaded by a UEFI bootloader). It write-protects the kexec kernel image and associated pages (including kexec trampoline and segments) only if the signature verification is successful. It does support both normal kexec and crash kexec (also known as kdump) as well as invalidation/removal of previously loaded kexec kernel data when kexec is called again.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/mem_integrity.rs
+++ b/litebox_platform_lvbs/src/mshv/mem_integrity.rs
@@ -448,13 +448,18 @@ fn extract_module_data_and_signature(
         })
         .ok_or(VerificationError::SignatureNotFound)?;
 
-    #[expect(clippy::cast_ptr_alignment)]
-    let module_signature = unsafe {
-        &*(signed_module
-            .as_ptr()
-            .add(module_signature_offset)
-            .cast::<ModuleSignature>())
-    };
+    let mut module_signature = core::mem::MaybeUninit::<ModuleSignature>::uninit();
+    unsafe {
+        core::ptr::copy_nonoverlapping(
+            signed_module
+                .as_ptr()
+                .add(module_signature_offset)
+                .cast::<u8>(),
+            module_signature.as_mut_ptr().cast::<u8>(),
+            core::mem::size_of::<ModuleSignature>(),
+        );
+    }
+    let module_signature = unsafe { module_signature.assume_init() };
     if !module_signature.is_valid() {
         return Err(VerificationError::InvalidSignature);
     }
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -665,12 +665,17 @@ pub fn mshv_vsm_kexec_validate(pa: u64, nranges: u64, crash: u64) -> Result<i64,
 
     // If this function is called for crash kexec, we protect its kimage segments as well.
     if is_crash {
-        let mut kimage_buf = vec![0u8; kexec_image.len()];
+        let mut kimage = core::mem::MaybeUninit::<Kimage>::uninit();
+        let kimage_slice: &mut [u8] = unsafe {
+            core::slice::from_raw_parts_mut(
+                kimage.as_mut_ptr().cast::<u8>(),
+                core::mem::size_of::<Kimage>(),
+            )
+        };
         kexec_image
-            .read_bytes(kexec_image.start().unwrap(), &mut kimage_buf)
+            .read_bytes(kexec_image.start().unwrap(), kimage_slice)
             .map_err(|_| Errno::EINVAL)?;
-        #[allow(clippy::cast_ptr_alignment)]
-        let kimage = unsafe { &*(kimage_buf.as_ptr().cast::<Kimage>()) };
+        let kimage = unsafe { kimage.assume_init() };
         if kimage.nr_segments > u64::try_from(KEXEC_SEGMENT_MAX).unwrap() {
             serial_println!("VSM: Invalid kexec image segments");
             return Err(Errno::EINVAL);


litebox_platform_lvbs/Cargo.toml
```diff
@@ -25,9 +25,13 @@ hashbrown = "0.15.2"
 elf = { version = "0.7.4", default-features = false }
 cms = { version = "0.2.3", default-features = false, features = ["alloc"] }
 rsa = { version = "0.9.8", default-features = false }
+sha1 = { version = "0.10.6", default-features = false, features = ["force-soft"] }
```
wdcui, line 28: I don't see the `sha1` crate is being used?

litebox_platform_lvbs/src/mshv/mem_integrity.rs
```diff
@@ -495,6 +500,104 @@ fn decode_signature(
     ))
 }
 
+/// This function verifies the signature of a Linux kernel blob PE for kexec. Unlike kernel module signing
```
wdcui: Why does Linux kernel use PE?

litebox_platform_lvbs/src/mshv/mem_integrity.rs
```diff
@@ -495,6 +500,104 @@ fn decode_signature(
...
+    let mut authenticode_signatures = Vec::new();
+
+    let Ok(Some(attr_cert_iter)) = AttributeCertificateIterator::new(&pe) else {
+        return Err(VerificationError::ParseFailed);
+    };
+    for attr_cert in attr_cert_iter {
+        if let Ok(acert) = attr_cert
+            && let Ok(authenticode_signature) = acert.get_authenticode_signature()
+        {
+            authenticode_signatures.push(authenticode_signature);
```
wdcui: Should we break out of the loop since we only need the first one (for now)?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -373,13 +383,20 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
...
                 .set_system_certificate(cert);
-            Ok(0)
         } else {
             serial_println!("VSM: Failed to parse system certificate");
-            Err(Errno::EINVAL)
+            return Err(Errno::EINVAL);
         }
     }
 
+    for kexec_trampoline_range in &kexec_trampoline_metadata {
```
wdcui, line 392: What's the trampoline for?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -599,14 +614,153 @@ pub fn mshv_vsm_copy_secondary_key(_pa: u64, _nranges: u64) -> Result<i64, Errno
...
+        }
+    }
+
+    // write protect the kexec memory ranges first to avoid the race condition during verification
+    for kexec_mem_range in &kexec_memory_metadata {
+        protect_physical_memory_range(kexec_mem_range.phys_frame_range, MemAttr::MEM_ATTR_READ)?;
+    }
+
+    // verify the signature of kexec blob
+    let kexec_kernel_blob_size = kexec_kernel_blob.len();
```
wdcui, line 701: What's the relationship between `kernel_image` and `kernel_blob`?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -599,14 +614,153 @@ pub fn mshv_vsm_copy_secondary_key(_pa: u64, _nranges: u64) -> Result<i64, Errno
...
+            protect_physical_memory_range(
+                kexec_mem_range.phys_frame_range,
+                MemAttr::MEM_ATTR_READ | MemAttr::MEM_ATTR_WRITE,
+            )?;
+        }
+        return Err(result.into());
+    }
+
+    // register the protected kexec memory ranges to support possible invalidation in the future
+    kexec_metadata_ref.register_memory(kexec_memory_metadata);
```
wdcui, line 729: Do we need to make the kexec image executable?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -599,14 +614,153 @@ pub fn mshv_vsm_copy_secondary_key(_pa: u64, _nranges: u64) -> Result<i64, Errno
...
+
     Ok(0)
 }
 
+/// This function copies data for kexec validation from VTL0 to VTL1. The physical address ranges of
```
wdcui, line 734: If we have already copied memory from VTL0 to VTL1, why do we need to call `read_bytes` again during validation?

litebox_platform_lvbs/Cargo.toml
sangho2: The `authenticode` crate depends on the `sha1` crate (even though our code doesn't use `sha1` at all). WIthout this explicit `force-soft` feature enabling, there is a build error.

litebox_platform_lvbs/src/mshv/mem_integrity.rs
sangho2: To work with UEFI firmware/bootloader, known as [EFI Boot Stub](https://docs.kernel.org/admin-guide/efi-stub.html). This EFI boot stub can be signed for UEFI Secure Boot.

litebox_platform_lvbs/src/mshv/mem_integrity.rs
sangho2: That would be better. I'll remove `Vec` and do `break`.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: kexec trampoline is like a bootloader for kexec. It supports transition and prepares an environment for kexec.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: `kernel_blob` contains the original kernel binary file (as both ELF and PE). `kernel_image` is its in-memory representation once it is loaded [struct kimage](https://elixir.bootlin.com/linux/v6.6.85/source/include/linux/kexec.h#L296)

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: `MemAttr::MEM_ATTR_READ` enables both read and executable permissions.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: `read_bytes` of `MemoryContainer` is for reading bytes in terms of VTL0 virtual addresses. Data copies below focus on copying guest physical pages from VTL0 to VTL1.

litebox_platform_lvbs/src/mshv/vsm.rs
wdcui: Why do we combine read and executable into one? Is this the existing LVBS code's choice?

litebox_platform_lvbs/src/mshv/vsm.rs
wdcui: As we discussed offline, please create an issue to track this memory optimization.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Sorry for the confusion. `MemAttr::MEM_ATTR_READ` only enables read and user executable permissions. This `mshv_vsm_kexec_validate` function is mainly for write protecting loaded kexec bytes. Make it executable should be done later like when we hook a reset signal (needs to be another PR).

PR #189, opened 2025-07-02
Support kernel text patching

sangho2:
This PR adds initial support for patching kernel/module text. It introduces the `mshv_vsm_patch_text` function that the VTL0 kernel calls when it needs to patch kernel or module text regions (which are write protected with SLAT). Currently, this function does not check whether a patching request potentially violates the security (which will be covered in another PR). It only does sanity checks.

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -186,9 +186,9 @@ impl<Host: HostInterface> LinuxKernel<Host> {
 
             unsafe {
                 core::ptr::copy_nonoverlapping(
-                    page_addr.wrapping_add(page_offset) as *const T,
-                    (*raw).as_mut_ptr(),
-                    1,
+                    page_addr.wrapping_add(page_offset).cast::<u8>(),
+                    (*raw).as_mut_ptr().cast::<u8>(),
+                    core::mem::size_of::<T>(),
                 );
             }
 
@@ -225,9 +225,9 @@ impl<Host: HostInterface> LinuxKernel<Host> {
                 usize::try_from(phys_addr - phys_addr.align_down(Size4KiB::SIZE)).unwrap();
             unsafe {
                 core::ptr::copy_nonoverlapping(
-                    core::ptr::from_ref::<T>(value),
-                    page_addr.wrapping_add(page_offset).cast::<T>(),
-                    1,
+                    core::ptr::from_ref::<T>(value).cast::<u8>(),
+                    page_addr.wrapping_add(page_offset).cast::<u8>(),
+                    core::mem::size_of::<T>(),
                 );
             }
             assert!(
@@ -264,9 +264,9 @@ impl<Host: HostInterface> LinuxKernel<Host> {
                 usize::try_from(phys_addr - phys_addr.align_down(Size4KiB::SIZE)).unwrap();
             unsafe {
                 core::ptr::copy_nonoverlapping(
-                    value.as_ptr(),
-                    page_addr.wrapping_add(page_offset).cast::<T>(),
-                    value.len(),
+                    value.as_ptr().cast::<u8>(),
+                    page_addr.wrapping_add(page_offset).cast::<u8>(),
+                    core::mem::size_of_val(value),
                 );
             }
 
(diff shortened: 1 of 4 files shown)


litebox_platform_lvbs/src/host/linux.rs
```diff
@@ -125,3 +125,11 @@ impl ModuleSignature {
             && self.key_id_len == 0
     }
 }
+
+/// `list_head` from [Linux](https://elixir.bootlin.com/linux/v6.6.85/source/include/linux/types.h#L190)
+#[derive(Clone, Copy, Debug)]
+#[repr(C)]
+pub struct ListHead {
```
wdcui, line 163: Is this for a double linked list? If so, calling it ListHead is a little confusing since next/prev points to a list entry.

litebox_platform_lvbs/src/lib.rs
```diff
@@ -225,6 +225,51 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
+    ///
+    /// Panics if phys_addr is invalid
+    pub unsafe fn copy_slice_to_vtl0_phys<T: Copy>(
+        &self,
+        phys_addr: x86_64::PhysAddr,
+        value: &[T],
+    ) -> bool {
+        if let Ok(addr) = self.map_vtl0_phys_range(
+            phys_addr,
+            phys_addr + u64::try_from(core::mem::size_of_val(value) + PAGE_SIZE).unwrap(),
```
wdcui: consider to use `next_multiple_of`?

wdcui: We assume `map_vtl0_phys_range` will align down. Should we have it to align up as well?

wdcui: We can have `map_vtl0_phys_range` return the size of the mapped range as well. This way we don't need to calculate it again when calling unmap.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -607,32 +609,135 @@ pub fn mshv_vsm_kexec_validate(_pa: u64, _nranges: u64, _crash: u64) -> Result<i
...
+        }) else {
+            return Err(Errno::EINVAL);
+        };
+        let Some(page1) = (unsafe {
+            crate::platform_low().copy_from_vtl0_phys::<[u8; PAGE_SIZE]>(PhysAddr::new(patch_pa_1))
+        }) else {
+            return Err(Errno::EINVAL);
+        };
+        let mut concatenated = Vec::with_capacity(PAGE_SIZE * 2);
+        concatenated.extend_from_slice(page0.as_slice());
```
wdcui: Does this involve another page copy? Can we allocate two pages at once then copy memory over from vtl0?

litebox_platform_lvbs/src/host/linux.rs
sangho2: True. This is for Linux kernel's internal double linked lists. We add it here mainly for maintaining the layout of other data structures (e.g., `HekiPatchType`). We can surely change its name, but someone who is familiar with Linux kernel's internal data structures might be confused.

litebox_platform_lvbs/src/lib.rs
sangho2: Let me improve this.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: `patch_pa_0` and `patch_pa_1` are not contiguous, so we require two VTL0-VTL1 memory copies or need to manipulate the VTL1 page table. `copy_from_vtl0_phys` is mainly written for strict types (with internal heap allocation). Let me write another simple copy function without internal heap allocation.

litebox_platform_lvbs/src/lib.rs
```diff
@@ -78,7 +79,7 @@ impl<Host: HostInterface> LinuxKernel<Host> {
     ) -> &'static Self {
         let pt = unsafe { mm::PageTable::new(init_page_table_addr) };
         let physframe_start = PhysFrame::containing_address(phys_start);
-        let physframe_end = PhysFrame::containing_address(phys_end);
+        let physframe_end = PhysFrame::containing_address(phys_end.align_up(Size4KiB::SIZE) + 1);
```
wdcui: Why do you need to have `+1`?

sangho2: Internally, it does use a range of page frames, and I want to ensure its end is the next page frame. Anyhow, it seems that this is unnecessary.

PR #191, opened 2025-07-02
Reserve space for trampoline code and load trampoline for ld

CvvT:
This PR addresses two issues:

1. The trampoline code should be loaded right after the executable file. To ensure there is sufficient space for the trampoline code, we need to allocate more memory when the first `mmap` is called. The elf loader we use would call `mmap` to load all segments with the `PT_LOAD` type at once and then invoke `mmap` again to change the memory protection.
2. `ld` lib is also rewritten and hence we need to load its trampoline code.

### DIFF
--- a/litebox_shim_linux/src/loader/elf.rs
+++ b/litebox_shim_linux/src/loader/elf.rs
@@ -300,9 +300,15 @@ fn get_trampoline_hdr(object: &mut ElfFile) -> Option<TrampolineHdr> {
 }
 
 fn load_trampoline(trampoline: TrampolineHdr, relo_off: usize, fd: i32) {
+    // Our rewriter ensures that both `trampoline.vaddr` and `trampoline.file_offset` are page-aligned.
+    // Otherwise, `ElfLoaderMmap::mmap` will fail and panic.
     assert!(
         trampoline.vaddr % PAGE_SIZE == 0,
         "trampoline address must be page-aligned"
+    );
+    assert!(
+        trampoline.file_offset % PAGE_SIZE == 0,
+        "trampoline file offset must be page-aligned"
     );
     let start_addr = relo_off + trampoline.vaddr;
     let end_addr = (start_addr + trampoline.size).next_multiple_of(0x1000);
@@ -330,7 +336,6 @@ fn load_trampoline(trampoline: TrampolineHdr, relo_off: usize, fd: i32) {
     unsafe {
         placeholder.write(litebox_platform_multiplex::platform().get_syscall_entry_point());
     }
-    // `mprotect` requires the address to be page-aligned
     let ptr = crate::MutPtr::from_usize(start_addr);
     let pm = litebox_page_manager();
     unsafe { pm.make_pages_executable(ptr, end_addr - start_addr) }
@@ -412,6 +417,7 @@ impl ElfLoader {
             let trampoline = get_trampoline_hdr(&mut object);
             if let Some(trampoline) = trampoline.as_ref() {
                 // store the trampoline size so that [`ElfLoaderMmap`] can use it
+                // `end_addr` is only used by the main ELF file to compute `brk` address and thus we don't need to reset it.
                 loader_manager()
                     .lock()
                     .get_mut(&tid)


litebox_shim_linux/src/loader/elf.rs
```diff
@@ -102,7 +102,15 @@ impl ElfLoaderMmap {
             // A default low address is used for the binary (which grows upwards) to avoid
             // conflicts with the kernel's memory mappings (which grows downwards).
             addr.unwrap_or(0x1000_0000),
-            len,
+            len + if addr.is_none() {
+                // The elf loader we use calls `mmap` to load all segments with the `PT_LOAD` type at once,
+                // and then invokes `mmap` again to change the memory protection. Thus, only the first
+                // call to `mmap` does not have the `MAP_FIXED` flag, and we need to reserve extra pages
+                // for trampoline code.
+                MAX_SIZE_OF_TRAMPOLINE
```
wdcui: How can we ensure the `MAX_SIZE_OF_TRAMPOLINE` is enough (or too much)?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -215,6 +223,8 @@ struct TrampolineHdr {
     size: usize,
 }
 
+const MAX_SIZE_OF_TRAMPOLINE: usize = 0x2000;
```
wdcui: Did you check the size of the trampoline for libc? That could be the worst case scenario.

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -266,6 +276,49 @@ fn get_trampoline_hdr(object: &mut ElfFile) -> Option<TrampolineHdr> {
...
+    assert_eq!(
+        unsafe { version_number.read() },
+        super::REWRITER_VERSION_NUMBER,
+        "trampoline section version number mismatch"
+    );
+    let placeholder = (start_addr + 8) as *mut usize;
+    unsafe {
+        placeholder.write(litebox_platform_multiplex::platform().get_syscall_entry_point());
+    }
+    // `mprotect` requires the address to be page-aligned
```
wdcui: Should we add a check to see if the start_addr is aligned (or we should have this enforcement in the syscall rewriter)?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -377,19 +392,24 @@ impl ElfLoader {
                 .downcast_ref::<AtomicUsize>()
                 .unwrap()
                 .load(Ordering::Relaxed);
-            let init_brk =
-                core::cmp::max((base + brk).next_multiple_of(PAGE_SIZE), end_of_trampoline);
+            let init_brk = (base + brk).next_multiple_of(PAGE_SIZE) + MAX_SIZE_OF_TRAMPOLINE;
```
wdcui: Is this value consistent with `len + MAX_SIZE_OF_TRAMPOLINE`?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -377,19 +392,24 @@ impl ElfLoader {
...
-                    .map_err(ElfLoaderError::LoaderError)?,
-            )
+            let mut object = ElfFile::new(interp_name).map_err(ElfLoaderError::OpenError)?;
+            let trampoline = get_trampoline_hdr(&mut object);
+            let file_fd = object.as_fd().unwrap();
+            let interp = loader
+                .easy_load(object)
+                .map_err(ElfLoaderError::LoaderError)?;
+
+            if let Some(trampoline) = trampoline {
```
wdcui, line 432: This code is confusing: we are not getting any trampoline returned?

litebox_shim_linux/src/loader/elf.rs
CvvT: For libc, the size is 8649. Updated it to 0x3000.

litebox_shim_linux/src/loader/elf.rs
CvvT: It's enforced by the rewriter. Also, it's checked by `make_pages_executable`. If it's not aligned, `make_pages_executable` would return an error and then the program would panic.

litebox_shim_linux/src/loader/elf.rs
CvvT: What does this `len` refer to?

litebox_shim_linux/src/loader/elf.rs
CvvT: This should be a conservative value and would be checked at runtime. Ideally, this should be `trampoline.size.next_multiple_of(0x1000)`, but I haven't figured out how to pass the value to this function.

litebox_shim_linux/src/loader/elf.rs
CvvT: You mean the return value from `load_trampoline`? We don't need the return value anymore as now the trampoline region has a fixed size.

litebox_shim_linux/src/loader/elf.rs
wdcui: Should we extend this comment to say that this is likely to be the reason for the following call to fail?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -300,96 +370,67 @@ impl ElfLoader {
...
-            )
+            let mut object = ElfFile::new(interp_name).map_err(ElfLoaderError::OpenError)?;
+            let trampoline = get_trampoline_hdr(&mut object);
+            if let Some(trampoline) = trampoline.as_ref() {
+                // store the trampoline size so that [`ElfLoaderMmap`] can use it
+                loader_manager()
+                    .lock()
+                    .get_mut(&tid)
+                    .expect("loader info not found")
+                    .trampoline_size = trampoline.size;
```
wdcui, line 425: Do we need to update the `end_addr`?

CvvT: No, it is only used to compute `brk` address for the main binary.

litebox_shim_linux/src/loader/elf.rs
CvvT: Added some comments at the beginning of the function.

litebox_shim_linux/src/loader/elf.rs
wdcui: Can you add this into the comment so it's clear for a future reader of this code?

CvvT: Sure, added.

PR #192, opened 2025-07-02
Add rewriter backend for runner

CvvT:
Add another unit test for runner to test the rewriter backend. Since we haven't got the dynamic linking with rewriter working, the test is commented out for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,7 @@ jobs:
           fi
       - name: Set up Rust
         run: |
-          rustup toolchain install $(awk -F'"' '/channel/{print $2}' rust-toolchain.toml) --profile minimal --no-self-update --component rustfmt,clippy
-          rustup target add x86_64-unknown-linux-musl
+          rustup toolchain install $(awk -F'"' '/channel/{print $2}' rust-toolchain.toml) --profile minimal --no-self-update --component rustfmt,clippy --target x86_64-unknown-linux-gnu,x86_64-unknown-linux-musl
       - name: Set up Nextest
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -58,10 +60,11 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
-      - run: sudo apt install -y gcc-multilib
+      - run: sudo apt install -y gcc-multilib musl-tools
       - name: Set up Rust
         run: |
           rustup toolchain install $(awk -F'"' '/channel/{print $2}' rust-toolchain.toml) --profile minimal --no-self-update --component rustfmt,clippy --target i686-unknown-linux-gnu
+          rustup target add i686-unknown-linux-musl
```
jaybosamiya-ms: I recommend combining these
```suggestion
          rustup toolchain install $(awk -F'"' '/channel/{print $2}' rust-toolchain.toml) --profile minimal --no-self-update --component rustfmt,clippy --target i686-unknown-linux-gnu,i686-unknown-linux-musl
```

.github/workflows/ci.yml
```diff
@@ -31,6 +32,7 @@ jobs:
       - name: Set up Rust
         run: |
           rustup toolchain install $(awk -F'"' '/channel/{print $2}' rust-toolchain.toml) --profile minimal --no-self-update --component rustfmt,clippy
+          rustup target add x86_64-unknown-linux-musl
```
jaybosamiya-ms: Similarly to the other one, it becomes a little more clear that we are pulling two targets if we explicitly have `--target x86_64-unknown-linux-gnu,x86_64-unknown-linux-musl`

CvvT: Fixed. Thanks!

PR #194, opened 2025-07-03
remove north litebox::fs dependency from the south platform

Icegrave0391:


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -140,7 +139,7 @@ impl LinuxUserland {
             syscalls::syscall3(
                 syscalls::Sysno::open,
                 path.as_ptr() as usize,
-                OFlags::RDONLY.bits() as usize,
+                libc::O_RDONLY as usize,
```
CvvT, line 142: We should probably redefine this constant in `litebox_common_linux`.

PR #195, opened 2025-07-06
Support host platform: freebsd_userland

Icegrave0391:
- Added a new platform under `litebox_platform_freebsd_userland`.
- Added a new testcase to run native Linux helloworld program on FreeBSD.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -587,6 +587,9 @@ impl RawMutex {
                         // when we attempted to wait, so we can do a proper check.
                     }
                 }
+                Err(syscalls::Errno::ETIMEDOUT) => {
+                    return Ok(UnblockedOrTimedOut::TimedOut);
+                }
                 Err(e) => {
                     panic!("Unexpected errno={e} for FUTEX_WAIT")
                 }
--- a/litebox_shim_linux/Cargo.toml
+++ b/litebox_shim_linux/Cargo.toml
@@ -30,8 +30,8 @@ elf_loader = { version = "=0.10.0", default-features = false, features = ["rel"]
 
 [features]
 
-[lints]
-workspace = true
-
 [dev-dependencies]
 spin = { version = "0.9.8", default-features = false, features = ["spin_mutex"] }
+
+[lints]
+workspace = true
--- a/litebox_shim_linux/src/syscalls/mod.rs
+++ b/litebox_shim_linux/src/syscalls/mod.rs
@@ -1,5 +1,6 @@
 //! Syscalls Handlers
 
+pub(crate) mod epoll;
 pub(crate) mod eventfd;
 pub mod file;
 pub(crate) mod misc;
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -985,6 +985,45 @@ where
 
         Ok(socket_handle.socket_metadata.insert(metadata))
     }
+
+    /// Get the [`Events`] for a socket.
+    pub fn check_events(&self, fd: &SocketFd<Platform>) -> Option<Events> {
+        let descriptor_table = self.litebox.descriptor_table_mut();
+        let mut table_entry = descriptor_table.get_entry_mut(fd);
+        let socket_handle = &mut table_entry.entry;
+        match socket_handle.protocol() {
+            Protocol::Tcp => {
+                let tcp_socket = self.socket_set.get::<tcp::Socket>(socket_handle.handle);
+                let mut events = Events::empty();
+                if tcp_socket.can_recv() {
+                    events |= Events::IN;
+                }
+                if tcp_socket.can_send() {
+                    events |= Events::OUT;
+                }
+                if !tcp_socket.is_open() {
+                    events |= Events::HUP;
+                } else if !events.contains(Events::IN) {
+                    // A server socket should have `Events::IN` if any of its listening sockets is connected.
+                    if let Some(server_socket) = socket_handle.specific.tcp().server_socket.as_ref()
+                    {
+                        server_socket
+                            .socket_set_handles
+                            .iter()
+                            .any(|&h| {
+                                let socket: &tcp::Socket = self.socket_set.get(h);
+                                socket.state() == tcp::State::Established
+                            })
+                            .then(|| events.insert(Events::IN));
+                    }
+                }
+                Some(events)
+            }
+            Protocol::Udp => unimplemented!(),
+            Protocol::Icmp => unimplemented!(),
+            Protocol::Raw { protocol: _ } => unimplemented!(),
+        }
+    }
 }
 
 /// Protocols for sockets supported by LiteBox
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1203,6 +1203,30 @@ impl<Platform: litebox::platform::RawPointerProvider> Clone for RobustListHead<P
     }
 }
 
+bitflags::bitflags! {
+    #[derive(Debug)]
+    pub struct EpollCreateFlags: core::ffi::c_uint {
+        const EPOLL_CLOEXEC = litebox::fs::OFlags::CLOEXEC.bits();
+        /// <https://docs.rs/bitflags/*/bitflags/#externally-defined-flags>
+        const _ = !0;
+    }
+}
+
+#[repr(i32)]
+#[derive(Debug, IntEnum, PartialEq, Eq)]
+pub enum EpollOp {
+    EpollCtlAdd = 1,
+    EpollCtlDel = 2,
+    EpollCtlMod = 3,
+}
+
+#[derive(Clone, Copy, Debug)]
+#[repr(C, packed)]
+pub struct EpollEvent {
+    pub events: u32,
+    pub data: u64,
+}
+
 /// Request to syscall handler
 #[non_exhaustive]
 pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
@@ -1373,6 +1397,23 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
     Getcwd {
         buf: Platform::RawMutPointer<u8>,
         size: usize,
+    },
+    EpollCtl {
+        epfd: i32,
+        op: EpollOp,
+        fd: i32,
+        event: Platform::RawConstPointer<EpollEvent>,
+    },
+    EpollPwait {
+        epfd: i32,
+        events: Platform::RawMutPointer<EpollEvent>,
+        maxevents: u32,
+        timeout: i32,
+        sigmask: Option<Platform::RawConstPointer<SigSet>>,
+        sigsetsize: usize,
+    },
+    EpollCreate {
+        flags: EpollCreateFlags,
     },
     ArchPrctl {
         arg: ArchPrctlArg<Platform>,
@@ -1871,6 +1912,53 @@ impl<'a, Platform: litebox::platform::RawPointerProvider> SyscallRequest<'a, Pla
             Sysno::getgid => SyscallRequest::Getgid,
             Sysno::geteuid => SyscallRequest::Geteuid,
             Sysno::getegid => SyscallRequest::Getegid,
+            Sysno::epoll_ctl => {
+                let op: i32 = ctx.syscall_arg(1).reinterpret_as_signed().truncate();
+                if let Ok(op) = EpollOp::try_from(op) {
+                    SyscallRequest::EpollCtl {
+                        epfd: ctx.syscall_arg(0).reinterpret_as_signed().truncate(),
+                        op,
+                        fd: ctx.syscall_arg(2).reinterpret_as_signed().truncate(),
+                        event: Platform::RawConstPointer::from_usize(ctx.syscall_arg(3)),
+                    }
+                } else {
+                    SyscallRequest::Ret(errno::Errno::EINVAL)
+                }
+            }
+            Sysno::epoll_wait => SyscallRequest::EpollPwait {
+                epfd: ctx.syscall_arg(0).reinterpret_as_signed().truncate(),
+                events: Platform::RawMutPointer::from_usize(ctx.syscall_arg(1)),
+                maxevents: ctx.syscall_arg(2).truncate(),
+                timeout: ctx.syscall_arg(3).reinterpret_as_signed().truncate(),
+                sigmask: None,
+                sigsetsize: 0,
+            },
+            Sysno::epoll_pwait => SyscallRequest::EpollPwait {
+                epfd: ctx.syscall_arg(0).reinterpret_as_signed().truncate(),
+                events: Platform::RawMutPointer::from_usize(ctx.syscall_arg(1)),
+                maxevents: ctx.syscall_arg(2).truncate(),
+                timeout: ctx.syscall_arg(3).reinterpret_as_signed().truncate(),
+                sigmask: if ctx.syscall_arg(4) == 0 {
+                    None
+                } else {
+                    Some(Platform::RawConstPointer::from_usize(ctx.syscall_arg(4)))
+                },
+                sigsetsize: ctx.syscall_arg(5),
+            },
+            Sysno::epoll_create => {
+                // the `size` argument is ignored, but must be greater than zero;
+                let size: i32 = ctx.syscall_arg(0).reinterpret_as_signed().truncate();
+                if size > 0 {
+                    SyscallRequest::EpollCreate {
+                        flags: EpollCreateFlags::empty(),
+                    }
+                } else {
+                    SyscallRequest::Ret(errno::Errno::EINVAL)
+                }
+            }
+            Sysno::epoll_create1 => SyscallRequest::EpollCreate {
+                flags: EpollCreateFlags::from_bits_truncate(ctx.syscall_arg(0).truncate()),
+            },
             Sysno::arch_prctl => {
                 let code: u32 = ctx.syscall_arg(0).truncate();
                 if let Ok(code) = ArchPrctlCode::try_from(code) {
--- a/litebox_shim_linux/src/channel.rs
+++ b/litebox_shim_linux/src/channel.rs
@@ -4,23 +4,35 @@
 use core::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 
 use alloc::sync::{Arc, Weak};
-use litebox::{LiteBox, fs::OFlags};
+use litebox::{LiteBox, event::Events, fs::OFlags};
 use litebox_common_linux::errno::Errno;
 use litebox_platform_multiplex::Platform;
 use ringbuf::{
     HeapCons, HeapProd, HeapRb,
-    traits::{Consumer as _, Producer as _, Split as _},
+    traits::{Consumer as _, Observer as _, Producer as _, Split as _},
 };
+
+use crate::event::Observer;
+
+/// The maximum number of bytes for atomic write.
+///
+/// See <https://man7.org/linux/man-pages/man7/pipe.7.html> for more details.
+#[cfg(not(test))]
+const PIPE_BUF: usize = 4096;
+#[cfg(test)]
+const PIPE_BUF: usize = 2;
 
 struct EndPointer<T> {
     rb: litebox::sync::Mutex<Platform, T>,
+    pollee: crate::event::Pollee,
     is_shutdown: AtomicBool,
 }
 
 impl<T> EndPointer<T> {
-    pub fn new(rb: T, litebox: &LiteBox<Platform>) -> Self {
+    pub fn new(rb: T, init_events: Events, litebox: &LiteBox<Platform>) -> Self {
         Self {
             rb: litebox.sync().new_mutex(rb),
+            pollee: crate::event::Pollee::new(init_events),
             is_shutdown: AtomicBool::new(false),
         }
     }
@@ -51,6 +63,20 @@ macro_rules! common_functions_for_channel {
                 true
             }
         }
+
+        pub fn poll(&self, mask: Events, observer: Option<Weak<dyn Observer<Events>>>) -> Events {
+            self.endpoint
+                .pollee
+                .poll(mask, observer, || self.check_io_events())
+        }
+
+        pub(crate) fn register_observer(
+            &self,
+            observer: alloc::sync::Weak<dyn Observer<Events>>,
+            filter: Events,
+        ) {
+            self.endpoint.pollee.register_observer(observer, filter);
+        }
     };
 }
 
@@ -64,10 +90,22 @@ pub(crate) struct Producer<T> {
 impl<T> Producer<T> {
     fn new(rb: HeapProd<T>, flags: OFlags, litebox: &LiteBox<Platform>) -> Self {
         Self {
-            endpoint: EndPointer::new(rb, litebox),
+            endpoint: EndPointer::new(rb, Events::OUT, litebox),
             peer: Weak::new(),
             status: AtomicU32::new((flags | OFlags::WRONLY).bits()),
         }
+    }
+
+    fn check_io_events(&self) -> Events {
+        let rb = self.endpoint.rb.lock();
+        let mut events = Events::empty();
+        if self.is_peer_shutdown() {
+            events |= Events::ERR;
+        }
+        if !self.is_shutdown() && !rb.is_full() {
+            events |= Events::OUT;
+        }
+        events
     }
 
     fn try_write(&self, buf: &[T]) -> Result<usize, Errno>
@@ -81,8 +119,20 @@ impl<T> Producer<T> {
             return Ok(0);
         }
 
-        let write_len = self.endpoint.rb.lock().push_slice(buf);
+        let write_len = {
+            let mut rb = self.endpoint.rb.lock();
+            let total_size = core::mem::size_of_val(buf);
+            if rb.vacant_len() < total_size && total_size <= PIPE_BUF {
+                // No sufficient space for an atomic write
+                0
+            } else {
+                rb.push_slice(buf)
+            }
+        };
         if write_len > 0 {
+            if let Some(peer) = self.peer.upgrade() {
+                peer.endpoint.pollee.notify_observers(Events::IN);
+            }
             Ok(write_len)
         } else {
             Err(Errno::EAGAIN)
@@ -96,14 +146,12 @@ impl<T> Producer<T> {
         if self.get_status().contains(OFlags::NONBLOCK) {
             self.try_write(buf)
         } else {
-            // TODO: use poll rather than busy wait
-            loop {
-                match self.try_write(buf) {
-                    Err(Errno::EAGAIN) => {}
-                    ret => return ret,
-                }
-                core::hint::spin_loop();
-            }
+            self.endpoint.pollee.wait_or_timeout(
+                Events::OUT,
+                None,
+                || self.try_write(buf),
+                || self.check_io_events(),
+            )
         }
     }
 
@@ -114,6 +162,12 @@ impl<T> Producer<T> {
 impl<T> Drop for Producer<T> {
     fn drop(&mut self) {
         self.shutdown();
+
+        if let Some(peer) = self.peer.upgrade() {
+            // when reading from a channel such as a pipe or a stream socket, this event
+            // merely indicates that the peer closed its end of the channel.
+            peer.endpoint.pollee.notify_observers(Events::HUP);
+        }
     }
 }
 
@@ -126,10 +180,22 @@ pub(crate) struct Consumer<T> {
 impl<T> Consumer<T> {
     fn new(rb: HeapCons<T>, flags: OFlags, litebox: &LiteBox<Platform>) -> Self {
         Self {
-            endpoint: EndPointer::new(rb, litebox),
+            endpoint: EndPointer::new(rb, Events::empty(), litebox),
             peer: Weak::new(),
             status: AtomicU32::new((flags | OFlags::RDONLY).bits()),
         }
+    }
+
+    fn check_io_events(&self) -> Events {
+        let rb = self.endpoint.rb.lock();
+        let mut events = Events::empty();
+        if self.is_peer_shutdown() {
+            events |= Events::HUP;
+        }
+        if !self.is_shutdown() && !rb.is_empty() {
+            events |= Events::IN;
+        }
+        events
     }
 
     fn try_read(&self, buf: &mut [T]) -> Result<usize, Errno>
@@ -145,6 +211,9 @@ impl<T> Consumer<T> {
 
         let read_len = self.endpoint.rb.lock().pop_slice(buf);
         if read_len > 0 {
+            if let Some(peer) = self.peer.upgrade() {
+                peer.endpoint.pollee.notify_observers(Events::OUT);
+            }
             Ok(read_len)
         } else {
             if self.is_peer_shutdown() {
@@ -163,14 +232,12 @@ impl<T> Consumer<T> {
         if self.get_status().contains(OFlags::NONBLOCK) {
             self.try_read(buf)
         } else {
-            // TODO: use poll rather than busy wait
-            loop {
-                match self.try_read(buf) {
-                    Err(Errno::EAGAIN) => {}
-                    ret => return ret,
-                }
-                core::hint::spin_loop();
-            }
+            self.endpoint.pollee.wait_or_timeout(
+                Events::IN,
+                None,
+                || self.try_read(buf),
+                || self.check_io_events(),
+            )
         }
     }
 
@@ -181,6 +248,12 @@ impl<T> Consumer<T> {
 impl<T> Drop for Consumer<T> {
     fn drop(&mut self) {
         self.shutdown();
+
+        if let Some(peer) = self.peer.upgrade() {
+            // This bit is also set for a file descriptor referring to the write end
+            // of a pipe when the read end has been closed.
+            peer.endpoint.pollee.notify_observers(Events::ERR);
+        }
     }
 }
 
@@ -224,17 +297,12 @@ impl<T> Channel<T> {
 #[cfg(test)]
 mod tests {
     use litebox_common_linux::errno::Errno;
-    use litebox_platform_multiplex::{Platform, set_platform};
 
     extern crate std;
 
-    fn init_platform() {
-        set_platform(Platform::new(None));
-    }
-
     #[test]
     fn test_blocking_channel() {
-        init_platform();
+        crate::syscalls::tests::init_platform(None);
 
         let (prod, cons) =
             super::Channel::<u8>::new(2, litebox::fs::OFlags::empty(), crate::litebox()).split();
@@ -264,7 +332,7 @@ mod tests {
 
     #[test]
     fn test_nonblocking_channel() {
-        init_platform();
+        crate::syscalls::tests::init_platform(None);
 
         let (prod, cons) =
             super::Channel::<u8>::new(2, litebox::fs::OFlags::NONBLOCK, crate::litebox()).split();
--- a/litebox_shim_linux/src/event.rs
+++ b/litebox_shim_linux/src/event.rs
@@ -1,0 +1,270 @@
+use alloc::collections::btree_map::BTreeMap;
+use litebox::platform::RawMutex as _;
+use litebox::platform::RawMutexProvider;
+use litebox::{
+    event::Events,
+    platform::{Instant as _, TimeProvider as _},
+};
+use litebox_common_linux::errno::Errno;
+
+trait EventsFilter<E>: Send + Sync + 'static {
+    fn filter(&self, event: &E) -> bool;
+}
+
+impl EventsFilter<Events> for Events {
+    fn filter(&self, events: &Events) -> bool {
+        self.intersects(*events)
+    }
+}
+
+pub(crate) trait Observer<E>: Send + Sync {
+    fn on_events(&self, events: &E);
+}
+
+struct ObserverKey<E> {
+    observer: alloc::sync::Weak<dyn Observer<E>>,
+}
+
+impl<E> PartialEq for ObserverKey<E> {
+    fn eq(&self, other: &Self) -> bool {
+        self.observer.ptr_eq(&other.observer)
+    }
+}
+
+impl<E> Eq for ObserverKey<E> {}
+
+impl<E> PartialOrd for ObserverKey<E> {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl<E> Ord for ObserverKey<E> {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        self.observer
+            .as_ptr()
+            .cast::<()>()
+            .cmp(&other.observer.as_ptr().cast::<()>())
+    }
+}
+
+impl<E> ObserverKey<E> {
+    fn upgrade(&self) -> Option<alloc::sync::Arc<dyn Observer<E>>> {
+        self.observer.upgrade()
+    }
+}
+
+/// A Subject notifies interesting events to registered observers.
+struct Subject<E, F: EventsFilter<E>> {
+    /// A table that maintains all interesting observers.
+    observers:
+        litebox::sync::Mutex<litebox_platform_multiplex::Platform, BTreeMap<ObserverKey<E>, F>>,
+    /// Number of observers.
+    nums: core::sync::atomic::AtomicUsize,
+}
+
+impl<E, F: EventsFilter<E>> Subject<E, F> {
+    fn new() -> Self {
+        Self {
+            observers: crate::litebox().sync().new_mutex(BTreeMap::new()),
+            nums: core::sync::atomic::AtomicUsize::new(0),
+        }
+    }
+
+    fn register_observer(&self, observer: alloc::sync::Weak<dyn Observer<E>>, filter: F) {
+        let mut observers = self.observers.lock();
+        if observers.insert(ObserverKey { observer }, filter).is_none() {
+            self.nums
+                .fetch_add(1, core::sync::atomic::Ordering::Relaxed);
+        }
+    }
+
+    fn unregister_observer(&self, observer: alloc::sync::Weak<dyn Observer<E>>) {
+        let mut observers = self.observers.lock();
+        if observers.remove(&ObserverKey { observer }).is_some() {
+            self.nums
+                .fetch_sub(1, core::sync::atomic::Ordering::Relaxed);
+        }
+    }
+
+    fn notify_observers(&self, events: E) {
+        if self.nums.load(core::sync::atomic::Ordering::Relaxed) == 0 {
+            return;
+        }
+
+        let mut observers = self.observers.lock();
+        observers.retain(|observer, filter| {
+            if let Some(observer) = observer.upgrade() {
+                if filter.filter(&events) {
+                    observer.on_events(&events);
+                }
+                true
+            } else {
+                self.nums
+                    .fetch_sub(1, core::sync::atomic::Ordering::Relaxed);
+                false
+            }
+        });
+    }
+}
+
+pub(crate) struct Pollee {
+    inner: alloc::sync::Arc<PolleeInner>,
+}
+
+struct PolleeInner {
+    subject: Subject<Events, Events>,
+}
+
+impl Pollee {
+    pub(crate) fn new(init_events: Events) -> Self {
+        let inner = alloc::sync::Arc::new(PolleeInner {
+            subject: Subject::new(),
+        });
+        Self { inner }
+    }
+
+    /// Poll the pollee with the given mask of events and register the poller.
+    pub(crate) fn poll<F>(
+        &self,
+        mask: Events,
+        observer: Option<alloc::sync::Weak<dyn Observer<Events>>>,
+        check: F,
+    ) -> Events
+    where
+        F: FnOnce() -> Events,
+    {
+        let mask = mask | Events::ALWAYS_POLLED;
+
+        if let Some(observer) = observer {
+            self.register_observer(observer, mask);
+        }
+
+        check() & mask
+    }
+
+    /// Wait until `try_op` returns a non-EAGAIN result or timeout.
+    /// If the `timeout` is None, it will wait indefinitely.
+    pub fn wait_or_timeout<F, C, R>(
+        &self,
+        mask: Events,
+        timeout: Option<core::time::Duration>,
+        mut try_op: F,
+        check: C,
+    ) -> Result<R, Errno>
+    where
+        F: FnMut() -> Result<R, Errno>,
+        C: FnOnce() -> Events,
+    {
+        // Try first without waiting
+        match try_op() {
+            Err(Errno::EAGAIN) => {}
+            ret => return ret,
+        }
+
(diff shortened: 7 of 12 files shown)


Cargo.toml
```diff
@@ -14,7 +15,7 @@ members = [
 # exclude binaries which need different toolchain and build target.
 exclude = [
     "litebox_platform_lvbs",
-    "litebox_runner_lvbs"
+    "litebox_runner_lvbs",
```
wdcui: Minor: we probably should avoid this unnecessary change until we need to actually change anything in this array.

litebox_platform_freebsd_userland/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "litebox_platform_freebsd_userland"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+libc = { version = "0.2.169", default-features = false }
+litebox = { path = "../litebox/", version = "0.1.0" }
+litebox_common_linux = { path = "../litebox_common_linux", version = "0.1.0" }
+nix = { version = "0.29.0", default-features = false, features = ["fs", "ioctl", "mman", "signal"] }
```
wdcui: Does nix support FreeBSD?

litebox_platform_freebsd_userland/README.md
```diff
@@ -0,0 +1,9 @@
+# A LiteBox Platform for running LiteBox on userland Linux
```
wdcui: Fix the comment: Linux -> FreeBSD?

litebox_platform_freebsd_userland/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+This crate provides an instantiation of the LiteBox `platform::Provider`, with
+parameterized punchthrough.
+
+It requires access to a TUN device that has been initialized. For convenience,
+[`./scripts/tun-setup.sh`](./scripts/tun-setup.sh) will help you initialize a
+TUN device and ready it for usage with for this platform. Passing `-h` as an
```
wdcui: "with for" => "with" or "for"?

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
wdcui, line 1: What's the difference between "freebsd_types/mod.rs" and "freebsd_types.rs"?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,873 @@
...
+// use std::os::fd::{AsRawFd as _, FromRawFd as _};
+use std::sync::atomic::AtomicU32;
+// use std::sync::atomic::Ordering::SeqCst;
+use std::time::Duration;
+
+use litebox::platform::ImmediatelyWokenUp;
+use litebox::platform::page_mgmt::MemoryRegionPermissions;
+use litebox::platform::trivial_providers::TransparentMutPtr;
+use litebox::platform::{ThreadLocalStorageProvider, UnblockedOrTimedOut};
+// use litebox::platform::{ImmediatelyWokenUp, RawConstPointer, ThreadLocalStorageProvider};
```
wdcui: Remove the lines that are commented out?

Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
     "litebox_common_linux",
     "litebox_platform_linux_kernel",
     "litebox_platform_linux_userland",
+    "litebox_platform_freebsd_userland",
```
wdcui, line 8: As we discussed in the meeting today, a more precise name could be "litebox_platform_linux_on_freebsd_userland"?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,873 @@
...
+use litebox::utils::ReinterpretUnsignedExt as _;
+use litebox_common_linux::{ProtFlags, PunchthroughSyscall};
+// use litebox_common_linux::{CloneFlags, MRemapFlags, MapFlags, ProtFlags, PunchthroughSyscall};
+
+pub mod syscall_raw;
+use syscall_raw::syscalls;
+
+mod freebsd_types;
+
+// todo(chuqi): we do not use systrap for now as there's no sccomp interception on FreeBSD.
```
wdcui: This is not really a todo since we will not do seccomp interception on FreeBSD at all?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,873 @@
...
+
+/// The syscall handler passed down from the shim.
+static SYSCALL_HANDLER: std::sync::RwLock<Option<SyscallHandler>> = std::sync::RwLock::new(None);
+
+/// The userland FreeBSD platform.
+///
+/// This implements the main [`litebox::platform::Provider`] trait, i.e., implements all platform
+/// traits.
+#[allow(dead_code)]
+pub struct FreeBSDUserland {
```
wdcui, line 35: We need to decide on the naming. "LinuxOnFreeBSDUserland" is a little long though it's more accurate. @jaybosamiya-ms @CvvT what's your preference?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,873 @@
...
+    tun_socket_fd: std::sync::RwLock<Option<std::os::fd::OwnedFd>>,
+    // seccomp_interception_enabled: std::sync::atomic::AtomicBool,
```
wdcui: This should just be removed since we will have seccomp interception on FreeBSD.

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,873 @@
...
+            tun_socket_fd,
+            // seccomp_interception_enabled: std::sync::atomic::AtomicBool::new(false),
+            reserved_pages: Self::read_proc_self_maps(),
+        };
+
+        platform.set_init_tls();
+        Box::leak(Box::new(platform))
+    }
+
+    /// Register the syscall handler (provided by the FreeBSD shim)
```
wdcui: "provided by the Linux shim"?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,873 @@
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the function has already been invoked earlier.
+    pub fn register_syscall_handler(&self, syscall_handler: SyscallHandler) {
```
wdcui, line 58: If we don't support seccomp interception, do we still need to register this syscall handler?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,873 @@
...
+        );
+    }
+
+    /// Enable seccomp syscall interception on the platform.
+    ///
+    /// # Panics
+    ///
+    /// Panics if this function has already been invoked on the platform earlier.
+    pub fn enable_seccomp_based_syscall_interception(&self) {
+        // todo(chuqi): we do not support seccomp for FreeBSD
```
wdcui: I saw the following code in litebox_platform_multiplex:

#[cfg(feature = "platform_linux_userland")]
pub type Platform = litebox_platform_linux_userland::LinuxUserland;

I guess we should add a feature for freebsd_userland so that the Linux shim won't call the functions related to seccomp interceptions?

litebox_platform_freebsd_userland/src/syscall_raw.rs
```diff
@@ -0,0 +1,755 @@
+// Automatically generated — do not edit.
```
wdcui: How do you generate this file?

litebox_platform_multiplex/Cargo.toml
```diff
@@ -6,11 +6,13 @@ edition = "2024"
 [dependencies]
 litebox = { path = "../litebox/", version = "0.1.0" }
 litebox_platform_linux_userland = { path = "../litebox_platform_linux_userland/", version = "0.1.0", optional = true }
+litebox_platform_freebsd_userland = { path = "../litebox_platform_freebsd_userland/", version = "0.1.0", optional = true }
 once_cell = { version = "1.20.2", default-features = false, features = ["alloc", "race"] }
 
 [features]
 default = ["platform_linux_userland"]
 platform_linux_userland = ["dep:litebox_platform_linux_userland"]
+platform_freebsd_userland = ["dep:litebox_platform_freebsd_userland"]
```
wdcui, line 15: We should use this feature to disable the code that calls functions for seccomp interception.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -1,5 +1,4 @@
 //! Implementation of memory management related syscalls, eg., `mmap`, `munmap`, etc.
-
```
wdcui: Please avoid this unnecessary change in your PR.

litebox_shim_linux/src/syscalls/mod.rs
```diff
@@ -10,6 +10,21 @@ pub(crate) mod process;
 #[cfg(test)]
 pub(crate) mod tests;
 
+macro_rules! nontest_log_println {
```
wdcui: I'm not sure we should check in this code.

litebox_shim_linux/tests/exec.rs
```diff
@@ -145,3 +145,52 @@ fn test_syscall_rewriter() {
     common::install_file(executable_data, executable_path);
     common::test_load_exec_common(executable_path);
 }
+
+#[test]
+#[cfg(all(target_arch = "x86_64", target_os = "freebsd"))]
+fn test_syscall_rewriter_curdir() {
```
wdcui: Do we need this test for FreeBSD?

wdcui:
I didn't see that this PR is marked as draft; but I hope you still find my comments useful.

Cargo.toml
Icegrave0391: Sure. I will not touch `exclude`

litebox_platform_freebsd_userland/Cargo.toml
Icegrave0391: Removed nix as well

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
Icegrave0391: Good question! I may consider:
(a) merge `freebsd_types.rs` into `freebsd_types/mod.rs`, or
(b) change `freebsd_types/mod.rs` to `freebsd_types.rs`

litebox_platform_freebsd_userland/src/syscall_raw.rs
Icegrave0391: I wrote a python script to parse this specification: https://cgit.freebsd.org/src/tree/sys/kern/syscalls.master

Will only keep host syscalls that we need. The full list of syscalls is just for convenience.

litebox_shim_linux/tests/exec.rs
Icegrave0391: This is a helloworld test. It uses pre-compiled linux native binary for rewriter and host freebsd_userland platform.

litebox_shim_linux/src/syscalls/mm.rs
Icegrave0391: Get it. Previously added some debug printing. Will check git diff to avoid unnecessary deletions.

litebox_platform_freebsd_userland/src/lib.rs
CvvT: I believe it's fine as Platform does not need to care about the upper layer (interface for applications).

litebox_platform_freebsd_userland/src/lib.rs
CvvT: Yes, the syscall entry function (to save and restore context) is defined in the platform, while the handler is defined in the shim layer. Syscall entry function calls the syscall handler provided by the shim layer.

litebox_platform_freebsd_userland/tests/syscalls.rs
```diff
@@ -0,0 +1,166 @@
...
+            println!("Successfully wrote {} bytes to stderr", bytes_written);
+        }
+        Err(errno) => {
+            panic!(
+                "Writing to stderr should have succeeded, but got errno: {}",
+                errno
+            );
+        }
+    }
+}
```
CvvT: Are those tests just testing the host syscalls? They don't test our code?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -140,7 +139,7 @@ impl LinuxUserland {
             syscalls::syscall3(
                 syscalls::Sysno::open,
                 path.as_ptr() as usize,
-                OFlags::RDONLY.bits() as usize,
+                libc::O_RDONLY as usize,
```
CvvT: You can redefine this constant in litebox_common_linux to avoid using libc

litebox_shim_linux/tests/exec.rs
CvvT: Could we merge this one with the existing one by using conditional compilation?

litebox_shim_linux/src/syscalls/mod.rs
CvvT: I added similar functionality to litebox yesterday, you don't need to reimplement it. Please refer to #196

litebox_platform_freebsd_userland/tests/syscalls.rs
Icegrave0391: yup they just test host syscalls

Cargo.toml
jaybosamiya-ms: I think the _platform_ is still `litebox_platform_freebsd_userland`.  The _runner_ is what would be "linux on freebsd"

Cargo.toml
jaybosamiya-ms: The `,` change is still in the diff right now. Please clean up before we can merge this.

litebox_platform_freebsd_userland/README.md
```diff
@@ -0,0 +1,4 @@
+# A LiteBox Platform for running LiteBox on userland FreeBSD
+
+This crate provides an instantiation of the LiteBox `platform::Provider`, with
+parameterized punchthrough.
```
jaybosamiya-ms: The FreeBSD userland platform does _not_ support parameterized punchthrough (that particular point was part of an older design of Linux userland platform that we have moved away from). It would be better to just drop the "with parameterized punchthrough" part here

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
jaybosamiya-ms: If you have only one `mod.rs` file in a directory, then the additional indirection of placing it into a directory is not helpful.  Please rename this file to just `freebsd_types.rs`.

Generally speaking, we use the `foo/mod.rs` pattern for a module `foo` only if we _also_ have sub-modules below it (thus having the files in a directory makes sense).

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
```diff
@@ -0,0 +1,62 @@
+#![allow(non_camel_case_types)]
+
+use bitflags;
+
+// pub const STDIN_FILENO: i32 = 0;
```
jaybosamiya-ms: Any reason this is commented out?  If it is not used, please remove it.

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+pub const STDOUT_FILENO: i32 = 1;
+pub const STDERR_FILENO: i32 = 2;
```
jaybosamiya-ms: I don't see why these are `pub`.  Crates outside this do not need to know these numbers, so this should _at max_ be `pub(crate)`.  Prefer fully private wherever possible, but where not, prefer `pub(crate)` instead.

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+
+bitflags::bitflags! {
+    /// Desired memory protection of a memory mapping.
+    #[derive(PartialEq, Debug)]
+    pub struct ProtFlags: core::ffi::c_int {
```
jaybosamiya-ms: I'm assuming you'll look over all the `pub` in this crate and consider which ones can be made `pub(crate)` or even private, but just marking an extra one here just as a reminder that you should look at all of them.

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+        /// Share this mapping. Mutually exclusive with `MAP_PRIVATE`.
+        const MAP_SHARED = 0x0001;
+        /// Changes are private
+        const MAP_PRIVATE = 0x0002;
+        /// Interpret addr exactly
+        const MAP_FIXED = 0x0010;
+        /// don't use a file (FreeBSD uses MAP_ANON)
+        const MAP_ANON = 0x1000;
+        /// Synonym for [`MAP_ANON`] (FreeBSD style)
+        const MAP_ANONYMOUS = 0x1000;
```
jaybosamiya-ms: You can say `= Self::MAP_ANON.bits()` to make it clearer, and make sure there isn't a mistake.

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+        /// Reserve address space without allocating memory
+        const MAP_GUARD = 0x00002000;
```
jaybosamiya-ms: All the ones before this are `0xABCD` (four hex digits) whereas this one onwards is `0xABCDEFGH` (eight hex digits).  I don't see any particular reason for this change.  Please write things out consistently one way or the other.

Also, Rust supports more readable constants than (say) C.  You can write this out as `0x0000_2000` which makes it much easier for humans to parse the value.

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+        /// Do not include this mapping in core dumps
+        const MAP_NOCORE = 0x00020000;
+        /// Prefault read pages (FreeBSD equivalent of MAP_POPULATE?)
+        const MAP_PREFAULT_READ = 0x00040000;
+        /// Don't sync to backing store
+        const MAP_NOSYNC = 0x00000800;
+        /// Use 2MB super pages if possible
+        const MAP_ALIGNED_SUPER = 0x01000000;
+        /// Align to specific boundary (used with MAP_ALIGNED_SUPER)
+        const MAP_ALIGNMENT_SHIFT = 24;
```
jaybosamiya-ms: Is this actually intended to be the _decimal_ number 24?  Why not write this out in hex?

litebox_platform_freebsd_userland/src/lib.rs
jaybosamiya-ms: Agreed, there is a lot of commented out code in general, please do not leave commented-out code in (anywhere in the PR, not just in this one spot), thanks!

litebox_platform_freebsd_userland/src/lib.rs
jaybosamiya-ms: Agreed, the platform is **not** Linux-specific.  `FreeBSDUserland` _is_ correct.

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+pub type SyscallHandler = fn(litebox_common_linux::SyscallRequest<FreeBSDUserland>) -> isize;
+
+/// The syscall handler passed down from the shim.
+static SYSCALL_HANDLER: std::sync::RwLock<Option<SyscallHandler>> = std::sync::RwLock::new(None);
+
+/// The userland FreeBSD platform.
+///
+/// This implements the main [`litebox::platform::Provider`] trait, i.e., implements all platform
+/// traits.
+#[allow(dead_code)]
```
jaybosamiya-ms: Why is this `allow(dead_code)`?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+    /// Create a new userland-FreeBSD platform for use in `LiteBox`.
+    ///
+    /// Takes an optional tun device name (such as `"tun0"` or `"tun99"`) to connect networking (if
+    /// not specified, networking is disabled).
+    ///
+    /// # Panics
+    ///
+    /// Panics if the tun device could not be successfully opened.
+    pub fn new(_tun_device_name: Option<&str>) -> &'static Self {
+        // todo(chuqi): ignore tun device for now
```
jaybosamiya-ms: Usually, when writing `todo` comments, you want to write what _should_ be done, not what the current state is.  Thus, this comment is better written as `todo(chuqi): implement tun` or similar.  Obviously, please apply this to anywhere you write `todo` comments.

litebox_platform_freebsd_userland/src/lib.rs
jaybosamiya-ms: This function should simply not be here.

This function is invoked by the runner, so no special feature is needed, we simply need to remove this.

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+        unimplemented!("seccomp interception is not supported on FreeBSD");
+    }
+
+    fn read_proc_self_maps() -> alloc::vec::Vec<core::ops::Range<usize>> {
+        // TODO: this function is not guaranteed to return all allocated pages, as it may
+        // allocate more pages after the mapping file is read. Missing allocated pages may
+        // cause the program to crash when calling `mmap` or `mremap` with the `MAP_FIXED` flag later.
+        // We should either fix `mmap` to handle this error, or let global allocator call this function
+        // whenever it get more pages from the host.
+
```
jaybosamiya-ms: This comment seems identical to the one in Linux userland.  Is this function identical to that?  If so, this comment might be misleading since it may or may not apply to FreeBSD (have you checked)?  If the function has been updated, then again, the comment can be misleading.

You may wish to write a comment saying something like "This function same as the one in `LinuxUserland` and might have similar issues to be resolved"

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+    const EXIT_FAILURE: Self::ExitCode = 1;
+
+    fn exit(&self, code: Self::ExitCode) -> ! {
+        let Self {
+            tun_socket_fd: _,
+            reserved_pages: _,
+        } = self;
+
+        // todo(chuqi): ignore tun device for now
+        // drop::<Option<std::os::fd::OwnedFd>>(tun_socket_fd.write().unwrap().take());
```
jaybosamiya-ms: If you are keeping around the `tun_socket_fd` field, then you should not remove this code.  I think this PR might want to remove the `tun_socket_fd` field anyways, since it is not actually supported at all on FreeBSD within this PR, but also I can see a reason to keep the field for future tun support.  But at the very least, if we are keeping _some_ tun support, then this line should *not* be commented out (in the `None` case, it will simply do a drop, which does not hurt anyways).

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+}
+
+impl litebox::platform::ThreadProvider for FreeBSDUserland {
+    type ExecutionContext = litebox_common_linux::PtRegs;
+    type ThreadArgs = litebox_common_linux::NewThreadArgs<FreeBSDUserland>;
+    type ThreadSpawnError = litebox_common_linux::errno::Errno;
+    type ThreadId = usize;
+
+    #[allow(unused_variables)]
+    #[allow(unused_mut)]
```
jaybosamiya-ms: Prefer using `expect` rather than `allow`, so that it will complain at you once you fix things up and you can actually remove it.

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+    // The uppermost two bits (1<<31, and 1<<30) act as a "lock bit" for the waker (we use two of
+    // them to make it easier to catch accidental integer wrapping bugs more easily, at the cost of
+    // supporting "only" 1-billion waiters being woken up at once), preventing multiple wakers from
+    // running at the same time.
+    //
+    // The lower 30 bits indicate how many waiters the waker wants to wake up. The waiters
+    // themselves will decrement this number as they wake up, but should make sure not to overflow
+    // (this is why we use two bits for the lock bit---to catch implementation bugs of this kind).
+    num_to_wake_up: AtomicU32,
+}
```
jaybosamiya-ms: This is a direct copy of the one in the Linux platform.  And this PR is not using it.  Please don't include comments and code unnecessarily.  Especially with the current `unimplemented!`s, there is no need for `num_to_wake_up` being in this PR.

Also, "more space than absolutely ideal" does not apply, because we don't know what the ideal is on FreeBSD.

Also, "However, the current design makes sure that spurious wake-ups do not actually occur, and that something that is `block`ed can only be woken up by a `wake`." is entirely false, because there is no actual block or wake being done at all.

Generally speaking, my philosophy is "good comment > no comment >> misleading comment".

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+            } | if populate_pages {
+                freebsd_types::MapFlags::MAP_PREFAULT_READ
+            } else {
+                freebsd_types::MapFlags::empty()
+            });
+
+        let ptr = unsafe {
+            syscalls::syscall6(
+                // todo(chuqi): add x86 support
+                syscalls::Sysno::Mmap,
```
jaybosamiya-ms: Does this mean that it fails on x86?  If so, it would be helpful to include a `if cfg!(target_arch = "x86") { unimplemented!() }` or similar.

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+                range.start,
+                range.len(),
+                prot_flags(initial_permissions)
+                    .bits()
+                    .reinterpret_as_unsigned() as usize,
+                map_flags.bits().reinterpret_as_unsigned() as usize,
+                usize::MAX, // -1 for anonymous mapping
+                0,
```
jaybosamiya-ms: Where is the `-1`?  If you intend to use `-1` as the value here, it is probably better written as `-1isize as usize, // anonymous mapping` instead.

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+                })
+                .unwrap(),
+                buf.as_ptr() as usize,
+                buf.len(),
+            )
+        } {
+            Ok(n) => Ok(n),
+            // todo(chuqi): handle EPIPE
+            // Err(syscalls::Errno::EPIPE) => Err(litebox::platform::StdioWriteError::Closed),
+            Err(err) => panic!("unhandled error {err}"),
```
jaybosamiya-ms: There is already a panic for unhandled error. Why is there a special TODO for `EPIPE`?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+    fn alloc(layout: &std::alloc::Layout) -> Option<(usize, usize)> {
+        let size = core::cmp::max(
+            layout.size().next_power_of_two(),
+            // Note `mmap` provides no guarantee of alignment, so we double the size to ensure we
+            // can always find a required chunk within the returned memory region.
+            core::cmp::max(layout.align(), 0x1000) << 1,
+        );
+        unsafe {
+            syscalls::syscall6(
+                syscalls::Sysno::Mmap, // todo(chuqi): add x86 support
```
jaybosamiya-ms: Similar

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
+//! A [LiteBox platform](../litebox/platform/index.html) for running LiteBox on userland FreeBSD.
+
+// Restrict this crate to only work on FreeBSD. For now, we are restricting this to only x86/x86-64
+// FreeBSD, but we _may_ allow for more in the future, if we find it useful to do so.
+#![cfg(all(
+    target_os = "freebsd",
+    any(target_arch = "x86_64", target_arch = "x86")
+))]
```
jaybosamiya-ms: We don't need this on both x86 and x86-64.  A lot of the code in this PR does not work on 32-bit FreeBSD _anyways_ (there are open TODOs).  It is more reasonable to just limit to 64-bit for this.

jaybosamiya-ms: Please also clean up all the comments related to x86 across this PR.

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+    use litebox::platform::ThreadLocalStorageProvider as _;
+
+    use crate::FreeBSDUserland;
+    use litebox::platform::{DebugLogProvider, PageManagementProvider};
+
+    extern crate std;
+
+    // #[test]
+    // fn test_raw_mutex() {
+    //     let mutex = std::sync::Arc::new(super::RawMutex {
```
jaybosamiya-ms: Please don't keep commented-out tests.  If you think there is a test that _currently_ cannot work, then you can mark it as `#[ignore]`; at least that way, the test runner will tell us that there are currently ignored tests, rather than silently being hidden away.

litebox_platform_freebsd_userland/src/syscall_raw.rs
jaybosamiya-ms: If we are keeping auto-generated version, please include the Python file in this repo.  Otherwise, how are future users supposed to maintain this file?

I agree with paring it down to a smaller set of only the ones we need though, there is no reason to have such a massive list.

litebox_platform_freebsd_userland/src/syscall_raw.rs
```diff
@@ -0,0 +1,755 @@
...
+// Since the Rust package "syscalls" does not support FreeBSD,
+// we define the syscall table manually here.
+
+#[repr(i32)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[allow(dead_code)] // Allow unused syscall variants as this is a comprehensive syscall table
```
jaybosamiya-ms: If you want to include reasoning for a lint, you can use the `reason = "<whatever you want>"` syntax with the lint.

As mentioned in the prior comment, you should remove most of these though, so you won't need this lint, but I am mentioning for future reference.

litebox_platform_freebsd_userland/tests/syscalls.rs
jaybosamiya-ms: If so, I don't really see why these tests are added.  These tests can probably be removed.

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: Yes, that could be done.  But _this_ PR should not change things in the Linux userland platform _at all_.  Please back out this particular change entirely.

litebox_platform_multiplex/Cargo.toml
jaybosamiya-ms: There is no need to look at the feature list; the seccomp interception is something triggered by the runner and the runner already knows which platform it is on.

Indeed, if you see the function that calls for interception, it is _already_ a runtime choice on the Linux runner.  For the Linux-on-FreeBSD runner, we can simply not call things (e.g., see what the rewriter backend does), rather than need to use any special conditional compilation for it:
https://github.com/microsoft/litebox/blob/38de8f609698b83ebbc7b65c0add8c278602cde8/litebox_runner_linux_userland/src/lib.rs#L202-L205

litebox_platform_multiplex/src/lib.rs
```diff
@@ -51,6 +59,10 @@ static PLATFORM: once_cell::race::OnceBox<&'static Platform> = once_cell::race::
     clippy::match_wild_err_arm,
     reason = "the platform itself is not Debug thus we cannot use `expect`"
 )]
+#[cfg(any(
+    all(feature = "platform_linux_userland", target_os = "linux"),
+    all(feature = "platform_freebsd_userland", target_os = "freebsd")
+))]
```
jaybosamiya-ms: Same

litebox_platform_multiplex/src/lib.rs
```diff
@@ -63,6 +75,10 @@ pub fn set_platform(platform: &'static Platform) {
 /// # Panics
 ///
 /// Panics if [`set_platform`] has not been invoked before this
+#[cfg(any(
+    all(feature = "platform_linux_userland", target_os = "linux"),
+    all(feature = "platform_freebsd_userland", target_os = "freebsd")
+))]
```
jaybosamiya-ms: Same

litebox_platform_multiplex/src/lib.rs
```diff
@@ -29,14 +29,22 @@ compile_error!(
...
+#[cfg(all(feature = "platform_linux_userland", target_os = "linux"))]
 pub type Platform = litebox_platform_linux_userland::LinuxUserland;
 
+#[cfg(all(feature = "platform_freebsd_userland", target_os = "freebsd"))]
+pub type Platform = litebox_platform_freebsd_userland::FreeBSDUserland;
+
+#[cfg(any(
+    all(feature = "platform_linux_userland", target_os = "linux"),
+    all(feature = "platform_freebsd_userland", target_os = "freebsd")
+))]
```
jaybosamiya-ms: I don't understand why this was added here; are we ever in a state where we don't have at least one of these in place _and_ the negation is not caught by the `compile_error` that already exists?

litebox_platform_multiplex/src/lib.rs
```diff
@@ -29,14 +29,22 @@ compile_error!(
...
-    r##"No platforms specified.  Please enable the feature for the platform you want."##
-);
+#[cfg(not(any(
+    all(feature = "platform_linux_userland", target_os = "linux"),
+    all(feature = "platform_freebsd_userland", target_os = "freebsd")
+)))]
+compile_error!(r##"No platforms specified. Please enable the feature for the platform you want."##);
 
-#[cfg(feature = "platform_linux_userland")]
+#[cfg(all(feature = "platform_linux_userland", target_os = "linux"))]
```
jaybosamiya-ms, line 31: Any reason we explicitly _also_ needed to add the `target_os` part to this?  It seems strange to me that it was needed

litebox_platform_freebsd_userland/src/syscall_raw.rs
Icegrave0391: Shrinking them to a small set of syscalls

litebox_shim_linux/tests/exec.rs
Icegrave0391: merged.

litebox_platform_freebsd_userland/tests/syscalls.rs
Icegrave0391: Removed.

litebox_platform_freebsd_userland/src/freebsd_types/mod.rs
Icegrave0391: My point is that this is a value indicating the bits to shift, so that using demical number is more intuitive.

.gitignore
```diff
@@ -73,7 +73,8 @@ flycheck_*.el
 .LSOverride
 
 # Icon must end with two \r
-Icon
+Icon
+
```
jaybosamiya-ms: I think this change was accidentally done by your editor; the `^M` is explicitly documented for the "two `\r`" on MacOS. Not particularly important, but it is important not to change unrelated things.

.gitignore
```diff
@@ -135,12 +140,7 @@ tags
 [._]*.un~
 
 ### VisualStudioCode ###
-.vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
-!.vscode/extensions.json
-!.vscode/*.code-snippets
+.vscode
```
jaybosamiya-ms: I presume the gdb and vscode changes you've done are due to issues you were having in your own setup?  If so, that should be a separate PR, not rolled up into a big PR.

Fwiw, to set up local gitignores, you can use `.git/info/exclude` if you have a local issue with some files.  In this particular case, I don't really see a downside to updating these, but yeah, should be a separate PR, not included in this one.

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,811 @@
+//! A [LiteBox platform](../litebox/platform/index.html) for running LiteBox on userland FreeBSD.
+
+// Restrict this crate to only work on FreeBSD. For now, we are restricting this to only x86/x86-64
+// FreeBSD, but we _may_ allow for more in the future, if we find it useful to do so.
+#![cfg(all(target_os = "freebsd", target_arch = "x86_64"))]
```
jaybosamiya-ms: Please update the comment to remove the `x86/` from it, to accurately reflect the code below it, thanks!

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,811 @@
...
+    ///
+    /// Panics if the tun device could not be successfully opened.
+    pub fn new(_tun_device_name: Option<&str>) -> &'static Self {
+        let platform = Self {
+            reserved_pages: Self::read_proc_self_maps(),
+        };
+
+        platform.set_init_tls();
+        Box::leak(Box::new(platform))
+    }
```
jaybosamiya-ms: This no longer has a tun device, and it is being ignored, please remove that argument, and remove the panic documentation for it, thanks!

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,811 @@
...
+    inner: std::time::Instant,
+}
+
+impl litebox::platform::Instant for Instant {
+    fn checked_duration_since(&self, earlier: &Self) -> Option<core::time::Duration> {
+        self.inner.checked_duration_since(earlier.inner)
+    }
+}
+
+#[allow(dead_code)]
```
jaybosamiya-ms: Please use `expect` rather than `allow`, thanks!

litebox_platform_freebsd_userland/src/syscall_raw.rs
```diff
@@ -0,0 +1,279 @@
...
+    use super::SyscallTable;
+
+    /// Syscall number alias for compatibility
+    pub(crate) type Sysno = SyscallTable;
+
+    /// Result type for syscalls
+    pub(crate) type SyscallResult = Result<usize, isize>;
+
+    /// Perform a syscall with no arguments
+    #[allow(
```
jaybosamiya-ms: Similar

litebox_platform_freebsd_userland/src/syscall_raw.rs
```diff
@@ -0,0 +1,279 @@
...
+        dead_code,
+        reason = "For comprehension we added all syscall interfaces for now. To be removed."
```
jaybosamiya-ms: Nit: I am not sure that this sentence is grammatically correct.  Grammatically correct alternative: "To be comprehensive, ..."

litebox_platform_multiplex/src/lib.rs
jaybosamiya-ms: Discussed with Chuqi, we need this to prevent the CI from complaining when it attempts its `--all-features` run.

@Icegrave0391, for future reference, especially if a decision is made IRL that is in possible disagreement with a review comment, it is helpful to write a quick comment, rather than just directly marking the comment as resolved; otherwise, there is no future documentation available for why a certain decision is made.  Thanks!

litebox_shim_linux/Cargo.toml
```diff
@@ -18,6 +18,7 @@ elf_loader = { version = "=0.10.0", default-features = false, features = [] }
 thiserror = { version = "2.0.6", default-features = false }
 ringbuf = { version = "0.4.8", default-features = false, features = ["alloc"] }
 syscalls = { version = "0.6", default-features = false }
+arrayvec = { version = "0.7.6", default-features = false }
```
jaybosamiya-ms: I believe this is no longer needed, yes?

litebox_shim_linux/tests/exec.rs
```diff
@@ -102,19 +102,84 @@ void _start() {
...
+                src_path.to_str().unwrap(),
+                path.to_str().unwrap(),
+                true,
+                true,
+            );
+            let hooked_path =
+                std::path::Path::new(dir_path.as_str()).join("hello_exec_nolibc.hooked");
+            (path, hooked_path)
+        }
+    };
```
jaybosamiya-ms: I think the `cfg` conditions on this need to be updated: rather than for freebsd and not-freebsd, instead it should be for not-linux and linux respectively.

litebox_shim_linux/tests/exec.rs
jaybosamiya-ms: This apparently is supposed to be merged, but looking at the diff, it doesn't really seem to be, and is re-doing the work.  Please check the diff for this, thanks!

litebox_platform_freebsd_userland/src/freebsd_types.rs
```diff
@@ -0,0 +1,61 @@
...
+        /// Do not include this mapping in core dumps
+        const MAP_NOCORE = 0x20000;
+        /// Prefault read pages (FreeBSD equivalent of MAP_POPULATE?)
+        const MAP_PREFAULT_READ = 0x40000;
+        /// Don't sync to backing store
+        const MAP_NOSYNC = 0x800;
+        /// Use 2MB super pages if possible
+        const MAP_ALIGNED_SUPER = 0x1000000;
+        /// Align to specific boundary (used with MAP_ALIGNED_SUPER)
+        const MAP_ALIGNMENT_SHIFT = 24;
```
jaybosamiya-ms: This was marked by GitHub as "outdated" due to the file rename and I can't directly connect it here, so I'm bringing back old comments manually:

@jaybosamiya-ms wrote:
> Is this actually intended to be the _decimal_ number 24?  Why not write this out in hex?

jaybosamiya-ms: I see what you mean now.  Basically, `MAP_ALIGNED_SUPER` is `1 << MAP_ALIGNMENT_SHIFT`.  I don't understand why `MAP_ALIGNMENT_SHIFT` is actually stored as a constant here in that case.  It can be trivially re-derived via [`ilog2`](https://doc.rust-lang.org/std/primitive.i32.html#method.ilog2), which the compiler will optimize since it is a constant, but that way we don't need to deal with understanding a separate constant.

I would recommend removing `MAP_ALIGNMENT_SHIFT`

.gitignore
Icegrave0391: Thank you for the reminder. My diff even didn't show the changes

.gitignore
Icegrave0391: Sure. Get it.

litebox_platform_freebsd_userland/src/lib.rs
Icegrave0391: Thanks for the reminder

litebox_platform_freebsd_userland/src/lib.rs
Icegrave0391: Sure, I would keep a placeholder _name (but unused)? The platform multiplexer can have a unified way to perform platform::new("somename")

litebox_shim_linux/Cargo.toml
```diff
@@ -29,12 +29,15 @@ elf_loader = { version = "=0.10.0", default-features = false, features = ["rel"]
...
+platform_freebsd_userland = ["litebox_platform_multiplex/platform_freebsd_userland"]
+platform_linux_userland = ["litebox_platform_multiplex/platform_linux_userland"]
 
 [lints]
 workspace = true
 
 [dev-dependencies]
 arrayvec = { version = "0.7.6", default-features = false }
-litebox_shim_linux = { path = ".", features = ["unstable-testing", "systrap_backend"] }
+litebox_shim_linux = { path = ".", default-features = false, features = ["unstable-testing"] }
```
Icegrave0391: @jaybosamiya-ms @CvvT 
I know here it is better not to edit what you have configured. However, the current setting `features = [..., systrap_backend]` will enforce the `linux_userland` and systrap and I must remove it to enable FreeBSD.

Any suggestions?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,818 @@
...
+            "Should not register more than one syscall_handler"
+        );
+    }
+
+    /// Enable seccomp syscall interception on the platform.
+    ///
+    /// # Panics
+    ///
+    /// FreeBSD does not support seccomp, so this function will panic if called.
+    pub fn enable_seccomp_based_syscall_interception(&self) {
```
Icegrave0391: @jaybosamiya-ms I agree that it is better to remove this method. 

However, simply removing this would cause compiler complaints when I wanted to execute the tests under `litebox_shim_linux`:

https://github.com/microsoft/litebox/blob/c1311b0af9dfbb55ec8bfa6ff71efb493c593517/litebox_shim_linux/tests/common/mod.rs#L93-L95

Icegrave0391:
Thank you @jaybosamiya-ms for detailed suggestions!

I fixed most of them and left two minor comments for you to check.

litebox_platform_freebsd_userland/src/lib.rs
jaybosamiya-ms: Platforms have differing constraints on how they are spun up, this is why the multiplexer does not expose a `new`.  It is the runner's job to actually spin up a platform, and the runner knows _exactly_ which platform it has.  Thus, we don't need a common `new` interface (otherwise, we would have added a trait for it).  Things that _require_ commonality come with traits.

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -0,0 +1,805 @@
...
+pub struct FreeBSDUserland {
+    /// Reserved pages that are not available for guest programs to use.
+    reserved_pages: Vec<core::ops::Range<usize>>,
+}
+
+const SELFPROC_MAPS_PATH: &str = "/proc/curproc/map";
+
+impl FreeBSDUserland {
+    /// Create a new userland-FreeBSD platform for use in `LiteBox`.
+    pub fn new(_name: Option<&str>) -> &'static Self {
```
jaybosamiya-ms: Please see attached comment: https://github.com/microsoft/litebox/pull/195#discussion_r2208930485

litebox_shim_linux/Cargo.toml
jaybosamiya-ms: I briefly mentioned this IRL; as a quick note: I think removing the `systrap_backend` from the `dev-dependencies` is the right move here.  Please let @CvvT confirm here before resolving this comment.

CvvT: Removing this would break the integration test in `litebox_shim_linux`. I just submitted a PR #214 which should fix this.

Icegrave0391:
Updated:

The helloworld test case now is moved to a separate PR: https://github.com/microsoft/litebox/pull/215#issue-3236989250

PR #196, opened 2025-07-08
Add debug print to litebox

CvvT:
Move `log_println` macro from litebox_shim_linux to litebox crate.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -443,9 +443,7 @@ pub(crate) fn sys_get_robust_list(
 mod tests {
     use core::mem::MaybeUninit;
 
-    use litebox::{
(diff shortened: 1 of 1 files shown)


litebox/src/platform/mod.rs
```diff
@@ -15,6 +15,16 @@ use thiserror::Error;
 
 pub use page_mgmt::PageManagementProvider;
 
+#[macro_export]
+macro_rules! log_println {
+    ($platform:expr, $($tt:tt)*) => {{
+        use core::fmt::Write;
```
jaybosamiya-ms: Since we are not using this to impl `Write`, it is cleaner to put the `as _`.  We should also invoke the `use` for `DebugLogProvider` so that it is guaranteed to be in scope.
```suggestion
        use core::fmt::Write as _;
        use $crate::platform::DebugLogProvider as _;
```

PR #198, opened 2025-07-09
Reserve space for trampoline code

CvvT:
As mentioned in #191, the trampoline code must be loaded right after the executable code, so our elf loader tries to `mmap` more memory at the beginning. However, for those libs loaded by `ld`, it is not aware of trampoline code and hence there might be no space for loading trampoline.

Instead of fixing it in the loader, this PR takes a different approach, which modifies `mmap` to make sure that whenever we `mmap` a file with zero file offset and no MAP_FIXED flag, there must be enough memory space (16MB) after the mapping. Note that this is done when searching for available memory regions, it does not actually reserve the 16MB space.

This PR also fixes the interface for `create_mapping` where `PageRange` is decoupled into `Option<NonZeroAddress>` and `NonZeroPageSize`.

### DIFF
--- a/litebox/src/mm/linux.rs
+++ b/litebox/src/mm/linux.rs
@@ -114,17 +114,6 @@ impl<const ALIGN: usize> IntoIterator for PageRange<ALIGN> {
     }
 }
 
-impl<const ALIGN: usize> From<PageRange<ALIGN>>
-    for (NonZeroAddress<ALIGN>, NonZeroPageSize<ALIGN>)
-{
-    fn from(val: PageRange<ALIGN>) -> Self {
-        (
-            NonZeroAddress::new(val.start).unwrap(),
-            NonZeroPageSize::new(val.len()).unwrap(),
-        )
-    }
-}
-
 impl<const ALIGN: usize> PageRange<ALIGN> {
     /// Create a new [`PageRange`].
     ///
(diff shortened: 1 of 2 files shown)


litebox/src/mm/linux.rs
```diff
@@ -115,6 +114,17 @@ impl<const ALIGN: usize> IntoIterator for PageRange<ALIGN> {
...
+    for (NonZeroAddress<ALIGN>, NonZeroPageSize<ALIGN>)
+{
+    fn from(val: PageRange<ALIGN>) -> Self {
+        (
+            NonZeroAddress::new(val.start).unwrap(),
+            NonZeroPageSize::new(val.len()).unwrap(),
+        )
+    }
+}
+
```
jaybosamiya-ms: Minor: I think this particular `From` declaration can lead to confusing indirection to figure out what a particular `.into()` means.  Might be worth considering replacing this with a `start_and_length()` implemented on `PageRange` instead?

litebox/src/mm/linux.rs
```diff
@@ -161,11 +171,38 @@ impl<const ALIGN: usize> NonZeroPageSize<ALIGN> {
...
+    fn add(self, rhs: usize) -> Self::Output {
+        NonZeroPageSize::new(self.size + rhs)
+    }
+}
+
+/// An address that is `ALIGN`-aligned.
+pub struct NonZeroAddress<const ALIGN: usize>(usize);
+
+impl<const ALIGN: usize> NonZeroAddress<ALIGN> {
+    /// Create a new `NonZeroAddress`, if the address is non-zero and aligned.
```
jaybosamiya-ms: It might be helpful to update the doc comment for the struct to also say "non-zero"?

litebox/src/mm/linux.rs
CvvT: Good point! Fixed.

litebox/src/mm/linux.rs
CvvT: Added.

PR #199, opened 2025-07-09
Add RawMutex for FreeBSD

Icegrave0391:
- Added `errno/` to indicate FreeBSD error numbers
- Added RawMutexProvider support for FreeBSD, based on `umtx_op` syscall

### DIFF
--- a/litebox_platform_freebsd_userland/src/lib.rs
+++ b/litebox_platform_freebsd_userland/src/lib.rs
@@ -362,7 +362,6 @@ impl RawMutex {
             ) {
                 Ok(0) => {
                     // Fallthrough: check if spurious.
-                    println!("UMTX_OP_WAIT_UINT returned 0, checking for spurious wake-up");
                 }
                 Err(e) if e == i32::from(errno::Errno::EAGAIN) as isize => {
                     // A wake-up was already in progress when we attempted to wait. Has someone
@@ -388,18 +387,12 @@ impl RawMutex {
             match self.num_to_wake_up.fetch_update(SeqCst, SeqCst, |n| {
                 if n & (1 << 31) == 0 {
                     // No waker in play, do nothing to the value
-                    println!("No waker in play. n = 0x{:x}", n);
                     None
                 } else if n & ((1 << 30) - 1) > 0 {
                     // There is a waker, and there is still capacity to wake up
-                    println!(
-                        "There is a waker, and there is still capacity to wake up. n = 0x{:x}",
-                        n
-                    );
                     Some(n - 1)
                 } else {
                     // There is a waker, but capacity is gone
-                    println!("There is a waker, but capacity is gone. n = 0x{:x}", n);
                     None
                 }
             }) {
@@ -448,13 +441,6 @@ impl litebox::platform::RawMutex for RawMutex {
             // until the other waker is done with their job and brings the value down.
             core::hint::spin_loop();
         }
-
-        println!(
-            "Waking up 0x{:x} threads, setting 'num_to_wake_up' to 0x{:x}",
-            n,
-            n | (0b11 << 30)
-        );
-
         // Now we can actually wake them up using FreeBSD's umtx_op and it always returns 0
         // on success, so we cannot ask the kernel how many were woken up.
         let num_woken_up = match umtx_op_operation_timeout(
@@ -470,11 +456,6 @@ impl litebox::platform::RawMutex for RawMutex {
         // Unlock the lock bits, allowing other wakers to run.
         let remain = n - num_woken_up;
 
-        println!(
-            "Woken up 0x{:x} threads, remaining to wake: 0x{:x}",
-            num_woken_up, remain
-        );
-
         while let Err(v) = self.num_to_wake_up.fetch_update(SeqCst, SeqCst, |v| {
             // Due to spurious or immediate wake-ups (i.e., unexpected wakeups that may decrease `num_to_wake_up`),
             // `num_to_wake_up` might end up being less than expected. Thus, we check `<=` rather than `==`.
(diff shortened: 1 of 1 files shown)


litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -438,6 +578,64 @@ impl litebox::platform::DebugLogProvider for FreeBSDUserland {
...
+// by passing either the address of struct timespec, or its
+// extended variant, struct _umtx_time, as the uaddr2 argu-
+// ment of _umtx_op(). They are distinguished by the uaddr
+// value, which must be equal to the size of the structure
+// pointed to by uaddr2, casted to uintptr_t.
+fn umtx_op_operation_timeout(
+    obj: &AtomicU32,
+    op: freebsd_types::UmtxOpOperation,
+    val: u32,
+    timeout: Option<Duration>,
```
CvvT, line 593: According to https://man.freebsd.org/cgi/man.cgi?query=_umtx_op&sektion=2&n=1, if op is `UMTX_OP_WAIT`, `obj` is a pointer to a variable of type long (not u32?).

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -438,6 +578,64 @@ impl litebox::platform::DebugLogProvider for FreeBSDUserland {
...
+    };
+
+    unsafe {
+        syscalls::syscall5(
+            syscalls::Sysno::UmtxOp,
+            if timeout_spec.is_some() {
+                obj_ptr
+            } else {
+                uaddr
+            },
```
CvvT, line 630: Looks like it's always `obj_ptr`.

litebox_platform_freebsd_userland/src/errno/generated.rs
```diff
@@ -0,0 +1,320 @@
+//! Generated code for the [`super::Errno`] constants.
```
jaybosamiya-ms, line 1: Can you include the code that was used to generate this please, thanks!

litebox_platform_freebsd_userland/src/errno/mod.rs
```diff
@@ -0,0 +1,94 @@
+use thiserror::Error;
+
+mod generated;
+
+/// Linux error numbers
+///
+/// This is a transparent wrapper around Linux error numbers (i.e., `i32`s) intended
+/// to provide some type safety by expecting explicit conversions to/from `i32`s.
```
jaybosamiya-ms, line 8: These are not Linux error numbers, please update the comment

litebox_platform_freebsd_userland/src/freebsd_types.rs
```diff
@@ -60,3 +60,13 @@ bitflags::bitflags! {
         const _ = !0;
     }
 }
+
+/// Operations currently supported by the safer variants of the FreeBSD _umtx_op syscall
+#[repr(i32)]
+#[allow(non_camel_case_types)]
+#[allow(dead_code)] // Allow unused variants as this is a comprehensive operation table
```
jaybosamiya-ms, line 67: allow->expect, and place this comment in the reason

litebox_platform_freebsd_userland/src/freebsd_types.rs
```diff
@@ -60,3 +60,13 @@ bitflags::bitflags! {
         const _ = !0;
     }
 }
+
+/// Operations currently supported by the safer variants of the FreeBSD _umtx_op syscall
+#[repr(i32)]
+#[allow(non_camel_case_types)]
```
jaybosamiya-ms, line 66: Do we need these to be non-camel-case?  This is a Rust enum, please use the correct camel casing for it, thanks!

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -24,8 +25,9 @@ use litebox_common_linux::{ProtFlags, PunchthroughSyscall};
 pub mod syscall_raw;
 use syscall_raw::syscalls;
 
-mod freebsd_types;
+pub mod errno;
```
jaybosamiya-ms, line 28: I don't believe this needs to be `pub mod errno`; I think `mod errno` should suffice, no?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -337,7 +417,67 @@ impl litebox::platform::RawMutex for RawMutex {
...
+        }
+        // Now we can actually wake them up using FreeBSD's umtx_op and it always returns 0
+        // on success, so we cannot ask the kernel how many were woken up.
+        let num_woken_up = match umtx_op_operation_timeout(
+            &self.num_to_wake_up,
+            freebsd_types::UmtxOpOperation::UMTX_OP_WAKE,
+            n,    // number of threads to wake
+            None, // no timeout for wake operations
+        ) {
+            Ok(_) => n,  // todo(chuqi): always assume all were woken up on success returns.
```
jaybosamiya-ms, line 452: Why is this a `todo`?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -337,7 +417,67 @@ impl litebox::platform::RawMutex for RawMutex {
...
+                None
+            }
+        }) {
+            // Confirm that no one has clobbered the lock bits (which would indicate an implementation
+            // failure somewhere).
+            debug_assert_eq!(v >> 30, 0b11, "lock bits should remain unclobbered");
+            core::hint::spin_loop();
+        }
+
+        // Return the number that were actually woken up
```
jaybosamiya-ms, line 479: At least based on prior comments, this comment appears to be directly contradicting them.  Can you please check through and make sure none of the comments are misleading, thanks!

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -438,6 +578,64 @@ impl litebox::platform::DebugLogProvider for FreeBSDUserland {
...
+            .checked_div(TEN_POWER_NINE)
+            .unwrap()
+            .try_into()
+            .unwrap();
+        let tv_nsec = nanos
+            .checked_rem(TEN_POWER_NINE)
+            .unwrap()
+            .try_into()
+            .unwrap();
+        libc::timespec { tv_sec, tv_nsec }
```
jaybosamiya-ms, line 610: You can use `t.as_secs()` and `t.subsec_nanos()` to get the values you want: https://doc.rust-lang.org/std/time/struct.Duration.html#method.subsec_nanos

litebox_platform_freebsd_userland/src/lib.rs
Icegrave0391: good point. a stupid and unnecessary condition here.

PR #200, opened 2025-07-09
add rtld_audit

CvvT:
Following #173 but made a few changes:

1. Remove dependency on libc to break circular dependency
2. Instead of reading syscall entry point from env, it directly reads it from the trampoline code for the main binary.

For now, it only works for x64, will add support for x86 in the next PR.

### DIFF
--- a/litebox_rtld_audit/rtld_audit.c
+++ b/litebox_rtld_audit/rtld_audit.c
@@ -1,4 +1,5 @@
 #define _GNU_SOURCE
+#include <assert.h>
 #include <elf.h>
 #include <link.h>
 #include <stdint.h>
@@ -16,6 +17,7 @@
 #define SYS_mmap 9
 #define SYS_mprotect 10
 #define SYS_munmap 11
+#define SYS_exit_group 231
 #define AT_FDCWD -100
 
 // Linux flags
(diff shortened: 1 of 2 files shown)


litebox_rtld_audit/rtld_audit.c
```diff
@@ -0,0 +1,291 @@
...
+                        uintptr_t *cookie __attribute__((unused))) {
+  syscall_print("[audit] la_objopen called\n", 26);
+  const char *path = map->l_name;
+
+  if (!path || path[0] == '\0') {
+    // main binary should be called first
+    set_syscall_entry(map);
+    syscall_print("[audit] main binary is patched by libOS\n", 40);
+    return 0; // main binary is patched by libOS
+  } else if (strcmp(path, "/lib64/ld-linux-x86-64.so.2") == 0) {
```
wdcui: Minor: Is there a better way to decide it's ld than using this fixed path?

litebox_shim_linux/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+arrayvec = { version = "0.7.6", default-features = false }
```
wdcui, line 7: Do we still need this `arrayvec` for this crate?

CvvT: I'd like to print out some debug message when it's built in debug mode. Cargo does not allow me to enable certain dependency only in debug mode. I suspect it won't be compiled into the binary if it is not used at all (i.e., in release mode).

litebox_rtld_audit/rtld_audit.c
CvvT: Yes, I just took a look at our loader to see how it gets the ld's path. It reads the segment header to get the path.

PR #203, opened 2025-07-10
Support vdso

CvvT:
This PR tries to support VDSO for linux userland, in which case VSDO is already loaded by host kernel and thus we only need to get its base address.

Common Methods to Get the vDSO Address:
1. Using AT_SYSINFO_EHDR from the auxv (auxiliary vector)
2. Reading /proc/self/auxv manually (if avoiding libc)
3. Inspecting the memory map via /proc/self/maps

Since we also need to read /proc/self/maps for reserved memory regions, I chose option 3.

One tricky part is in CI we don't have access to the `/proc/self/map`. I added a patch to use option 1 (call `getauxval`) as a fallback in testing mode. We also need to handle the difference between x64 and x86 because `getauxval` would panic if the given key does not exist (see #142).

### DIFF
--- a/litebox_shim_linux/src/loader/auxv.rs
+++ b/litebox_shim_linux/src/loader/auxv.rs
@@ -63,6 +63,7 @@ pub enum AuxKey {
 
 pub type AuxVec = alloc::collections::btree_map::BTreeMap<AuxKey, usize>;
 
+/// Initialize the auxiliary vector with user information and VDSO address.
 pub fn init_auxv() -> AuxVec {
     let mut aux = AuxVec::new();
     let platform = litebox_platform_multiplex::platform();


litebox_shim_linux/Cargo.toml
```diff
@@ -5,6 +5,8 @@ edition = "2024"
 
 [dependencies]
 bitflags = "2.9.0"
+# libc should only be enabled in tests
```
wdcui: If it's only for tests, we should put it in `[dev-dependencies]`?

CvvT: `[dev-dependencies]` only works for unit test but not integration test. That's why we have a feature `unstable-testing`.

wdcui: I see.

litebox/src/platform/mod.rs
```diff
@@ -585,6 +585,9 @@ pub trait SystemInfoProvider {
     /// The entry point address is typically used by the runtime or kernel to save/restore
     /// execution context and transfer control to the syscall handler.
     fn get_syscall_entry_point(&self) -> usize;
+
+    /// Get the address of the VDSO (Virtual Dynamic Shared Object).
+    fn get_vdso_address(&self) -> Option<usize>;
```
jaybosamiya-ms: We should include some documentation about what `None` means here.  Is it just "Get VDSO if it exists"?

CvvT: Fixed

CvvT:
Update: previously, the loader creates the `auxvec` and initialize it. Now its caller is responsible for creating it and pass that to the loader. Loader only needs to set the keys related to loaded binaries. This allows us to get around the restriction on reading `/proc/self/map` as we could rely on `libc::getauxval` to set vdso in test only.

PR #206, opened 2025-07-11
Handle non-pie binary in audit lib

CvvT:
Just found that we may not get the base address from `ld` if the object is not position independent because it would be loaded at its default position. To address it, we could parse `ld` (which must be PIC) if we fail to parse the main binary.

The code assumes the order of the loading of the binaries
1. main object first
2. `ld`
3. other libs


CvvT:
Looks like I accidentally merge this PR into #200, which is now merged into main.

PR #207, opened 2025-07-15
Initial support for VTL1 user space

sangho2:
This PR adds initial support for VTL1 user space. It introduces a new global data structure (`UserContextMap`) to maintain each user context (`UserContext`) which contains a unique top-level page table to enable address space separation. It also maintains the kernel page table which to support kernel-only functionalities including VTL switch. Also, there is a new trait (`UserSpaceProvider`) to create, delete, and enter the user space as well as load a (dummy) binary into the user space. It also enables 64-bit system calls (`syscall/sysret`), and registers a system call handler to capture every system call request from the user space. Note that other features including handlers for each TA syscall and TA ELF loader will be provided in different PRs.

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -6,7 +6,7 @@
 
 use crate::{
     mshv::{vsm::Vtl0KernelInfo, vtl1_mem_layout::PAGE_SIZE},
-    syscall_callback::SyscallHandler,
+    syscall_entry::SyscallHandler,
     user_context::UserContextMap,
 };
 
@@ -38,7 +38,7 @@ pub mod mm;
 pub mod mshv;
 pub mod ptr;
 
-pub mod syscall_callback;
+pub mod syscall_entry;
 pub(crate) mod user_context;
 
 static CPU_MHZ: AtomicU64 = AtomicU64::new(0);
@@ -348,10 +348,10 @@ impl<Host: HostInterface> LinuxKernel<Host> {
         pt
     }
 
-    /// Register a callback for `syscall_handler` passed down from the shim. This function must be called
+    /// Register `syscall_handler` passed down from the shim. This function must be called
     /// for each core to program its MSRs.
     pub fn register_syscall_handler(syscall_handler: SyscallHandler) {
-        syscall_callback::init(syscall_handler);
+        syscall_entry::init(syscall_handler);
     }
 }
 
--- a/litebox_platform_lvbs/src/syscall_entry.rs
+++ b/litebox_platform_lvbs/src/syscall_entry.rs
@@ -88,7 +88,7 @@ impl SyscallContext {
 
 #[allow(clippy::similar_names)]
 #[allow(unreachable_code)]
-fn syscall_callback(sysnr: u64, ctx: *const SyscallContext) -> isize {
+fn syscall_entry(sysnr: u64, ctx: *const SyscallContext) -> isize {
     let syscall_handler: SyscallHandler = *SYSCALL_HANDLER
         .get()
         .expect("Syscall handler should be initialized");
@@ -145,7 +145,7 @@ fn syscall_callback(sysnr: u64, ctx: *const SyscallContext) -> isize {
 }
 
 #[unsafe(naked)]
-unsafe extern "C" fn syscall_callback_wrapper() {
+unsafe extern "C" fn syscall_entry_wrapper() {
     naked_asm!(
         "push rsp",
         "push r11",
@@ -161,14 +161,14 @@ unsafe extern "C" fn syscall_callback_wrapper() {
         "mov rdi, rax",
         "mov rsi, rsp",
         "and rsp, {stack_alignment}",
-        "call {syscall_callback}",
+        "call {syscall_entry}",
         "add rsp, {register_space}",
         "pop rcx",
         "pop r11",
         "pop rbp",
         "sysretq",
         stack_alignment = const STACK_ALIGNMENT,
-        syscall_callback = sym syscall_callback,
+        syscall_entry = sym syscall_entry,
         register_space = const core::mem::size_of::<SyscallContext>() - core::mem::size_of::<u64>() * NUM_REGISTERS_TO_POP,
     );
 }
@@ -188,8 +188,8 @@ pub(crate) fn init(syscall_handler: SyscallHandler) {
     efer.insert(EferFlags::SYSTEM_CALL_EXTENSIONS);
     unsafe { Efer::write(efer) };
 
-    let syscall_callback_addr = syscall_callback_wrapper as *const () as u64;
-    LStar::write(VirtAddr::new(syscall_callback_addr));
+    let syscall_entry_addr = syscall_entry_wrapper as *const () as u64;
+    LStar::write(VirtAddr::new(syscall_entry_addr));
 
     let rflags = RFlags::INTERRUPT_FLAG;
     SFMask::write(rflags);


litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -322,6 +322,76 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
...
+    /// It preserves the CR3 flags.
+    ///
+    /// # Safety
+    /// The caller must ensure that the page table is valid and maps the entire VTL1 kernel address space.
+    /// Currently, we do not support KPTI-like kernel/user space page table separation.
+    ///
+    /// # Panics
+    /// Panics if the page table is invalid
+    #[allow(clippy::similar_names)]
+    pub(crate) fn switch_address_space(&self) -> PhysFrame {
```
wdcui: Minor: `switch_to_address_space`?

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -322,6 +322,76 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
...
+        let p4_pa = M::va_to_pa(VirtAddr::new(p4_va as u64));
+        PhysFrame::containing_address(p4_pa)
+    }
+
+    /// Deallocate physical frames of all page tables except for the top-level page table.
+    ///
+    /// # Safety
+    /// The caller is expected to unmap all non-page-table pages before calling this function.
+    /// Also, the caller must ensure no page table frame is shared with other page tables.
+    /// TODO: consider merging this function with `Drop` if it is reliable enough.
```
wdcui: Does this function do the same thing as `Drop` below?

litebox_platform_lvbs/src/syscall_handle.rs
wdcui, line 1: Should we call this file "syscall_handler" instead of "syscall_handle"?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -319,6 +325,30 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
+            .is_err()
+        {
+            panic!("Failed to map VTL1 physical memory");
+        }
+
+        pt
+    }
+
+    /// Register `syscall_handler` to handle syscalls. This function must be called for each core.
+    pub fn register_syscall_handler(syscall_handler: syscall_handle::SyscallHandler) {
```
wdcui: A better function name?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -210,34 +212,48 @@ pub fn vtl_switch_loop_entry(platform: Option<&'static crate::Platform>) -> ! {
 /// the VTL1 kernel stack to discard any leftovers (e.g., unwind, panic, ...).
 #[allow(clippy::inline_always)]
 #[inline(always)]
-pub(crate) fn jump_vtl_switch_loop_with_stack_cleanup() -> ! {
+fn jump_vtl_switch_loop_with_stack_cleanup() -> ! {
```
wdcui: Minor: `jump_to_vtl_switch_loop_with_stack_cleanup`?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -210,34 +212,48 @@ pub fn vtl_switch_loop_entry(platform: Option<&'static crate::Platform>) -> ! {
...
             in("rax") stack_top, loop = sym vtl_switch_loop,
+            stack_alignment = const STACK_ALIGNMENT,
             options(noreturn)
         );
     }
 }
 
+/// expose `vtl_switch_loop` to the outside (e.g., the syscall handler)
+#[unsafe(naked)]
+pub(crate) unsafe extern "C" fn jump_vtl_switch_loop() -> ! {
```
wdcui: `jump_to_vtl_swtich_loop`?

litebox_platform_lvbs/src/syscall_handle.rs
```diff
@@ -0,0 +1,202 @@
...
+use core::arch::{asm, naked_asm};
+use x86_64::{
+    VirtAddr,
+    registers::{
+        model_specific::{Efer, EferFlags, LStar, SFMask, Star},
+        rflags::RFlags,
+    },
+};
+
+// Generic x86_64 syscall support with a minor extension for realizing OP-TEE's
```
wdcui: This code can eventually be moved to the shim layer?

litebox_platform_lvbs/src/syscall_handle.rs
```diff
@@ -0,0 +1,202 @@
...
+    );
+
+    // placeholder for the syscall handler
+    let sysret = syscall_handler();
+
+    // TODO: check syscall number and its result to determine whether it should
+    // return to the user space or switch to VTL0
+
+    // placeholder for returning to the user space
+    if sysret != 0 {
```
wdcui: What if `sysret` equals to 0?

litebox_platform_lvbs/src/user_context.rs
```diff
@@ -0,0 +1,320 @@
...
+use litebox::mm::linux::{PageRange, VmFlags};
+use litebox_common_linux::errno::Errno;
+use x86_64::{
+    VirtAddr,
+    registers::{control::Cr3, rflags::RFlags},
+};
+
+// Let us confine the size of the VTL1 user space for now (1 GiB).
+// This would be fine because we only run tiny programs in the VTL1 user space.
+const VTL1_USER_BASE: u64 = 0x1_0000_0000;
```
wdcui: As we discussed today, LiteBox's mm will use the same virtual address space for different user contexts before it adds support for multiple processes.

litebox_platform_lvbs/src/user_context.rs
```diff
@@ -0,0 +1,320 @@
...
+const VTL1_USER_TOP: u64 = 0x1_4000_0000;
+// fixed-size user stack for an OP-TEE TA
+pub const VTL1_USER_STACK_SIZE: usize = 16 * PAGE_SIZE;
+
+/// UserSpace provider trait for OP-TEE Trusted Applications (TAs) with fixed-size heap and stack.
+#[allow(dead_code)]
+pub trait UserSpaceProvider {
```
wdcui: Why do we need a trait for this? Do we expect multiple implementations for this?

litebox_platform_lvbs/src/user_context.rs
```diff
@@ -0,0 +1,320 @@
...
+        rflags: RFlags,
+    ) -> Result<(), Errno>;
+}
+
+/// Data structure to hold user context information.
+pub struct UserContext {
+    pub page_table: crate::mm::PageTable<PAGE_SIZE>,
+    pub rip: VirtAddr,
+    pub rsp: VirtAddr,
+    pub rflags: RFlags,
```
wdcui, line 66: Do we need to store other registers?

litebox_platform_lvbs/src/user_context.rs
```diff
@@ -0,0 +1,320 @@
...
+
+impl UserContextMap {
+    pub fn new() -> Self {
+        UserContextMap {
+            inner: spin::mutex::SpinMutex::new(HashMap::new()),
+        }
+    }
+}
+
+impl<Host: HostInterface> UserSpaceProvider for LinuxKernel<Host> {
```
wdcui: Why do we have `LinuxKernel<Host>` here?

litebox_platform_lvbs/src/user_context.rs
```diff
@@ -0,0 +1,320 @@
...
+            // Since we cannot copy memory pages without mapping, we use the user page table for this memory copy.
+            user_ctx.page_table.switch_address_space();
+            unsafe {
+                core::ptr::copy_nonoverlapping(
+                    (dummy_syscall_fn as *const ()).cast::<u8>(),
+                    u64::try_from(entry_point).unwrap() as *mut u8,
+                    PAGE_SIZE,
+                );
+            }
+            self.page_table.switch_address_space(); // reload the kernel page table
```
wdcui: Why do we switch to the kernel page table before the following operations on the user space page table?

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: `clean_up()` is `MappedPageTable`'s member function to deallocate physical frames of level 1-3 page tables but not the top level one. We still need to separately deallocate the physical frame of the top level page table.

litebox_platform_lvbs/src/lib.rs
sangho2: I borrowed this function name from `litebox_platform_linux_userland` :) We can consider a better name later.

litebox_platform_lvbs/src/syscall_handle.rs
sangho2: `handler` might be better. Since both LiteBox Platform and Shim do have `syscall_handler`, it's a bit confusing to have a name `handler` here.

litebox_platform_lvbs/src/syscall_handle.rs
sangho2: A portion of this file might be, but I think most stuff should be here because this file is mainly implementing architecture/kernel-level syscall logic to capture syscalls and invoke corresponding handlers within a shim. Other platforms do this separation as well. In theory, we could have this logic inside a shim, but this is a bit concerning because it might imply that a shim works with a specific system call handling/interception mechanism.

litebox_platform_lvbs/src/syscall_handle.rs
sangho2: This function is partially implemented. We need to add actual logic later.
Anyway, based on syscall number and return value, it determines whether it should resume the TA's execution (`return` -> `sysretq`) or switch to VTL0 to support some an RPC from a TA to its client or service daemon (within VTL0 userspace).

litebox_platform_lvbs/src/user_context.rs
sangho2: Let me or us consider it more (I'll create an issue). Perhaps we need a function call to reserve a virtual address range for each TA. most TAs are simple enough and do not have strict virtual address requirements (since it doesn't rely on shared libraries).

litebox_platform_lvbs/src/user_context.rs
sangho2: Currently, TAs/utee use the user stack to save/restore all other registers.

litebox_platform_lvbs/src/user_context.rs
sangho2: `UserSpaceProvider` should be able to access page tables and these are stored in `LinuxKernel<Host>`.

litebox_platform_lvbs/src/user_context.rs
sangho2: `load_program` loads a program into user space memory with kernel mode execution. All these are still about populating a user page table. It uses the user page table for a while just for memory copy.

sangho2: Instead of doing this, we can use the ELF loader or similar to do memory copy with user mode execution. One concern is that, if we use the LiteBox's ELF loader for this, we might need to either handle both Linux and OP-TEE system calls or modify the LiteBox's ELF loader to make it use OP-TEE syscalls.

litebox_platform_lvbs/src/user_context.rs
sangho2: I wrote it as a trait because this implementation is like extending the features of the LVBS platform's main struct (`LinuxKernel<Host>`). The SEV-SNP platform/host might need something like this or better one.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
wdcui: I see. Thanks.

litebox_platform_lvbs/src/lib.rs
wdcui: I think it makes sense for the userland to call it `register`.  Here it's more like setting up the syscall on each core?

litebox_platform_lvbs/src/syscall_handle.rs
wdcui: Maybe add a TODO here (to include your comment above)?

litebox_platform_lvbs/src/user_context.rs
wdcui: Got it. Thanks. Maybe add a comment here?

sangho2: Let me do that.

litebox_platform_lvbs/src/user_context.rs
sangho2: `delete_userspace` shares the same problem with `load_program`. `delete_userspace` should unmap/dellocate memory pages assigned to a process through `PageManager` before deallocating any page tables.

litebox_platform_lvbs/src/syscall_handler.rs
```diff
@@ -0,0 +1,209 @@
...
+#[cfg(target_arch = "x86_64")]
+pub(crate) fn init(syscall_handler: SyscallHandler) {
+    SYSCALL_HANDLER.call_once(|| syscall_handler);
+
+    // enable 64-bit syscall/sysret
+    let mut efer = Efer::read();
+    efer.insert(EferFlags::SYSTEM_CALL_EXTENSIONS);
+    unsafe { Efer::write(efer) };
+
+    let dispatcher_addr = syscall_dispatcher_wrapper as *const () as u64;
```
wdcui: What's the difference between syscall_handler and syscall_dispatcher?

sangho2: I've checked `litebox_platform_linux_userland`, `litebox_shim_linux`, and `litebox_common_linux` again. `syscall_handler` and `syscall_dispatcher` are technically the same thing and I should use `syscall_callback` here. Let me fix this.

wdcui: Is `syscall_callback` more like a `syscall_entry` in the platform which then calls the `syscall_handler` in the shim?

wdcui:
Please feel free to merge it.

PR #209, opened 2025-07-15
fix CI

CvvT:


jaybosamiya-ms:
Oh oops, I didn't notice this when I opened #212.  Practically the same change.  Sorry Weiteng I should have checked the PR list before pushing a new PR.

Closing this since it no longer applies.

PR #211, opened 2025-07-15
Provide more data for FS stat

### LINKED ISSUE #68: File status should return dev/ino/uid/gid fields too

jaybosamiya-ms:
Closes #68 

This PR adds uid/gid/dev/ino/rdev/blksize to the `FileStatus` provided by the file systems in LiteBox.  This removes some of the specialization that exists within the shim for stdio, and paves the way towards switching the descriptor interfaces to being fully within LiteBox (as part of the series begun by #180).

---

Note: as a drive-by change, the `FileSystem` trait had become unwieldy with the lack of spaces between functions; this PR also just adds a bunch of newlines there to improve readability.


litebox/src/fs/in_mem.rs
```diff
@@ -51,6 +61,7 @@ impl<Platform: sync::RawSyncPrimitivesProvider> FileSystem<Platform> {
                 group: 1000,
             },
             current_working_dir: "/".into(),
+            unique_id_freshness: 1.into(), // the root dir gets unique ID of 0
```
wdcui, line 64: uniqe ID of `0` or `1`?

litebox/src/fs/layered.rs
```diff
@@ -14,7 +14,11 @@ use super::errors::{
     ChmodError, ChownError, CloseError, FileStatusError, MkdirError, OpenError, PathError,
     ReadError, RmdirError, SeekError, UnlinkError, WriteError,
 };
-use super::{FileStatus, FileType, Mode, OFlags, SeekWhence};
+use super::{FileStatus, FileType, Mode, NodeInfo, OFlags, SeekWhence};
+
+/// Just a random constant that is distinct from other file systems. In this case, it is
+/// `b'Lyrs'.hex()`.
```
wdcui, line 20: This sounds like `liars` ;-)

litebox/src/fs/in_mem.rs
```diff
@@ -13,9 +13,17 @@ use super::errors::{
...
+use super::{FileStatus, Mode, NodeInfo, SeekWhence, UserInfo};
 use crate::utilities::anymap::AnyMap;
 
+/// Just a random constant that is distinct from other file systems. In this case, it is
+/// `b'IMem'.hex()`.
+const DEVICE_ID: usize = 0x494d656d;
+
+/// Block size for file system I/O operations
+// TODO(jayb): Determine appropriate block size
+const BLOCK_SIZE: usize = 0;
```
wdcui, line 25: Does any code use the block size? Should we make it 4096 for now?

litebox/src/fs/in_mem.rs
jaybosamiya-ms: Root gets 0, that's why this is set to 1 rather than 0.

litebox/src/fs/layered.rs
jaybosamiya-ms: Haha, the intention was "layers" :)

litebox/src/fs/in_mem.rs
jaybosamiya-ms: As far as I am aware (@Cvvt can confirm), nothing except stdio uses blocksize. I want it to be 0 so that any place that uses it makes it more noisy.

litebox/src/fs/layered.rs
```diff
@@ -296,6 +303,19 @@ impl<Platform: sync::RawSyncPrimitivesProvider, Upper: super::FileSystem, Lower:
...
+        let mut node_info_lookup = self.node_info_lookup.write();
+        let rdev = node_info.rdev;
+        let new_id = node_info_lookup.len();
+        let ino = *node_info_lookup.entry(node_info).or_insert(new_id);
+        NodeInfo {
+            dev: DEVICE_ID,
+            ino,
+            rdev,
+        }
+    }
```
CvvT, line 335: why not just have one `unique_id_freshness` in Litebox that's shared across fs? Every fs calls `fresh_id` to get a unique id so that the layered fs does not need this convertion.

jaybosamiya-ms: Good point, indeed we can set up a globally-unique freshness counter.  Will update with that, thanks!

litebox/src/fs/in_mem.rs
CvvT: We don't have code that uses this. It should be fine with 0.

jaybosamiya-ms:

jaybosamiya-ms:
@CvvT for a globally unique freshness counter, I started implementing a version of it on the [`jayb/wip/freshness-cleanup`](https://github.com/microsoft/litebox/tree/jayb/wip/freshness-cleanup) branch, built on top of this, but there are complexities that show up regarding long-range effects due to the (currently) not-very-accurate handling of inodes in (say) the tar-ro FS.  We _can_ clean this up, but I don't think it is the most important thing to fix at the moment, and I don't actually expect the lookup table at the layered FS to be a particularly major bottleneck (if it is, we can clean it up later).  If we need to revive it, I already have begun some of the changes on that `jayb/wip/freshness-cleanup` branch, so we will be able to do so then.

Nonetheless, looking into details about that _did_ lead me to figure out that there was a missing element to the layered FS.  I have fixed that in 6295891 which is the only added change since you last reviewed it.

If it seems reasonable to you, I'd like to merge this PR and leave the global-freshness counter thing for later after we fix up an underlying issue in tar-ro (I will open an issue to track it).

PR #214, opened 2025-07-16
Move tests

CvvT:
As mentioned in https://github.com/microsoft/litebox/pull/195#discussion_r2208324798, `litebox_shiim_linux` enables `seccomp` for tests, which is only available on Linux userand and thus breaks other platforms. This PR address it by moving the tests that require `seccomp` to `litebox_runner_linux_userland`.

### DIFF
--- a/litebox_shim_linux/Cargo.toml
+++ b/litebox_shim_linux/Cargo.toml
@@ -29,12 +29,9 @@ syscalls = { version = "0.6", default-features = false }
 elf_loader = { version = "=0.10.0", default-features = false, features = ["rel"] }
 
 [features]
-unstable-testing = []
-systrap_backend = ["litebox_platform_multiplex/systrap_backend"]
 
 [lints]
 workspace = true
 
 [dev-dependencies]
-litebox_shim_linux = { path = ".", features = ["unstable-testing"] }
 spin = { version = "0.9.8", default-features = false, features = ["spin_mutex"] }


litebox_shim_linux/Cargo.toml
```diff
@@ -29,12 +29,9 @@ syscalls = { version = "0.6", default-features = false }
 elf_loader = { version = "=0.10.0", default-features = false, features = ["rel"] }
 
 [features]
-unstable-testing = []
```
wdcui, line 32: You don't need to move these removed lines to `litebox_runner_linux_userland`?

CvvT: I searched this and somehow we don't have any places that use feature. I forgot where we used this before.

CvvT: Turns out we need this feature in #203, but we should be able to get rid of it by refactoring loader's interface. I will update #203 to fix it.

Icegrave0391:
Hey @CvvT,

I saw you moved all test cases from `litebox_shim_linux` to `litebox_runner_linux_userland`.
Doe this mean I have to copy all same testcases to another runner (e.g., FreeBSD's runner) to test all of them？

CvvT:
> Hey @CvvT,
> 
> I saw you moved all test cases from `litebox_shim_linux` to `litebox_runner_linux_userland`. Doe this mean I have to copy all same testcases to another runner (e.g., FreeBSD's runner) to test all of them？

Probably only `test_syscall_rewriter`, the other two requrie `seccomp`. I remember you already that for freebsd: https://github.com/microsoft/litebox/blob/5ab94334163087e1186b89865cde80391b84b0f0/litebox_shim_linux/tests/exec.rs#L151

PR #215, opened 2025-07-16
Add a FreeBSD runner with helloworld testcase

Icegrave0391:
- Added `litebox_runner_linux_on_freebsd_userland`

Under `tests/` directory, we are able to put all testcases that uses `platform_freebsd_userland`. Currently a helloworld (native linux binary) is added to test the loader.

### DIFF
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -985,6 +985,45 @@ where
 
         Ok(socket_handle.socket_metadata.insert(metadata))
     }
+
+    /// Get the [`Events`] for a socket.
+    pub fn check_events(&self, fd: &SocketFd<Platform>) -> Option<Events> {
+        let descriptor_table = self.litebox.descriptor_table_mut();
+        let mut table_entry = descriptor_table.get_entry_mut(fd);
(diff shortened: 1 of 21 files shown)


litebox_runner_linux_on_freebsd_userland/tests/hello.c
```diff
@@ -0,0 +1,14 @@
+#include <stdio.h>
```
CvvT: You don't need this file.

Icegrave0391:
> Where is the source code for the runner?

The runner just has a testcase for a compiled helloworld program with syscall rewriter.

litebox_runner_linux_on_freebsd_userland/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    // Just to make sure cargo sets up the `OUT_DIR` environment variable
+}
```
jaybosamiya-ms, line 3: Do we need the `OUT_DIR` variable for this runner?

jaybosamiya-ms:
Closing since this is missing an actual runner.  There are some separate plans to merge runners together, at which point this will automatically get resolved too.

PR #216, opened 2025-07-16
Enable Multi-threading for FreeBSD userland 

Icegrave0391:
Implemented ThreadProvider for the platform `freebsd_userland`. The test case is under `litebox_shim_linux/src/syscalls/process.rs`.


Icegrave0391:

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -217,14 +217,69 @@ impl litebox::platform::ExitProvider for FreeBSDUserland {
...
+/// Thread start trampoline function for FreeBSD.
+/// This is called by FreeBSD's thr_new and it unpacks the thread arguments.
+extern "C" fn thread_start(arg: *mut ThreadStartArgs) {
+    // SAFETY: The arg pointer is guaranteed to be valid and point to a ThreadStartArgs
+    // that was created via Box::into_raw in spawn_thread.
+    let thread_start_args = unsafe { Box::from_raw(arg) };
+
+    let pt_regs = thread_start_args.pt_regs;
+    let thread_args = thread_start_args.thread_args;
+    let entry_point = thread_start_args.entry_point;
```
jaybosamiya-ms: Any reason you are unpacking these like this?  It won't affect the final assembly by this seems like an unnecessary set of lines of code

litebox_shim_linux/litebox_shim_linux-.core
jaybosamiya-ms, line 1: Why is this file included here?  Please make sure not to accidentally commit binary files, thanks!

litebox_platform_freebsd_userland/src/lib.rs
Icegrave0391: Thanks! Now removed these unnecessary moves

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -233,14 +284,77 @@ impl litebox::platform::ThreadProvider for FreeBSDUserland {
...
 
-    fn terminate_thread(&self, code: Self::ExitCode) -> ! {
-        unimplemented!(
-            "terminate_thread is not implemented for FreeBSD yet. code: {}",
-            code
-        );
+    fn terminate_thread(&self, _code: Self::ExitCode) -> ! {
+        // Use thr_exit to terminate the current thread
+        unsafe {
+            syscalls::syscall1(syscalls::Sysno::ThrExit, core::ptr::null::<()>() as usize)
```
CvvT, line 353: The second argument should be `void *code`?

Icegrave0391: I think this is a common trait and the parameter is `Self::ExitCode` (same as the method in `linux_userland`)?
Anyways for FreeBSD the syscall does not accept such a error code parameter so it is a placeholder.

CvvT: https://man.freebsd.org/cgi/man.cgi?query=thr_exit&sektion=2&n=1

```
If the state argument is not NULL, the location pointed to by the argument  is updated
with an arbitrary non-zero value, and an _[umtx_op(2)](https://man.freebsd.org/cgi/man.cgi?query=umtx_op&sektion=2&apropos=0&manpath=FreeBSD+14.3-RELEASE+and+Ports)
UMTX_OP_WAKE operation is consequently performed on the location.
```

I just took a look. It looks like the meaning of the argument is different from what I thought.

PR #217, opened 2025-07-17
Support madvise

CvvT:
Support syscall `madvise` with command `DontNeed`, which basically frees physical pages without unmap it. We could emulate it by `unmap` and `mmap` it again.

Currently, file-backed mapping is not supported for this syscall and would cause panic.

### DIFF
--- a/litebox/src/mm/linux.rs
+++ b/litebox/src/mm/linux.rs
@@ -337,13 +337,18 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
         range: PageRange<ALIGN>,
     ) -> Result<(), VmemUnmapError> {
         let range: Range<usize> = range.into();
+        // Any unmapped regions in the original range will result in this function returning `DeallocationError::AlreadyUnallocated`
+        // while still resetting all of the existing vmas in the range.
         let unmapped_error = self.vmas.gaps(&range).next().is_some();
         let overlapping_ranges: Vec<(Range<usize>, VmArea)> = self
             .overlapping(range.clone())
             .map(|(r, vma)| (r.clone(), *vma))
             .collect();
         for (r, vma) in overlapping_ranges {
-            assert!(!vma.is_file_backed(), "Cannot reset file-backed mappings");
+            assert!(
+                !vma.is_file_backed(),
+                "resetting file-backed mappings is not supported yet"
+            );
             let start = r.start.max(range.start);
             let end = r.end.min(range.end);
             let new_range = PageRange::new(start, end).unwrap();
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1227,6 +1227,7 @@ pub struct EpollEvent {
     pub data: u64,
 }
 
+#[non_exhaustive]
 #[repr(i32)]
 #[derive(Debug, IntEnum)]
 pub enum MadviseBehavior {
@@ -1718,7 +1719,7 @@ impl<'a, Platform: litebox::platform::RawPointerProvider> SyscallRequest<'a, Pla
             Sysno::madvise => {
                 let behavior: i32 = ctx.syscall_arg(2).reinterpret_as_signed().truncate();
                 let behavior =
-                    MadviseBehavior::try_from(behavior).expect("Invalid madvise behavior");
+                    MadviseBehavior::try_from(behavior).expect("unsupported madvise behavior");
                 SyscallRequest::Madvise {
                     addr: Platform::RawMutPointer::from_usize(ctx.syscall_arg(0)),
                     length: ctx.syscall_arg(1),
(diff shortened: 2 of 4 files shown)


CvvT:
> * d I am not making Submissions in the course of work for my employer

litebox/src/mm/linux.rs
```diff
@@ -304,6 +319,45 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
...
+    /// # Safety
+    ///
+    /// The caller must ensure that the memory contents in the affected region are no longer accessed or
+    /// relied upon. Any pointers or references to the previous contents become invalid.
+    pub(super) unsafe fn reset_pages(
+        &mut self,
+        range: PageRange<ALIGN>,
+    ) -> Result<(), VmemUnmapError> {
+        let range: Range<usize> = range.into();
+        let unmapped_error = self.vmas.gaps(&range).next().is_some();
```
wdcui, line 342: I don't understand why we get the `unmapped_error` here then check it after inserting new mappings?

CvvT: I should add some comment here:

See https://elixir.bootlin.com/linux/v5.19.17/source/mm/madvise.c#L1185
```
/*
 * Walk the vmas in range [start,end), and call the visit function on each one.
 * The visit function will get start and end parameters that cover the overlap
 * between the current vma and the original range.  Any unmapped regions in the
 * original range will result in this function returning -ENOMEM while still
 * calling the visit function on all of the existing vmas in the range.
 * Must be called with the mmap_lock held for reading or writing.
 */
```

litebox/src/mm/linux.rs
```diff
@@ -304,6 +319,45 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
...
+        let overlapping_ranges: Vec<(Range<usize>, VmArea)> = self
+            .overlapping(range.clone())
+            .map(|(r, vma)| (r.clone(), *vma))
+            .collect();
+        for (r, vma) in overlapping_ranges {
+            assert!(!vma.is_file_backed(), "Cannot reset file-backed mappings");
```
jaybosamiya-ms: Rather than "Cannot", it is more "Current unsupported", yes?

litebox_common_linux/src/lib.rs
```diff
@@ -1227,6 +1227,19 @@ pub struct EpollEvent {
     pub data: u64,
 }
 
+#[repr(i32)]
+#[derive(Debug, IntEnum)]
+pub enum MadviseBehavior {
+    /// Normal behavior, no special treatment
```
jaybosamiya-ms, line 1234: `#[non_exhaustive]`?

litebox_common_linux/src/lib.rs
```diff
@@ -1697,6 +1715,16 @@ impl<'a, Platform: litebox::platform::RawPointerProvider> SyscallRequest<'a, Pla
                 pipefd: Platform::RawMutPointer::from_usize(ctx.syscall_arg(0)),
                 flags: litebox::fs::OFlags::from_bits_truncate(ctx.syscall_arg(1).truncate()),
             },
+            Sysno::madvise => {
+                let behavior: i32 = ctx.syscall_arg(2).reinterpret_as_signed().truncate();
+                let behavior =
+                    MadviseBehavior::try_from(behavior).expect("Invalid madvise behavior");
```
jaybosamiya-ms: I don't think it is invalid, just unimplemented, right?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -286,6 +290,53 @@ pub(crate) fn sys_brk(addr: MutPtr<u8>) -> Result<usize, Errno> {
...
+            // Note we do not support shared memory yet, so this is just to discard the pages without removing the mapping.
+            unsafe { litebox_page_manager().reset_pages(addr, len) }.map_err(|err| match err {
+                VmemUnmapError::UnAligned => Errno::EINVAL,
+                VmemUnmapError::UnmapError(e) => match e {
+                    DeallocationError::Unaligned => Errno::EINVAL,
+                    DeallocationError::AlreadyUnallocated => Errno::ENOMEM,
+                    _ => unimplemented!(),
+                },
+            })
+        }
```
jaybosamiya-ms: Conversion to errno is via the `From` declarations usually, e.g., https://github.com/microsoft/litebox/blob/b6a2ea8043b1d706630bb2c9502f1ac3dbc8d72a/litebox_common_linux/src/errno/mod.rs#L154-L162

litebox/src/mm/linux.rs
CvvT: Added some comment here.

litebox/src/mm/linux.rs
CvvT: Fixed

litebox_common_linux/src/lib.rs
CvvT: Added.

litebox_common_linux/src/lib.rs

litebox_shim_linux/src/syscalls/mm.rs
CvvT: Right! Fixed.

PR #218, opened 2025-07-17
Add dummy vdso skeleton for FreeBSD

Icegrave0391:
Added a placeholder trait implementation for FreeBSD's vdso.


jaybosamiya-ms:
@CvvT I think this was missed when https://github.com/microsoft/litebox/pull/203 was merged, because the CI is not actually checking on FreeBSD, do you think we should spin up a FreeBSD CI task too (to make sure such stuff is caught), or do you think it is ok to leave it for now?

PR #219, opened 2025-07-18
Fix LVBS workspace and platform multiplex issues

sangho2:
This mini PR addresses the workspace and platform multiplex issues of `litebox_runner_lvbs` and `litebox_platform_lvbs`. It uses `default-members` to exclude `litebox_runner_lvbs` from default build target. It also uses  `cargo clippy/doc ... --workspace --exclude litebox_platform_lvbs --exclude litebox_runner_lvbs` to exclude LVBS from default clippy/doc target. Additionally, it adds a patch to work around the `x86_64` crate's bug due to a recently introduced restriction: Rust no longer allows functions with custom ABI to return a value.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,8 +80,8 @@ jobs:
           # aren't included in nextest at the moment. See relevant discussion at
           # https://github.com/nextest-rs/nextest/issues/16
 
-  build_and_test_nightly:
-    name: Build and Test LVBS (nightly)
+  build_and_test_lvbs:
+    name: Build and Test LVBS
     runs-on: ubuntu-latest
     env:
       RUSTFLAGS: -Dwarnings
--- a/litebox_platform_lvbs/src/mshv/mem_integrity.rs
+++ b/litebox_platform_lvbs/src/mshv/mem_integrity.rs
@@ -437,7 +437,7 @@ impl RsaPkcs1v15Verifier {
 
 /// This function extracts the module data and signature from a signed kernel module.
 /// A signed kernel module has the following layout:
-/// [module data (ELF)][signature (PKCS#7/DER)][`ModuleSignature`][`MODULE_SIGNATURE_MAGIC`]
+/// <`module_data` (ELF)|`signature_der` (PKCS#7/DER)|`ModuleSignature`|`MODULE_SIGNATURE_MAGIC`>
 fn extract_module_data_and_signature(
     signed_module: &[u8],
 ) -> Result<(&[u8], &[u8]), VerificationError> {
--- a/litebox_platform_lvbs/src/mshv/mod.rs
+++ b/litebox_platform_lvbs/src/mshv/mod.rs
@@ -100,7 +100,7 @@ pub const HV_REGISTER_PENDING_EVENT0: u32 = 0x0001_0004;
 
 pub const HV_SECURE_VTL_BOOT_TOKEN: u8 = 0xdc;
 
-/// VTL call parameters (param[0]: function ID, param[1-3]: parameters)
+/// VTL call parameters (`param[0]`: function ID, `param[1..4]`: parameters)
 pub const NUM_VTLCALL_PARAMS: usize = 4;
 
 pub const VSM_VTL_CALL_FUNC_ID_ENABLE_APS_VTL: u32 = 0x1_ffe0;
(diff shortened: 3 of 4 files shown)


sangho2:
> LGTM. Please wait for Jay's approval before you merge it.

Thanks. I'll make a few minor changes to deal with CI's clippy.

sangho2:
@jaybosamiya-ms `cargo clippy` with `--all-targets` and `--all-features` are annoying. `cfg-if` seems to work for `--all-targets` as you suggested before, but `--all-features` is still problematic due to `abi_x86_interrupt`. One workaround is using `cargo +nightly clippy` in CI but not so sure whether this is a good idea.

sangho2:
`clippy --all-targets --all-features --workspace --exclude litebox_platform_lvbs --exclude litebox_runner_lvbs` seems to work.

litebox_platform_multiplex/src/lib.rs
```diff
@@ -23,16 +23,19 @@ extern crate alloc;
...
+cfg_if::cfg_if! {
+    if #[cfg(all(feature = "platform_linux_userland", target_os = "linux"))] {
+        pub type Platform = litebox_platform_linux_userland::LinuxUserland;
+    } else if #[cfg(all(feature = "platform_freebsd_userland", target_os = "freebsd"))] {
+        pub type Platform = litebox_platform_freebsd_userland::FreeBSDUserland;
+    } else if #[cfg(feature = "platform_lvbs")] {
+        pub type Platform = litebox_platform_lvbs::host::LvbsLinuxKernel;
+    } else {
+        compile_error!(
+            r##"Too many platforms specified. Are you sure you have marked 'default-features = false'?"##
```
jaybosamiya-ms: I don't think "too many" is the correct thing any more (the old one also should have been updated once freebsd was added).  The particular situation here is more subtle due to the addition of the `target_os` stuff and that `cfg_if` actually does an if-elseif-elseif-... style thing.

Anyways, for the message just something like "Hint: you might have forgotten to mark 'default-features = false'." might be sufficient here.

Cargo.toml
```diff
@@ -6,17 +6,26 @@ members = [
...
+    "litebox_platform_linux_userland",
+    "litebox_platform_freebsd_userland",
     "litebox_platform_lvbs",
-    "litebox_runner_lvbs"
+    "litebox_platform_multiplex",
+    "litebox_runner_linux_userland",
+    "litebox_shim_linux",
+    "litebox_syscall_rewriter"
 ]
+# exclude `litebox_runner_lvbs` from default-members because it requires a nightly toolchain.
```
jaybosamiya-ms: Something like "We exclude `litebox_runner_lvbs` from default-members because it _currently_ requires a nightly toolchain (due to X, Y, Z)" might be helpful to provide context for future cleanup here.

wdcui:
@jaybosamiya-ms I added you as a reviewer for this PR.

sangho2:
> Thanks Sangho, generally seems reasonable to me. I've added a couple of minor comments.
> 
> Regarding options: we could either do an (additional) cargo-nightly run on the CI _without_ the excludes (so we are not missing them), or we could switch to not using `--all-features`. Personally, I prefer the additional nightly run, if you could add that in too (as a separate job, maybe `build_and_test_nightly`), that'd be awesome, thanks!

That's good idea. I think we can replace the LVBS build pipeline with a nightly build pipeline. Let me try it.

Cargo.toml
sangho2: Let me add more context.

.github/workflows/ci.yml
```diff
@@ -78,50 +80,43 @@ jobs:
           # aren't included in nextest at the moment. See relevant discussion at
           # https://github.com/nextest-rs/nextest/issues/16
 
-  build_and_test_lvbs:
-    name: Build and Test LVBS
+  build_and_test_nightly:
+    name: Build and Test LVBS (nightly)
```
jaybosamiya-ms: I didn't realize that there were still some LVBS specific things being checked (i.e., the target-specific build).  In that case, I don't think the name `build_and_test_nightly` or `Build and Test LVBS (nightly)` is particularly accurate.  This is also _especially_ true because we are not testing on _actual_ nightly, but on a specific fixed nightly (so for example, we won't be able to detect breakages due to Rust updates sooner than it coming to stable).

We might want to revert just these two lines of change back to their old names (i.e., `build_and_test_lvbs` and `Build and Test LVBS`).  We should keep the rest of the changes here, since it is simplifying things, and running more tests and such.

This will also fix up the CI "required checks" thing without us needing to coordinate things more.

sangho2: make sense. let me revert it.

PR #221, opened 2025-07-22
Fix memory manager

CvvT:
If the memory manger runs in user mode, it is not aware of the memory used by global allocator and litebox itself (which is allocated by the host) and thus may choose an unavailable memory address. This PR makes the host to decide on the address to be allocated.

One tricky problem is how do we make sure the trampoline code can be loaded at the expected address. Previously, we assumed our memory manager knows all allocated memory and thus can pick a suitable address for the binary with extra space after it. One potential solution is to always `mmap` more pages when mapping a file and later on, use `MAP_FIXED` when loading the trampoline code. In this PR, I just added an assertion so that it would panic if loading the trampoline code fails. The chance that the trampoline code collides with global allocator's memory should be small as all the trampoline code is loaded at the beginning.

### DIFF
--- a/litebox/src/mm/tests.rs
+++ b/litebox/src/mm/tests.rs
@@ -28,14 +28,14 @@ impl crate::platform::RawPointerProvider for DummyVmemBackend {
 impl crate::platform::PageManagementProvider<PAGE_SIZE> for DummyVmemBackend {
     fn allocate_pages(
         &self,
-        range: Range<usize>,
+        suggested_range: Range<usize>,
         initial_permissions: crate::platform::page_mgmt::MemoryRegionPermissions,
         can_grow_down: bool,
-        populate_pages: bool,
+        populate_pages_immediately: bool,
         fixed_address: bool,
     ) -> Result<Self::RawMutPointer<u8>, crate::platform::page_mgmt::AllocationError> {
         Ok(TransparentMutPtr {
-            inner: range.start as *mut u8,
+            inner: suggested_range.start as *mut u8,
         })
     }
 


litebox/src/mm/linux.rs
```diff
@@ -382,14 +383,19 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
...
+        fixed_address: bool,
     ) -> Option<Platform::RawMutPointer<u8>> {
         let (start, end) = (range.start, range.end);
         if start < Self::TASK_ADDR_MIN || end > Self::TASK_ADDR_MAX {
             return None;
         }
-        for (r, _) in self.vmas.overlapping(start..end) {
-            let intersection = r.start.max(start)..r.end.min(end);
-            unsafe { self.platform.deallocate_pages(intersection) }.ok()?;
+        if !fixed_address {
```
wdcui: I don't follow the comment and the code here. The comment is about finding a suitable address but the code is about deallocating pages in overlapped ranges? If we deallocate these pages, what happens to the existing vma?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -576,12 +576,16 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Fr
         initial_permissions: MemoryRegionPermissions,
         can_grow_down: bool,
         populate_pages: bool,
+        fixed_address: bool,
```
wdcui, line 707: Should we change the variable name from `range` to `suggested_range` in line 575 (and `populate_pages_immediately`)?

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -354,12 +354,14 @@ pub trait HostInterface {
 impl<Host: HostInterface, const ALIGN: usize> PageManagementProvider<ALIGN> for LinuxKernel<Host> {
     fn allocate_pages(
         &self,
-        range: core::ops::Range<usize>,
+        suggested_range: core::ops::Range<usize>,
         initial_permissions: litebox::platform::page_mgmt::MemoryRegionPermissions,
         can_grow_down: bool,
-        populate_pages: bool,
+        populate_pages_immediately: bool,
+        /* ignored because the suggested address is guaranteed to be available when running in kernel mode */
```
wdcui, line 361: It's guaranteed in the kernel mode because the caller of this function handles vm overlapping?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -291,18 +291,23 @@ fn load_trampoline(trampoline: TrampolineHdr, relo_off: usize, fd: i32) -> usize
...
     let end_addr = (start_addr + trampoline.size).next_multiple_of(0x1000);
     let mut need_copy = false;
-    unsafe {
+    let ret = unsafe {
         ElfLoaderMmap::mmap(
             Some(start_addr),
             end_addr - start_addr,
             elf_loader::mmap::ProtFlags::PROT_READ | elf_loader::mmap::ProtFlags::PROT_WRITE,
-            elf_loader::mmap::MapFlags::MAP_PRIVATE | elf_loader::mmap::MapFlags::MAP_FIXED,
+            elf_loader::mmap::MapFlags::MAP_PRIVATE,
```
wdcui, line 299: Why not setting the `fixed_address` to be true?

litebox/src/mm/linux.rs
CvvT:
```
/// If the inserted range partially or completely overlaps any
/// existing range in the map, then the existing range (or ranges) will be
/// partially or completely replaced by the inserted range.
```

Overlapping mappings should be replaced by deallocating them first and then allocating again.

litebox_platform_freebsd_userland/src/lib.rs
CvvT: Fixed

litebox_platform_linux_kernel/src/lib.rs
CvvT: Yes, if it is `true`, the caller would deallocate the overlapping mapping. If it is false, memory manager would choose an available range with no overlapping. When running in user mode, it is different because memory manager is not aware of the user space memory used by global allocator.

litebox_shim_linux/src/loader/elf.rs
CvvT: There is no guarantee this address is available. `MAP_FIEXED` would replace overlapping mappings. Using `MAP_FIXED_NOREPLACE` is more accurate but we don't support it yet. I would expect this works without using `MAP_FIEXED` due to how the memory manager carefully choose the address. If it doesn't work, as least we can detect it. With `MAP_FIXED`, `mmap` always succeeds.

litebox/src/mm/linux.rs
CvvT: Good catch on the issue. Thanks!

litebox/src/mm/tests.rs
```diff
@@ -32,6 +32,7 @@ impl crate::platform::PageManagementProvider<PAGE_SIZE> for DummyVmemBackend {
         initial_permissions: crate::platform::page_mgmt::MemoryRegionPermissions,
```
wdcui, line 32: Change this variable to `suggested_range`?

CvvT: Fixed

Issue #186, opened 2025-06-28
Fix `test_blocking_eventfd`


CvvT:
The unit test `test_blocking_eventfd` seems to be broken from time to time (i.e., running for too long).

CvvT:
I believe the root cause lies on the implementation of the raw mutex. In pollee and poller, I used raw mutex in the following way:

https://github.com/microsoft/litebox/blob/83f02707487cae912ed4c2105b4a9b7bcb139bd5/litebox_shim_linux/src/event.rs#L254

https://github.com/microsoft/litebox/blob/83f02707487cae912ed4c2105b4a9b7bcb139bd5/litebox_shim_linux/src/event.rs#L265-L268

If `block` happens after `wake_one`, the calling thread would be waiting forever. This is because of the two-queue design we use:

https://github.com/microsoft/litebox/blob/83f02707487cae912ed4c2105b4a9b7bcb139bd5/litebox_platform_linux_userland/src/lib.rs#L655-L670

Future waiters that show up after the `Requeue` operation would wait on `self.num_to_wake_up` instead of `self.inner`, which is not going to be changed.

PR #222, opened 2025-07-22
Fix lock

CvvT:
This is the PR I made last week to fix the issue mentioned in #186. This also contains a test case that could reliably reproduce the issue.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -609,8 +609,8 @@ impl litebox::platform::RawMutex for RawMutex {
             &self.inner,
             FutexOperation::Wake,
             /* number to wake up */ n,
-            0,
-            None,
+            /* val2: ignored */ 0,
+            /* uaddr2: ignored */ None,
             /* val3: ignored */ 0,
         )
         .expect("failed to wake up waiters")
--- a/litebox_shim_linux/src/syscalls/eventfd.rs
+++ b/litebox_shim_linux/src/syscalls/eventfd.rs
@@ -162,7 +162,7 @@ mod tests {
     }
 
     #[test]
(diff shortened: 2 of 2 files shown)


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -648,71 +605,15 @@ impl litebox::platform::RawMutex for RawMutex {
...
-            FutexOperation::Requeue,
+        futex_val2(
+            &self.inner,
+            FutexOperation::Wake,
             /* number to wake up */ n,
-            /* number to requeue */ i32::MAX as u32,
-            Some(&self.num_to_wake_up),
+            0,
+            None,
             /* val3: ignored */ 0,
```
jaybosamiya-ms: It might be helpful to include `/* val2: ignored */` and `/* uaddr2: ignored */` comments here, similar to `val3`, just to keep things more locally readable.

litebox_shim_linux/src/syscalls/eventfd.rs
```diff
@@ -161,6 +161,28 @@ mod tests {
         assert_eq!(ret, u64::MAX - 1);
     }
 
+    #[test]
+    fn test_blocking_eventfd_race_conditions() {
```
jaybosamiya-ms: Nit: naming wise, we don't want to imply that there should be a race condition. Instead, this is a regression test, so possibly: `test_blocking_eventfd_no_race_on_massive_readwrite` or similar?

litebox_platform_linux_userland/src/lib.rs

litebox_shim_linux/src/syscalls/eventfd.rs
CvvT: Sure, fixed.

PR #224, opened 2025-07-22
Update `elf_loader` to 0.12.0: disable lazy binding for Windows

Icegrave0391:
In `elf_loader` v0.12, we can disable the feature `lazy` to enable it on Windows.

Tested the binary loading functionality on FreeBSD.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -520,7 +520,6 @@ impl litebox::platform::RawMutexProvider for LinuxUserland {
     fn new_raw_mutex(&self) -> Self::RawMutex {
         RawMutex {
             inner: AtomicU32::new(0),
-            num_to_wake_up: AtomicU32::new(0),
         }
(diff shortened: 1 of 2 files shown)


litebox_shim_linux/Cargo.toml
```diff
@@ -15,7 +15,7 @@ elf = { version = "0.7.4", default-features = false }
 # https://github.com/weizhiao/rust-elfloader/ causes trouble on the latest
 # released version of elf-loader, thus we limit ourselves to the released
 # version right before it.
-elf_loader = { version = "=0.10.0", default-features = false, features = [] }
+elf_loader = { version = "=0.12.0", default-features = false, features = [] }
```
jaybosamiya-ms: Is the comment above this no longer applicable?  If so, we should remove that comment, and not leave in misleading comments.

Icegrave0391: I'm not sure about that @CvvT

PR #225, opened 2025-07-23
Migrate and fix design of events interfaces into LiteBox crate

jaybosamiya-ms:
Our existing interface was a clean-up/generalization of the one in VSBox. However, we were not perfectly happy with it.  Thus, @CvvT (thanks!) recognized this and implemented a more manageable design into the Linux shim.  However, this meant that we had multiple different interfaces for events, with differing functionalities and interfaces.

This PR _effectively_ is a migration of (a version of) that into the main LiteBox crate. Specifically, it removes the old event-manager design, and instead sets things up purely using observer and polling interfaces.

Along the way, I have not only added documentation, but also cleaned up the interfaces to have more easy-to-use ergonomics and functionality.  Additionally, there was a performance bug that existed in the shim that (in the process of making sure there is a good API specification) that was discovered---this has been fixed too.

I think that the design can still be improved a bit more, but is good enough for now, and already cleans some things up to the point that I would be happy enough to have it be part of the explicit public interface of LiteBox.  

---

Note: This PR also unblocks the _next_ step towards switching the descriptor interfaces to being fully within LiteBox (as part of the series begun by https://github.com/microsoft/litebox/pull/180).

### DIFF
--- a/litebox_shim_linux/src/channel.rs
+++ b/litebox_shim_linux/src/channel.rs
@@ -131,7 +131,10 @@ impl<T> Producer<T> {
             self.endpoint.pollee.wait_or_timeout(
                 None,
                 || self.try_write(buf),
-                || self.check_io_events().contains(Events::OUT),
+                || {
+                    self.check_io_events()
+                        .intersects(Events::OUT | Events::ALWAYS_POLLED)
+                },
             )
         }?)
     }
@@ -240,7 +243,10 @@ impl<T> Consumer<T> {
             self.endpoint.pollee.wait_or_timeout(
                 None,
                 || self.try_read(buf),
-                || self.check_io_events().contains(Events::IN),
+                || {
+                    self.check_io_events()
+                        .intersects(Events::IN | Events::ALWAYS_POLLED)
+                },
             )
         }?)
     }


litebox/src/event/polling.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        // Return immediately if the timeout is zero.
+        if timeout.is_some_and(|d| d.is_zero()) {
+            return Err(TryOpError::TimedOut);
+        }
+
+        let start_time = self.litebox.x.platform.now();
+        let poller = Arc::new(Poller::new(&self.litebox));
+        self.register_observer(Arc::downgrade(&poller) as _, Events::all());
```
CvvT, line 83: Why the filter here is always `Events::all`?

litebox/src/event/polling.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        loop {
+            if !check() {
+                let remaining_time = timeout.map(|t| {
+                    t.saturating_sub(self.litebox.x.platform.now().duration_since(&start_time))
+                });
+                poller
+                    .wait_or_timeout(remaining_time)
+                    .map_err(|TimedOut| TryOpError::TimedOut)?;
+            }
```
CvvT, line 93: Poller does not reset the lock to zero, so it should only call `wait_or_timeout` once (otherwise only the first call would actually wait). We can move lines 82-83 into the loop (before check).

litebox_shim_linux/src/channel.rs
```diff
@@ -24,15 +32,16 @@ const PIPE_BUF: usize = 2;
...
 struct EndPointer<T> {
     rb: litebox::sync::Mutex<Platform, T>,
-    pollee: crate::event::Pollee,
+    pollee: Pollee<Platform>,
     is_shutdown: AtomicBool,
 }
 
 impl<T> EndPointer<T> {
     pub fn new(rb: T, init_events: Events, litebox: &LiteBox<Platform>) -> Self {
+        let _ = init_events; // TODO: The `init_events` were being ignored before, what are we actually supposed to do with them?
```
CvvT: We can remove `init_events`. I don't think it is used anymore.

litebox/src/event/polling.rs
jaybosamiya-ms: Poller _does_ actually set it to 0 using the swap (I noticed this after our IRL discussion, which is why I kept the above lines 82-83 outside): https://github.com/microsoft/litebox/blob/a064bcd083cee097082c99e3f7e53cff157960aa/litebox/src/event/polling.rs#L140-L145

If we move the 82-83 lines down, then we are constantly allocating/deallocating things, which seems unnecessary.

litebox_shim_linux/src/channel.rs
jaybosamiya-ms: Thanks, will remove it.

litebox/src/event/polling.rs
jaybosamiya-ms: Good question: I got rid of the `mask` (from the arguments), with now only a simpler `check` predicate. If we get woken up on other events, we will still only run `check` and if it returns false, we'll immediately go back to waiting.  I don't think this should lead to a perf bottleneck, but do let me know if you believe it might, and we can discuss more IRL.

litebox/src/event/polling.rs
CvvT: Right, I forgot that.

litebox/src/event/polling.rs
CvvT: I see. It should be fine to leave it.

litebox_shim_linux/src/channel.rs
```diff
@@ -135,30 +117,47 @@ impl<T> Producer<T> {
...
-        if self.get_status().contains(OFlags::NONBLOCK) {
+        Ok(if self.get_status().contains(OFlags::NONBLOCK) {
             self.try_write(buf)
         } else {
             self.endpoint.pollee.wait_or_timeout(
-                Events::OUT,
                 None,
                 || self.try_write(buf),
-                || self.check_io_events(),
+                || self.check_io_events().contains(Events::OUT),
```
CvvT: One more thing I just noticed: `check` should also return `true` if there is any error. Here it should be `intersects(Events::OUT | Events::ALWAYS_POLLED )`.

litebox_shim_linux/src/channel.rs
```diff
@@ -221,24 +226,23 @@ impl<T> Consumer<T> {
...
-        if self.get_status().contains(OFlags::NONBLOCK) {
+        Ok(if self.get_status().contains(OFlags::NONBLOCK) {
             self.try_read(buf)
         } else {
             self.endpoint.pollee.wait_or_timeout(
-                Events::IN,
                 None,
                 || self.try_read(buf),
-                || self.check_io_events(),
+                || self.check_io_events().contains(Events::IN),
```
CvvT: Same here.

litebox_shim_linux/src/channel.rs
jaybosamiya-ms: Good point, fixed.

PR #226, opened 2025-07-23
Add the skeleton of Windows userland platform

Icegrave0391:
Finished a basic Windows userland platform + runner's loader tests (for syscall rewriter). The platform currently:
- marks all traits as unimplemented
- does not support a global allocator; it rather uses the default allocator.

### DIFF
--- a/litebox/src/mm/linux.rs
+++ b/litebox/src/mm/linux.rs
@@ -352,7 +352,8 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
             let start = r.start.max(range.start);
             let end = r.end.min(range.end);
             let new_range = PageRange::new(start, end).unwrap();
-            unsafe { self.insert_mapping(new_range, vma, false) }.expect("failed to reset pages");
+            unsafe { self.insert_mapping(new_range, vma, false, true) }
+                .expect("failed to reset pages");
         }
         if unmapped_error {
             Err(VmemUnmapError::UnmapError(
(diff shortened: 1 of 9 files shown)


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -0,0 +1,424 @@
...
+/// The syscall handler passed down from the shim.
+static SYSCALL_HANDLER: std::sync::RwLock<Option<SyscallHandler>> = std::sync::RwLock::new(None);
+
+/// The userland Windows platform.
+///
+/// This implements the main [`litebox::platform::Provider`] trait, i.e., implements all platform
+/// traits.
+pub struct WindowsUserland {
+    /// Reserved pages that are not available for guest programs to use.
+    reserved_pages: Vec<core::ops::Range<usize>>,
```
jaybosamiya-ms: Since this is not being used, I think within the `PageManagementProvider` you should return an `unimplemented!()` and you should not actually be storing this value here (at least as part of this PR).

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -0,0 +1,424 @@
...
+        }
+    }
+}
+
+// A skeleton of a raw mutex for Windows.
+#[expect(dead_code)]
+pub struct RawMutex {
+    // The `inner` is the value shown to the outside world as an underlying atomic.
+    inner: AtomicU32,
+    num_to_wake_up: AtomicU32,
```
jaybosamiya-ms: We don't need `num_to_wake_up` here yet

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -0,0 +1,424 @@
...
+    fn with_thread_local_storage_mut<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&mut Self::ThreadLocalStorage) -> R,
+    {
+        unimplemented!("Windows TLS access not implemented yet");
+    }
+}
+
+#[cfg(test)]
+mod tests {}
```
jaybosamiya-ms: This empty module can be removed

PR #229, opened 2025-07-23
Add sysinfo

CvvT:
Add support for syscall sysinfo. Most of fields are filled with bogus values for now as they are not critical.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -58,7 +58,7 @@ static BOOT_TIME: once_cell::race::OnceBox<<Platform as litebox::platform::TimeP
 pub(crate) fn boot_time() -> &'static <Platform as litebox::platform::TimeProvider>::Instant {
     BOOT_TIME
         .get()
-        .expect("set_platform should have already been called before this point")
+        .expect("litebox() should have already been called before this point")
 }
 
 /// Get the global litebox object


litebox_platform_multiplex/src/lib.rs
```diff
@@ -65,3 +72,14 @@ pub fn platform() -> &'static Platform {
...
+/// Get the `Instant` representing the boot time of the platform.
+///
+/// # Panics
+///
+/// Panics if [`set_platform`] has not been invoked before this
+pub fn boot_time() -> &'static <Platform as litebox::platform::TimeProvider>::Instant {
+    BOOT_TIME
+        .get()
+        .expect("set_platform should have already been called before this point")
+}
```
jaybosamiya-ms: I don't think boot time is something that should sit in the multiplexer.  The multiplexer is meant to be a _trivial_ monomorphization point, we don't really want any other functionality here.  This particular constant can be stored in the shim, no?

litebox_shim_linux/src/syscalls/misc.rs
```diff
@@ -77,6 +80,34 @@ pub(crate) fn sys_uname(buf: crate::MutPtr<litebox_common_linux::Utsname>) -> Re
...
+        sharedram: 0, // We don't support shared memory
+        bufferram: 0,
+        totalswap: 0,
+        freeswap: 0,
+        procs: super::process::NR_THREADS.load(core::sync::atomic::Ordering::Relaxed),
+        pad: 0,
+        totalhigh: 0,
+        freehigh: 0,
+        mem_unit: 1,
+        _f: [0; 20 - 2 * core::mem::size_of::<usize>() - core::mem::size_of::<u32>()],
```
jaybosamiya-ms: `Default::default()` should be equivalent, and you don't need to explicitly re-write this particular constant.

CvvT: Good point! Fixed.

litebox_platform_multiplex/src/lib.rs
CvvT: I see. Moved it back to the shim layer.

litebox_shim_linux/src/lib.rs
```diff
@@ -48,11 +48,28 @@ type LinuxFS = litebox::fs::layered::FileSystem<
...
+
+/// Get the `Instant` representing the boot time of the platform.
+///
+/// # Panics
+///
+/// Panics if [`litebox()`] has not been invoked before this
+pub(crate) fn boot_time() -> &'static <Platform as litebox::platform::TimeProvider>::Instant {
+    BOOT_TIME
+        .get()
+        .expect("set_platform should have already been called before this point")
```
jaybosamiya-ms: Nit: "`litebox` should have already been caled before this point"

CvvT: Good catch! Thanks!

PR #230, opened 2025-07-23
Fix ino for tar fs

CvvT:
`ld` requires each dependency has unique `ino` to distinguish between each other. We can simply use index as `ino` for each entry in the tar file.

### DIFF
--- a/litebox/src/fs/tar_ro.rs
+++ b/litebox/src/fs/tar_ro.rs
@@ -40,6 +40,11 @@ use super::{
 /// `b'Taro'.hex()`.
 const DEVICE_ID: usize = 0x5461726f;
 
+/// TODO(jayb): Replace this proper auto-incrementing inode number storage (although that will
+/// require migrating to the hashmap based tar entry storage). This is ok for now, until something
+/// is actually checking for real inode numbers.
+const TEMPORARY_DEFAULT_CONSTANT_INODE_NUMBER: usize = 0xFACE;
+
 /// Block size for file system I/O operations
 // TODO(jayb): Determine appropriate block size
 const BLOCK_SIZE: usize = 0;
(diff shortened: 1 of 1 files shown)


litebox/src/fs/tar_ro.rs
```diff
@@ -155,6 +150,7 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
         } else {
             // it is a dir
             Ok(self.litebox.descriptor_table_mut().insert(Descriptor::Dir {
+                idx,
```
jaybosamiya-ms: This spot actually causes a silent aliasing of files and directories, which will end up being quite hard to debug in the future when it causes trouble. That's the main reason I used the `TEMPORARY_DEFAULT_CONSTANT_INODE_NUMBER` rather than simply using indices everywhere.

If you want to use incrementing numbers for the files (which is what I expect the loader needs) and keep `TEMPORARY_DEFAULT_CONSTANT_INODE_NUMBER` (and old comment) for all the directories, then I think I'd be ok with merging this.

litebox/src/fs/tar_ro.rs
```diff
@@ -313,35 +309,34 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
-            Some(p) if p.filename().as_str().unwrap() != path => Ok(super::FileStatus {
+            Some((idx, p)) if p.filename().as_str().unwrap() != path => Ok(super::FileStatus {
                 file_type: super::FileType::Directory,
                 mode: DEFAULT_DIR_MODE,
                 size: super::DEFAULT_DIRECTORY_SIZE,
                 owner: owner_from_posix_header(p.posix_header()),
                 node_info: NodeInfo {
                     dev: DEVICE_ID,
-                    ino: TEMPORARY_DEFAULT_CONSTANT_INODE_NUMBER,
+                    ino: idx,
```
jaybosamiya-ms: This is the other spot where silent aliasing would occur if you are directly using indices.

litebox/src/fs/tar_ro.rs
CvvT: Right, directories have no entries in the tar file. Using `TEMPORARY_DEFAULT_CONSTANT_INODE_NUMBER` for dir should work. Let me update and check it. Thanks for the suggestion!

PR #231, opened 2025-07-24
Add TLS for WindowsUserland

Icegrave0391:
Implemented LiteBox's TLS by leveraging Windows default TLS (`gs:`) with a set of Win32 APIs (e.g., `TlsAlloc`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,8 @@ jobs:
       - name: Check out repo
         uses: actions/checkout@v4
       - run: sudo apt update && sudo apt install -y musl-tools
+      - name: Use Node.js
+        uses: actions/setup-node@v4
       - name: "Ratchet: ensure number of 'transmute's in codebase is not increasing"
         run: |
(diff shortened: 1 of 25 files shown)


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -27,16 +31,51 @@ pub type SyscallHandler = fn(litebox_common_linux::SyscallRequest<WindowsUserlan
...
+        let tls_index = unsafe { TlsAlloc() };
+        if tls_index == Win32_Threading::TLS_OUT_OF_INDEXES {
+            Err(unsafe { GetLastError() })
+        } else {
+            Ok(Self {
+                dwtlsindex: tls_index,
+            })
+        }
+    }
+}
```
CvvT, line 49: If we directly use windows API, how do we initialize tls for a new thread. In Linux, tls is initialized for a new thread by `pthread_create`. We may need to directly modify `gs` if there is no API to initialize `tls` for a new thread.

CvvT: According to our discussion, seems like the windows API we use for creating a new thread already takes care of it.

Issue #41, opened 2025-03-19
Introduce randomness source to the platform


jaybosamiya-ms:
We need external non-determinism to come from the platform. Otoh, LiteBox itself should itself contain a CSPRNG (possibly rand-chacha?) that is seeded from the non-determinism coming from the platform, rather than depending _entirely_ on cryptographic security of the platform's randomness provider.

PR #233, opened 2025-07-24
Initial support for OP-TEE system calls

sangho2:
This PR adds initial support for OP-TEE system calls. A new crate, `litebox_common_optee`, defines OP-TEE's TEE system call table and other essential data structures like `TeeObjectInfo`. This separation is mainly for avoiding a cyclic dependency between `litebox_platform_lvbs` and `litebox_shim_optee`. We populate `SyscallRequest` with an essential set of syscall placeholders which are needed to run our examples (actual code will be included in subsequent PRs). We add a few syscall function implementation into `litebox_shim_optee` mostly for demonstration and test purposes.

### DIFF
--- a/litebox_shim_optee/src/syscalls/cryp.rs
+++ b/litebox_shim_optee/src/syscalls/cryp.rs
@@ -1,10 +1,9 @@
+use litebox::utils::rng::FastRng;
 use litebox_common_optee::TeeResult;
 
-// Use RDRAND instruction for now for testing. We should decide whether we want to use this RNG or Rust one.
-// This heavily depends on the source of entropy we have (and whether we can trust it).
-#[cfg(target_arch = "x86_64")]
 pub fn sys_cryp_random_number_generate(buf: &mut [u8]) -> Result<(), TeeResult> {
-    use core::arch::x86_64::_rdrand64_step as rdrand64_step;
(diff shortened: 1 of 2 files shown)


litebox_shim_optee/src/syscalls/cryp.rs
```diff
@@ -0,0 +1,37 @@
+use litebox_common_optee::TeeResult;
+
+// Use RDRAND instruction for now for testing. We should decide whether we want to use this RNG or Rust one.
```
wdcui: I thought we have a random number provider in LiteBox?

sangho2: There is `FastRng` (`litebox/src/utils/rng.rs`). We could use it but it's not cryptographically secure. Perhaps we might be able to ignore the security of nonces for now.

sangho2:
Refer #41.
`litebox_shim_optee` also uses `FastRng` for now. I think using RDRAND or RDSEED might be a reasonable option (if we can trust the processor hardware).

litebox_shim_optee/src/syscalls/cryp.rs
sangho2: revised the function. Now it uses LiteBox's `FastRng`.

wdcui: Thanks!

PR #234, opened 2025-07-24
Add getpid

CvvT:
Support syscall `getpid`

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -293,8 +293,9 @@ pub(crate) fn sys_clone(
     }
 
     let platform = litebox_platform_multiplex::platform();
-    let (credentials, pid) =
-        platform.with_thread_local_storage_mut(|tls| (tls.current_task.credentials.clone(), tls.current_task.pid));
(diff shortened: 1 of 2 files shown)


litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -447,6 +449,11 @@ pub(crate) fn sys_get_robust_list(
     unsafe { head_ptr.write_at_offset(0, head) }.ok_or(Errno::EFAULT)
 }
 
+/// Handle syscall `getpid`.
+pub(crate) fn sys_getpid() -> i32 {
+    litebox_platform_multiplex::platform().with_thread_local_storage_mut(|tls| tls.current_task.pid)
```
wdcui, line 454: Does this mean each thread will have a different `pid`?

CvvT: No, new thread would copy `pid` from old thread:

https://github.com/microsoft/litebox/blob/381ecb5df6b91f477ae2053f9b1d4e36276a99d9/litebox_shim_linux/src/syscalls/process.rs#L296-L314

so, all threads share the same `pid`.

wdcui: Got it. Thanks!

PR #236, opened 2025-07-25
Support capget

CvvT:
Add support for `capget` which simply returns empty capability for now.

### DIFF
--- a/litebox_shim_linux/src/syscalls/misc.rs
+++ b/litebox_shim_linux/src/syscalls/misc.rs
@@ -118,23 +118,23 @@ pub(crate) fn sys_capget(
     let hdr = unsafe { header.read_at_offset(0) }.ok_or(Errno::EFAULT)?;
     match hdr.version {
         _LINUX_CAPABILITY_VERSION_1 => {
-            let cap = litebox_common_linux::CapData {
-                effective: 0,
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/syscalls/misc.rs
```diff
@@ -104,6 +104,62 @@ pub(crate) fn sys_sysinfo() -> litebox_common_linux::Sysinfo {
...
+///
+/// Note we don't support capabilities in LiteBox, so this returns empty capabilities.
+pub(crate) fn sys_capget(
+    header: crate::MutPtr<litebox_common_linux::CapHeader>,
+    data: Option<crate::MutPtr<litebox_common_linux::CapData>>,
+) -> Result<(), Errno> {
+    let hdr = unsafe { header.read_at_offset(0) }.ok_or(Errno::EFAULT)?;
+    match hdr.version {
+        _LINUX_CAPABILITY_VERSION_1 => {
+            let cap = litebox_common_linux::CapData {
```
wdcui: Minor optimization: if data_ptr is None, then we don't need to instantiate `cap`?

CvvT: Fixed

PR #237, opened 2025-07-25
Enable `RawMutex` for `WindowsUserland`

Icegrave0391:
Note: Windows does not tell us how many threads we actually woke up as well. For now, I always return a minimum number between `n` and the actual number of `waiters`.

### DIFF
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -499,7 +499,6 @@ fn do_prefetch_on_range(start: usize, size: usize) {
     });
 }
 
-#[expect(unused, reason = "Will be added for PageManagementProvider soon.")]
 impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for WindowsUserland {
     fn allocate_pages(
         &self,
@@ -509,8 +508,8 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
         populate_pages_immediately: bool,
         _fixed_address: bool,
     ) -> Result<Self::RawMutPointer<u8>, litebox::platform::page_mgmt::AllocationError> {
-        let mut base_addr = suggested_range.start as *mut c_void;
-        let mut size = suggested_range.len();
+        let base_addr = suggested_range.start as *mut c_void;
+        let size = suggested_range.len();
         // TODO: For Windows, there is no MAP_GROWDOWN features so far.
 
         // 1) In case we have a suggested VA range, we first check and deal with the case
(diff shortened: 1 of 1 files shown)


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -242,16 +246,81 @@ impl litebox::platform::RawMutexProvider for WindowsUserland {
...
+            let timeout_ms = match timeout {
+                None => Win32_Threading::INFINITE, // no timeout
+                Some(timeout) => match timeout.checked_sub(start.elapsed()) {
+                    None => {
+                        // Already timed out
+                        break Ok(UnblockedOrTimedOut::TimedOut);
+                    }
+                    Some(remaining_time) => {
+                        let ms = remaining_time.as_millis();
+                        ms.min(u32::MAX as u128) as u32
```
jaybosamiya-ms: This should be `(u32::MAX - 1) as u128` rather than `u32::MAX` because `Win32_Threading::INFINITE == u32::MAX`, so if someone sets a long (but still finite) remaining time, then it'd accidentally mean infinite.  Unlikely to actually be a problem in practice (`u32::MAX` ~= 50 days lol), but nonetheless, would not hurt to add a `assert_eq!(Win32_Threading::INFINITE, u32::MAX);` and then do the subtraction.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -261,7 +330,28 @@ impl litebox::platform::RawMutex for RawMutex {
     }
 
     fn wake_many(&self, n: usize) -> usize {
-        unimplemented!("wake_many is not implemented for Windows yet. n: {}", n);
+        assert!(n > 0, "wake_many should be called with n > 0");
+        let n: u32 = n.try_into().unwrap();
+        let waiting = self.waiter_count.load(SeqCst);
+
+        unsafe {
+            if n as usize == usize::MAX {
```
jaybosamiya-ms: changing this to `n >= waiting` would be probably better?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -261,7 +330,28 @@ impl litebox::platform::RawMutex for RawMutex {
...
+                    Win32_Threading::WakeByAddressSingle(
+                        self.underlying_atomic().as_ptr() as *const c_void
+                    );
+                }
+            }
+        }
+
+        // For windows, the OS kernel does not tell us how many threads were actually woken up,
+        // so we just return the minimum of `n` and the current number of waiters
+        waiting.min(n as usize)
```
CvvT: This may have race condition issue. I wonder if we can just call `WakeByAddressSingle` when n == 1 and call `WakeByAddressAll` when  n > 1. Always return `n` as the result. The current implementation is not accurate anyway.

I want to avoid adding another field into `RawMutex` because in #239 we need to convert `*mut u32` to `*mut RawMutex`. Any thoughts @jaybosamiya-ms

CvvT: It should be fine to leave it as it is for now. Just need to take care of the edge case where `n == 1`.

jaybosamiya-ms: Regarding converting `*mut u32` to a `*mut RawMutex`, we should not be doing it via a raw conversion, and should not be depending on any particular internal shape of `RawMutex`, so for this, I would say it is fine, and we should fix things up in the #239 side.

PR #239, opened 2025-07-28
Support syscall futex

CvvT:
Add support for syscall futex with some limitations:

1. Shared futex is treated as private futex as we don't support multi-processing
2. Only support WAIT/WAKE/WAIT_BITSET
3. For WAIT_BITSET, only support bitmask `FUTEX_BITSET_MATCH_ANY`, which essentially is `WAIT`. Supporting arbitrary bitmask is possible but require some more work.
4. For Linux userland, our `TimeProvider` calls `libc::clock_gettime(CLOCK_MONOTONIC)` (which in turn calls into vDSO). Thus, only support monotonic time but not real time for `timeout` parameter in this PR.

This PR also adds test for running node using rewriter backend. However, the test becomes really slow (5-10min) because rewriting node (>100MB) takes time. Also, it may trigger OOM when I tested it locally. Thus, this is also disabled for now.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -10,10 +10,10 @@ use std::sync::atomic::Ordering::SeqCst;
 use std::time::Duration;
 
 use litebox::fs::OFlags;
-use litebox::platform::UnblockedOrTimedOut;
 use litebox::platform::page_mgmt::MemoryRegionPermissions;
-use litebox::platform::trivial_providers::TransparentMutPtr;
-use litebox::platform::{ImmediatelyWokenUp, RawConstPointer, ThreadLocalStorageProvider};
+use litebox::platform::trivial_providers::{TransparentConstPtr, TransparentMutPtr};
+use litebox::platform::{RawConstPointer, ThreadLocalStorageProvider};
+use litebox::platform::{RawMutexBlockError, UnblockedOrTimedOut};
 use litebox::utils::{ReinterpretSignedExt, ReinterpretUnsignedExt as _, TruncateExt};
 use litebox_common_linux::{CloneFlags, MRemapFlags, MapFlags, ProtFlags, PunchthroughSyscall};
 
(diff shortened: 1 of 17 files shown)


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -525,6 +525,19 @@ impl litebox::platform::RawMutexProvider for LinuxUserland {
...
+    fn try_from(
+        addr: litebox::platform::trivial_providers::TransparentConstPtr<u32>,
+    ) -> Result<Self, Self::Error> {
+        if addr.as_usize() % core::mem::align_of::<RawMutex>() != 0 {
+            return Err(litebox_common_linux::errno::Errno::EINVAL);
+        }
+        Ok(unsafe { &*(addr.as_usize() as *const RawMutex) })
+    }
+}
+
```
jaybosamiya-ms: Fwiw, I don't think this is a good interface for us to expect out of raw mutexes.  This essentially assumes that a raw mutex is equivalent to an atomic u32, which is not always guaranteed to be true.  Indeed, the core design of the raw mutex API _explicitly_ aims to never make this assumption, because I think on the Windows and FreeBSD platforms, it is impossible to set up a true futex-like interface.

We can chat offline as to how we might be able to implement futex API on top of the existing interfaces (or how we can tweak the existing interfaces), but yeah, simply adding this `TryFrom` is a workaround that has non-trivial future-compatibility hazards.

CvvT: I added another trait `UserRawMutex` to litebox. Now each platform can implement this to support futex.

CvvT:
> Thanks, Weiteng! Unfortunately, I don't think I agree with the (current) design of the `UserRawMutex` trait. At a high level, it causes confusion about when one should use `RawMutex` and when one should use `UserRawMutex`; as far as I can tell, it is simply used as a distinction of "cannot support atomic->T" or "can support atomic->T", which is a valid distinction, but I think it leads to fragmentation in functionality above it. Specifically, is makes it unclear what north-functionality a platform would lose out on if it is only able to implement `RawMutex` but not `UserRawMutex`. It also makes it difficult for north functionality to pick whether to use `UserRawMutex` or `RawMutex`.
> 
> I believe we can have a cleaner design, but it will take me at least a few days to implement (I won't be able to get a final version ready until I'm back from DEFCON).
> 
> For now, I think I would suggest _against_ merging this PR because it leads to non-trivial subtle differences (due to the mild variation between `RawMutex` and `UserRawMutex`)---I expect this will end up becoming hard to maintain later on. Instead, I think with the cleaner design I have in mind, we should have a more maintainable implementation, and the interface required from platforms should remain quite small.

I didn't follow what you meant. One is for AtomicU32->T, the other one is for ConstPtr<u32> --> T'. Both share some common functions: `block` and `wake`.

jaybosamiya-ms:
While still using the original platform interfaces, I believe the following design is feasible to implement (and I have an almost fully working implementation of it); there are still some details I am sketching out and finalizing (mostly related to handling the bitsets, and optimizing the implementation), but basically, the shim should be able to use a `litebox::sync::futex::FutexManager` to provide whatever futex functionalities it needs: 
image

I believe this _should_ be sufficient for you to build the same shim functionalities you've described here.  Can you let me know if the functionality I describe above would be sufficient? I will be opening a PR with the above interface soon (once I'm happy with some of the implementation details within it), but yeah, knowing if there are other interface constraints might be good at this point.

CvvT:
> While still using the original platform interfaces, I believe the following design is feasible to implement (and I have an almost fully working implementation of it); there are still some details I am sketching out and finalizing (mostly related to handling the bitsets, and optimizing the implementation), but basically, the shim should be able to use a `litebox::sync::futex::FutexManager` to provide whatever futex functionalities it needs: image
> 
> I believe this _should_ be sufficient for you to build the same shim functionalities you've described here. Can you let me know if the functionality I describe above would be sufficient? I will be opening a PR with the above interface soon (once I'm happy with some of the implementation details within it), but yeah, knowing if there are other interface constraints might be good at this point.

Looks good to me! Thanks!

One minor issue with the comment: 
"Test if the futex word still contains the execpected value. If it does, return ..." -> It should be the opposite: wait if it does.

jaybosamiya-ms:
> One minor issue with the comment

Ah, good catch. The code does the correct thing, but the comment needs an update. There's also a typo in the next line that I've fixed up.

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: This concern is superseded by later comments and then handled by #280

jaybosamiya-ms:
Thanks again Weiteng for your patience with this!  #280 is now merged (thanks for catching the potential subtle issue with the original draft of that during your review).  

This PR should now be considered unblocked, and can now be based on top of the new `FutexManager`.

CvvT:
Replaced by #304

PR #241, opened 2025-07-29
Fix freebsd unit tests

CvvT:
1. Freebsd and Linux have different TASK_ADDR_MIN, and thus move it from `Vmem` to `PageManagementProvider`.
2. Disable some tests for Freebsd as some functionalities are not supported yet.

### DIFF
--- a/litebox/src/platform/page_mgmt.rs
+++ b/litebox/src/platform/page_mgmt.rs
@@ -26,8 +26,12 @@ bitflags::bitflags! {
 /// platform has only one canonical alignment.
 pub trait PageManagementProvider<const ALIGN: usize>: RawPointerProvider {
     /// The lower bound (inclusive) for virtual addresses that can be allocated for task memory.
+    ///
+    /// Note it must be aligned to `ALIGN`.
     const TASK_ADDR_MIN: usize;
     /// The upper bound (exclusive) for virtual addresses that can be allocated for task memory.
+    ///
+    /// Note it must be aligned to `ALIGN`.
     const TASK_ADDR_MAX: usize;
 
     /// Allocates new memory pages at the specified `suggested_range` with the given `initial_permissions`.
--- a/litebox/src/mm/linux.rs
+++ b/litebox/src/mm/linux.rs
@@ -768,6 +768,8 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
             Platform::TASK_ADDR_MIN,
             Platform::TASK_ADDR_MAX - length.as_usize(),
         );
+        debug_assert!(Platform::TASK_ADDR_MIN % ALIGN == 0);
+        debug_assert!(Platform::TASK_ADDR_MAX % ALIGN == 0);
         let last_end = self.vmas.last_range_value().map_or(low_limit, |r| r.0.end);
         if last_end <= high_limit {
             return Some(high_limit);


litebox_shim_linux/cargo-nextest.core
jaybosamiya-ms, line 1: Accidentally included a 73 MB file, please remove, thanks!

litebox/src/mm/tests.rs
```diff
@@ -26,6 +26,12 @@ impl crate::platform::RawPointerProvider for DummyVmemBackend {
 
 #[expect(unused_variables, reason = "dummy/mock backend")]
 impl crate::platform::PageManagementProvider<PAGE_SIZE> for DummyVmemBackend {
+    const TASK_ADDR_MIN: usize = 0x1_0000; // default linux config
+    #[cfg(target_arch = "x86_64")]
+    const TASK_ADDR_MAX: usize = 0x7FFF_FFFF_F000; // (1 << 47) - PAGE_SIZE;
+    #[cfg(target_arch = "x86")]
+    const TASK_ADDR_MAX: usize = 0xC000_0000; // 3 GiB (see arch/x86/include/asm/page_32_types.h)
+
```
jaybosamiya-ms: Do we need to also include a `target_os = "linux"` here? Otherwise, I think some mock tests _may_ fail if we run on FreeBSD, no?

litebox/src/platform/page_mgmt.rs
```diff
@@ -25,6 +25,9 @@ bitflags::bitflags! {
 /// `ALIGN` as a parameter. In the future, this may be changed to an associated constant, since each
 /// platform has only one canonical alignment.
 pub trait PageManagementProvider<const ALIGN: usize>: RawPointerProvider {
+    const TASK_ADDR_MIN: usize;
+    const TASK_ADDR_MAX: usize;
+
```
jaybosamiya-ms: Please add doc comments to these, thanks!

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -398,6 +398,7 @@ mod tests {
         sys_close(fd).unwrap();
     }
 
+    #[cfg(not(feature = "platform_freebsd_userland"))]
```
jaybosamiya-ms, line 402: Does this mean that this test is ok on the Windows platform?  Why is it not ok on the FreeBSD platform?

CvvT: It's just because we didn't implement it for freebsd: https://github.com/microsoft/litebox/blob/9bb35776d10df6ec183595826c2533b3c4d1ef57/litebox_platform_freebsd_userland/src/lib.rs#L762. I added some comment here.

It looks like freebsd does not have an equivalent syscall to support this directly. We can emulate this via `munmap` and `mmap`. I will create an issue for this.

litebox/src/platform/page_mgmt.rs
CvvT: Fixed

litebox_shim_linux/cargo-nextest.core
CvvT: Good catch! Removed.

litebox/src/mm/tests.rs
CvvT: Good point! Fixed.

litebox/src/platform/page_mgmt.rs
```diff
@@ -25,6 +25,11 @@ bitflags::bitflags! {
 /// `ALIGN` as a parameter. In the future, this may be changed to an associated constant, since each
 /// platform has only one canonical alignment.
 pub trait PageManagementProvider<const ALIGN: usize>: RawPointerProvider {
+    /// The lower bound (inclusive) for virtual addresses that can be allocated for task memory.
+    const TASK_ADDR_MIN: usize;
+    /// The upper bound (exclusive) for virtual addresses that can be allocated for task memory.
+    const TASK_ADDR_MAX: usize;
```
jaybosamiya-ms: Brief reminder to add doc comments to say that they must be page aligned, thanks!

CvvT: Right, I forgot it. Thanks!

litebox_shim_linux/src/syscalls/mm.rs
CvvT: See #257

PR #244, opened 2025-07-30
Support multi-threading for WindowsUserland

Icegrave0391:
Add multi-threading support based on 
- CreateThread and trampoline code,
- Platform-maintained `FS` register base and per-thread storage to capture the `FS` base.


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -19,17 +21,63 @@ use litebox_common_linux::PunchthroughSyscall;
...
-        self as Win32_Threading, GetCurrentProcess, TlsAlloc, TlsFree, TlsGetValue, TlsSetValue,
+        self as Win32_Threading, CreateThread, GetCurrentProcess, TlsAlloc, TlsFree, TlsGetValue,
+        TlsSetValue,
     },
 };
 
 extern crate alloc;
 
+/// Per-thread FS base storage structure
+#[derive(Debug, Clone, Copy)]
```
jaybosamiya-ms, line 43: Does this actually need `Clone` and `Copy`?  At least based on the type that this is storing, I don't see a reason it needs either of the two.  If you actually need it, then do let me know here in the comment, rather than changing things.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -98,6 +199,14 @@ impl WindowsUserland {
         };
         platform.set_init_tls();
 
+        // Initialize it's own fs-base (for the main thread)
+        WindowsUserland::init_thread_fs_base();
+
+        // Register the exception handler
+        unsafe {
+            let _ = AddVectoredExceptionHandler(0, Some(exception_handler));
+        }
```
jaybosamiya-ms: As a comment, this is a "what" type comment, rather than "why".  A slightly more helpful comment would be something like "Windows sets fsbase to 0 regularly upon scheduling; we register an exception handler to set fsbase back to a reasonable value whenever we notice that it has become 0".

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -19,17 +21,63 @@ use litebox_common_linux::PunchthroughSyscall;
...
+struct ThreadFsBaseState {
+    /// The current FS base value for this thread
+    fs_base: usize,
+    /// Whether the FS base has been explicitly set by the guest
+    is_set: bool,
+}
+
+impl ThreadFsBaseState {
+    fn new() -> Self {
+        let current_fs_base = unsafe { litebox_common_linux::rdfsbase() };
```
CvvT, line 51: We should probably move this function to `litebox` or another crate as it supposes to be linux specific, but we can leave it for now.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -19,17 +21,63 @@ use litebox_common_linux::PunchthroughSyscall;
...
 
 extern crate alloc;
 
+/// Per-thread FS base storage structure
+#[derive(Debug, Clone, Copy)]
+struct ThreadFsBaseState {
+    /// The current FS base value for this thread
+    fs_base: usize,
+    /// Whether the FS base has been explicitly set by the guest
+    is_set: bool,
```
CvvT: Do we really need this boolean value? Is it always `true`? When read from current fsbase for the main thread, it should be already set by libc. For newly spawned threads, we always set it before the thread returns to guest program.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -849,7 +1075,14 @@ unsafe extern "C" fn syscall_handler(
             syscall_handler(d)
         }
         Err(err) => err.as_neg() as isize,
-    }
+    };
+
+    // Note: For Windows x64, the OS kernel will clear userspace fs register at every
+    // context switch (e.g., task scheduling). Therefore, we must manually restore it
+    // before returning to the guest, as the guest thread's fs register is used for thread-local storage
+    WindowsUserland::restore_thread_fs_base();
```
CvvT: I thought it only resets fs when switching tasks. Context switch (syscall does not switch tasks) should be fine.

Icegrave0391: The problem is that syscall may not be atomic. A lot of syscalls (e.g., mmap, write...) can be preempted. In such cases context switches still happen. For example, during memory allocation, the kernel can schedule to another thread (if I remembered correctly).

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: We are using `std::cell::Cell`: which requires `Copy` and `Clone`:

```rust
thread_local! {
    static THREAD_FS_BASE: Cell<ThreadFsBaseState> = Cell::new(ThreadFsBaseState::new());
}
```

litebox_platform_windows_userland/src/lib.rs
CvvT: If switching task may or may not happen, why not just leave it to the exception handler.

Icegrave0391: Agreed

litebox_platform_windows_userland/src/lib.rs
jaybosamiya-ms: I see, the issue is not `Cell` requires it, but that `Cell::get` requires it.

litebox_platform_windows_userland/src/lib.rs
jaybosamiya-ms: Leaving it here for now is fine I think.  I think fsbase-specific things are going to be Linux-specific anyways, and we do need to refactor things in the future to better handle Linux-specific things in the platform layers, but for now it is fine to leave it here in this PR.

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: removed

PR #245, opened 2025-07-30
Refine the testcase `process` for WindowsUserland

Icegrave0391:
In `litebox_shim_linux`'s syscall testcases, also added TLS content read check for WindowsUserland to make sure it can handle the memory access violation (caused by un-saved `FS` at OS context switches).

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -636,19 +636,18 @@ mod tests {
                     unsafe { current_fs_base.assume_init() },
                     "FS base should match TLS pointer"
                 );
-            }
 
-            // Check the TLS value from FS base
-            let mut fs_0: u8;
-            unsafe {
-                core::arch::asm!("mov {0}, fs:0", out(reg_byte) fs_0);
+                // Check the TLS value from FS base
+                let mut fs_0: u8;
+                unsafe {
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -524,7 +524,7 @@ mod tests {
         sys_arch_prctl(ArchPrctlArg::SetFs(ptr.as_usize())).expect("Failed to restore FS base");
     }
 
-    static mut TLS: [u8; PAGE_SIZE] = [0; PAGE_SIZE];
+    static mut TLS: [u8; PAGE_SIZE] = [1; PAGE_SIZE];
```
jaybosamiya-ms, line 529: You should make a note of why this is a `1`, otherwise, it is just a magic number and will be hard to figure out.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -636,6 +636,20 @@ mod tests {
                 );
             }
 
+            // For Windows: check the TLS value from FS base
+            #[cfg(target_os = "windows")]
```
jaybosamiya-ms: Why is this tested only on Windows?  This should be true on other platforms too, no?

Icegrave0391:
@jaybosamiya-ms Could you kindly check again for the updates?

PR #246, opened 2025-07-30
ELF loader and Linux-userland runner for OP-TEE TAs

sangho2:
This PR adds initial implementation of the ELF loader and the Linux-userland runner for OP-TEE TAs.

This TA ELF loader and `mmap*` functions are largely from `litebox_shim_linux` except that
this loader does not assume a filesystem (everything is in memory) and it invokes the `easy_relocate`
function (because there is no `ld`). It is able to reuse the `mmap*` functions/syscalls in `litebox_shim_linux`
because OP-TEE OS does not directly expose memory management system calls to TAs.
They are for LDELF (OP-TEE's ELF loader) and Pseudo TAs (i.e., kernel-mode TAs). To this end, it does not have to
implement separate `mmap*` functions/syscalls for TAs.

It also introduce a runner for OP-TEE TAs on the Linux userland (`litebox_runner_optee_on_linux_userland`).
This runner is expected to ease development and debugging. Currently, this runner does not support 
end-to-end TA operation (i.e., open, invoke, close). It can load and start a TA and intercept its syscalls, but
its interaction with TAs is incomplete (e.g., it lacks an event loop to control the end-to-end operation) which
will be covered in the next PR.

### DIFF
--- a/litebox_platform_multiplex/Cargo.toml
+++ b/litebox_platform_multiplex/Cargo.toml
@@ -13,7 +13,7 @@ once_cell = { version = "1.20.2", default-features = false, features = ["alloc",
 cfg-if = "1.0.0"
 
 [features]
-default = ["platform_linux_userland"]
+default = ["platform_linux_userland", "litebox_platform_linux_userland/linux_syscall"]
 platform_linux_userland = ["dep:litebox_platform_linux_userland"]
 platform_freebsd_userland = ["dep:litebox_platform_freebsd_userland"]
 platform_windows_userland = ["dep:litebox_platform_windows_userland"]
--- a/litebox_shim_linux/Cargo.toml
+++ b/litebox_shim_linux/Cargo.toml
@@ -8,7 +8,7 @@ arrayvec = { version = "0.7.6", default-features = false }
 bitflags = "2.9.0"
 litebox = { path = "../litebox/", version = "0.1.0" }
 litebox_common_linux = { path = "../litebox_common_linux/", version = "0.1.0" }
-litebox_platform_multiplex = { path = "../litebox_platform_multiplex/", version = "0.1.0", default-features = false }
+litebox_platform_multiplex = { path = "../litebox_platform_multiplex/", version = "0.1.0" }
 once_cell = { version = "1.20.2", default-features = false, features = ["alloc", "race"] }
 elf = { version = "0.7.4", default-features = false }
 elf_loader = { version = "0.12.0", default-features = false, features = [] }


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1341,32 +1349,68 @@ unsafe extern "C" {
...
+                Ok(d) => {
+                    let syscall_handler: SyscallHandler = SYSCALL_HANDLER
+                        .read()
+                        .unwrap()
+                        .expect("Should have run `register_syscall_handler` by now");
+                    syscall_handler(d)
+                }
+                Err(err) => (err.as_neg() as isize).try_into().unwrap(),
+            }
         }
```
CvvT: It feels like the translation from `pt_regs` to `SyscallRequest` should be done in the shim layer where we handle the syscalls. We could change type `SyscallHandler` to 
`pub type SyscallHandler = fn(syscall_number: usize, ctx: &'a PtRegs) -> isize;`
Then, different shim can still share the same code here without using feature configuration.

litebox_shim_optee/src/syscalls/mm.rs
CvvT, line 1: I'm thinking maybe we should move the common code to `litebox_common_linux`. We only need to tweak the interface a little bit. For example for mmap, we can have 
```
pub(crate) fn sys_mmap(
    pm: &PageManager,
    addr: usize,
    len: usize,
    prot: ProtFlags,
    flags: MapFlags,
    fd: i32,
    offset: usize,
) -> Result<MutPtr<u8>, Errno> 
```

litebox_platform_linux_userland/src/lib.rs
sangho2: Let me consider how to simplify this. Another difference between Linux and OP-TEE syscalls is their return type (`long` vs `uint32_t`).

litebox_shim_optee/src/syscalls/mm.rs
sangho2: Agree with you. It would be great if we can remove `mm.rs` from `litebox_shim_optee`.

litebox_platform_linux_userland/src/lib.rs
sangho2: Simplified

litebox_shim_optee/src/syscalls/mm.rs
sangho2: A problem is that `PageManager` requires `Platform` such that we might not be able to move it to `litebox_common_linux` due to cyclic dependency. To avoid this, we might require another crate like `litebox_common_linux_mm`.

CvvT: We can have generic function like the following:

```Rust
pub fn do_mmap<
    Platform: litebox::platform::RawPointerProvider
        + litebox::sync::RawSyncPrimitivesProvider
        + litebox::platform::PageManagementProvider<{ litebox::mm::linux::PAGE_SIZE }>,
>(
    pm: &litebox::mm::PageManager<Platform, { litebox::mm::linux::PAGE_SIZE }>,
    suggested_addr: Option<usize>,
    len: usize,
    prot: ProtFlags,
    flags: MapFlags,
    ensure_space_after: bool,
    op: impl FnOnce(Platform::RawMutPointer<u8>) -> Result<usize, litebox::mm::linux::MappingError>,
) -> Result<Platform::RawMutPointer<u8>, litebox::mm::linux::MappingError> {
```

sangho2: #256 it seems to be working. Of course, we can't do these for file-backed ones, but we don't need to worry about them at least for now.

litebox_platform_linux_userland/src/lib.rs
CvvT: My main concern is that platform should not be aware of the shim layer. Adding another shim layer shouldn't affect the underlying platform. I believe it should be fine to use long as the return type and then do a conversion in optee shim layer. It's fine to leave it for now and we can decide later.

sangho2: Yes. Making `litebox_patform_linux_userland` work with two different `syscall_handler`s results in this messy conditional build (because the shim layer cannot fully handle syscalls by itself). Maybe, we need a separate syscall handler or multiplex crate to avoid this issue.

sangho2: Now we are doing PtRegs conversion at `litebox_shim_optee` (this introduces some perf overhead because our current optee and lvbs crates don't use PtRegs, but it would be fine for now). Speaking of the return type, I am (or was) a bit afraid of sign extension because every OP-TEE syscalls error code sets the MSB (of u32).

PR #247, opened 2025-07-31
Implement `read_dir` for the file systems

jaybosamiya-ms:
This PR introduces `read_dir` to all the file systems.  The layered file system is the most interesting one in terms of implementation, because you need to account for directories with files existing across the layers.

Along the way, this PR also fixes up a minor issue that existed in the `tar-ro` file system where trying to open the root directory would not work (I discovered it when writing tests for `read_dir`).

### DIFF
--- a/litebox/src/fs/layered.rs
+++ b/litebox/src/fs/layered.rs
@@ -4,7 +4,7 @@ use alloc::string::String;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::sync::atomic::{AtomicUsize, Ordering::SeqCst};
-use hashbrown::HashMap;
+use hashbrown::{HashMap, HashSet};
 
 use crate::LiteBox;
 use crate::fd::{InternalFd, TypedFd};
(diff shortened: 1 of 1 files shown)


litebox/src/fs/mod.rs
```diff
@@ -107,6 +108,11 @@ pub trait FileSystem: private::Sealed + FdEnabledSubsystem {
     /// Remove a directory
     fn rmdir(&self, path: impl path::Arg) -> Result<(), RmdirError>;
 
+    /// Read directory entries from a directory file descriptor.
+    ///
+    /// Returns a list of file/directory names (explicitly _not_ including `.` or `..`).
+    fn read_dir(&self, fd: &TypedFd<Self>) -> Result<Vec<DirEntry>, ReadDirError>;
```
CvvT, line 114: I just remembered Linux has a weird behavior about `getdent`, which is not documented in the man page. Similar to regular file that internally maintains an offset, directory also needs to maintain an `offset`. The interface should look like `fn read_dir(&self, fd: &TypedFd<Self>, max_num: usize) -> Result<Vec<DirEntry>, ReadDirError>` in which `max_num` indicates how many entries it should return at most.

jaybosamiya-ms: I think this should be handled on the shim layer, not here.  Specifically, the shim layer can grab the full `Vec`, and store it away in the fd entry metadata if it wants to maintain an offset-based behavior.  I don't think we would want to force the Linux-specific offsetting behavior into the general case.

CvvT: I see. Makes sense

PR #248, opened 2025-07-31
Fix time provider

CvvT:
This is part of the PR #239 that fixes time provider so that we could convert a `Timespec` to `Instant`, which is needed for supporting futex timeout arguments. The `std::time::Instant` does not expose this interface and internally it uses either `Duration` (for windows) or `Timespec` to represent it. I copied and adjusted those code from `libc` to corresponding platform.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -626,7 +626,7 @@ cfg_if::cfg_if! {
     }
 }
 
-/// timespec from [Linux](https://elixir.bootlin.com/linux/v5.19.17/source/include/uapi/linux/time.h#L11)
+/// timespec from [Linux](https://elixir.bootlin.com/linux/v5.19.17/source/include/uapi/linux/time_types.h#L7)
 #[derive(Debug, Clone, Copy, PartialOrd, PartialEq, Eq)]
 #[repr(C)]
 pub struct Timespec {
--- a/litebox_platform_freebsd_userland/src/lib.rs
+++ b/litebox_platform_freebsd_userland/src/lib.rs
@@ -540,8 +540,7 @@ impl litebox::platform::TimeProvider for FreeBSDUserland {
         unsafe { libc::clock_gettime(libc::CLOCK_MONOTONIC, t.as_mut_ptr()) };
         let t = unsafe { t.assume_init() };
         Instant {
-            inner: std::time::Instant::now(),
-            #[allow(clippy::useless_conversion, reason = "conversion is needed for 32bit")]
+            #[cfg_attr(target_arch = "x86_64", expect(clippy::useless_conversion))]
             inner: litebox_common_linux::Timespec {
                 tv_sec: i64::from(t.tv_sec),
                 tv_nsec: u64::from(t.tv_nsec.reinterpret_as_unsigned()),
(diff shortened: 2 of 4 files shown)


litebox_common_linux/src/lib.rs
```diff
@@ -623,16 +623,54 @@ cfg_if::cfg_if! {
...
+/// timespec from [Linux](https://elixir.bootlin.com/linux/v5.19.17/source/include/uapi/linux/time.h#L11)
+#[derive(Debug, Clone, Copy, PartialOrd, PartialEq, Eq)]
+#[repr(C)]
+pub struct Timespec {
+    /// Seconds.
+    pub tv_sec: i64,
+
+    /// Nanoseconds. Must be less than 1_000_000_000.
+    pub tv_nsec: u64,
+}
```
jaybosamiya-ms: Timespec from that link actually appears to be different on 32-bit and on 64-bit: it essentially boils down to a pair of `long`s which I think is equivalent to a pair of `isize`s in Rust, no?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -536,19 +536,32 @@ impl litebox::platform::TimeProvider for FreeBSDUserland {
     type Instant = Instant;
 
     fn now(&self) -> Self::Instant {
+        let mut t = core::mem::MaybeUninit::<libc::timespec>::uninit();
+        unsafe { libc::clock_gettime(libc::CLOCK_MONOTONIC, t.as_mut_ptr()) };
+        let t = unsafe { t.assume_init() };
         Instant {
-            inner: std::time::Instant::now(),
+            #[allow(clippy::useless_conversion, reason = "conversion is needed for 32bit")]
```
jaybosamiya-ms: Minor: something like `#[cfg_attr(target_arch = "x86_64", expect(clippy::useless_conversion))]` would both document and check this more accurately.  Not particularly important to update this, just mentioning this for future situations where you want to specifically have a warning only on one of many configurations.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -398,19 +400,33 @@ impl litebox::platform::TimeProvider for WindowsUserland {
...
-    inner: std::time::Instant,
+    inner: core::time::Duration,
 }
 
 impl litebox::platform::Instant for Instant {
     fn checked_duration_since(&self, earlier: &Self) -> Option<core::time::Duration> {
-        self.inner.checked_duration_since(earlier.inner)
+        // On windows there's a threshold below which we consider two timestamps
+        // equivalent due to measurement error. For more details + doc link,
+        // check the docs on epsilon.
```
jaybosamiya-ms: Nit: `[epsilon](perf_counter::PerformanceCounterInstant::epsilon)` will also set up the correct link for rustdoc

litebox_common_linux/src/lib.rs
CvvT: My bad. The correct link should be https://elixir.bootlin.com/linux/v5.19.17/source/include/uapi/linux/time_types.h#L7

litebox_platform_freebsd_userland/src/lib.rs
CvvT: Cool, nice to learn this. Thanks!

litebox_platform_windows_userland/src/lib.rs
CvvT: Fixed. Thanks!

PR #249, opened 2025-07-31
Fix rewriter

CvvT:
The trampoline section is different from the trampoline code, which does not have to reside in the executable segment. In some rare cases where the executable segment has no enough space, we need to look for a different segment.

### DIFF
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -71,6 +71,9 @@ pub const TRAMPOLINE_SECTION_NAME_PREFIX: &str = ".trampolineLB";
 /// to know that they have a trampoline that satisfies the expected version.
 pub const TRAMPOLINE_SECTION_NAME: &str = ".trampolineLB0";
 
+/// The size of the trampoline section, which includes the magic number, the trampoline address, and the size of the trampoline data.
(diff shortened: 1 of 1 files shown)


litebox_syscall_rewriter/src/lib.rs
```diff
@@ -133,20 +133,40 @@ pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<usize>) -> R
...
+                    return false;
+                }
+                // ensure the segment has enough space for the trampoline section (0x18 bytes)
+                let end_offset = seg[0].p_offset + seg[0].p_filesz;
+                if seg[1].p_type == object::elf::PT_LOAD && seg[1].p_offset - end_offset < 0x18 {
+                    return false;
+                }
+                if end_offset.next_multiple_of(0x1000) - end_offset < 0x18 {
+                    return false;
+                }
```
jaybosamiya-ms: Minor: just to prevent trouble later in the future, we should place this 0x18 into a `const TRAMPOLINE_SECTION_SIZE: usize = 0x18;`, and also add an assertion after making `trampoline_vec` that says `assert_eq!(trampoline_vec.len(), TRAMPOLINE_SECTION_SIZE);`.

CvvT: Good point! Fixed.

PR #251, opened 2025-08-01
Support flag O_DIRECTORY for open

CvvT:
```
O_DIRECTORY
              If pathname is not a directory, cause the open to fail.
              This flag was added in Linux 2.1.126, to avoid denial-of-
              service problems if [opendir(3)](https://www.man7.org/linux/man-pages/man3/opendir.3.html) is called on a FIFO or tape
              device.
```

Note all tests were generated by copilot with minor fix.

### DIFF
--- a/litebox/src/fs/tar_ro.rs
+++ b/litebox/src/fs/tar_ro.rs
@@ -144,10 +144,10 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
         }
         assert!(flags.contains(OFlags::RDONLY));
         if entry.filename().as_str().unwrap() == path {
+            // it is a file
             if flags.contains(OFlags::DIRECTORY) {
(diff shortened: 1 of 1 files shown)


litebox/src/fs/tar_ro.rs
```diff
@@ -144,6 +144,9 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
         }
         assert!(flags.contains(OFlags::RDONLY));
         if entry.filename().as_str().unwrap() == path {
+            if flags.contains(OFlags::DIRECTORY) {
+                return Err(OpenError::PathError(PathError::ComponentNotADirectory));
+            }
             // it is a file
```
jaybosamiya-ms: Nit: this comment should be before the added flag check

CvvT: Fixed. Thanks!

PR #252, opened 2025-08-01
Support O_NDELAY

CvvT:
Support O_NDELAY for all FS except stdio.


litebox/src/fs/devices/stdio.rs
```diff
@@ -123,6 +124,7 @@ impl<Platform: crate::sync::RawSyncPrimitivesProvider + crate::platform::StdioPr
         if open_directory {
             return Err(OpenError::PathError(PathError::ComponentNotADirectory));
         }
+        assert!(!nodelay, "NDELAY is not supported for stdio streams");
```
jaybosamiya-ms: This is _currently_ unsupported, but might be supported in the future, yes?  In that case, the better way to document it would be a:
```rs
if nodelay {
  unimplemented!()
}
```

CvvT: Good point! Fixed now. Thanks!

PR #255, opened 2025-08-04
Remove `target=x86_64-unknown-linux-musl` when using syscall_rewriter 

Icegrave0391:
We can use default gnu as the building target (for syscall_rewriter and rtld.so).


jaybosamiya-ms:
Discussed IRL: please update the GitHub CI action to remove the musl part, thanks!

PR #258, opened 2025-08-05
`litebox_runner_linux_on_windows_userland`: Support static / dynamic linking with syscall_rewriter for WindowsUserland

Icegrave0391:
- Added testcases
- Added litebox_runner_linux_on_windows
- Supported static + dynamic linking binary (helloworld) with syscall rewriter


CvvT:
#241 is merged to main.

litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+            );
+            // Always use root privileges for directory creation to avoid permission issues
+            in_mem.with_root_privileges(|fs| {
+                fs.mkdir(path.to_str().unwrap(), mode_and_user.0).unwrap();
+                if mode_and_user.1 != 0 {
+                    // This file is owned by a non-root user, so we need to set the ownership to our default user
+                    fs.chown(path.to_str().unwrap(), Some(1000), Some(1000))
+                        .unwrap();
+                }
+            });
```
CvvT: litebox only accepts unix style path. I implemented a fix for it: https://github.com/microsoft/litebox/commit/5acecfd458d65eee4ddbfc940bd9a3923079350e

litebox_runner_linux_on_windows_userland/tests/test-bins/hello_world_dyn.hooked
CvvT, line 1: We don't need this file (and other *.hooked files) ?

litebox_runner_linux_on_windows_userland/tests/loader.rs
```diff
@@ -145,7 +145,7 @@ fn test_syscall_rewriter() {
     // Use the already compiled executable from the tests folder (same dir as this file)
     let mut test_dir = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
     test_dir.push("tests/test-bins");
-    let path = test_dir.join("hello_exec_nolibc");
+    let path: std::path::PathBuf = test_dir.join("hello_exec_nolibc");
     let hooked_path = test_dir.join("hello_exec_nolibc.hooked");
```
CvvT: We probably do not need this test `test_syscall_rewriter` as we have `test_static_linked_prog_with_rewriter`.  We used `hello_exec_nolibc` before just because some syscalls were not supported back then. It's fine to leave it here.

Icegrave0391: Good point.

litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+
+use anyhow::{Result, anyhow};
+use clap::Parser;
+use litebox::LiteBox;
+use litebox::fs::FileSystem as _;
+use litebox_platform_multiplex::Platform;
+use std::os::windows::fs::MetadataExt;
+use std::path::{Component, PathBuf};
+
+/// Get file permissions and owner ID in a cross-platform way
```
jaybosamiya-ms: Minor: this function is not cross platform, but instead converting Windows permissions to litebox permissions, yes?

litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+        // Directories need full permissions to allow creating subdirectories and files
+        // Even if marked read-only, we need write access for directory operations
+        mode |= litebox::fs::Mode::RWXU | litebox::fs::Mode::RWXG | litebox::fs::Mode::RWXO;
+    } else if !is_readonly {
+        // If not read-only, grant full permissions to user/group/other
+        mode |= litebox::fs::Mode::RWXU | litebox::fs::Mode::RWXG | litebox::fs::Mode::RWXO;
+    } else {
+        // If read-only file, grant only read and execute permissions
+        mode |= litebox::fs::Mode::RUSR | litebox::fs::Mode::RGRP | litebox::fs::Mode::ROTH;
+        mode |= litebox::fs::Mode::XUSR | litebox::fs::Mode::XGRP | litebox::fs::Mode::XOTH;
```
jaybosamiya-ms, line 38: I'm not sure if this is ok, should we not check whether it is actually executable or not?

litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+    if components.is_empty() || components == ["/"] {
+        "/".to_string()
+    } else if components[0] == "/" {
+        format!("/{}", components[1..].join("/"))
+    } else {
+        components.join("/")
+    }
+}
+
+/// Run Linux programs with LiteBox on unmodified Linux
```
jaybosamiya-ms: Comment needs to be updated

litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+            })
+            .collect();
+        let data = std::fs::read(prog).unwrap();
+        let data = if cli_args.rewrite_syscalls {
+            // capstone declares a global allocator in conflict with our own.
+            // https://github.com/capstone-rust/capstone-rs/blob/14e855ca58400f454cb7ceb87d2c5e7b635ce498/capstone-rs/src/lib.rs#L16
+            // litebox_syscall_rewriter::hook_syscalls_in_elf(&data, None).unwrap()
+            // Might be a bug in `capstone-rs`: https://github.com/capstone-rust/capstone-rs/pull/171
+            todo!()
+        } else {
```
jaybosamiya-ms: We might want to just remove `rewrite_syscalls` from this specific runner. This comment already disabled this `--rewrite-syscalls` flag on Linux, it seems incorrect to copy over the disabled feature over to the Windows side _especially_ since the bug in `capstone-rs` has now been fixed.  For now, simply removing this conditional, and removing the `rewrite_syscalls` from this specific file makes sense. We can leave the Linux runner as-is for now, and fix it up separately, but at least this PR should not introduce this into this new file.

litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+            .map_err(|e| anyhow!("Could not read tar file at {}: {}", tar_file.display(), e))?
+    } else {
+        litebox::fs::tar_ro::empty_tar_file()
+    };
+
+    // TODO(jb): Clean up platform initialization once we have https://github.com/MSRSSP/litebox/issues/24
+    //
+    // TODO: We also need to pick the type of syscall interception based on whether we want
+    // systrap/sigsys interception, or binary rewriting interception. Currently
+    // `litebox_platform_linux_userland` does not provide a way to pick between the two.
```
jaybosamiya-ms: There is no systrap/sigsys interception on the Windows platform.

litebox_runner_linux_on_windows_userland/src/lib.rs
Icegrave0391: The problem is Windows does not give us a way (e.g., file attribute) to distinguish if a file is READ_ONLY or EXECUTABLE. We may use a file's suffix but this is pretty hand-waivy.

Icegrave0391: Left a TODO in the loader for now.

PR #263, opened 2025-08-09
Support for OP-TEE TA command loop and queues

sangho2:
This PR adds support for OP-TEE command loop and queues to enable TA's
event-driven execution. In particular, it executes a TA within a loop while
consuming commands from the submission queue and storing outcomes in
the completion queue. Input and output exchange is performed through
a TA-specific stack. All input commands to a TA can be written in a JSON
file which acts as a client app (mainly for testing and debugging). This PR
contains two example TAs (hello world and RNG) and JSON files to interact
with them.


litebox_runner_optee_on_linux_userland/src/lib.rs
```diff
@@ -81,47 +97,137 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
...
+    register_session_id_elf_load_info(session_id, loaded_program);
+
+    populate_optee_command_queue(session_id, &ta_commands);
+    optee_command_loop_entry(session_id);
+}
+
+/// OP-TEE/TA message command (base64 encoded). It consists of a function ID,
+/// command ID, and up to four arguments. This is base64 encoded to enable
+/// JSON-formatted input files.
+/// TODO: use JSON Schema if we need to validate JSON or we could use Protobuf instead
```
wdcui, line 121: How does OPTEE do this on ARM64 Linux?

litebox_shim_optee/src/lib.rs
```diff
@@ -67,3 +76,298 @@ pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> u32 {
...
+}
+
+fn session_id_elf_load_info_map() -> &'static SessionIdElfLoadInfoMap {
+    static SESS_ID_ELF_LOAD_INFO_MAP: OnceBox<SessionIdElfLoadInfoMap> = OnceBox::new();
+    SESS_ID_ELF_LOAD_INFO_MAP.get_or_init(|| alloc::boxed::Box::new(SessionIdElfLoadInfoMap::new()))
+}
+
+/// Register ELF load information for a session ID
+pub fn register_session_id_elf_load_info(session_id: u32, elf_load_info: ElfLoadInfo) {
+    session_id_elf_load_info_map().insert(session_id, elf_load_info);
```
wdcui: Do we need to check if there exists an entry for the given session_id?

litebox_shim_optee/src/lib.rs
```diff
@@ -67,3 +76,298 @@ pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> u32 {
...
+}
+
+/// OP-TEE TA command structure for the command submission queue
+#[derive(Clone)]
+pub(crate) struct OpteeCommand {
```
wdcui, line 225: What's the difference between TA command/param and Optee command and Utee param?

litebox_shim_optee/src/lib.rs
```diff
@@ -67,3 +76,298 @@ pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> u32 {
...
+
+/// OP-TEE command loop that dequeues commands from the command queue and handles each of them
+/// by interacting with loaded TAs.
+/// For now, it terminates the thread if there is no commands left in the queue.
+/// Instead, it can be an infinite loop with sleep to continously handle commands
+/// (i.e., `UteeEntryFunc::InvokeCommand`) until it gets `UteeEntryFunc::CloseSession`
+/// from the queue.
+/// # Panics
+/// This function panics if it cannot allocate a stack
+pub fn optee_command_loop_entry(session_id: u32) -> ! {
```
wdcui: Where is the loop?

litebox_runner_optee_on_linux_userland/src/lib.rs
sangho2: On an Arm64 Linux environment, a client application interacts with OP-TEE OS and TAs through a TEE driver at Linux kernel [TEE Subsystem](https://www.kernel.org/doc/html/v6.6/staging/tee.html). The TEE driver exposes several ioctl interfaces and the client application uses them via TEE Client API. This JSON-based command file is mainly for avoiding all these internals and directly interacting with TAs.

litebox_shim_optee/src/lib.rs
sangho2: That should be better. Let me add a check.

litebox_shim_optee/src/lib.rs
sangho2: Let me see whether we can have better names. Currently, the TA command is about the commands we extract from the JSON files. The OP-TEE command is about OP-TEE OS/kernel level messaging. UTEE param is about data to be passed to TA's libutee in the end.

litebox_shim_optee/src/lib.rs
sangho2: This is more like an imaginary loop through iretq and syscall. Better name might be needed.

wdcui: It might help with a better name. Thanks.

PR #264, opened 2025-08-11
Add syscall getppid

CvvT:
1. support syscall getppid
2. syscall [thr_new](https://man.freebsd.org/cgi/man.cgi?query=thr_new&sektion=2&n=1) on freebsd uses a different type (long) for `child_tid`. For consistency, changed `child_tid` from `i32` to `i64`.

### DIFF
--- a/litebox_platform_freebsd_userland/src/lib.rs
+++ b/litebox_platform_freebsd_userland/src/lib.rs
@@ -12,7 +12,7 @@ use litebox::platform::page_mgmt::MemoryRegionPermissions;
 use litebox::platform::trivial_providers::TransparentMutPtr;
 use litebox::platform::{ImmediatelyWokenUp, RawConstPointer};
 use litebox::platform::{ThreadLocalStorageProvider, UnblockedOrTimedOut};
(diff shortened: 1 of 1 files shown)


litebox_common_linux/src/lib.rs
```diff
@@ -1073,8 +1073,13 @@ impl<Platform: litebox::platform::RawPointerProvider> ThreadLocalStorage<Platfor
 pub struct Task<Platform: litebox::platform::RawPointerProvider> {
     /// Process ID
     pub pid: i32,
+    /// Parent Process ID
+    pub ppid: i32,
     /// Thread ID
-    pub tid: i32,
+    ///
+    /// Note tid is i32 on Linux but long on freebsd (though it always fits into i32).
```
wdcui: I'm not sure which approach is better. If it's the case that the tid on freebsd can always fit into an i32, should we do the conversion there locally instead of having to do truncation in several other places?

CvvT: Good point! Changed it back to `i32`.

PR #265, opened 2025-08-12
Support getdents64

CvvT:
1. add support for syscall `getdents64`
2. fix `ino`, which should not use 0 (see https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/sysdeps/posix/readdir.c#L110)
3. add a unit test for `getdents64` and an end-to-end test that runs `ls`: to make it run, also added syscall `statfs` but it returns `ENOSYS` directly. We may want to have actual implementation for it in the future.
4. add `.` and `..` and fix relevant tests.
5. fix `tar_ro` fs for metadata support
6. add `ino` to the result of `read_dir` (i.e., `DirEntry`)

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1366,7 +1366,7 @@ pub struct CapData {
     pub inheritable: u32,
 }
 
-#[repr(C, packed)]
+#[repr(C)]
 #[derive(Clone)]
 pub struct LinuxDirent64 {
     /// Inode number
@@ -1382,7 +1382,8 @@ pub struct LinuxDirent64 {
     ///
     /// This is a flexible array member (FAM) with variable length. The actual name data
     /// follows immediately after this struct in memory.
-    pub name: [u8; 0],
+    #[allow(clippy::pub_underscore_fields)]
+    pub __name: [u8; 0],
 }
 
 /// Request to syscall handler
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -955,7 +955,7 @@ pub fn sys_dup(oldfd: i32, newfd: Option<i32>, flags: Option<OFlags>) -> Result<
 struct Diroff(usize);
 
 const DIRENT_STRUCT_BYTES_WITHOUT_NAME: usize =
-    core::mem::offset_of!(litebox_common_linux::LinuxDirent64, name);
+    core::mem::offset_of!(litebox_common_linux::LinuxDirent64, __name);
 
 /// Handle syscall `getdents64`
 pub(crate) fn sys_getdirent64(fd: i32, dirp: MutPtr<u8>, count: usize) -> Result<usize, Errno> {
@@ -989,7 +989,7 @@ pub(crate) fn sys_getdirent64(fd: i32, dirp: MutPtr<u8>, count: usize) -> Result
             off: dir_off as u64,
             len: len.truncate(),
             typ: litebox_common_linux::DirentType::from(entry.file_type.clone()) as u8,
-            name: [0; 0],
+            __name: [0; 0],
         };
         let hdr_ptr = crate::MutPtr::from_usize(dirp.as_usize() + nbytes);
         unsafe { hdr_ptr.write_at_offset(0, dirent64) }.ok_or(Errno::EFAULT)?;
--- a/litebox_shim_linux/src/syscalls/tests.rs
+++ b/litebox_shim_linux/src/syscalls/tests.rs
@@ -180,13 +180,22 @@ fn test_getdent64() {
     let mut found_entries = alloc::vec::Vec::new();
 
     while offset < bytes_read {
-        let dirent_ptr = unsafe {
-            buffer
-                .as_ptr()
-                .add(offset)
-                .cast::<litebox_common_linux::LinuxDirent64>()
+        assert!(
+            unsafe { buffer.as_ptr().add(offset) }.addr()
+                & (core::mem::align_of::<litebox_common_linux::LinuxDirent64>() - 1)
+                == 0,
+            "Pointer at offset {} is not aligned for LinuxDirent64 (requires {}-byte alignment)",
+            offset,
+            core::mem::align_of::<litebox_common_linux::LinuxDirent64>()
+        );
+        let dirent = unsafe {
+            core::ptr::read_unaligned(
+                buffer
+                    .as_ptr()
+                    .add(offset)
+                    .cast::<litebox_common_linux::LinuxDirent64>(),
+            )
         };
-        let dirent = unsafe { &*dirent_ptr };
 
         // Validate the entry length
         assert!(dirent.len > 0, "Directory entry length must be positive");
@@ -198,10 +207,10 @@ fn test_getdent64() {
         let name_ptr = unsafe {
             buffer
                 .as_ptr()
-                .add(offset + core::mem::offset_of!(litebox_common_linux::LinuxDirent64, name))
+                .add(offset + core::mem::offset_of!(litebox_common_linux::LinuxDirent64, __name))
         };
-        let name_len =
-            dirent.len as usize - core::mem::offset_of!(litebox_common_linux::LinuxDirent64, name);
+        let name_len = dirent.len as usize
+            - core::mem::offset_of!(litebox_common_linux::LinuxDirent64, __name);
         let name_bytes = unsafe { core::slice::from_raw_parts(name_ptr, name_len) };
 
         // Find the null terminator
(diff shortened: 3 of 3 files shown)


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -950,3 +950,65 @@ pub fn sys_dup(oldfd: i32, newfd: Option<i32>, flags: Option<OFlags>) -> Result<
...
+            .next_multiple_of(align_of::<litebox_common_linux::LinuxDirent64>());
+        if nbytes + len > count {
+            // not enough space
+            break;
+        }
+        let stat = litebox_fs()
+            .file_status(entry.name.as_str())
+            .expect("failed to get file status");
+        let dirent64 = litebox_common_linux::LinuxDirent64 {
+            ino: stat.node_info.ino as u64,
```
CvvT: I just realized one issue here. `file_status` requires a full path but `entry.name` only tells the last component. Either `DirEntry` needs to have the `ino` information or we can get the path for the given `fd` and then construct the full path. @jaybosamiya-ms any thoughts?

jaybosamiya-ms: Yes, this should be updated at `DirEntry` itself to have an additional `ino: usize` field.  The initial PR that made `DirEntry` was with the understanding that we might extend it to have a small number of additional fields as needed, thus `#[non_exhaustive]`: https://github.com/microsoft/litebox/blob/4aae2c26572bdc92ba8900e14df6d10914045d4e/litebox/src/fs/mod.rs#L342-L343

CvvT: I have added `ino` into `DirEntry` as suggested and fixed some other issues.

litebox/src/fs/in_mem.rs
```diff
@@ -519,15 +519,69 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
+            let root = self.root.read();
+            let mut path = root
+                .entries
+                .iter()
+                .find_map(|(path, entry)| match entry {
+                    Entry::Dir(d) if alloc::sync::Arc::ptr_eq(d, dir) => Some(path.clone()),
+                    _ => None,
+                })
+                .unwrap_or(String::new());
+            path.push('/');
```
jaybosamiya-ms: We seem to push `/` and then trim it out later, might be able to skip this?

litebox/src/platform/mod.rs
```diff
@@ -490,6 +490,17 @@ where
...
+    /// Write a slice of values at the offset.
+    ///
+    /// Returns `None` if the provided pointer is invalid, or such an offset is known (in advance)
+    /// to be invalid.
+    ///
+    /// # Safety
+    ///
+    /// The offset must be valid location for the pointer.
+    #[must_use]
+    unsafe fn write_slice_at_offset(self, count: isize, value: &[T]) -> Option<()>;
```
jaybosamiya-ms: A few nits:
1. We should probably call the argument `values`.
2. Safety should be "All `values.len()` positions starting from the offset must be valid locations for the pointer."
3. I think this is a place where a default implementation is actually a good and valid thing:
```rs
unsafe fn write_slice_at_offset(self, count: isize, value: &[T]) -> Option<()> {
  for (offset, v) in (count..).zip(value) {
    // SAFETY: from the requirements of this function.
    unsafe { self.write_at_offset(offset, v.clone()) }?;
  }
  Some(())
}
```
4. We almost definitely want to also state in the doc comment that "if this function returns `None`, there are no guarantees on how many values have been written" or similar.

litebox/src/platform/trivial_providers.rs
```diff
@@ -184,6 +184,18 @@ impl<T: Clone> RawMutPointer<T> for TransparentMutPtr<T> {
...
+            return None;
+        }
+        if value.is_empty() {
+            return Some(());
+        }
+        unsafe {
+            core::ptr::copy_nonoverlapping(value.as_ptr(), self.inner.offset(count), value.len());
+        }
+        Some(())
+    }
```
jaybosamiya-ms: We can skip this entirely by using the blanket implementation.  The blanket implementation should be almost equivalent in performance (due to inlining), but would have the advantage of not having the foot-gun of potential UB due to the non-overlapping in this version.

litebox_common_linux/src/lib.rs
```diff
@@ -1333,6 +1364,25 @@ pub struct CapData {
...
+    pub off: u64,
+    /// Length of this dirent (including the following name and padding)
+    pub len: u16,
+    /// File type
+    pub typ: u8,
+    /// File name (null-terminated)
+    ///
+    /// This is a flexible array member (FAM) with variable length. The actual name data
+    /// follows immediately after this struct in memory.
+    pub name: [u8; 0],
```
jaybosamiya-ms: Minor (please ignore if either of these suggestions are inconvenient):

This is mildly difficult to get right if we just naively use the `[u8; 0]`, so we might wish to do it via what bindgen does:
```rs
#[repr(C)]
#[derive(Default)]
pub struct __IncompleteArrayField<T>(::std::marker::PhantomData<T>, [T; 0]);
impl<T> __IncompleteArrayField<T> {
    #[inline]
    pub const fn new() -> Self {
        __IncompleteArrayField(::std::marker::PhantomData, [])
    }
    #[inline]
    pub fn as_ptr(&self) -> *const T {
        self as *const _ as *const T
    }
    #[inline]
    pub fn as_mut_ptr(&mut self) -> *mut T {
        self as *mut _ as *mut T
    }
    #[inline]
    pub unsafe fn as_slice(&self, len: usize) -> &[T] {
        ::std::slice::from_raw_parts(self.as_ptr(), len)
    }
    #[inline]
    pub unsafe fn as_mut_slice(&mut self, len: usize) -> &mut [T] {
        ::std::slice::from_raw_parts_mut(self.as_mut_ptr(), len)
    }
}
impl<T> ::std::fmt::Debug for __IncompleteArrayField<T> {
    fn fmt(&self, fmt: &mut ::std::fmt::Formatter<'_>) -> ::std::fmt::Result {
        fmt.write_str("__IncompleteArrayField")
    }
}
```

Then, this field would be a `pub name: __IncompleteArrayField<u8>,`.

The other possible approach is to say `pub name: core::ffi::CStr`, which is the actual correct via of this, but I don't know how usable that would be wrt construction (DST support still is not _amazing_ in Rust).

litebox_platform_linux_kernel/src/ptr.rs
```diff
@@ -80,6 +80,10 @@ impl<T: Clone> RawMutPointer<T> for UserMutPtr<T> {
         todo!()
     }
 
+    unsafe fn write_slice_at_offset(self, _count: isize, _value: &[T]) -> Option<()> {
+        todo!()
+    }
+
```
jaybosamiya-ms: Could be removed with the blanked impl

litebox_platform_lvbs/src/ptr.rs
```diff
@@ -73,6 +73,10 @@ impl<T: Clone> RawMutPointer<T> for UserMutPtr<T> {
         todo!()
     }
 
+    unsafe fn write_slice_at_offset(self, _count: isize, _value: &[T]) -> Option<()> {
+        todo!()
+    }
+
```
jaybosamiya-ms: Could be removed with the blanked impl

litebox_platform_linux_kernel/src/ptr.rs
CvvT: Removed

litebox_platform_lvbs/src/ptr.rs
CvvT: Removed

litebox/src/platform/trivial_providers.rs
CvvT: Good point! It's removed.

litebox/src/platform/mod.rs
CvvT: Got it. Thanks for the suggestion! I thought it might be slow, but compiler should be able to optimize this as you mentioned below.

litebox/src/fs/in_mem.rs
CvvT: Good catch. This trailing `/` is needed for getting the full path of a child. I moved this line so that we don't need to trim it.

litebox_common_linux/src/lib.rs
CvvT: What's wrong with `[u8; 0]`? This field `name` is only used to get its offset: 

https://github.com/microsoft/litebox/blob/564802bb4e891ee1a7b0bfaa71a0b7b2a6ece929/litebox_shim_linux/src/syscalls/file.rs#L957-L958

It should not be accessed directly. I changed it to `__name`.

jaybosamiya-ms: Fair enough; nothing wrong with just `[u8; 0]`.  Please ignore the `__IncompleteArrayField` stuff.

PR #266, opened 2025-08-12
Support for OP-TEE's `check_access_rights` syscall

sangho2:
This PR adds support for OP-TEE's `check_access_rights` syscall which checks
whether a given memory range is with a given access rights.

### DIFF
--- a/litebox_shim_optee/src/syscalls/tee.rs
+++ b/litebox_shim_optee/src/syscalls/tee.rs
@@ -86,9 +86,14 @@ pub fn sys_check_access_rights(
 
     let start = NonZeroAddress::<PAGE_SIZE>::new(align_down(buf.as_usize(), PAGE_SIZE))
         .ok_or(TeeResult::AccessConflict)?;
-    let len = NonZeroPageSize::<PAGE_SIZE>::new(align_up(len, PAGE_SIZE))
-        .ok_or(TeeResult::AccessConflict)?;
-    if let Some(perms) = litebox_page_manager().get_memory_permissions(start, len) {
+    let aligned_len = {
+        let len = len
+            .checked_add(buf.as_usize() - align_down(buf.as_usize(), PAGE_SIZE))
+            .ok_or(TeeResult::AccessConflict)?;
+        NonZeroPageSize::<PAGE_SIZE>::new(align_up(len, PAGE_SIZE))
+            .ok_or(TeeResult::AccessConflict)?
+    };
+    if let Some(perms) = litebox_page_manager().get_memory_permissions(start, aligned_len) {
         if (flags.contains(TeeMemoryAccessRights::TEE_MEMORY_ACCESS_READ)
             && !perms.contains(MemoryRegionPermissions::READ))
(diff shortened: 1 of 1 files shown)


litebox/src/mm/linux.rs
```diff
@@ -732,6 +755,29 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
...
+        &self,
+        page_range: PageRange<ALIGN>,
+    ) -> Option<MemoryRegionPermissions> {
+        let (range_start, range_end) = (page_range.start, page_range.end);
+        let range: core::ops::Range<usize> = page_range.into();
+        if let Some(iter) = self.overlapping(range).next() {
+            if iter.0.start > range_start || iter.0.end < range_end {
+                // partial overlap implies that the given range contains unmapped pages or
+                // consists of memory pages with different permissions.
+                return None;
```
wdcui, line 772: Should we return an error or None? In other words, do we need to differentiate between unmapped pages and pages of different permissions?

Does the mm always merge two neighboring ranges into one if they have the same permission?

litebox_shim_optee/src/syscalls/tee.rs
```diff
@@ -55,3 +71,39 @@ pub fn sys_panic(code: usize) -> ! {
...
+    if flags.contains(TeeMemoryAccessRights::TEE_MEMORY_ACCESS_NONSECURE)
+        && flags.contains(TeeMemoryAccessRights::TEE_MEMORY_ACCESS_SECURE)
+    {
+        // `TEE_MEMORY_ACCESS_NONSECURE` and `TEE_MEMORY_ACCESS_SECURE` are mutually exclusive.
+        return Err(TeeResult::AccessDenied);
+    }
+
+    let start = NonZeroAddress::<PAGE_SIZE>::new(align_down(buf.as_usize(), PAGE_SIZE))
+        .ok_or(TeeResult::AccessConflict)?;
+    let len = NonZeroPageSize::<PAGE_SIZE>::new(align_up(len, PAGE_SIZE))
```
wdcui: Do we align up the len or buf+len?

litebox/src/mm/linux.rs
sangho2: I think we don't need to distinguish unmapped pages from pages with different permissions (at least for now) because OP-TEE TAs only check whether a certain memory range is with expected memory permissions. Let me reconfirm this by checking the OP-TEE OS's source code.

`self.overlapping()` is based on `RangeMap` which associates addresses with `VmArea/VmFlags`. It automatically coalesces addresses with the same `VmArea/VmFlags` into a single range.

litebox_shim_optee/src/syscalls/tee.rs
sangho2: Thanks. I missed that corner case. We should align up buf+len. Let me fix it.

wdcui: I will resolve this conversation after you push the fix.

PR #268, opened 2025-08-20
Support for OP-TEE's AES TA

sangho2:
This PR adds support for running the AES TA from the `optee_examples` written by Linaro.
It implements several system calls and data structures for the AES TA. Also, to support
the syscall rewriter's trampoline page, it fixes a `mmap` issue by specifying a default
ELF load address.

### DIFF
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -471,7 +471,7 @@ pub(crate) fn optee_command_loop_return(session_id: u32) -> ! {
     optee_command_loop_entry(session_id)
 }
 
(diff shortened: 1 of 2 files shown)


litebox_shim_optee/src/lib.rs
```diff
@@ -371,3 +470,285 @@ pub(crate) fn optee_command_loop_return(session_id: u32) -> ! {
 
     optee_command_loop_entry(session_id)
 }
+
+/// A data structure to represent a TEE object refered by `TeeObjHandle`.
```
wdcui: Minor: "referred by" => "referenced by"?

sangho2: fixed. Thanks!

PR #272, opened 2025-08-20
Support Linux guest `vDSO` syscall family on WindowsUserland 

Icegrave0391:
- Modified TimeProvider
- Add methods for Timespec and emulated TimeZone
- Add syscalls

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1737,6 +1737,11 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
         dirp: Platform::RawMutPointer<u8>,
         count: usize,
     },
+    SchedGetAffinity {
+        pid: Option<i32>,
+        len: usize,
+        mask: Platform::RawMutPointer<u8>,
+    },
     /// A sentinel that is expected to be "handled" by trivially returning its value.
     Ret(errno::Errno),
 }
@@ -2333,6 +2338,14 @@ impl<'a, Platform: litebox::platform::RawPointerProvider> SyscallRequest<'a, Pla
                 dirp: Platform::RawMutPointer::from_usize(ctx.syscall_arg(1)),
                 count: ctx.syscall_arg(2),
             },
+            Sysno::sched_getaffinity => {
+                let pid = ctx.syscall_arg(0).reinterpret_as_signed().truncate();
+                SyscallRequest::SchedGetAffinity {
+                    pid: if pid == 0 { None } else { Some(pid) },
+                    len: ctx.syscall_arg(1),
+                    mask: Platform::RawMutPointer::from_usize(ctx.syscall_arg(2)),
+                }
+            }
             // TODO: support syscall `statfs`
             Sysno::statx | Sysno::io_uring_setup | Sysno::rseq | Sysno::statfs => {
                 SyscallRequest::Ret(errno::Errno::ENOSYS)
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -551,6 +551,34 @@ pub(crate) fn sys_getegid() -> usize {
         .with_thread_local_storage_mut(|tls| tls.current_task.credentials.egid)
 }
 
+/// Number of CPUs
+const NR_CPUS: usize = 2;
+
+pub(crate) struct CpuSet {
+    bits: bitvec::vec::BitVec<u8>,
+}
+
+impl CpuSet {
+    pub(crate) fn len(&self) -> usize {
+        self.bits.len()
+    }
+    pub(crate) fn as_bytes(&self) -> &[u8] {
+        self.bits.as_raw_slice()
+    }
+}
+
+/// Handle syscall `sched_getaffinity`.
+///
+/// Note this is a dummy implementation that always returns the same CPU set
+pub(crate) fn sys_sched_getaffinity(pid: Option<i32>) -> CpuSet {
+    if pid.is_some() {
+        unimplemented!("Getting CPU affinity for a specific PID is not supported yet");
+    }
+    let mut cpuset = bitvec::bitvec![u8, bitvec::order::Lsb0; 0; NR_CPUS];
+    cpuset.iter_mut().for_each(|mut b| *b = true);
+    CpuSet { bits: cpuset }
+}
+
 #[cfg(test)]
 mod tests {
     use core::mem::MaybeUninit;
@@ -790,5 +818,20 @@ mod tests {
             result,
             "Parent TID mismatch"
         );
+    }
+
+    #[test]
+    fn test_sched_getaffinity() {
+        crate::syscalls::tests::init_platform(None);
+
+        let cpuset = super::sys_sched_getaffinity(None);
+        assert_eq!(cpuset.bits.len(), super::NR_CPUS);
+        cpuset.bits.iter().for_each(|b| assert!(*b));
+        let ones: usize = cpuset
+            .as_bytes()
+            .iter()
+            .map(|b| b.count_ones() as usize)
+            .sum();
+        assert_eq!(ones, super::NR_CPUS);
     }
 }
--- a/litebox_shim_linux/Cargo.toml
+++ b/litebox_shim_linux/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2024"
 
 [dependencies]
 arrayvec = { version = "0.7.6", default-features = false }
+bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
 bitflags = "2.9.0"
 litebox = { path = "../litebox/", version = "0.1.0" }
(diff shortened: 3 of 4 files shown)


litebox/src/platform/mod.rs
```diff
@@ -303,11 +303,24 @@ pub enum ReceiveError {
     WouldBlock,
 }
 
+/// A non-exhaustive of clock types that can be used with [`TimeProvider`].
+/// These types correspond to the various clocks available in the system.
+#[derive(Debug, Clone)]
+pub enum ClockType {
```
jaybosamiya-ms: Non-exhaustive should come with a `#[non_exhaustive]` to the enum.

litebox/src/platform/mod.rs
```diff
@@ -303,11 +303,24 @@ pub enum ReceiveError {
...
+    /// The real-time clock
+    Realtime,
+    /// The monotonic clock
+    Monotonic,
```
jaybosamiya-ms: This is not sufficient documentation/explanation, it is just repeating the names.  Here's a concrete suggestion:

```suggestion
    /// The real-time clock is the machine's best-guess as to the current wall-clock, time-of-day time. It can jump forwards and backwards as time is updated (e.g., via NTP).
    Realtime,
    /// The monotonic clock represents some elapsed wall-clock time since some arbitrary, fixed point in the past. It isn't affected by changes in the system time-of-day clock. It may or may not count time in a suspended state, but is guaranteed never to jump "backwards" in time.
    Monotonic,
```

litebox/src/platform/mod.rs
```diff
@@ -303,11 +303,24 @@ pub enum ReceiveError {
...
+}
+
 /// An interface to understanding time.
 pub trait TimeProvider {
     type Instant: Instant;
+    type Timespec;
```
jaybosamiya-ms: This generic `Timespec` is not sufficiently useful to create generically-helpful north interfaces above.  If you are curious about how to design this, we can chat more IRL about how this should be done.

As it currently stands, it makes it extremely unclear what the relation between `Instant` and `Timespec` is. Specifically, it makes it quite unclear whether `Instant` is supposed to be the monotonic one or the realtime one. Also, since the `Timespec` type is unrestricted, it is just a generic existential type to North shims (at least until monomorphization occurs), due to which, this is not directly usable, and not a good fit for the platform interface.

litebox_common_linux/src/lib.rs
```diff
@@ -700,6 +706,46 @@ impl TryFrom<TimeVal> for core::time::Duration {
     }
 }
 
+#[allow(clippy::cast_possible_wrap)]
```
jaybosamiya-ms: Please don't allow wraps like this across a large function, they should be specifically targeted to the single statement where such a wrap might occur, and should be marked as an `expect`

litebox/src/platform/mod.rs
Icegrave0391: removed new enum from the platform

litebox/src/platform/mod.rs
Icegrave0391: Thanks! Also this is removed from the platform

litebox/src/platform/mod.rs
Icegrave0391: removed

litebox_common_linux/src/lib.rs
Icegrave0391: Targeting a single statement is experimental. I changed it to `try_from` and now we don't need this `expect`.

Icegrave0391:
@jaybosamiya-ms Changed all implementation by using `PunchThrough` for now. Could you review again?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -872,6 +873,9 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
                 _ => panic!("unexpected error {err}"),
             })
             .map_err(litebox::platform::PunchthroughError::Failure),
+            _ => {
+                unimplemented!("PunchthroughToken for LinuxUserland is not fully implemented yet");
+            }
```
jaybosamiya-ms: Please include the implementation here as part of this PR, thanks!

litebox_common_linux/src/lib.rs
jaybosamiya-ms: Targeting a single statement is doable by adding a separate curly braces.

The `try_from` is silently saturating values, which is probably ok in this case?

Icegrave0391:
@jaybosamiya-ms Added `LinuxUserland` punchthrough. Could you approve the PR when you have time?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -867,6 +868,27 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
...
+            PunchthroughSyscall::Gettimeofday { tv, tz } => {
+                unreachable!("Due to vdso, this would not be triggered on LinuxUserland.");
+            }
+            #[expect(
+                unused_variables,
+                reason = "due to vdso, this would not be triggered on LinuxUserland"
+            )]
+            PunchthroughSyscall::Time { tloc } => {
+                unreachable!("Due to vdso, this would not be triggered on LinuxUserland.");
+            }
```
jaybosamiya-ms: You don't need these `expect`s here.  The `unreachable` makes sense, but you don't need to suppress the warnings this way.

The following should work:
```rs
            PunchthroughSyscall::ClockGettime { .. }
            | PunchthroughSyscall::Gettimeofday { .. }
            | PunchthroughSyscall::Time { .. } => {
                unreachable!("Due to vdso, this would not be triggered on LinuxUserland.");
            }
```

Icegrave0391: Thanks for showing that!

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -675,6 +673,34 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
                 Win32_Threading::WakeByAddressAll(addr.as_usize() as *const c_void);
                 Ok(0)
             },
+            PunchthroughSyscall::ClockGettime { clockid, tp } => {
+                let ts = perf_counter::get_timespec(clockid);
+                let clock_timespec = litebox_common_linux::Timespec {
+                    tv_sec: ts.0,
+                    tv_nsec: ts.1 as u64,
+                };
+                let _ = unsafe { tp.write_at_offset(0, clock_timespec) };
```
CvvT, line 682: why is the result ignored?

If you look at the code above, you can find how we handle the error:
```
unsafe { addr.write_at_offset(0, fs_base) }.ok_or(
    litebox::platform::PunchthroughError::Failure(
        litebox_common_linux::errno::Errno::EFAULT,
    ),
)?;
```

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -675,6 +673,34 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
...
+                let ts = perf_counter::get_timespec(litebox_common_linux::CLOCK_REALTIME);
+                let timeval = litebox_common_linux::TimeVal::from(litebox_common_linux::Timespec {
+                    tv_sec: ts.0,
+                    tv_nsec: ts.1 as u64,
+                });
+                let _ = unsafe { tv.write_at_offset(0, timeval) };
+                // Handle timezone parameter (usually NULL and deprecated)
+                // Return timezone as UTC (0 minutes west, no DST)
+                let timezone = litebox_common_linux::TimeZone::new(0, 0);
+                let _ = unsafe { tz.write_at_offset(0, timezone) };
```
CvvT, line 695: Similar

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -675,6 +673,34 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
...
+                Ok(0)
+            }
+            PunchthroughSyscall::Time { tloc } => {
+                let ts = perf_counter::get_timespec(litebox_common_linux::CLOCK_REALTIME);
+                let seconds = ts.0 as litebox_common_linux::time_t;
+                let _ = unsafe { tloc.write_at_offset(0, seconds) };
```
CvvT, line 701: Similar

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -454,6 +454,69 @@ pub(crate) fn sys_get_robust_list(
...
+    // Return the resolution of the clock
+    // For most modern systems, the resolution is typically 1 nanosecond
+    // This is a reasonable default for high-resolution timers
+    let resolution = litebox_common_linux::Timespec {
+        tv_sec: 0,
+        tv_nsec: 1, // 1 nanosecond resolution
+    };
+
+    unsafe {
+        res.write_at_offset(0, resolution);
```
CvvT, line 483: We should handler the error here.

Issue #273, opened 2025-08-20
Refine the memory bookkeeping system for `WindowsUserland`


Icegrave0391:
On the host platform `WindowsUserland`, memory allocation is performed via `VirtualAlloc()`. This function distinguishes between `RESERVE` and `COMMIT` operations. Apart from that, Windows memory reservation granularity is significantly larger (e.g., 64KB).
Similarly, `VirtualFree()` also differentiates between `RELEASE` and `DECOMMIT`.

To better support running guest Linux ELF binaries (with mmap style), we currently implemented a bookkeeping solution, which:
- tracks each individual `VirtualAlloc(RESERVE)` as "reserved memory ranges".
- track each 4KB granularity page's state within reserved ranges (COMMIT).

The current implementation uses a HashMap. However, we separate each reserved memory ranges to multiple "keys", where each key represents a minimal memory RESERVE granularity in Windows (e.g., 64KB). 
The current hashmap looks like: `{ key = (granularity_aligned_addr: u64): value = used_subregion: RangeSet }`
This increases both memory overhead and look-up complexity.

There are two potential improvement:

1. Use {`RangeSet`: `RangeSet`} as both key and value, as discussed @CvvT @jaybosamiya-ms.

2. We don't distinguish `DECOMMIT` and `RELEASE`. Once a memory is `RESERVED`, we never release it but only de-commit (as suggested by @wdcui). By doing so, we may potentially not need a complex bookkeeping system. Instead, we can always rely on `VirtualQuery()` to look up the memory status and then perform memory allocation/deallocation.

I would try out the second way later on.

CvvT:
Closed by #452 and #456

PR #274, opened 2025-08-20
Fix memory bookkeeping for `WindowsUserland`

Icegrave0391:
As pointed in https://github.com/microsoft/litebox/issues/273#issue-3339624311, we would refine this implementation later on.


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -774,6 +917,110 @@ fn do_query_on_region(mbi: &mut Win32_Memory::MEMORY_BASIC_INFORMATION, base_add
...
+        let mut buf: *mut u16 = std::ptr::null_mut();
+        let flags = Win32_Debug::FORMAT_MESSAGE_ALLOCATE_BUFFER
+            | Win32_Debug::FORMAT_MESSAGE_FROM_SYSTEM
+            | Win32_Debug::FORMAT_MESSAGE_IGNORE_INSERTS;
+        let len = Win32_Debug::FormatMessageW(
+            flags,
+            std::ptr::null_mut(),
+            err,
+            0,
+            (&mut buf) as *mut *mut u16 as *mut u16,
```
jaybosamiya-ms, line 925: This looks incredibly suspicious to me, but reading the docs, I think this is correct.  A comment here about "If `dwFlags` includes `FORMAT_MESSAGE_ALLOCATE_BUFFER`, the function allocates a buffer using the `LocalAlloc` function, and places the pointer to the buffer at the address specified in `lpBuffer`." would help.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -774,6 +917,110 @@ fn do_query_on_region(mbi: &mut Win32_Memory::MEMORY_BASIC_INFORMATION, base_add
...
+            state_to_str(mbi.State),
+            protect_to_str(mbi.Protect),
+            type_to_str(mbi.Type)
+        );
+        if next <= addr {
+            break;
+        }
+        addr = next;
+    }
+}
```
jaybosamiya-ms: Was this meant to be removed before making the PR?  If not, when is the actual time to remove it?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -920,11 +1181,12 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
...
             !addr.is_null(),
-            "VirtualAlloc2 failed. Address: {:p}, Size: {}, Permissions: {:?}. Error: {}",
+            "VirtualAlloc2 failed. Address: {:p}, Size: {}, Permissions: {:?}. Error: {} str: {}",
             aligned_base_addr,
-            aligned_size,
+            region_free_size,
             initial_permissions,
-            unsafe { GetLastError() }
+            unsafe { GetLastError() },
+            werr_text(unsafe { GetLastError() })
```
jaybosamiya-ms: Is it ok to invoke `GetLastError` twice? Are you guaranteed to get the same value both times?  Wouldn't it be a better idea to simply place the error number as part of the string in `werr_text` so that you don't need to invoke it twice?

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: This is just to help me on debugging stuff. After overhaul this implementation i will remove it.

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: Good point!

Icegrave0391:
@jaybosamiya-ms Thanks for pointing this out.
I restored this branch to a stable memory bookkeeper solution, and removed the unrelated `.so` files in the testcases.

Could you review it again?

PR #275, opened 2025-08-20
Implement ThreadLocalStorageProvider for SnpLinuxKernel

CvvT:
This PR also requires changes to [litebox_platform_linux_kernel/src/host/snp/snp-sandbox.h](https://github.com/microsoft/litebox/compare/weiteng/init_snp_runner...weiteng/tls_for_linux_kernel?expand=1#diff-1f47f08b1e220e786277e0b2af2bbf509e304ec1627b22542119185f04bc8c6e) which was copied from `sandbox_driver` from `VSBox`.

I have a working branch for [`sandbox_driver`](https://github.com/MSRSSP/sandbox_driver/tree/weiteng/snp), in which I only made minor changes instead of starting from scratch (to remove all unnecessary code). Should we move `sandbox_driver` to this repo or keep it separate?

### DIFF
--- a/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
+++ b/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
@@ -72,6 +72,7 @@ mod alloc {
     }
 }
 
+/// Get the current task
 fn current() -> Option<&'static mut bindings::vsbox_task> {
     let task: u64;
     unsafe {


litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -72,6 +72,76 @@ mod alloc {
...
+        let tls = litebox_common_linux::ThreadLocalStorage::new(task);
+        self.set_thread_local_storage(tls);
+    }
+}
+
+impl litebox::platform::ThreadLocalStorageProvider for SnpLinuxKernel {
+    type ThreadLocalStorage = litebox_common_linux::ThreadLocalStorage<SnpLinuxKernel>;
+
+    fn set_thread_local_storage(&self, value: Self::ThreadLocalStorage) {
+        let current_task = current().expect("Current task must be available");
```
wdcui, line 114: If we use `expect` here, will we receive an error message when the kernel crashes here?

CvvT: Yes, it would trigger our panic handler in which it prints out the error message.

PR #276, opened 2025-08-20
Implement fallible memcpy for Linux Kernel platform

CvvT:
This PR implements `RawConstPointer` and `RawMutPointer` for `LinuxKernel`.

Also fixed an issue where we use `mutate_subslice_with` on user space pointer, which assumes the underlying memory is present (i.e., no page faults).

Note that read from/write to user space pointer in kernel mode should not panic. Instead, it should be able to resume the execution and return an error. Linux achieves this through building an exception table: https://elixir.bootlin.com/linux/v5.19.17/source/arch/x86/mm/fault.c#L1327, which I would like to implement in the next PR.

### DIFF
--- a/litebox_platform_linux_kernel/src/ptr.rs
+++ b/litebox_platform_linux_kernel/src/ptr.rs
@@ -139,7 +139,7 @@ impl<T: Clone> RawMutPointer<T> for UserMutPtr<T> {
         _range: impl core::ops::RangeBounds<isize>,
         _f: impl FnOnce(&mut [T]) -> R,
     ) -> Option<R> {
-        todo!()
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
The `Raw*Pointer` APIs use `isize` for the counts to try to account for the (rare) case where we might need to reason about negative offsets.  Lengths are always non-negative, so `usize` makes sense.

wdcui:
> The `Raw*Pointer` APIs use `isize` for the counts to try to account for the (rare) case where we might need to reason about negative offsets. Lengths are always non-negative, so `usize` makes sense.

I see. Thanks!

PR #277, opened 2025-08-20
Implement StdioProvider for LinuxKernel

CvvT:
1. For stdio read/write, redirect it to host.
2. Also fixed the `StdioProvider` implementation for `LinuxUserland` to avoid `std` (use raw syscalls instead).

### DIFF
--- a/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
+++ b/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
@@ -2,7 +2,10 @@
 
 use core::arch::asm;
 
-use litebox::platform::{RawConstPointer, RawMutPointer, ThreadLocalStorageProvider};
+use litebox::{
+    platform::{RawConstPointer, RawMutPointer, ThreadLocalStorageProvider},
+    utils::ReinterpretUnsignedExt as _,
+};
 use litebox_common_linux::{SigSet, SigmaskHow};
 
 use super::ghcb::ghcb_prints;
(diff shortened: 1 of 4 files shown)


litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -323,4 +328,34 @@ impl HostInterface for HostSnpInterface {
...
+                        litebox::platform::StdioOutStream::Stdout => {
+                            litebox_common_linux::STDOUT_FILENO
+                        }
+                        litebox::platform::StdioOutStream::Stderr => {
+                            litebox_common_linux::STDERR_FILENO
+                        }
+                    }
+                    .reinterpret_as_unsigned(),
+                ),
+                buf.as_ptr() as u64,
```
wdcui, line 426: Where do we convert the va in vmpl2 to a va accessible by the host in vmpl0?

CvvT: At lines: https://github.com/MSRSSP/sandbox_driver/blob/f28f28012fc640aa177e515e8c39f4bac7251ae3/src/syscalls.c#L96-L97

It is converted to physical address first, which is then converted into virtual address (we create a one-to-one mapping from physical address to virtual address for each process). To ensure the kernel is not tricked to access arbitrary memory, there is a check on the physical address to ensure the physical page is assigned to VMPL2: https://github.com/MSRSSP/sandbox_driver/blob/f28f28012fc640aa177e515e8c39f4bac7251ae3/src/syscalls.c#L103

PR #280, opened 2025-08-21
Implement futex-like subsystem `FutexManager`

jaybosamiya-ms:
This PR introduces a new `FutexManager` subsystem to LiteBox, which is designed to make it simpler to provide a futex-like shim north of LiteBox, without needing to expand the south platform interface.  Importantly, this removes the need for #239 to expand the platform interface in order to support futexes, by instead providing library-OS-level futexes.

A natural question one might ask is "why do we need this, when we are expecting futexes from the platform below?".  The actual interface we get from the layer below is actually narrower than the full Linux futex interface.  Indeed, this is exactly why the (current) status of #239 was to expand the interface below to get the wider interface needed above.  This would not be sustainable or too maintainable in the long run, since obviously, not all platforms we support are able to provide the full futex API.  This means that we need to be able to provide a richer futex API to the north, while getting a narrower futex-like API from the south; and this means that we need to implement our own version of futexes.

Another natural question is: should this sit within LiteBox, or should this sit within the Linux shim.  The correct positioning of this is arguable, but I believe that there is value (especially after some possible future cleanups) to keep this at the LiteBox layer, since then our higher-level primitives could depend on this new `FutexManager` interface---indeed, this might allow us to tighten up our south interface even more.  More design thought is needed, but especially since that is more of a nice-to-have, I did not want to block things on that at this point.

In short: this PR provides a richer futex interface to the north, _without_ changing the south interface for raw mutexes.

Note: there are some optimization opportunities in this implementation that I have not yet fully squeezed out (there are TODO comments left for some of these).  The reason for this is that some of the details to even get it working were _very_ subtle, and took a lot of nasty debugging.  Indeed, debugging this is what led to #279, which needs to be merged before this PR is merged.


litebox/src/sync/futex.rs
```diff
@@ -0,0 +1,466 @@
...
+                    Ok(()) => {
+                        // fallthrough
+                    }
+                    Err(ImmediatelyWokenUp) => unreachable!(),
+                }
+            }
+
+            // We've just been woken up, we might need to go back to sleep if we are not in the
+            // bitset to be woken up, but otherwise, we are free to exit!
+            let Some(bitset) = bitset else {
```
wdcui, line 180: Is `bitset` a bit mask? If so, it's unclear to me where the bit mask is checked.

jaybosamiya-ms: Yes, it is a mask, it is checked a few lines later: https://github.com/microsoft/litebox/blob/d88b1c1d6f2078301824fcb86aea64891141cd0f/litebox/src/sync/futex.rs#L186-L189

wdcui: Ah, I see. I missed that line 168 is `else {`.

litebox/src/sync/futex.rs
```diff
@@ -0,0 +1,466 @@
...
+        let addr: usize = futex_addr.as_usize();
+        if !addr.is_multiple_of(align_of::<AtomicU32>()) {
+            return Err(FutexError::NotAligned);
+        }
+        let ptr = addr as *mut u32;
+        // SAFETY: we've ensured that it is aligned. The read/write lifetimes of `ptr` are going to
+        // be valid as long as we don't actually expose the created `AtomicU32` lifetime outside
+        // this module. And for the memory model, we are explicitly using it only on things that are
+        // supposed to be for futex operations.
+        Ok(unsafe { AtomicU32::from_ptr(ptr) })
```
CvvT, line 81: Here it only ensures `addr` is aligned, it does not check if it is a valid user space address. The following load/store could still fail, which cause the guest program to crash. For kernel, we'd like to handle this exception and return -EFAULT instead.

litebox/src/sync/futex.rs
```diff
@@ -0,0 +1,466 @@
...
+        // value does not drop, we know that nothing has been woken up yet. If the `lockable` for it
+        // has been GC'd out, that means all wakers got woken up, which is fine too to quit out.
+        //
+        // XXX(jayb): This check may not be ideal if there are no waiters that have any overlap in
+        // terms of bitset masks, in which case this might spin forever until at least someone with
+        // that mask goes to sleep.
+        loop {
+            let lockables = self.lockables.read();
+            if let Some(lockable) = lockables.get(&addr)
+                && lockable.num_waiters >= num_to_wake_up
```
CvvT, line 296: Is it possible that `num_waiter` increments because of a new waiter?

litebox/src/sync/futex.rs
jaybosamiya-ms: This is a valid point; checking whether it is a valid user-space address is mildly out of scope of this particular functionality though, and should be checked at the shim layer _before_ it is made into a `RawMutPointer`.  I think it is an orthogonal concern where we should probably set up better constraints on what is ok to be a `RawMutPointer` (i.e., if we made sure that every place that converts something to a `RawMutPointer` does it only for valid pointers, returning `-EFAULT` ASAP) then we would not need to worry about such things in the guts of wherever we are dealing with this.  Specifically, I think this is better handled by updating `from_usize`: https://github.com/microsoft/litebox/blob/bd48201a2c0d01cd5a007e141ff1cf4a1e9467e7/litebox/src/platform/mod.rs#L368-L373

jaybosamiya-ms: Opened #283 to track this

CvvT: We can only check if a pointer is valid or not when we use it. If you check it beforehand, the user program can `unmap` it after the check.

litebox/src/sync/futex.rs
jaybosamiya-ms: Yes, a new waiter can show up in that time when we request things to wake up.  The `>=` is meant to account for that, so that any new ones will be able to back out quickly.  If we did a strict `==` check, then we can accidentally be left in a state where the `wake` operation thinks that it woke someone up, when in reality, someone started waiting.  By setting this condition to `>=`, it ensures that we know that _at least_ someone woke up only _after_ we have evidence that someone did wake up.

CvvT: If it wakes up one waiter while a new one is joining, `num_waiters` would remain the same?

litebox/src/sync/futex.rs
jaybosamiya-ms: I see what you mean, do you have a suggestion on how to fix the check here?  We can discuss offline if needed, but at least as it currently stands, our pointer API design does not expose an atomic "check if valid at time of use", no?

litebox/src/sync/futex.rs
jaybosamiya-ms: I recall having thought of this scenario in the implementation, but I can't immediately point to what prevents it from being an issue, I'll give it more thought and let you know.  Thanks for pointing this out.

litebox/src/sync/futex.rs
CvvT: No, the API to use could be https://github.com/microsoft/litebox/blob/38f30867cf055117369fba7743972e3642c47858/litebox/src/platform/mod.rs#L384

Actually we don't check it at all, instead we rely on page fault: see https://github.com/microsoft/litebox/blob/38f30867cf055117369fba7743972e3642c47858/litebox_platform_linux_kernel/src/ptr.rs#L14-L33

and #276

jaybosamiya-ms: Those give us a read, but it is not an atomic read.  We are forced to convert the pointer to an atomic pointer.  And if we're ready relying on page fault, that would already be taken care of with the existing version I have here, no?

litebox/src/sync/futex.rs
jaybosamiya-ms: 9316de0df9eab7eb9bdb8acd743e7260f361d943 updates the implementation a little to take care of this. Specifically, such a new waiter would not even go to sleep, which means that the wake call would still wake up others, which would ensure that the value eventually drops below `num_waiters`.

jaybosamiya-ms: Good catch, thanks!

litebox/src/sync/futex.rs
CvvT: The difference is that page fault handler knows how to resume the execution within `__memcpy_fallible`, because we create a special exception table that records its IP address (which is not done yet). If we don't do the same thing here, it would just trigger panic.

CvvT: I agree it is not an atomic read, here is the implementation from Linux: https://elixir.bootlin.com/linux/v6.17-rc3/source/kernel/futex/futex.h#L301 We may need another interface.

CvvT: According to ChatGPT: 1. Futex value is a 32-bit word

By definition, futexes are int (32-bit) aligned words in user memory. The kernel side (e.g., get_futex_key()) already verified:

The pointer is user space,

The address is naturally aligned (4-byte aligned for 32-bit futex).

This alignment matters because on x86 (and almost all architectures Linux supports), a naturally aligned 32-bit load is guaranteed to be a single atomic bus transaction.

litebox/src/sync/futex.rs
CvvT: LGTM, thanks!

litebox/src/sync/futex.rs
jaybosamiya-ms: Hmm, I think such a new interface would be out of scope for this specific PR (i.e., should be a separate PR), since the details are not futex-specific, but just generally about how we handle EFAULT for pointers in general for kernel platforms.  Shall we open an issue to track it independent of this PR, and fix it up after this PR, because at least on user-mode platforms, this is not an issue.  I'm happy to leave this PR open until we figure it out, but at the same time, I think it is largely an orthogonal concern to the futex design.

Specifically, `futex_addr_as_atomic` is a private-only function that is used only in one place, so should be a pretty easy cleanup to replace with a new interface (the only primitive required for this implementation is the ability to do a single 32-bit atomic load).  I'm happy to add in comments to that effect if you think it'd be enough.

Note: we cannot assume x86-semantics if we ever want to support running LiteBox on ARM, in which case, we don't want to bake in the assumption that a naturally aligned 32-bit read is atomic _inside_ LiteBox itself.

CvvT: I see, it makes sense.

jaybosamiya-ms: Opened #296 to track this

PR #282, opened 2025-08-22
Add HOME env

CvvT:
See one example from node where the HOME environment variable is used: https://github.com/nodejs/node/blob/f993fca4e476c514f5a97f8a61287fc7f94df26f/deps/uv/src/unix/core.c#L1203

If it is not set, it would use unix socket which we don't support yet.


CvvT:
> Just to clarify: this change is just for running our tests with Node.js. When we actually run Node.js over LiteBox, we use the command line to control what environment variables to pass to Node.js, right?

Correct. The test passes the env through command line to the runner.

Issue #285, opened 2025-08-22
Signal handling on `WindowsUserland`


Icegrave0391:
For now, we place dummy implementation at `PunchThrough` whenever Linux guest-side signal syscalls happen:

```rust
// in litebox_platform_windows_userland.rs:
PunchthroughSyscall::RtSigaction { .. } |
PunchthroughSyscall::RtSigprocmask { .. } => {
    Ok(0)
}
```

Actual implementation to be added.

PR #286, opened 2025-08-22
Add dummy signal handling on `WindowsUserland`

Icegrave0391:
This placeholder for `PunchThrough` ensures that we can successfully test guest ELF binaries like node and Python.

Refer to this issue: https://github.com/microsoft/litebox/issues/285#issue-3346760100

### DIFF
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -701,10 +701,11 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
                 let _ = unsafe { tloc.write_at_offset(0, seconds) };
                 Ok(usize::try_from(seconds).unwrap_or(0))
             }
-            _ => {
-                unimplemented!(
-                    "PunchthroughToken for WindowsUserland is not fully implemented yet"
-                );
+            // TODO(chuqi): Currently, we just add a placeholder (dummy) implementation for signals. This is to
+            // allow temporal testing on target binaries like Node and Python. Will add actual implementation.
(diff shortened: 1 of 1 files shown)


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -701,10 +701,11 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
                 let _ = unsafe { tloc.write_at_offset(0, seconds) };
                 Ok(usize::try_from(seconds).unwrap_or(0))
             }
-            _ => {
-                unimplemented!(
-                    "PunchthroughToken for WindowsUserland is not fully implemented yet"
-                );
+            // TODO(chuqi): Currently, we just add a placeholder (dummy) implementation for signals. This is to
+            // allow temporal testing on target binaries like Node and Python. Will add actual implementation.
```
CvvT, line 705: "temporary"?

jaybosamiya-ms:
Does this PR need to be merged into `main`, or can it just be used locally, especially since you're using it for temporary testing?  I'm not seeing a reason we need to merge this version into `main` right now.

I'm not _completely_ opposed to merging it, it is just that this is a tiny amount of changes, and if it is just for testing, then I'd rather have the version on `main` actually report the `unimplemented!` rather than silently ignore it (I know it leaves in a TODO, but that is harder to find than a hard error).

jaybosamiya-ms:
Closing this since we don't plan to merge this one in.  We need signal handling on Windows Userland, but this PR is not how we'll be doing it.

PR #290, opened 2025-08-25
Implement thread provider for SNP

CvvT:
1. Implement `ThreadProvider` for SNP
2. parse `argv` and `envp` from host
3. add a multi-threading test to `test.tar` (it doesn't run completely because of the lack of support for syscall `futex`)

A few issues I noticed while developing this PR:
1. support vDSO: do we want to expose/map VMPL0's vDSO to VMPL2? We could also implement a new vDSO once we have the support for related syscalls.
2. how to pass a large tar file to VMPL2: for now, I just use `include_bytes` to embed it into kernel. However, when it is loaded into memory, we use `__get_free_pages` to allocate physically contiguous pages, which has limit on the size of memory you can allocate. Similarly, the allocator in litebox is also backed by `__get_free_pages` so it cannot allocate large chunk of memory for kernel platform.

### DIFF
--- a/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
+++ b/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
@@ -157,7 +157,8 @@ const EFLAGS_OFFSET: usize = core::mem::offset_of!(litebox_common_linux::PtRegs,
 
 /// Callback function for a new thread
 ///
-/// This is called by `sandbox_process_ret_from_fork` from `entry.S`
+/// This is called by `sandbox_process_ret_from_fork` from `entry.S`.
+/// Arguments should be set up by host.
 #[unsafe(no_mangle)]
 extern "C" fn thread_start(
     pt_regs: *mut litebox_common_linux::PtRegs,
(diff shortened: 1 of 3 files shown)


litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -137,14 +137,140 @@ impl litebox::platform::ThreadLocalStorageProvider for SnpLinuxKernel {
...
+    thread_start_args.thread_args.task.tid = regs.rax.reinterpret_as_signed().truncate();
+    // Child's pt_regs should have the same registers as parent's except rax, rip, and rsp.
+    *regs = thread_start_args.parent_ctx;
+    regs.rax = 0;
+    regs.rip = thread_start_args.entry_point;
+    regs.rsp = thread_start_args.stack_top;
+
+    // Set up thread-local storage for the new thread. This is done by
+    // calling the actual thread callback with the unpacked arguments
+    (thread_start_args.thread_args.callback)(*(thread_start_args.thread_args));
```
wdcui, line 184: Is this `callback` required or optional?

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -137,14 +137,140 @@ impl litebox::platform::ThreadLocalStorageProvider for SnpLinuxKernel {
...
+        // it sets `r15` to the thread start argument pointer right before vmexit.
+        // Host should ensure new thread's r15 is the same as the parent thread's.
+        // In `sandbox_process_ret_from_fork` from `entry.S` (called by host in the new thread's
+        // context), it moves `r15` to `rsi` (the second argument).
+        let mut req = bindings::SnpVmplRequestArgs::new_request(
+            bindings::SNP_VMPL_SYSCALL_REQ,
+            NR_SYSCALL_CLONE3, // repurpose size field to syscall id
+            [flags.bits(), 0, 0, 0, 0, 0],
+        );
+        unsafe {
```
wdcui: Why do we invoke this low-level call directly here instead of calling a wrapper function?

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
CvvT: It is required. Also, it is required to set `thread_start_args.thread_args.task.tid` properly. I added some comment to `litebox_common_linux::NewThreadArgs`.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
CvvT: I updated the driver code so now the argument `thread_start_arg_ptr` is passed through the high-level function instead of low-level register.

PR #291, opened 2025-08-25
Reinvent `PageManagementProvider` for `WindowsUserland` (no bookkeeping)

Icegrave0391:
This is the PR for the discussion here: https://github.com/microsoft/litebox/issues/273#issue-3339624311

### DIFF
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -862,10 +862,6 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
                 let mut query_region_end = query_addr + mbi.RegionSize;
                 query_region_end = core::cmp::min(query_region_end, suggested_range.end);
 
-                println!(
-                    "Querying addr: 0x{query_addr:x}, end: 0x{query_region_end:x}, suggested end: {:p} state: {:?}",
-                    suggested_range.end as *mut c_void, mbi.State
-                );
                 // Deal with collision with already committed memory outside of the tracker.
                 if mbi.State == Win32_Memory::MEM_COMMIT
                     && !is_range_within_committed(&(query_addr..query_region_end))


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -771,8 +766,13 @@ fn do_prefetch_on_range(start: usize, size: usize) {
         };
         PrefetchVirtualMemory(GetCurrentProcess(), 1, &raw const prefetch_entry, 0) != 0
     };
-    assert!(ok, "PrefetchVirtualMemory failed with error: {}", unsafe {
-        GetLastError()
+    assert!(ok, "{}", {
+        let last_error = unsafe { GetLastError() };
+        format!(
```
jaybosamiya-ms, line 774: Nit: a `format!` that's already inside a formatting thing is better written as a `format_args!`.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -918,29 +966,50 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
...
+            );
+            available_size = core::cmp::min(
+                available_size,
+                available_end_addr - aligned_base_addr as usize,
+            );
+        }
 
-        // Reserve and commit the memory.
+        // Do the real VirtualAlloc to reserve and commit the memory.
+        // TODO(chuqi): here we may also break down the RESERVE and COMMIT.
```
jaybosamiya-ms, line 1036: I'm not sure I understand this TODO here, can you explain more?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -865,43 +911,45 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
...
                         _ => {
                             panic!("Unexpected memory state: {:?}", mbi.State);
                         }
                     }
+                };
+
+                // Prefetch the memory range if requested
+                if populate_pages_immediately {
+                    do_prefetch_on_range(ptr as usize, size_within_region);
                 }
```
jaybosamiya-ms, line 983: I'm not sure I see why this portion moved up in the code

jaybosamiya-ms:
Generally seems reasonable to me, a few comments added inline.  Please wait for @CvvT's review before merging, thanks!

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: This is just a minor optimization. Basically, due to we have to `round_up()` the size for `VirtualAlloc(RESERVE)`, the actually allocated size may larger than what we need. Thus, we may not need to `COMMIT` all of them (just a part of them).

Nonetheless, this is very minor.

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: This is to make sure the prefetch will operate on a same "reserved" memory region. 
We can also modify the code in `do_prefetch_on_range()` to perform `VirtualQuery()`, before prefetching a large contiguous range (but this seems to be unnecessary).

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -1014,15 +1130,43 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
...
-                &raw mut old_protect,
-            ) != 0
-        };
-        assert!(ok, "VirtualProtect failed: {}", unsafe { GetLastError() });
+
+        let mut perm_update_base = range.start;
+        let mut remain_size = range.len();
+        let mut mbi = Win32_Memory::MEMORY_BASIC_INFORMATION::default();
+
+        // Decommit the region at mbi boundary
```
CvvT: Is this comment correct?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -868,13 +902,25 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
...
+                                assert!(
+                                    ptr == base_addr,
+                                    "VirtualAlloc2(COMMIT) returned address {:p} which is not the expected fixed address {:p}",
+                                    ptr,
+                                    base_addr
+                                );
+                            }
+                            ptr
                         }
                         // In case the region is already committed, we just need to change its permissions.
```
CvvT: Memory manager shouldn't ask for already committed memory range unless it is not aware of this range (e.g., memory allocated to global allocator is not tracked). If `fixed_address` is not true, we should reverse and commit a new range.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -985,14 +1074,41 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
...
-        };
-        assert!(ok, "VirtualFree failed: {}", unsafe { GetLastError() });
+        let mut decommit_base = range.start;
+        let mut remain_size: usize = range.len();
+
+        // Decommit the region at mbi boundary
+        while decommit_base < range.end {
+            let mut mbi = Win32_Memory::MEMORY_BASIC_INFORMATION::default();
+            do_query_on_region(&mut mbi, decommit_base as *mut c_void);
+            assert!(mbi.BaseAddress as usize == decommit_base);
```
CvvT, line 1110: Is it possible that the range contains some unallocated memory? If so, this assertion may fail? Normally, application would not do it, but it seems to be acceptable to Linux.

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: fixed comment

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: According to Win32 API (https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualfree): "The function does not fail if you attempt to decommit an uncommitted page. This means that you can decommit a range of pages without first determining the current commitment state."

It is safe without first determining the current commitment state. Let me add a comment

Icegrave0391: Also added a condition check

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: fixed

Icegrave0391: changed back to `format!` for now to pass the clippy

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: This case exists, and only exists due to we will `RESERVE | COMMIT` a larger range (than the size we required) at the beginning. I added a comment.

Icegrave0391: Also fixed that case.

Icegrave0391:
@jaybosamiya-ms @CvvT Could you review this PR again? Thanks.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -41,6 +42,9 @@ mod perf_counter;
 
 extern crate alloc;
 
+/// Global memory tracker to maintain all committed memory ranges from the PageManagementProvider
+static MEMORY_TRACKER: std::sync::RwLock<RangeSet<usize>> = std::sync::RwLock::new(RangeSet::new());
```
jaybosamiya-ms, line 46: This could be stored within the `WindowsUserland` struct, no?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -83,6 +87,23 @@ pub type SyscallHandler = fn(litebox_common_linux::SyscallRequest<WindowsUserlan
 /// The syscall handler passed down from the shim.
 static SYSCALL_HANDLER: std::sync::RwLock<Option<SyscallHandler>> = std::sync::RwLock::new(None);
 
+/// Helper functions for managing the global memory tracker
```
jaybosamiya-ms: This doc string applies only to the first function and not to the rest. If you are writing doc strings, then it should be specific to the function, not to a group of functions.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -829,61 +881,103 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
...
+                                (ptr as usize)..(ptr as usize + size_within_region),
                             );
+
+                            ptr
                         }
-                        // In case the region is already committed, we just need to change its permissions.
+
+                        // In case the region is already committed, we make sure the already committed region is in the tracker.
+                        // We simply change the protection flags of the region (if in the tracker). Otherwise it means we are
+                        // colluding with the global memory allocator and we have to handle new allocation.
```
jaybosamiya-ms: "colliding"?

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: I agree. We can either use static variable or merge it into the `WindowsUserland` struct.

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: fixed

CvvT:
I believe this can be closed by #452 and #456.

PR #292, opened 2025-08-26
SystemTime via the platform without punchthrough

jaybosamiya-ms:
#272 introduced various time/clock related functionality to the punchthrough. This PR cleans that up by properly introducing a system-time interface for the `TimeProvider`, and then cleans up the shim and platforms to use and implement these interfaces (thereby removing the punchthrough).

This PR is best reviewed by first looking at the changes in `litebox/src/platform/mod.rs` first before looking at other parts impacted by it.


litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -489,32 +503,26 @@ pub(crate) fn sys_gettimeofday(
...
-    })
+    if tz.as_usize() != 0 {
+        // `man 2 gettimeofday`: The use of the timezone structure is obsolete; the tz argument
+        // should normally be specified as NULL.
+        unimplemented!()
+    }
+    let timeval = litebox_common_linux::Timespec::try_from(real_time_as_duration_since_epoch())
+        .or(Err(Errno::EOVERFLOW))?
+        .into();
+    unsafe { tv.write_at_offset(0, timeval) }.ok_or(Errno::EFAULT)
```
CvvT, line 517: `tv` could be NULL, in which case we do nothing: https://elixir.bootlin.com/linux/v6.17-rc3/source/kernel/time/time.c#L143

We could either check `tv.as_usize() == 0` or change the type to Option<MutPtr>.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -489,32 +503,26 @@ pub(crate) fn sys_gettimeofday(
...
-            // Safe conversion from usize to time_t (i64)
-            litebox_common_linux::time_t::try_from(seconds_usize)
-                .unwrap_or(litebox_common_linux::time_t::MAX)
-        })
-        .unwrap_or(0)
+) -> Result<litebox_common_linux::time_t, Errno> {
+    let time = real_time_as_duration_since_epoch();
+    let seconds: u64 = time.as_secs();
+    let seconds: litebox_common_linux::time_t = seconds.try_into().or(Err(Errno::EOVERFLOW))?;
+    unsafe { tloc.write_at_offset(0, seconds) }.ok_or(Errno::EFAULT)?;
```
CvvT: Similarly, `tloc` could be NULL: https://elixir.bootlin.com/linux/v6.17-rc3/source/kernel/time/time.c#L66

litebox_shim_linux/src/syscalls/process.rs
jaybosamiya-ms: Ah, good catch!  Will update shortly

PR #295, opened 2025-08-26
Construct tar_ro fs with slice

CvvT:
Add a new constructor to tar_ro fs: `pub fn new_borrowed(litebox: &LiteBox<Platform>, tar_data: &'static [u8]) -> Self`, which helps avoid allocating extra memory. The tar file might be larger than what the allocator can allocate.

### DIFF
--- a/litebox/src/fs/tar_ro.rs
+++ b/litebox/src/fs/tar_ro.rs
@@ -25,7 +25,6 @@ use alloc::borrow::ToOwned as _;
 use alloc::string::String;
 use alloc::vec::Vec;
 use hashbrown::HashMap;
-use tar_no_std::TarArchive;
 
 use crate::{
     LiteBox,
@@ -61,6 +60,15 @@ enum TarData {
     Borrowed(tar_no_std::TarArchiveRef<'static>),
 }
 
+impl TarData {
+    fn entries(&self) -> tar_no_std::ArchiveEntryIterator<'_> {
+        match self {
+            TarData::Owned(ar) => ar.entries(),
+            TarData::Borrowed(ar_ref) => ar_ref.entries(),
+        }
+    }
+}
+
 /// A backing implementation for [`FileSystem`](super::FileSystem), storing all files in-memory, via
 /// a read-only `.tar` file.
 pub struct FileSystem<Platform: sync::RawSyncPrimitivesProvider> {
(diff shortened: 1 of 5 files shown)


litebox/src/fs/tar_ro.rs
```diff
@@ -88,11 +93,36 @@ impl<Platform: sync::RawSyncPrimitivesProvider> FileSystem<Platform> {
...
+
+    /// Construct a new `FileSystem` instance from a borrowed/static tar data slice.
+    ///
+    /// This is suitable for use with `include_bytes!()` (which yields a `'static` slice).
+    ///
+    /// # Panics
+    ///
+    /// Panics if the provided `tar_data` is found to be an invalid `.tar` file.
+    #[must_use]
+    pub fn new_borrowed(litebox: &LiteBox<Platform>, tar_data: &'static [u8]) -> Self {
```
jaybosamiya-ms: Instead of this, if you updated `new` to take an `alloc::borrow::Cow<'static, [u8]>`, we wouldn't need a new constructor, and would be able to work with either owned or borrowed data.  Sadly, the `tar_no_std` authors have not set up the ownership relation trait between `TarArchive` and `TarArchiveRef` so we can't just store a `Cow` and we need the `TarData` enum that you've set up, but yeah, at least the constructor could be done with the cleaner version.

litebox/src/fs/tar_ro.rs
```diff
@@ -88,11 +93,36 @@ impl<Platform: sync::RawSyncPrimitivesProvider> FileSystem<Platform> {
...
             current_working_dir: "/".into(),
         }
     }
 
+    fn entries(&self) -> tar_no_std::ArchiveEntryIterator<'_> {
+        match &self.tar_data {
+            TarData::Owned(ar) => ar.entries(),
+            TarData::Borrowed(ar_ref) => ar_ref.entries(),
+        }
+    }
```
jaybosamiya-ms: I recommend placing this as an impl on `TarData`, which actually reduces the scope of the borrow to _clearly_ only be from `tar_data`.  That also would completely remove most of the rest of the diff in this PR.

litebox/src/fs/tar_ro.rs
CvvT: Good point! Fixed

litebox/src/fs/tar_ro.rs
CvvT: Fixed as suggested

PR #297, opened 2025-08-26
Support `open(o_flag=O_EXCL)` for in_mem/layered/tar_ro fs

Icegrave0391:


litebox/src/fs/errors.rs
```diff
@@ -18,6 +18,8 @@ pub enum OpenError {
     NoWritePerms,
     #[error("write access requested for a file on a read-only filesystem")]
     ReadOnlyFileSystem,
+    #[error("file already exists and O_EXCL was specified")]
```
jaybosamiya-ms: Nit: I don't think the "and O_EXCL was specified" is needed here in the error message.

litebox/src/fs/layered.rs
```diff
@@ -377,10 +381,38 @@ impl<
...
+                        }
+                        EntryX::Upper { .. } => unreachable!(),
+                    }
+                } else if self.ensure_lower_contains(&path).is_ok() {
+                    // File exists in lower layer but not tracked, fail with EXCL
+                    return Err(OpenError::AlreadyExists);
+                }
+                // File doesn't exist anywhere, proceed with creation at upper layer only (fall through)
+                else {
+                }
```
jaybosamiya-ms: I think this is unnecessary added complexity; you can simply directly check `self.file_status(path.as_str()).is_ok()` which will take care of checking both layers; there is no special extra handling needed.

litebox/src/fs/layered.rs
```diff
@@ -377,10 +381,38 @@ impl<
...
+            } else {
+                // Normal O_CREAT behavior: try to open existing file first
```
jaybosamiya-ms: I am not sure why this extra comment was added, the existing comment for this condition already explains things enough, no?

litebox/src/fs/layered.rs
Icegrave0391: fixed

litebox/src/fs/errors.rs
Icegrave0391: fixed

litebox/src/fs/layered.rs
Icegrave0391: This is the previous comments and I did not delete them. I feel keeping them can improve the readability?

Icegrave0391:
@jaybosamiya-ms Could you also review this PR again?

litebox/src/fs/layered.rs
jaybosamiya-ms: We can keep the previous existing comment, there is no need to include the new "Normal O_CREAT behavior: try to open existing file first" comment.

Not a major issue to keep the comment, I just don't think adding the new text is helping.

Issue #240, opened 2025-07-29
`test_runner_with_nodejs` breaks


CvvT:
The test `test_runner_with_nodejs` breaks from time to time on CI (It never breaks locally though). For now, we can just rerun the failed test to pass CI.

PR #300, opened 2025-08-27
Simplify test code

CvvT:
This PR refactors the testing code so that it is easier to add a new end-to-end test.

1. Use `ldd` to find all dependencies instead of manually specifying them.
2. Add a new test `test_static_exec_with_rewriter` for syscall rewriter backend (which can run on both 64bit and 32bit platform as it does not rely on `rtld_audit`).
3. Add a new end-to-end test `thread.c`. For a new test, we just need to put its C code into the test folder.
4. I also re-enabled `test_runner_with_nodejs` and it seems to work fine now (I manually triggered CI multiple times). I haven't figure out the root cause of it #240 as I cannot reproduce it locally. #309 may have fixed it, but I'm not sure.
5. Add a test `test_node_with_rewriter` for node with rewriter backend. It's disabled for now as it takes > 5min to rewrite the binary. Alternatively, we could upload rewritten binaries to avoid it.

### DIFF
--- a/litebox_runner_linux_userland/tests/common/mod.rs
+++ b/litebox_runner_linux_userland/tests/common/mod.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 
 /// Find all dependencies of a given binary via `ldd`
+#[allow(dead_code, reason = "not used by loader.rs for x86")]
 pub fn find_dependencies(prog: &str) -> Vec<String> {
     let output = std::process::Command::new("ldd")
         .arg(prog)
--- a/litebox_runner_linux_userland/tests/loader.rs
+++ b/litebox_runner_linux_userland/tests/loader.rs
@@ -156,6 +156,7 @@ fn test_load_exec_common(executable_path: &str) {
     unsafe { trampoline(info.entry_point, info.user_stack_top) };
 }
 
+#[cfg(target_arch = "x86_64")]
 #[test]
 fn test_load_exec_dynamic() {
     let path = common::compile("./tests/hello.c", "hello_dylib", false, false);
@@ -179,6 +180,7 @@ fn test_load_exec_dynamic() {
     test_load_exec_common(executable_path);
 }
 
+#[cfg(target_arch = "x86_64")]
 #[test]
 fn test_load_exec_static() {
     let path = common::compile("./tests/hello.c", "hello_exec", true, false);
--- a/litebox_runner_linux_userland/tests/run.rs
+++ b/litebox_runner_linux_userland/tests/run.rs
@@ -242,6 +242,7 @@ fn test_static_exec_with_rewriter() {
     }
 }
 
+#[cfg(target_arch = "x86_64")]
 #[test]
 fn test_dynamic_lib_with_seccomp() {
     for path in find_c_test_files("./tests") {


CvvT:
```
cargo run -p litebox_syscall_rewriter /lib/ld-linux.so.2 -o ld.hooked
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/litebox_syscall_rewriter /lib/ld-linux.so.2 -o ld.hooked`
Error: insufficient bytes before or after syscall at 0x1c810
```

Fixed by #305

litebox_runner_linux_userland/tests/common/mod.rs
```diff
@@ -1,104 +1,58 @@
...
+            // Skip "not found"
+            if right.starts_with("not found") {
+                continue;
+            }
+            // Extract token before whitespace or '('
+            if let Some(token) = right.split_whitespace().next()
+                && token.starts_with('/')
+            {
+                paths.push(token.to_string());
+            }
```
wdcui, line 37: Should we add a warning message for the else branch?

litebox_runner_linux_userland/tests/common/mod.rs
```diff
@@ -1,104 +1,60 @@
...
-        &litebox,
-        if tar_data.is_empty() {
-            litebox::fs::tar_ro::empty_tar_file().into()
+        if let Some(idx) = line.find("=>") {
+            // Format: "libc.so.6 => /lib/.../libc.so.6 (0x...)"
+            let right = line[idx + 2..].trim();
+            // Skip "not found"
+            if right.starts_with("not found") {
+                continue;
+            }
```
jaybosamiya-ms: We probably want a warning here, too, no?

litebox_runner_linux_userland/tests/common/mod.rs
CvvT: Added

CvvT:
@wdcui I added a new test for node with rewriter backend. You can test it via `cargo nextest run --package litebox_runner_linux_userland --test run --nocapture -- test_node_with_rewriter --exact --ignored`

litebox_runner_linux_userland/tests/common/mod.rs
CvvT: Added, thanks!

CvvT:
> Overall seems reasonable to me, thanks for doing the simplification and cleanup!
> 
> Regarding the rewriting taking a long time, we might be able to set up some sort of cache-checking (do not attempt to rewrite if the rewritten binary already exists and the sha256sum of the input matches the checksum we've stored beside the rewritten binary). This would work pretty well on the CI _without_ checking large binaries in, because the CI _already_ has a caching step for anything that sits inside the `target/` directory (well technically, it caches a subset of even `target/` but we can tweak that quite easily). Anyways, I think that is an orthogonal concern, and need not block merging this PR in. Thanks again!

Good point! I created an issue #325 to track it.

CvvT:
I just realized the crate we use, `seccompiler`, does not support x86 arch: see https://github.com/rust-vmm/rust-vmm/issues/93, thus, those tests requiring seccomp only run on x86_64.

PR #303, opened 2025-08-27
Fix futex manager

CvvT:
I found a few issues while testing against node.

1. Waker has to wait until all waiters it asks to wake up finish waking up. Otherwise, it is possible that only the first woken thread sees `latest_wake_bitset` as it is would be reset to 0 by the waker soon afterwards.
2. `num_claimed_woken_up` might be less than `lockable.num_waiters` because some waiter might immediately wake up, so kernel doesn't count it. Thus, we don't check it at all.

### DIFF
--- a/litebox/src/sync/futex.rs
+++ b/litebox/src/sync/futex.rs
@@ -269,10 +269,8 @@ impl<Platform: RawSyncPrimitivesProvider + RawPointerProvider + TimeProvider>
         let old_num_waiters = lockable.num_waiters;
         let num_to_wake_up = num_to_wake_up.get();
         // Now we can actually trigger things to start waking up.
-        // Note that the number of waiters that actually woke up may be less than `old_num_waiters`,
-        // because some waiter might immediately wake up so kernel doesn't count it.
         let num_to_wake_up = if old_num_waiters >= num_to_wake_up {
-            let _ = lockable.raw_mutex.wake_many(
+            let num_claimed_woken_up = lockable.raw_mutex.wake_many(
                 #[expect(
                     clippy::missing_panics_doc,
                     reason = "this conversion should never fail"
@@ -281,10 +279,15 @@ impl<Platform: RawSyncPrimitivesProvider + RawPointerProvider + TimeProvider>
                     num_to_wake_up.try_into().unwrap()
                 },
             );
+            // Note that `num_claimed_woken_up` may be less than what we requested even if there are enough waiters present.
+            // This is because some waiter might immediately wake up due to a value mismatch so the kernel doesn't count it.
+            debug_assert!(num_claimed_woken_up <= num_to_wake_up as usize);
             num_to_wake_up
         } else {
             // Wake up all
-            let _ = lockable.raw_mutex.wake_all();
+            let num_claimed_woken_up = lockable.raw_mutex.wake_all();
+            // The number of woken threads cannot exceed the number of waiters that were present before waking up.
+            debug_assert!(num_claimed_woken_up <= old_num_waiters as usize);
             old_num_waiters
         };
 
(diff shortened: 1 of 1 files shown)


CvvT:
I didn't find where node calls futex, but I found this: https://github.com/bminor/glibc/blob/c49a32d7eb0466544cce089cd2affab15694dc5a/sysdeps/nptl/futex-internal.h#L210
It seems that applications usually don't use the return value from futex wake at all (probably they just assume futex wake always work as intended, e.g., wake(n) should wake up as many waiters as possible up to n, although the man page does not specify whether the kernel could wake up less waiters).

I also tried to change the return value of wake operation to always zero, node can still run fine, which seems to confirm my guess.

litebox/src/sync/futex.rs
```diff
@@ -266,34 +266,46 @@ impl<Platform: RawSyncPrimitivesProvider + RawPointerProvider + TimeProvider>
...
-            {
-                num_to_wake_up.try_into().unwrap()
-            },
-        );
-        debug_assert!(num_claimed_woken_up > 0);
+        let old_num_waiters = lockable.num_waiters;
+        let num_to_wake_up = num_to_wake_up.get();
+        // Now we can actually trigger things to start waking up.
+        // Note that the number of waiters that actually woke up may be less than `old_num_waiters`,
+        // because some waiter might immediately wake up so kernel doesn't count it.
```
jaybosamiya-ms: I am not sure if this comment is correct (or I may be misunderstanding it).  Number actually woken up may be _more_ than `old_num_waiters`, no?

litebox/src/sync/futex.rs
```diff
@@ -266,34 +266,46 @@ impl<Platform: RawSyncPrimitivesProvider + RawPointerProvider + TimeProvider>
...
+
         // Releasing the lock allows those woken threads to proceed with their wake-up sequence.
         drop(lockables);
-        // Now, we spin until at least one of the waiters has woken up and finished letting us know
-        // it has woken up. We know there must have been at least one before, so as long as the
-        // value does not drop, we know that nothing has been woken up yet. If the `lockable` for it
-        // has been GC'd out, that means all wakers got woken up, which is fine too to quit out.
+        // Now, we spin until all `num_to_wake_up` waiters have woken up and finished letting us know
+        // it has woken up. If the `lockable` for it has been GC'd out, that means all wakers got woken up,
+        // which is fine too to quit out.
```
jaybosamiya-ms, line 298: Minor: might be helpful to add a comment saying something like 'while our documentation of this function says "at most", we attempt to wake up "as many as", since this is closer to what some applications seem to expect; this is similar to how Linux also says "at most" but then attempts "as many as"' or similar

jaybosamiya-ms:
I know you know this already (since we've discussed this before) but just because it showed up in https://github.com/microsoft/litebox/pull/303#issuecomment-3235160870
> although the man page does not specify whether the kernel could wake up less waiters

The [man page](https://www.man7.org/linux/man-pages/man2/futex.2.html) states (emphasis mine):
> FUTEX_WAIT  (since Linux 2.6.0)
> This operation wakes _at most_ val of the waiters [...]. No guarantee is provided about which waiters are awoken [...]

This means that the kernel _can_ wake up fewer.  Technically node might be wrong to assume it'll wake up all, but I guess they're depending on all being woken up.

Also, as I mentioned on our Teams chat, the fact that the return value is ignored is fascinating!

litebox/src/sync/futex.rs
CvvT: Moved and updated the comment to make it clear. I meant the return value from `wake_many` might be less than what we requested `num_to_wake_up`due to immediate wakeup.

litebox/src/sync/futex.rs
CvvT: Added. Thanks!

PR #304, opened 2025-08-27
Support syscall futex

CvvT:
Add support for syscall futex with some limitations:

1. Shared futex is treated as private futex as we don't support multi-processing
2. Only support WAIT/WAKE/WAIT_BITSET
3. wait with `CLOCK_REALTIME` flag is not supported yet (unless `timeout` is NULL, which means blocking indefinitely).
4. Remove `WakeByAddress` punchthrough.
5. Simplify the implementation of RawMutex for windows/linux/freebsd platforms. Basically, they just need to call the corresponding syscalls without tracking anything.

### DIFF
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -17,7 +17,7 @@ use litebox::platform::UnblockedOrTimedOut;
 use litebox::platform::page_mgmt::MemoryRegionPermissions;
 use litebox::platform::trivial_providers::TransparentMutPtr;
 use litebox::platform::{ImmediatelyWokenUp, RawMutPointer};
-use litebox::utils::{TruncateExt as _, ReinterpretUnsignedExt as _};
+use litebox::utils::{ReinterpretUnsignedExt as _, TruncateExt as _};
 use litebox_common_linux::PunchthroughSyscall;
 
 use windows_sys::Win32::Foundation::{self as Win32_Foundation, FILETIME};


litebox_common_linux/src/lib.rs
```diff
@@ -1442,6 +1442,52 @@ pub struct LinuxDirent64 {
...
+        /// <https://docs.rs/bitflags/*/bitflags/#externally-defined-flags>
+        const _ = !0;
+
+        const FUTEX_CMD_MASK = !(FutexFlags::PRIVATE.bits() | FutexFlags::CLOCK_REALTIME.bits());
+    }
+}
+
+#[non_exhaustive]
+pub enum FutexArgs<Platform: litebox::platform::RawPointerProvider> {
+    WAIT {
```
jaybosamiya-ms: Nit: we might want to continue using regular CamelCase enum variant names for things that are regular Rust enums.  This one sits inside `SyscallRequest`, so suddenly going UPPER_SNAKE_CASE is somewhat inconsistent.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -269,6 +268,23 @@ impl Instant {
...
+        else {
+            return Instant(u64::MAX);
+        };
+        let Some(counter) = micros.checked_mul(CPU_MHZ.load(core::sync::atomic::Ordering::Relaxed))
+        else {
+            return Instant(u64::MAX);
+        };
+        Instant(counter)
+    }
+}
```
jaybosamiya-ms: I recommend against adding this `From<...> for Instant` interface.  It essentially places an "escape hatch" outside the range of the regular platform interfaces.

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
```diff
@@ -86,6 +86,16 @@ fn register_seccomp_filter() {
...
 
+    fn backdoor_cond_on_arg(arg_number: u8) -> SeccompCondition {
+        SeccompCondition::new(
+            arg_number,
+            SeccompCmpArgLen::Qword,
+            SeccompCmpOp::Eq,
+            super::SYSCALL_ARG_MAGIC as u64,
+        )
+        .unwrap()
+    }
```
jaybosamiya-ms: This is almost identical to the function above, so we might want to update `backdoor_on_arg` to use `backdoor_cond_on_arg` (i.e., keep this, but also update the one above to use it)

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -470,64 +469,41 @@ impl RawMutex {
...
-        let result = loop {
-            // Check if value changed before waiting
-            if self.inner.load(SeqCst) != val {
-                break Err(ImmediatelyWokenUp);
+        // Compute timeout in ms
+        let timeout_ms = match timeout {
+            None => Win32_Threading::INFINITE, // no timeout
+            Some(timeout) => {
+                let ms = timeout.as_millis();
+                ms.min(u128::from(u32::MAX - 1)).truncate()
```
jaybosamiya-ms: Nit: Technically, in this case, you need to make sure you have not accidentally let the value be equivalent to `Win32_Threading::INFINITE` (I don't expect this to be an issue in practice but nonetheless).

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -592,11 +583,89 @@ pub(crate) fn sys_sched_getaffinity(pid: Option<i32>) -> CpuSet {
...
+    }
+
+    let res = match arg {
+        FutexArgs::WAKE { addr, flags, count } => {
+            warn_shared_futex!(flags);
+            let Some(count) = core::num::NonZeroU32::new(count) else {
+                return Ok(0);
+            };
+            let futex_manager = crate::litebox_futex_manager();
+            futex_manager.wake(addr, count, Some(core::num::NonZeroU32::MAX))? as usize
```
jaybosamiya-ms: While it is not _incorrect_ to send the `Some(max)` value, the expected value here is actually a `None`

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -592,11 +583,89 @@ pub(crate) fn sys_sched_getaffinity(pid: Option<i32>) -> CpuSet {
...
+        FutexArgs::WAIT {
+            addr,
+            flags,
+            val,
+            timeout,
+        } => {
+            warn_shared_futex!(flags);
+            let futex_manager = crate::litebox_futex_manager();
+            let timeout = timeout.map(get_timeout).transpose()?.map(Into::into);
+            futex_manager.wait(addr, val, timeout, Some(core::num::NonZeroU32::MAX))?;
```
jaybosamiya-ms: Similar bitset

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -592,11 +583,89 @@ pub(crate) fn sys_sched_getaffinity(pid: Option<i32>) -> CpuSet {
...
+            {
+                unimplemented!("FUTEX_CLOCK_REALTIME is not supported yet");
+            }
+            let timeout = timeout
+                .map(get_timeout)
+                .transpose()?
+                .map(|ts| {
+                    let abs_time: <litebox_platform_multiplex::Platform as litebox::platform::TimeProvider>::Instant = ts.into();
+                    let now = litebox_platform_multiplex::platform().now();
+                    abs_time.duration_since(&now)
```
jaybosamiya-ms: Rather than doing it this way, I think the conversion should happen the other direction, specifically, convert `now()` to a timespec (we already have support via the stuff implemented for clock/time), and then subtract.  That way, we do not depend on the ability to convert _to_ an `Instant`.  I do not think we want to force platforms to be able to create an `Instant` from scratch, purely from a design perspective.

The other advantage to this subtraction-based approach is that it will trivially also help support `FUTEX_CLOCK_REALTIME` since we already have that conversion built in.

jaybosamiya-ms:
Generally looks good to me, the biggest comment is the `From<...> for Instant`, which we don't actually want our shim layer to depend upon, since it is a silent interface that is not exposed at the platform layer.  I've suggested an alternative implementation for it in the inline review comment which should work instead

litebox_common_linux/src/lib.rs
CvvT: Good point! fixed.

litebox_platform_linux_userland/src/syscall_intercept/systrap.rs
CvvT: Good catch, fixed!

litebox_platform_windows_userland/src/lib.rs
CvvT: I just checked: `Win32_Threading::INFINITE == u32::MAX`. I replaced `u32::MAX` with `Win32_Threading::INFINITE` so that it is clearer.

litebox_shim_linux/src/syscalls/process.rs
CvvT: Oh, I missed that part. Fixed.

litebox_shim_linux/src/syscalls/process.rs
CvvT: Fixed, thanks!

litebox_platform_linux_kernel/src/lib.rs
CvvT: Removed

litebox_shim_linux/src/syscalls/process.rs
CvvT: Good point! That's much easier. I also added support for CLOCK_REALTIME.

Issue #301, opened 2025-08-27
Syscall rewriter failed on /lib/ld-linux.so.2


CvvT:
```
cargo run -p litebox_syscall_rewriter /lib/ld-linux.so.2 -o ld.hooked
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/litebox_syscall_rewriter /lib/ld-linux.so.2 -o ld.hooked`
Error: insufficient bytes before or after syscall at 0x1c810
```

_Originally posted by @CvvT in https://github.com/microsoft/litebox/issues/300#issuecomment-3226199089_

CvvT:
Here is the related asm code:

```
1c809:       e9 19 fb ff ff          jmp    1c327 <_dl_catch_error@@GLIBC_PRIVATE+0x2887>
   1c80e:       66 90                   xchg   %ax,%ax
   1c810:       cd 80                   int    $0x80
   1c812:       c3                      ret
   1c813:       8d b4 26 00 00 00 00    lea    0x0(%esi,%eiz,1),%esi
```

wdcui:
I searched 1c80e and 1c810 in the assembly code, and didn't find any instruction that would reach these two instructions. My current plan is to modify the syscall_rewriter code to ignore this case (just like how we ignore __libc_start_call_main).  An alternative would be to patch the 3 instructions together which has exactly 5 bytes, but I'm not sure if we need to do it now. Thoughts?

PR #305, opened 2025-08-27
Handle special cases in syscall rewriter

wdcui:
This PR resolves the issue #301 by moving instructions before and after a syscall instruction.  It also addresses a previous case where there is a relative jump back to the instruction before the syscall instruction.

### DIFF
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -631,7 +631,7 @@ fn hook_syscall_and_after(
 #[allow(clippy::too_many_arguments)]
 fn hook_syscall_before_and_after(
     arch: Arch,
-    _control_transfer_targets: &HashSet<u64>,
+    control_transfer_targets: &HashSet<u64>,
     section_base_addr: u64,
     section_data: &mut [u8],
     trampoline_base_addr: u64,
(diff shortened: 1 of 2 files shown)


litebox_syscall_rewriter/src/lib.rs
```diff
@@ -628,3 +627,126 @@ fn hook_syscall_and_after(
 
     Ok(())
 }
+
+#[allow(clippy::too_many_arguments)]
+fn hook_syscall_before_and_after(
+    arch: Arch,
+    _control_transfer_targets: &HashSet<u64>,
```
CvvT: Why we don't need to use it to check if any instruction is a control transfer target?

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -628,3 +627,126 @@ fn hook_syscall_and_after(
...
+
+    // Copy the next inst
+    trampoline_data.extend_from_slice(next_inst.bytes());
+
+    // Add jmp back to original after syscall
+    let return_addr = next_inst.address() + next_inst.bytes().len() as u64;
+    let jmp_back_offset = i64::try_from(return_addr).unwrap()
+        - i64::try_from(trampoline_base_addr + trampoline_data.len() as u64 + 5).unwrap();
+    trampoline_data.push(0xE9);
+    trampoline_data.extend_from_slice(&(i32::try_from(jmp_back_offset).unwrap().to_le_bytes()));
```
CvvT: If the next instruction is `jmp` or `ret`, we don't need to jump back. It doesn't hurt to add it though.

PR #306, opened 2025-08-27
Support for KMPP TA

sangho2:
This PR adds initial support for the KMPP TA. It partially implements five OP-TEE syscalls
(`GetProperty`, `GetPropertyNameToIndex`, `OpenTaSession`, `CloseTaSession`, `InvokeTaCommand`)
and related data structures to provide some essential functionalities that the KMPP TA assumes
like `PTA_SYSTEM_DERIVE_TA_UNIQUE_KEY` to generate a TA unique key to protect the keys that
the KMPP TA generates or imports from a client app.
We confirm that this PR works for a KMPP TA built with `--no-asm` option and
SymCrypt with a minor patch.

### DIFF
--- a/litebox_runner_optee_on_linux_userland/tests/kmpp-ta-cmds.json
+++ b/litebox_runner_optee_on_linux_userland/tests/kmpp-ta-cmds.json
@@ -44,3 +44,4 @@
     "func_id": "close_session"
   }
 ]
+


litebox_common_optee/src/lib.rs
```diff
@@ -412,13 +476,21 @@ pub enum TeeAttributeType {
...
 /// TAs, cryptographic keys, and more.
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, PartialEq, Default, Debug)]
 #[repr(C)]
 pub struct TeeUuid {
-    time_low: u32,
-    time_mid: u16,
-    time_hi_and_version: u16,
-    clock_seq_and_node: [u8; 8],
+    pub time_low: u32,
```
wdcui, line 482: Why is it called `*time*_low`?

litebox_shim_optee/src/syscalls/tee.rs
```diff
@@ -76,6 +90,152 @@ pub fn sys_panic(code: usize) -> ! {
...
+        // `open_ta_session` syscall lets a user-mode TA open a session with a PTA which provides
+        // several import services (it works as a proxy for extra system calls).
+        unsafe {
+            ta_sess_id
+                .write_at_offset(0, get_pta_session_id())
+                .ok_or(TeeResult::AccessDenied)?;
+        }
+        Ok(())
+    } else {
+        // `open_ta_session` syscall lets a user-mode TA open a session another user-mode TA
```
wdcui: "open a session *to* another user-mode TA"?

litebox_common_optee/src/lib.rs
sangho2: OP-TEE uses the field names from [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.2). For example, `time_low` and `time_mid` stand for the low and middle fields of the timestamp.

PR #308, opened 2025-08-29
Add `mremap()` syscall support

Icegrave0391:
Supported on Linux


Icegrave0391:
@jaybosamiya-ms @CvvT Could you review this PR again?

According to the last discussion with @CvvT , the method `remap_pages` in platform's PageManagementProvider would be better changed to `move_and_expand_pages` for clarity.

PR #309, opened 2025-08-29
Fix 32 bit arch

CvvT:
1. Similar to #307, fix stack alignment issue for x86. Basically, stack pointer has to be 16-byte alignment before the call instruction.
2. Our rtld_audit implementation does not support 32 bits yet, add some check to give us warning.
3. Change `SyscallReturnType` from isize to usize as `mmap` may return an address outside the isize range.


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -24,7 +24,7 @@ extern crate alloc;
 cfg_if::cfg_if! {
     if #[cfg(feature = "linux_syscall")] {
         use litebox_common_linux::SyscallRequest;
-        pub type SyscallReturnType = isize;
+        pub type SyscallReturnType = usize;
```
wdcui, line 27: Why do you need this change in this PR?

PR #310, opened 2025-08-29
Enable `debug` format string printing for SyscallRequest

Icegrave0391:


jaybosamiya-ms:
Closing as superseded by #312, but thanks for pointing out this bit that we could improve!

PR #312, opened 2025-08-29
Add some more `Debug` impls/derives

jaybosamiya-ms:
This PR introduces `Debug` to `SyscallRequest` and related enums (thanks @Icegrave0391 for pointing out that this would be helpful in #310)

To be able to do this nicely, we actually need make a minor update the platform interface for the raw pointer types. Specifically, we need those pointers to support `Debug`.  Not including this was partly an oversight.  Having it around makes it nicer to to debug things.


Icegrave0391:
Thanks! That is really helpful.

PR #314, opened 2025-08-30
Revise KMPP TA unit test

sangho2:
This PR revises the unit test for the KMPP TA. It
- Generate and import a new RSA key pair
- Open the generated private RSA key
- Sign a message with the key

In addition, we slightly modify the KMPP TA binary to work
around one sanity check which can be fixed with
end-to-end implementation later.


litebox_runner_optee_on_linux_userland/src/lib.rs
```diff
@@ -231,6 +231,17 @@ fn populate_optee_command_queue(session_id: u32, ta_commands: &[TaCommandBase64]
             TaEntryFunc::InvokeCommand => UteeEntryFunc::InvokeCommand,
         };
 
+        // special handling for the KMPP TA whose `OpenSession` expects the session ID
```
wdcui, line 234: Why is the session ID not in the params in the first place?

sangho2: `litebox_runner_optee_on_linux_userland` currently uses the PID as the session ID and we cannot pre-embed such a dynamic value in our static command sequence for unit testing.

PR #315, opened 2025-08-30
Multiple clippy fixups for Windows + update exit handling

jaybosamiya-ms:
This PR is largely focused around clippy fixups, but along the way (due to the clippy warnings that showed up on Windows), I noticed that our Linux shim exit usage was accidentally assuming that it knows the actual underlying type of the `ExitCode`.  Thus, I've also fixed that up in this PR.  Well, partly---this actually adds a TODO on how we actually want to expose the exit code to the outside world; the `ExitProvider` design explicitly does not support arbitrary exit codes, since there are platforms where this might even exist, but we then need to think about how we handle this, or if it even matters, if we are assuming single-process world). Either way, that TODO should allow us to look into this later.

This PR is a blocker to us enabling Windows CI support (#298).

### DIFF
--- a/litebox_runner_linux_on_windows_userland/tests/loader.rs
+++ b/litebox_runner_linux_on_windows_userland/tests/loader.rs
@@ -147,7 +147,7 @@ fn test_static_linked_prog_with_rewriter() {
     test_dir.push("tests/test-bins");
 
     let prog_name = "hello_world_static";
-    let prog_name_hooked = format!("{}.hooked", prog_name);
+    let prog_name_hooked = format!("{prog_name}.hooked");
 
     let path = test_dir.join(prog_name);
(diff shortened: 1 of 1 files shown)


wdcui:
Why didn't our CI detect these clippy warnings before?

jaybosamiya-ms:
Our CI was not running on Windows.  After #298 is merged, it will add Windows to our CI, but I wanted to separate those changes from the actual changes here.  

Since things have changed since then, I'll still need to do a few more fixes as part of #298 but the main reason I split this was due to the exit handling part, which I didn't want to keep in the same commit on `main` as part of the Windows CI stuff mostly from an understandability perspective for future stuff.

jaybosamiya-ms:
These changes are primarily semantics-preserving (except the exit thing I mention in the main PR description) and are blocking #298 so am merging.

PR #317, opened 2025-09-03
Support `mkdir()` syscall

Icegrave0391:


jaybosamiya-ms:
I think you've manually copied the changes, which is why _despite_ `jayb/tar-ro-file-status` being the base, GitHub is confused about the diff for the review since it sees separate commits on it.  You may wish to rebase on top of the branch to have it work correctly.

Icegrave0391:
@jaybosamiya-ms only showing correct changes now

litebox_shim_linux/src/lib.rs
```diff
@@ -330,6 +330,11 @@ pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> usize {
         SyscallRequest::Lseek { fd, offset, whence } => {
             syscalls::file::sys_lseek(fd, offset, whence)
         }
+        SyscallRequest::Mkdir { pathname, mode } => {
+            pathname.to_cstring().map_or(Err(Errno::EFAULT), |path| {
```
jaybosamiya-ms: I think this should actually be an `EINVAL` not `EFAULT` if the `pathname.to_cstring()` fails.

Icegrave0391: Fixed.

PR #313, opened 2025-08-29
Support `open(O_TRUNC)` with testcases

Icegrave0391:

PR #318, opened 2025-09-04
Add file truncation support

jaybosamiya-ms:
This PR implements support for `O_TRUNC` at the shim layer, by introducing a new `truncate` function at the LiteBox layer.  This leads to a simpler design for truncation at the layered FS, and makes sure there are no entry leaks upon truncation.

This is similar to #313 by @Icegrave0391 but takes the different approach I was mentioning, in order to keep things simpler.

### DIFF
--- a/litebox/src/fs/in_mem.rs
+++ b/litebox/src/fs/in_mem.rs
@@ -132,6 +132,7 @@ impl<Platform: sync::RawSyncPrimitivesProvider> FileSystem<Platform> {
 }
 
 impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem<Platform> {
+    #[expect(clippy::too_many_lines)]
     fn open(
         &self,
         path: impl crate::path::Arg,


litebox/src/fs/devices/stdio.rs
```diff
@@ -127,7 +132,11 @@ impl<Platform: crate::sync::RawSyncPrimitivesProvider + crate::platform::StdioPr
         if nonblocking {
             unimplemented!("Non-blocking I/O is not supported for stdio streams");
         }
-        Ok(self.litebox.descriptor_table_mut().insert(stream))
+        let fd = self.litebox.descriptor_table_mut().insert(stream);
+        if truncate {
```
CvvT, line 136: According to the man page, O_TRUNC should be ignored for terminal device file

```
O_TRUNC
              If the file already exists and is a regular file and the
              access mode allows writing (i.e., is O_RDWR or O_WRONLY) it
              will be truncated to length 0.  If the file is a FIFO or
              terminal device file, the O_TRUNC flag is ignored.
              Otherwise, the effect of O_TRUNC is unspecified.
```

litebox/src/fs/devices/stdio.rs
```diff
@@ -189,6 +198,13 @@ impl<Platform: crate::sync::RawSyncPrimitivesProvider + crate::platform::StdioPr
         unimplemented!()
     }
 
+    fn truncate(&self, fd: &FileFd<Platform>) -> Result<(), TruncateError> {
+        match self.litebox.descriptor_table().get_entry(fd).entry {
+            StdioStream::Stdin => Err(TruncateError::NotForWriting),
+            StdioStream::Stdout | StdioStream::Stderr => Ok(()),
+        }
+    }
```
CvvT: I asked ChatGPT to generate a test case that calls `truncate` on `stdout`. It seems that `open` STDOUT with `O_TRUNC` is fine (because the flag is ignored) but `truncate` returns `EINVAL`.

```
#include <stdio.h>
#include <unistd.h>
#include <fcntl.h>

int main(void) {
    // Write something to stdout first
    printf("Hello, world!\n");
    fflush(stdout);

    // Try to truncate stdout (fd = 1) to zero length
    if (ftruncate(STDOUT_FILENO, 0) == -1) {
        perror("ftruncate");
    } else {
        printf("Successfully truncated stdout\n");
    }
    return 0;
}
```

litebox/src/fs/in_mem.rs
```diff
@@ -212,27 +213,30 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
             }
-            Entry::Dir(dir) => Ok(self.litebox.descriptor_table_mut().insert(Descriptor::Dir {
+            Entry::Dir(dir) => self.litebox.descriptor_table_mut().insert(Descriptor::Dir {
                 dir: dir.clone(),
                 metadata: AnyMap::new(),
-            })),
+            }),
+        };
+        if flags.contains(OFlags::TRUNC) {
+            self.truncate(&fd)?;
```
CvvT: Should we close it if it fails?

litebox/src/fs/layered.rs
```diff
@@ -485,12 +498,20 @@ impl<
...
             entry,
             position: 0.into(),
-        }))
+        });
+        if original_flags.contains(OFlags::TRUNC) {
+            // The only scenario where we need to manually trigger truncation is when a file does
+            // not exist at the upper level but exists at the lower level; in that case, our
+            // `truncate` functionality (at the layered FS itself) should correctly migrate things
+            // over and handle them.
+            self.truncate(&fd)?;
```
CvvT: Similar, should we close it if it fails?

litebox/src/fs/tar_ro.rs
```diff
@@ -183,26 +184,29 @@ impl<Platform: sync::RawSyncPrimitivesProvider> super::FileSystem for FileSystem
...
             // it is a dir
-            Ok(self.litebox.descriptor_table_mut().insert(Descriptor::Dir {
+            self.litebox.descriptor_table_mut().insert(Descriptor::Dir {
                 path: path.to_owned(),
                 metadata: AnyMap::new(),
-            }))
+            })
+        };
+        if flags.contains(OFlags::TRUNC) {
+            self.truncate(&fd)?;
```
CvvT: Similar, should we close it if it fails?

litebox/src/fs/devices/stdio.rs
jaybosamiya-ms: Updated, thanks!

litebox/src/fs/devices/stdio.rs
jaybosamiya-ms: Good catch!  I've introduced a new error code for this, so that we can correctly handle this; also planning ahead for `ftruncate`, it is likely helpful to add partial length truncations, so I'm adding that too.

litebox/src/fs/in_mem.rs
jaybosamiya-ms: Good catch!  Our `panic_on_unclosed_fd_drop` feature would have caught this during testing, but we don't actually test that often with it (maybe we should enable that at some point for CI at least).  Updated here and in other spots.

litebox/src/fs/layered.rs
jaybosamiya-ms: Fixed

litebox/src/fs/tar_ro.rs
jaybosamiya-ms: Fixed

PR #320, opened 2025-09-05
Use the same GDT entry number for all threads

CvvT:
I have a simple test that failed at `set_thread_local_storage` because of no free GDT entry:

```
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <unistd.h>

#define NUM_THREADS 2

// Thread function
void* thread_func(void* arg) {
    int thread_id = *(int*)arg;  // get thread ID passed in
    printf("Hello from thread %d (TID: %lu)\n", thread_id, pthread_self());
}

int main() {
    pthread_t threads[NUM_THREADS];
    int thread_ids[NUM_THREADS];
    int rc;

    for (int i = 0; i < NUM_THREADS; i++) {
        thread_ids[i] = i;
        rc = pthread_create(&threads[i], NULL, thread_func, &thread_ids[i]);
        if (rc) {
            fprintf(stderr, "Error creating thread %d\n", i);
            exit(EXIT_FAILURE);
        }
    }

    // Wait for all threads to finish
    for (int i = 0; i < NUM_THREADS; i++) {
        pthread_join(threads[i], NULL);
    }

    printf("All threads finished!\n");
    return 0;
}
```

It seems that a child thread would inherit its parent's GDT. Thus, instead of requesting for a free GDT entry, we should overwrite the same entry.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -1510,6 +1510,7 @@ impl litebox::platform::ThreadLocalStorageProvider for LinuxUserland {
         };
         set_thread_area(user_desc_ptr).expect("Failed to set thread area for TLS");
 
+        assert!(user_desc.entry_number <= 0xfff);
         self.tls_entry_number.store(
             user_desc.entry_number & 0xfff,
             core::sync::atomic::Ordering::Relaxed,


CvvT:
> Huh, it runs out of free entries after just trying to spin up two threads?
> 
> Fix seems reasonable to me, fwiw.
> 
> Minor comment would be to also add a quick comment about the `0xfff` constant. Or actually even a `assert!(user_desc.entry_number < 0xfff)` might work. The core worry is that otherwise it might silently mask things off (specifically, the first time around, it would be sending it a `-1`, which as per the man page of `set_thread_area`, if no free entry is found, it might not update the `entry_number`, which would then get masked by the `0xfff` to just silently be 0xfff, no?). I don't understand enough to know if this is even possible in practice, but nonetheless, a simple `assert` is a safer bet as I understand it.

Each thread only has three slots available: https://elixir.bootlin.com/linux/v6.16.4/source/arch/x86/include/asm/segment.h#L229

Previously, I thought these per-thread slots (https://elixir.bootlin.com/linux/v6.16.4/source/arch/x86/include/asm/processor.h#L450) would be initialized to zero for a new thread, but it looks like Linux kernel copies the entries from parent to its child.

If no free entry is found, `set_thread_area` would return an error. Currently, I use `expect` to assume it always succeed so I didn't check `entry_number`. I forgot why I need to do `entry_number & 0xfff`. For now, I just added the `assert` as you suggested.

PR #322, opened 2025-09-07
Implement IPInterfaceProvider for WindowsUserland

Icegrave0391:
- Use wintun to implement IPInterfaceProvider
- Added testcases

### DIFF
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -219,7 +219,8 @@ impl WindowsUserland {
             let adapter = match wintun::Adapter::open(&wintun, adapter_name) {
                 Ok(adapter) => adapter,
                 Err(_) => {
-                    // If opening failed (adapter doesn't exist), create a new one
+                    // If opening failed (adapter doesn't exist), create a new one.
+                    // TODO: The tunnel_type "LiteBox" is a placeholder name for now.
                     wintun::Adapter::create(&wintun, adapter_name, "LiteBox", None)
                         .expect("Failed to create WinTun adapter")
                 }
@@ -601,9 +602,8 @@ impl litebox::platform::IPInterfaceProvider for WindowsUserland {
 
         // Copy the IP packet data into the WinTun packet
         let packet_bytes = wintun_packet.bytes_mut();
-        if packet_bytes.len() != packet.len() {
-            unimplemented!("unexpected size {}", packet_bytes.len())
-        }
+
+        assert_eq!(packet_bytes.len(), packet.len());
         packet_bytes[..packet.len()].copy_from_slice(packet);
 
         // Send the packet through WinTun
@@ -1401,35 +1401,6 @@ mod tests {
     }
 
     #[test]
-    fn test_ip_interface_without_wintun() {
-        // Test IP interface functionality when WinTun is not enabled
-        let platform = WindowsUserland::new(None);
-
-        // Create a sample IP packet (minimal IPv4 header)
-        let mut test_packet = [0u8; 64];
-        test_packet[0] = 0x45; // Version 4, Header length 20 bytes
-        test_packet[9] = 0x06; // Protocol: TCP
-
-        // Test send_ip_packet without WinTun - should panic with unimplemented
-        let result = std::panic::catch_unwind(|| platform.send_ip_packet(&test_packet));
-        assert!(
-            result.is_err(),
-            "send_ip_packet should panic when WinTun is not available"
-        );
-
-        // Test receive_ip_packet without WinTun - should panic with unimplemented
-        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
-            let mut receive_buffer = [0u8; 1500];
-            platform.receive_ip_packet(&mut receive_buffer)
-        }));
-        assert!(
-            result.is_err(),
-            "receive_ip_packet should panic when WinTun is not available"
-        );
-    }
-
-    #[test]
-    #[ignore] // This test requires Administrator privileges and WinTun driver to be installed
     fn test_ip_interface_with_wintun() {
         // Test IP interface functionality with WinTun enabled
         // Note: This test requires Administrator privileges to create network adapters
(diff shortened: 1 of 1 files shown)


Icegrave0391:
It seems that `cargo clippy` fails due to the source code of the used package `wintun`.

Icegrave0391:
@jaybosamiya-ms I fixed the CI by only including the dependency into `target=Windows`

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -117,6 +118,8 @@ pub struct WindowsUserland {
     tls_slot: TlsSlot,
     reserved_pages: alloc::vec::Vec<core::ops::Range<usize>>,
     sys_info: std::sync::RwLock<Win32_SysInfo::SYSTEM_INFO>,
+    /// WinTun session for networking
+    wintun_session: std::sync::RwLock<Option<Arc<wintun::Session>>>,
```
jaybosamiya-ms, line 122: I am not sure I understand why this is an `RwLock<Option<Arc<`.  Is this solely because of the exit handling?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -204,10 +210,34 @@ impl WindowsUserland {
...
+        let wintun_session = wintun_adapter_name.map(|adapter_name| {
+            // Load the WinTun library from the system
+            let wintun = unsafe { wintun::load().expect("Failed to load WinTun library") };
+
+            // Try to open an existing adapter, or create a new one
+            let adapter = match wintun::Adapter::open(&wintun, adapter_name) {
+                Ok(adapter) => adapter,
+                Err(_) => {
+                    // If opening failed (adapter doesn't exist), create a new one
+                    wintun::Adapter::create(&wintun, adapter_name, "LiteBox", None)
```
jaybosamiya-ms, line 224: Not sure why the tunner type is `LiteBox` here, please include a comment

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -547,20 +585,70 @@ impl litebox::platform::RawMutex for RawMutex {
...
+                // Since SendError is an empty enum, we can't create values of it
+                // We'll panic if allocation fails for now, similar to Linux implementation
+                panic!("Failed to allocate WinTun send packet")
+            })?;
+
+        // Copy the IP packet data into the WinTun packet
+        let packet_bytes = wintun_packet.bytes_mut();
+        if packet_bytes.len() != packet.len() {
+            unimplemented!("unexpected size {}", packet_bytes.len())
+        }
```
jaybosamiya-ms: I think this should just be an `assert_eq!(..., ...);` unless you have good reason to believe that the two might differ

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -547,20 +585,70 @@ impl litebox::platform::RawMutex for RawMutex {
...
-        );
+        let session_guard = self.wintun_session.read().unwrap();
+        let Some(session) = session_guard.as_ref() else {
+            unimplemented!("networking without WinTun is unimplemented")
+        };
+
+        // Try to receive a packet from WinTun (non-blocking)
+        session
+            .try_receive()
+            .map_err(|_| litebox::platform::ReceiveError::WouldBlock)
```
jaybosamiya-ms, line 627: I don't think this is correct; the errors that can happen here are not a "this would block" but instead these: https://docs.rs/wintun/latest/wintun/enum.Error.html

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -1311,4 +1399,406 @@ mod tests {
...
+        // Test receive_ip_packet without WinTun - should panic with unimplemented
+        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+            let mut receive_buffer = [0u8; 1500];
+            platform.receive_ip_packet(&mut receive_buffer)
+        }));
+        assert!(
+            result.is_err(),
+            "receive_ip_packet should panic when WinTun is not available"
+        );
+    }
```
jaybosamiya-ms: I don't see how this negative test is particularly helpful.  In the future, if we ever did stuff without wintun, then this test would fail, but that would not be a bad thing, no?  Probably can be deleted

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -1311,4 +1399,406 @@ mod tests {
...
+
+    #[test]
+    #[ignore] // This test requires Administrator privileges and WinTun driver to be installed
```
jaybosamiya-ms: Does this actually fail on CI?  I'm not fully sure how it helps to add a whole bunch of ignored tests.  You can see the Linux one we actually do extra things on the CI to make sure we can run the CI tests.  Is that possible here?

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: fixed

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: fixed

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: I think currently we only have one WouldBlock error in `litebox::platform::ReceiveError`. We may want to add more typt to the platform's error

litebox_platform_windows_userland/src/lib.rs
Icegrave0391: removed

litebox_platform_windows_userland/src/lib.rs
jaybosamiya-ms: For now, you can just throw in that case

CvvT:
It looks like `wintun` requires admin privileges to create an adapter and it's per-process. I don't think we'd like to let untrusted code run with admin privileges. We could create a service as a proxy with admin privilege and then LiteBox can communicate with it via IPC. Alternatively, we could use [tap](https://github.com/OpenVPN/tap-windows6) on Windows, which is cleaner IMO. Also, we have discussed switching to macvlan + tap for Linux. I will try to see how much work we need for using tap.

PR #324, opened 2025-09-08
Support syscall execve

CvvT:
This PR makes the following changes to support `execve`:

1. Similar to Linux, `PageManager` is moved to per-thread `Task` and shared across threads. Network and Filesystem remain as global static singleton, which might be changed in the future.
2. `execve` supposes to clear all TLS slots on x86, we reserve two for LiteBox and only the last one (only 3 slots are available).
3. Split `execve` into two parts: 1) the shim layer only prepares all the arguments and resets some states (e.g., unmap memory, close files); and 2) the runner is responsible for loading the target program (runner already has this functionality; it just needs to register it to the shim layer).

### DIFF
--- a/litebox/src/mm/mod.rs
+++ b/litebox/src/mm/mod.rs
@@ -345,6 +345,11 @@ where
         Ok(brk)
     }
 
+    pub unsafe fn reset_brk(&self) {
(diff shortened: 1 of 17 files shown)


CvvT:
Replaced by #328

PR #328, opened 2025-09-09
Support syscall execve

CvvT:
This PR makes the following changes to support execve:

1. Similar to Linux, PageManager is moved to per-thread Task and shared across threads. Network and Filesystem remain as global static singleton, which might be changed in the future.
2. execve supposes to clear all TLS slots on x86, we reserve two for LiteBox and only the last one (only 3 slots are available) is cleared on `execve`.
4. Split execve into two parts: 1) the shim layer only prepares all the arguments and resets some states (e.g., unmap memory, close files); and 2) the runner is responsible for loading the target program (runner already has this functionality; it just needs to register it to the shim layer).

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1857,7 +1857,7 @@ impl<'a, Platform: litebox::platform::RawPointerProvider> SyscallRequest<'a, Pla
     /// is allowed to panic upon receiving a syscall number (or arguments) that it does not know how
     /// to handle.
     #[expect(clippy::too_many_lines)]
-    pub fn try_from_raw(syscall_number: usize, ctx: &'a mut PtRegs) -> Result<Self, errno::Errno> {
+    pub fn try_from_raw(syscall_number: usize, ctx: &'a PtRegs) -> Result<Self, errno::Errno> {
         // sys_req! is a convenience macro that automatically takes the correct numbered arguments
         // (in the order of field specification); due to some Rust restrictions, we need to manually
         // specify pointers by adding the `:*` to that field, but otherwise everything else about
--- a/litebox_platform_freebsd_userland/src/lib.rs
+++ b/litebox_platform_freebsd_userland/src/lib.rs
@@ -47,6 +47,7 @@ impl FreeBSDUserland {
         let platform = Self {
             reserved_pages: Self::read_proc_self_maps(),
         };
+
         platform.set_init_tls();
         Box::leak(Box::new(platform))
     }
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -510,6 +510,8 @@ impl Descriptor {
     }
 
     pub(crate) fn get_file_descriptor_flags(&self) -> FileDescriptorFlags {
+        // Currently, only one such flag is defined: FD_CLOEXEC, the close-on-exec flag.
+        // See https://www.man7.org/linux/man-pages/man2/F_GETFD.2const.html
         match self {
             Descriptor::File(file) => litebox_fs()
                 .with_metadata(file, |flags: &FileDescriptorFlags| *flags)


litebox/src/mm/mod.rs
```diff
@@ -345,6 +345,33 @@ where
...
     }
 
+    /// Release memory mappings that satisfy the given condition and reset the program break.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the released memory regions are no longer used.
+    pub unsafe fn release_memory(
+        &self,
+        f: fn(Range<usize>, VmFlags) -> bool,
```
jaybosamiya-ms: Nit: calling this `release_condition` or `releasable` or similar might make the polarity clearer.

litebox_common_linux/src/lib.rs
```diff
@@ -1119,27 +1119,37 @@ pub type ThreadLocalDescriptor = u8;
 #[cfg(target_arch = "x86")]
 pub type ThreadLocalDescriptor = UserDesc;
 
-pub struct NewThreadArgs<Platform: litebox::platform::RawPointerProvider> {
+pub struct NewThreadArgs<Platform, const ALIGN: usize = { litebox::mm::linux::PAGE_SIZE }>
```
jaybosamiya-ms: I'm not fully sure why all the structs here have the default assigned to them.  Are we using the non-default anywhere?  If so, why not just keep `const ALIGN: usize` _without_ the default, so that users are forced to pick the default.  Or if we are always using the default, then why provide this as an argument here?

Is it the case that we _almost_ always (but not always) want `ALIGN = PAGE_SIZE` but there are some cases where it might not be?

litebox/src/platform/mod.rs
```diff
@@ -671,4 +671,10 @@ pub trait ThreadLocalStorageProvider {
     /// Panics if TLS is not set yet.
     /// Panics if TLS is being used by [`Self::with_thread_local_storage_mut`].
     fn release_thread_local_storage(&self) -> Self::ThreadLocalStorage;
+
+    /// Clear any guest thread-local storage state for the current thread.
+    ///
+    /// This is used to help emulate certain syscalls (e.g., `execve`) that clear TLS,
+    /// which is necessary for some platforms (e.g., x86 Linux).
+    fn clear_guest_thread_local_storage(&self);
```
jaybosamiya-ms: We might handle this better by making it a `#[cfg(target_arch = "x86")]` maybe?  In that case it is very clear that x64 targets don't need to care about this?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -1017,6 +1019,8 @@ impl litebox::platform::ThreadLocalStorageProvider for FreeBSDUserland {
         tls.borrowed = false; // Mark as not borrowed anymore
         res
     }
+
+    fn clear_guest_thread_local_storage(&self) {}
```
jaybosamiya-ms: Should this body be `unreachable!()` instead of just empty?

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -141,6 +146,8 @@ impl litebox::platform::ThreadLocalStorageProvider for SnpLinuxKernel {
         current_task.tls = ::core::ptr::null_mut();
         *tls
     }
+
+    fn clear_guest_thread_local_storage(&self) {}
```
jaybosamiya-ms: Similar

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -172,8 +178,9 @@ impl LinuxUserland {
             // u32::MAX (i.e., -1) means not allocated yet
             tls_entry_number: AtomicU32::new(u32::MAX),
         };
-        platform.set_init_tls();
-        Box::leak(Box::new(platform))
+        let platform = Box::leak(Box::new(platform));
+        platform.set_init_tls(&litebox::LiteBox::new(platform));
```
jaybosamiya-ms: This drops the `LiteBox` object right after making it, which seems weird to me.  The platform itself should not be constructing a `LiteBox` object imho.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1556,6 +1589,26 @@ impl litebox::platform::ThreadLocalStorageProvider for LinuxUserland {
         tls.borrowed = false; // mark as not borrowed anymore
         ret
     }
+
+    #[cfg(target_arch = "x86_64")]
+    fn clear_guest_thread_local_storage(&self) {}
```
jaybosamiya-ms: Similar.  Although if we have added the `#[cfg(arch ... x86)]` at the trait that I mentioned, then we wouldn't even need this.

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -62,6 +64,9 @@ pub enum InterceptionBackend {
     Rewriter,
 }
 
+static REQUIRE_RTLD_AUDIT: core::sync::atomic::AtomicBool =
+    core::sync::atomic::AtomicBool::new(false);
```
jaybosamiya-ms, line 76: Why not just pass this as an argument to `load_program`?

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -231,6 +239,24 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
...
+}
+
+fn load_program_wrapper(
+    _ctx: &mut litebox_common_linux::PtRegs,
+    path: &str,
+    argv: alloc::vec::Vec<alloc::ffi::CString>,
+    envp: alloc::vec::Vec<alloc::ffi::CString>,
+) -> Result<(), litebox_common_linux::errno::Errno> {
+    load_program(path, argv, envp)
+}
```
jaybosamiya-ms: This is just ignoring the `PtRegs`, so I am not sure why this needed the mutable ptregs and why the syscall side needed to update it to mutable either.

litebox_shim_linux/src/lib.rs
```diff
@@ -97,9 +97,10 @@ pub fn litebox_fs<'a>() -> &'a LinuxFS {
 }
 
 /// Get the global page manager
-pub fn litebox_page_manager<'a>() -> &'a PageManager<Platform, PAGE_SIZE> {
-    static VMEM: OnceBox<PageManager<Platform, PAGE_SIZE>> = OnceBox::new();
-    VMEM.get_or_init(|| alloc::boxed::Box::new(PageManager::new(litebox())))
+pub fn litebox_page_manager() -> alloc::sync::Arc<PageManager<Platform, { PAGE_SIZE }>> {
+    use litebox::platform::ThreadLocalStorageProvider;
+    litebox_platform_multiplex::platform()
+        .with_thread_local_storage_mut(|tls| tls.current_task.page_manager.clone())
```
jaybosamiya-ms: I think the comment might need to be updated to "Get the thread-local page manager"?

litebox_shim_linux/src/lib.rs
```diff
@@ -758,6 +773,12 @@ pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> usize {
             }
         }
         SyscallRequest::Futex { args } => syscalls::process::sys_futex(args),
+        SyscallRequest::Execve {
+            pathname,
+            argv,
+            envp,
+            ctx,
+        } => syscalls::process::sys_execve(pathname, argv, envp, ctx).map(|()| 0),
```
jaybosamiya-ms: Execve should never return, right?  Might be good to place an `unreachable!()` for the `Ok` case?

litebox_shim_linux/src/lib.rs
```diff
@@ -97,9 +97,10 @@ pub fn litebox_fs<'a>() -> &'a LinuxFS {
 }
 
 /// Get the global page manager
-pub fn litebox_page_manager<'a>() -> &'a PageManager<Platform, PAGE_SIZE> {
-    static VMEM: OnceBox<PageManager<Platform, PAGE_SIZE>> = OnceBox::new();
-    VMEM.get_or_init(|| alloc::boxed::Box::new(PageManager::new(litebox())))
+pub fn litebox_page_manager() -> alloc::sync::Arc<PageManager<Platform, { PAGE_SIZE }>> {
```
jaybosamiya-ms: I know it was `pub` before, but since we're updating things here anyways, we might want to switch this to `pub(crate)` instead?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -675,6 +680,144 @@ pub(crate) fn sys_futex(
     Ok(res)
 }
 
+pub type ExecveCallback = fn(
+    ctx: &mut litebox_common_linux::PtRegs,
+    path: &str,
+    argv: alloc::vec::Vec<alloc::ffi::CString>,
+    envp: alloc::vec::Vec<alloc::ffi::CString>,
+) -> Result<(), Errno>;
```
jaybosamiya-ms: This should return a `Result<!, Errno>`, no?

litebox/src/platform/mod.rs
wdcui, line 680: Nit: why not calling it `clear_thread_local_storage`? Not sure why we use `guest` here.

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: I believe that the core reason for this update you've made is because of making `PageManager` be part of the tls struct, which is something I am not fully sure I understand.  Is this meant to be thread-local or process-local or global?  Your explanation of "Similar to Linux, PageManager is moved to per-thread Task and shared across threads" is mildly confusing, since that implies it is both per-thread but also not per-thread.  I believe you mean for it to be per-(guest)-process; is that correct?  If so, I don't think placing it into TLS is the ideal way, but instead, we might just wish to have the global with a "re-initialize" or "re-create" functionality that is run by the execve functionality?

jaybosamiya-ms:
Thanks Weiteng!  I've added a bunch of comments, but if it is easier to explain offline, we can do that too.  The biggest concerns I have are related to the created-and-dropped `LiteBox` object (the core problem of which is that it breaks some of our layering) and the `&mut PtRegs` which seem superfluous.  Other comments are minor.

litebox_platform_linux_userland/src/lib.rs
wdcui: We don't need this function for 64bit?

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -511,6 +511,26 @@ impl Descriptor {
         };
         Ok(fstat)
     }
+
+    pub(crate) fn get_file_descriptor_flags(&self) -> FileDescriptorFlags {
```
wdcui, line 512: Does this function need a comment saying it currently only supports close_on_exec?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -675,6 +680,144 @@ pub(crate) fn sys_futex(
...
+        };
+        unsafe { tls.current_task.page_manager.release_memory(release) }
+            .expect("failed to release memory mappings");
+    });
+    #[cfg(target_arch = "x86")]
+    litebox_platform_multiplex::platform().clear_guest_thread_local_storage();
+
+    if let Some(callback) = EXECVE_CALLBACK.get() {
+        callback(ctx, path, argv_vec, envp_vec).expect("we already released memory above");
+    }
```
wdcui: Do we need to return an error in the else branch?

litebox/src/mm/mod.rs
CvvT: Good point, fixed.

litebox/src/platform/mod.rs
CvvT: I want to distinguish between host and guest TLS. Host (i.e., LiteBox itself) should call `release_thread_local_storage` to clear its TLS.

CvvT: I'm not sure if any other arch needs to do this as well. Restrict it to x86; we can revisit it later when we support more architectures.

litebox_shim_linux/src/lib.rs
CvvT: It is used by `litebox_runner_snp`:

https://github.com/microsoft/litebox/blob/0c70e44dc669c9fb1f3e4e2227b528c28538e749/litebox_runner_snp/src/main.rs#L25

litebox_common_linux/src/lib.rs
CvvT: We don't have a case for non-default value for now (I'm thinking maybe windows platform should use a different value, e.g., 64KB, but it requires more changes).  For simplicity, just use a fixed value for now.

litebox_platform_freebsd_userland/src/lib.rs
CvvT: Removed it as we restrict it to x86 only and we only support x64 on FreeBSD.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
CvvT: Removed it as we restrict it to x86 only and we only support x64 on snp.

litebox_platform_linux_userland/src/lib.rs
CvvT: Fair, it is unnecessary to put it into per-thread Task. I did it because I would like to run multiple programs concurrently (e.g., on snp), which means we cannot have just one as a static variable. For this PR, I reverted it back; any change should be separate.

litebox_platform_linux_userland/src/lib.rs
CvvT: Removed

CvvT: Because x64 uses `gs.base`, overwriting it never fails. x86 instead use gs selector which requires modifying GDT, and it only has three slots available.

litebox_runner_linux_userland/src/lib.rs
CvvT: This function is registered to `litebox_shim_linux` and would be called when calling `execve`. `litebox_shim_linux` does not save this argument.

litebox_runner_linux_userland/src/lib.rs
CvvT: You're right. I was considering it running in kernel mode. For example, on snp:

https://github.com/microsoft/litebox/blob/eff8ac4d62e313300e12cf194fd9f046abff3d16/litebox_runner_snp/src/main.rs#L164-L166

But on a second thought, probably I can avoid it. So removed it from this PR. I will give more thoughts when implementing it for snp.

litebox_shim_linux/src/lib.rs
CvvT: Reverted the change.

litebox_shim_linux/src/lib.rs

litebox_shim_linux/src/syscalls/process.rs
CvvT: Looks like it is not supported yet?

```
error[E0658]: the `!` type is experimental
   --> litebox_shim_linux/src/syscalls/process.rs:706:13
    |
706 | ) -> Result<!, Errno> {
    |             ^
    |
    = note: see issue #35121 <https://github.com/rust-lang/rust/issues/35121> for more information
```

litebox_shim_linux/src/syscalls/process.rs
CvvT: At this point, it is not possible to recover as we have already unmapped everything. We should add some basic checks before we unmap everything. Left a TODO here.

litebox_shim_linux/src/syscalls/file.rs
CvvT: Currently, only one such flag is defined: FD_CLOEXEC, the close-on-exec flag. See https://www.man7.org/linux/man-pages/man2/F_GETFD.2const.html

litebox_common_linux/src/lib.rs
```diff
@@ -1852,7 +1857,7 @@ impl<'a, Platform: litebox::platform::RawPointerProvider> SyscallRequest<'a, Pla
     /// is allowed to panic upon receiving a syscall number (or arguments) that it does not know how
     /// to handle.
     #[expect(clippy::too_many_lines)]
-    pub fn try_from_raw(syscall_number: usize, ctx: &'a PtRegs) -> Result<Self, errno::Errno> {
+    pub fn try_from_raw(syscall_number: usize, ctx: &'a mut PtRegs) -> Result<Self, errno::Errno> {
```
jaybosamiya-ms: Nit: this mutability change is no longer needed

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -47,7 +47,6 @@ impl FreeBSDUserland {
         let platform = Self {
             reserved_pages: Self::read_proc_self_maps(),
         };
-
```
jaybosamiya-ms: Nit: unnecessary change

litebox_shim_linux/src/syscalls/file.rs
jaybosamiya-ms: I agree with adding a comment (you can point to the manpage too)

litebox_shim_linux/src/syscalls/process.rs
jaybosamiya-ms: Ah, good point, technically https://doc.rust-lang.org/core/convert/enum.Infallible.html is what we use for `!` until `!` gets stabilized.  `!` is actually usable in stable if you just did `-> !`, but seems like `Result<!, ...>` is not supported yet. Oh well.

litebox_shim_linux/src/syscalls/file.rs
CvvT: Added

litebox_common_linux/src/lib.rs
CvvT: Removed

litebox_platform_freebsd_userland/src/lib.rs
CvvT: Fixed

PR #330, opened 2025-09-09
Remove thread test

CvvT:
Just ran into a problem with the following thread test:

```
// Thread function
void* thread_func(void* arg) {
    int thread_id = *(int*)arg;  // get thread ID passed in
    printf("Hello from thread %d (TID: %lu)\n", thread_id, pthread_self());
}

int main() {
    pthread_t threads[NUM_THREADS];
    int thread_ids[NUM_THREADS];
    int rc;

    for (int i = 0; i < NUM_THREADS; i++) {
        thread_ids[i] = i;
        rc = pthread_create(&threads[i], NULL, thread_func, &thread_ids[i]);
        if (rc) {
            fprintf(stderr, "Error creating thread %d\n", i);
            exit(EXIT_FAILURE);
        }
    }

    // Wait for all threads to finish
    for (int i = 0; i < NUM_THREADS; i++) {
        pthread_join(threads[i], NULL);
    }

    printf("All threads finished!\n");
    return 0;
}
```

The new thread may not call exit, resulting in `clear_child_tid` not being cleared (see `CLONE_CHILD_CLEARTID` from the [man page](https://www.man7.org/linux/man-pages/man2/clone.2.html)). This causes the parent thread to hang in `futex_wait` forever (called by `thread_join`). I'm not sure why it calls `exit` sometimes but not always. I will take a look at the implementation [`start_thread`](https://elixir.bootlin.com/glibc/glibc-2.35/source/nptl/pthread_create.c#L335). Before I figure out the root cause, remove the test for now.


CvvT:
ah, I guess the problem may not be caused by the new test but rather a race condition issue on futex implementation. I will experiment more to see if I can reproduce it.

CvvT:
Might be fixed by #335, will re-open if I run into the same problem.

PR #332, opened 2025-09-10
Enable tun for testing

CvvT:
Add another command line option to the runner to allow to enable tun.

### DIFF
--- a/litebox_runner_linux_userland/src/lib.rs
+++ b/litebox_runner_linux_userland/src/lib.rs
@@ -50,11 +50,11 @@ pub struct CliArgs {
         default_value = "seccomp"
     )]
     pub interception_backend: InterceptionBackend,
+    /// Connect to a TUN device with this name
     #[arg(
         long = "tun-device-name",
         requires = "unstable",
-        help_heading = "Unstable Options",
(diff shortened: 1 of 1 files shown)


litebox_runner_linux_userland/src/lib.rs
```diff
@@ -50,6 +50,13 @@ pub struct CliArgs {
         default_value = "seccomp"
     )]
     pub interception_backend: InterceptionBackend,
+    #[arg(
+        long = "tun-device-name",
+        requires = "unstable",
+        help_heading = "Unstable Options",
+        help = "If provided, connect to a TUN device with this name inside the LiteBox instance"
+    )]
+    pub tun_device_name: Option<String>,
```
jaybosamiya-ms: Nit: you don't need to provide a `help = ...` if you give it a doc-string (see the ones above it).  Also, you can make it mildly shorter docstring "connect to a TUN device with this name".  The "If provided" is redundant because `clap` will show optional arguments with the optional flags stuff anyways.

CvvT: I see, thanks!

PR #337, opened 2025-09-11
Support clone syscall

jstarks:
Add support for the legacy clone syscall, where the arguments are passed in registers instead of in memory.


jstarks:
OK, yeah, I wasn't sure about that. Happy to make the change or for you to push on top.

PR #339, opened 2025-09-12
Fix panic

CvvT:


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -38,6 +38,44 @@ pub type SyscallHandler = fn(SyscallRequest<LinuxUserland>) -> SyscallReturnType
...
+                $s.as_ptr() as usize,
+                $s.len(),
+                // Unused by the syscall but would be checked by Seccomp filter if enabled.
+                syscall_intercept::SYSCALL_ARG_MAGIC,
+            )
+        };
+    };
+}
+
+/// The standard `panic!` macro may not work as expected because it uses TLS, which may have
```
jstarks, line 56: This is pretty scary. Is there no appropriate point to restore TLS to what the runtime expects, instead?

CvvT: That's doable. I also tried saving and restoring TLS, but it doesn't give more information (as stack unwinding is also broken). Also, any new thread spawned by LiteBox does not involve std (we use raw syscalls); thus, calling any std functions in a new thread may break. 

Ultimately, we should have no_std for all crates except those runners (which are responsible for setting up the environment before starting LiteBox).

jaybosamiya-ms:
Hmm, I'm not too particularly excited with this approach, since it leaves a bunch of holes where a regular panic might still happen, and get into trouble (as @jstarks correctly pointed out, the fact that the regular panic uses the TLS makes it kinda scary).  The approaches I have in mind instead are:
1. Use `catch_unwind` and `resume_unwind` to help safely carry the panic across the TLS changes needed.  We should need to catch such unwinds only around the syscall handling, so the scope of these should be fairly targeted, and would also mean that we stop depending on nice unwinding to carry through into the guest app.  I _think_ this needs `std` but I might be mistaken and there might be a no-std variant of this too.
2. We add a custom `#[panic_handler]` (which can be used for _pure_ `no_std` targets) or we can set up a hook via `set_hook` (which can be used for targets where we get `std`).

jaybosamiya-ms:
Since the runner itself has `std`, `set_hook` is probably the correct move. https://doc.rust-lang.org/std/panic/fn.set_hook.html

jaybosamiya-ms:
@CvvT, @wdcui and I had a fairly detailed discussion IRL.  Summarizing things a bit below, but please feel free to chime in if I missed anything.

Fixing panics properly is something we _should_ do, but to do that properly, we definitely need to do the FS restoration.  Without that, _even if_ you change out the hook, Rust `std` included _anywhere_ (such as with the runner) will trigger some code during panics that _must_ hit FS before we have a chance to hop in.

Thus, we must do FS restoration/resetting.  We actually have a small number of good places for it, which should be exhaustive (syscall entry/exit).  However, there are some drawbacks with threading which will _still_ prevent proper FS support for the panic handling on anything except the main thread.

To handle threads properly, we actually need to spin up an "`std`-like thread", which the move then is `std::thread::spawn`, which requires a change to how we set up threads on Linux Userland platform.  With the Windows Userland platform, we already perform some trampolining to translate the semantics over, we need to do similar semantics translation to get support.  On the Windows side though, the use of GS for regular userland, and FS for the guest app means that we don't need the shenanigans necessary for Linux-on-Linux, so this wasn't a big issue there, but there is more care needed on the Linux side to handle it correctly.

Interestingly, once we have proper thread spawning that can support FS, we no longer need to restrict ourselves to no-std platforms for _userland_ platforms.  Indeed, this leads to a simpler/cleaner design that actually improves portability to new platforms too.  Basically, the `std` folks have already handled most of the portability concerns for us.

There is a mild downside, which is that our systrap backend may no longer be applicable (at least not directly).  Our rewriter backend would then become the only properly feasible way to run on Linux userland.  This only impacts Linux-on-Linux, so we felt it was not a big deal, _especially_ since the systrap backend was mostly meant for early development, rather than our main focus.

Overall, the migration then is to start saving/restoring FS at the correct choke points, then move over the threading to the better-handled FS world, and _then_ we will have nice panics.

There are some higher priority tasks that are taking precedence over cleaning up panics at the moment, so sadly, we will sadly have to live with somewhat not-great panics, but thankfully, `gdb` helps when it happens.

Note: all of the above is for _userland_ platforms.  For kernel/baremetal platforms, we obviously cannot use `std`, in which case, even our runner is `no_std`, which means that we don't have the above constraints, and we can actually just install a `#[panic_handler]`, and there is no trouble; all the trouble above comes purely from needing `std` in the runner, which forces installation of its own `#[panic_handler]`.  We _could_ make the runner `no_std` on userland platforms too, but that is a massive can of worms wrt complexity, compared to the approach written above.

jaybosamiya-ms:
Closing this PR (we can reopen or open a new one in future) due to above^

PR #340, opened 2025-09-12
lvbs: Move max cpus and VTL1 memory parsing to one time during boot

tgopinath-microsoft:
Parse command line and bootparams to extract maximum possible cpus and the VTL1 memory ranges once during initial boot and store them. Use these values from then on. This will later allow boot page to be used as regular memory. It also paves way to use the max cpus to dynamically allocate per cpu memory and to support different memory ranges for VTL1.

### DIFF
--- a/litebox_platform_lvbs/src/host/bootparam.rs
+++ b/litebox_platform_lvbs/src/host/bootparam.rs
@@ -8,6 +8,7 @@ use crate::{
 };
 use core::ffi::{CStr, c_char};
 use num_enum::TryFromPrimitive;
+use spin::Once;
 
 // This module defines a simplified Linux boot params structure
 // (based on arch/x86/include/uapi/asm/bootparam.h and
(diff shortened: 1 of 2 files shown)


sangho2:
Thanks @tgopinath-microsoft!

It is recommended to run `cargo fmt` and `cargo clippy` to maintain code style and minimize potential bugs.

I have a question regarding the VTL1 memory range. While it is understood that the number of CPUs can dynamically change  due to hot plugging, should dynamic changes to the VTL1 memory range be supported without requiring a reboot? If a reboot remains required, the use of `Once` might be more appropriate than `Atomic`.

tgopinath-microsoft:
@sangho2 I will run cargo clippy. My first time dealing with a rust code base and so you will hve to be patient while I learn.
So no memory range should not change throught the lifecycle of the boot., Neither should the possible cpus (this is the absolute max that can be supported this boot). I did try using once_cell apis but with my limited understanding it requires dynamic allocation of memory and this code is so early on in the boot that dynamic allocation is not possible. Can once_ be used with static memory

sangho2:
@tgopinath-microsoft  I also had and still have issues with `cargo clippy` :) 

Our IDT does use `spin::Once` for static memory.

https://github.com/microsoft/litebox/blob/c62aef46043b48fcb893777705d5bfb03add3283/litebox_platform_lvbs/src/arch/x86/interrupts.rs#L13-L37

PR #343, opened 2025-09-12
Fix large read

CvvT:
The current value we set for `MAX_KERNEL_BUF_SIZE` is not enough for running node (version v22.19.0), which read ~4MB data from a file. Instead of keeping updating this value (also 4MB happens to be the limit our snp kernel allow to allocate), this PR tries to read data by chunks.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -314,7 +314,7 @@ fn pread_with_user_buf(
     count: usize,
     offset: i64,
 ) -> Result<usize, Errno> {
-    let mut kernel_buf = vec![0u8; 0x1000];
+    let mut kernel_buf = vec![0u8; count.min(MAX_KERNEL_BUF_SIZE)];
     let mut read_total = 0;
     while read_total < count {
         let to_read = (count - read_total).min(kernel_buf.len());
@@ -332,6 +332,7 @@ fn pread_with_user_buf(
             Err(e) => return Err(e),
         }
     }
+    assert!(read_total <= count);
     Ok(read_total)
 }
 
@@ -361,18 +362,29 @@ pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> usize {
                 // We read data in chunks and update the file offset ourselves only if the read succeeds.
                 syscalls::file::sys_lseek(fd, 0, litebox::fs::SeekWhence::RelativeToCurrentOffset)
                     .inspect_err(|e| {
-                        if *e == Errno::ESPIPE {
-                            unimplemented!("read on non-seekable fds with large buffers");
+                        match *e {
+                            Errno::EBADF => (), // safe errors to return
+                            Errno::ESPIPE => {
+                                unimplemented!("read on non-seekable fds with large buffers");
+                            }
+                            Errno::EINVAL => {
+                                unreachable!("seekable file should not return EINVAL when getting current offset");
+                            }
+                            _ => {
+                                unimplemented!("unexpected error from lseek: {}", e);
+                            }
                         }
                     })
                     .and_then(|cur_loc| {
                         pread_with_user_buf(fd, buf, count, i64::try_from(cur_loc).unwrap())
                             .inspect(|read_total| {
+                                // Update the file offset to reflect the read we just did.
                                 syscalls::file::sys_lseek(
                                     fd,
                                     (cur_loc + read_total).reinterpret_as_signed(),
                                     litebox::fs::SeekWhence::RelativeToBeginning,
                                 )
+                                // Given that previous lseek and pread succeeded, this lseek should also succeed.
                                 .expect("lseek failed");
                             })
                     })


CvvT:
@jaybosamiya-ms I remembered I implemented a similar version before, but you mentioned it messes up the file offset on the error path. Add you as reviewer, please take a look to see if this PR makes sense to you.

jaybosamiya-ms:
Why not just bump up `MAX_KERNEL_BUF_SIZE` to ~4MiB?  We just need _some_ reasonable value and it doesn't hurt if we have it even in the megabytes range, no?  Our allocations never _actually_ get that big unless a user asks for something that big.

Indeed, there were some issues with an older variant of the implementation.  Reviewing this one shortly.

However, in #125, you bumped the value up to 512KiB from the 64KiB that I'd originally set it to in #83.  I see no harm in bumping it to 4 or even 8MiB.  All the allocations do a `.min` anyways, so only situations like node are impacted by this.

litebox_shim_linux/src/lib.rs
```diff
@@ -307,6 +307,34 @@ pub extern "C" fn close(fd: i32) -> i32 {
...
 const MAX_KERNEL_BUF_SIZE: usize = 0x80_000;
 
+/// A wrapper function around `sys_pread64` that copies data in chunks to avoid OOMing.
+fn pread_with_user_buf(
+    fd: i32,
+    buf: MutPtr<u8>,
+    count: usize,
+    offset: i64,
+) -> Result<usize, Errno> {
+    let mut kernel_buf = vec![0u8; 0x1000];
```
jaybosamiya-ms: Why not use the `MAX_KERNEL_BUF_SIZE` here?

litebox_shim_linux/src/lib.rs
```diff
@@ -307,6 +307,34 @@ pub extern "C" fn close(fd: i32) -> i32 {
...
+            Ok(0) => break, // EOF
+            Ok(size) => {
+                buf.copy_from_slice(read_total, &kernel_buf[..size])
+                    .ok_or(Errno::EFAULT)?;
+                read_total += size;
+            }
+            Err(e) => return Err(e),
+        }
+    }
+    Ok(read_total)
```
jaybosamiya-ms: I would add an `assert!(read_total <= count)` here just as a sanity check

litebox_shim_linux/src/lib.rs
```diff
@@ -321,21 +349,48 @@ pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> usize {
...
+                        .ok_or(Errno::EFAULT)
+                })
+            } else {
+                // If the read size is too large, we need to do some extra work to avoid OOMing.
+                // We read data in chunks and update the file offset ourselves only if the read succeeds.
+                syscalls::file::sys_lseek(fd, 0, litebox::fs::SeekWhence::RelativeToCurrentOffset)
+                    .inspect_err(|e| {
+                        if *e == Errno::ESPIPE {
+                            unimplemented!("read on non-seekable fds with large buffers");
+                        }
```
jaybosamiya-ms: Can any other error occur?  Are other errors safe?  Probably just good to not have the `if` condition here and directly trigger the `unimplemented`, no?  If there are some safe error cases, then we should explicitly list the safe error cases, and fail on all the others.

litebox_shim_linux/src/lib.rs
```diff
@@ -321,21 +349,48 @@ pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> usize {
...
+                    })
+                    .and_then(|cur_loc| {
+                        pread_with_user_buf(fd, buf, count, i64::try_from(cur_loc).unwrap())
+                            .inspect(|read_total| {
+                                syscalls::file::sys_lseek(
+                                    fd,
+                                    (cur_loc + read_total).reinterpret_as_signed(),
+                                    litebox::fs::SeekWhence::RelativeToBeginning,
+                                )
+                                .expect("lseek failed");
```
jaybosamiya-ms, line 388: Nit: given that the previous `lseek` succeeded, this one should not fail, correct?  If so, putting that in the text might be helpful?

CvvT:
> Why not just bump up `MAX_KERNEL_BUF_SIZE` to ~4MiB? We just need _some_ reasonable value and it doesn't hurt if we have it even in the megabytes range, no? Our allocations never _actually_ get that big unless a user asks for something that big.
> 
> Indeed, there were some issues with an older variant of the implementation. Reviewing this one shortly.
> 
> However, in #125, you bumped the value up to 512KiB from the 64KiB that I'd originally set it to in #83. I see no harm in bumping it to 4 or even 8MiB. All the allocations do a `.min` anyways, so only situations like node are impacted by this.

The max size we can allocate for snp platform is 4MB: https://github.com/microsoft/litebox/blob/main/litebox_platform_linux_kernel/src/host/snp/snp_impl.rs#L400

litebox_shim_linux/src/lib.rs
CvvT: Changed to `count.min(MAX_KERNEL_BUF_SIZE)`. This code is also used to handle syscall `pread64` so `count` could be smaller than `MAX_KERNEL_BUF_SIZE`

litebox_shim_linux/src/lib.rs
CvvT: Added

litebox_shim_linux/src/lib.rs
CvvT: Good point! Listed all possible errors here.

litebox_shim_linux/src/lib.rs
CvvT: Added

PR #344, opened 2025-09-12
Support poll and ppoll syscalls

jstarks:
Reuse and extend the polling infrastructure to implement `poll` and `ppoll`.

Lots of `Arc`s and allocations in the runtime path... probably could be flattened a bit.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1444,7 +1444,6 @@ pub struct Pollfd {
     pub revents: i16,
 }
 
-#[non_exhaustive]
 #[repr(i32)]
 #[derive(Debug, IntEnum)]
 pub enum MadviseBehavior {


jstarks:
Hitting some picky cast prohibitions, probably I just need to figure out the conventions. No problem.

The fd i32 -> u32 thing is annoying, though. Should the fd type be changed to i32 throughout, perhaps? Although it's true that file descriptors can't be less than 0, it's also true that they can't be bigger than `i32::MAX`. Seems easier to just use the "natural" type rather than have redundant checks.

E.g., look at `sys_read`. It first converts the `u32` into `i32` with an explicit failure path, then has another failure path if the `fd` is invalid. Seems better to just accept `i32` in the `get_fd` call and have it do the appropriate checks in one place.

jaybosamiya-ms:
fd i32 / u32 issue: there's a long-running refactor where I'm cleaning up all the fd-related stuff.  It essentially began with #31 (but descriptions on GitHub are kinda out of date with the design, as it has evolved).  With the final version of the descriptor system I have in mind, most of the difficulties with FDs should disappear.  It is mild temporary ugliness (I don't quite like it as it is either) that we sadly have to deal with for now, but yeah, within the next few weeks (hopefully! assuming other high-priority tasks don't keep bumping it) we should have a nicer system to deal with it.  I agree with you that the "natural" fd type within `litebox_shim_linux` _is_ `i32`.

Most of the other clippy cast prohibitions are indeed to maintain a consistent convention indeed.

litebox_shim_linux/src/syscalls/epoll.rs
```diff
@@ -398,13 +402,148 @@ impl ReadySet {
...
+                pollee: self.pollee().clone(),
+            });
+            // TODO: unregister interest when the PollSet is dropped to avoid a
+            // leak for long-lived files that never wake up. Or update the
+            // Observer infra to garbage collect dead observers periodically.
+            let revents = file.poll(mask, Some(Arc::downgrade(&observer) as _));
+            if !revents.is_empty() {
+                // Race. Accumulate the events.
+                self.is_ready = true;
+                observer.on_events(&revents);
```
CvvT: Why not just call `add_ready_interest` and `continue` here?

litebox_shim_linux/src/syscalls/epoll.rs
```diff
@@ -398,13 +402,148 @@ impl ReadySet {
...
+        let Some(pollable) = file.io_pollable() else {
+            self.add_empty_interest();
+            return;
+        };
+        // Check first before allocating an observer.
+        let revents = pollable.check_io_events() & mask;
+        // Skip monitoring if this or any previous interest is already ready.
+        if !revents.is_empty() || self.is_ready {
+            self.is_ready = true;
+            self.entries.push(PollEntry::Ready(revents));
```
CvvT: Just call `add_ready_interest` instead?

litebox_common_linux/src/lib.rs
```diff
@@ -1709,6 +1717,18 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
...
+        nfds: usize,
+        timeout: i32,
+    },
+    Ppoll {
+        fds: Platform::RawMutPointer<Pollfd>,
+        nfds: usize,
+        timeout: Option<Platform::RawConstPointer<Timespec>>,
+        sigmask: Option<Platform::RawConstPointer<SigSet>>,
+        sigsetsize: usize,
+    },
```
CvvT, line 1972: Similar to `clock_nanosleep`, 32 bit uses `TimeVal` instead of `Timespec`: https://elixir.bootlin.com/linux/v6.16.4/source/fs/select.c#L1121

litebox_shim_linux/src/syscalls/epoll.rs
jstarks: Yeah that seems reasonable. I think in my head we needed to keep the observer around to unregister it, but that's not implemented yet anyway... I'll just add a comment so we don't forget that at some point.

litebox_common_linux/src/lib.rs
jstarks: Good catch, again.

jstarks:
> Just out of curiosity, what would happend if a `fd` is closed during polling? Its corresponding `revents` should be `POLLNVAL` or `0`?

Hmm, that's a good point. In this implementation, we'll set it to 0, but Linux seems to set it to `POLLNVAL` (only after a timeout, so the close doesn't wake up the poll). Perhaps to match Linux behavior we need to rescan the fds (which we now need to capture).

A nice thing about this is that we may be able to get away with just a single observer.

jstarks:
To match Linux behavior, I had to rethink this a bit.

I wasn't really able to reuse `Pollee` in the new impl, because the split between `try_op` and `check` is rather awkward and race prone here (notably, `check` must actually check, because `try_op` is called before the inner poller is created and registered). Possibly this infra could be reconsidered a bit.

I also added a TODO because of the annoyance of having to allocate multiple observers that all contain the same state and do the same thing. This could be fixed by having observer masks get ORed at registration time rather than overwritten.

litebox_shim_linux/src/syscalls/epoll.rs
```diff
@@ -398,13 +402,148 @@ impl ReadySet {
...
+                        Some(weak as _)
+                    };
+                    // TODO: add machinery to unregister the observer to avoid leaks.
+                    file.poll(entry.mask, observer)
+                } else {
+                    Events::NVAL
+                };
+                if !entry.revents.is_empty() {
+                    is_ready = true;
+                    register = false;
```
CvvT, line 494: Is `register` unnecessary? It is set together with `is_ready`.

CvvT:
> To match Linux behavior, I had to rethink this a bit.
> 
> I wasn't really able to reuse `Pollee` in the new impl, because the split between `try_op` and `check` is rather awkward and race prone here (notably, `check` must actually check, because `try_op` is called before the inner poller is created and registered). Possibly this infra could be reconsidered a bit.
> 
> I also added a TODO because of the annoyance of having to allocate multiple observers that all contain the same state and do the same thing. This could be fixed by having observer masks get ORed at registration time rather than overwritten.

The previous version that uses `Pollee` also looks fine to me. I agree it is a little bit awkward because it is meant to do `try_op` when there is any io event (i.e., return true from `check`). In the `ppoll` case, `try_op` is essentially the same as `check` as follows:

```
for fd in fds:
    if fd.check_io_events():
        return true
return false
```

Then, we collect all the events afterwards. It definitely causes some redundant work. We could refactor the code for improvement. 

In terms of the leak of observer, we could also add another API to `IOPollable` for cleanup.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -961,6 +961,83 @@ pub fn sys_epoll_pwait(
...
+        );
+        let revents: u16 = revents.bits().truncate();
+        unsafe {
+            revents_ptr
+                .write_at_offset(0, revents.reinterpret_as_signed())
+                .ok_or(Errno::EFAULT)
+        }?;
+        if revents != 0 {
+            ready_count += 1;
+        }
```
CvvT, line 1190: I encountered the same issue before. My previous solution was writing the whole struct to userspace instead of just one field. It would be nice to have an interface that could directly write a field to userspace. I will create an issue to track it.

litebox_shim_linux/src/syscalls/epoll.rs
jstarks: Good catch, register should be cleared after the first time through the loop. Fixed.

PR #346, opened 2025-09-13
Support tgkill

CvvT:
Currently all signal related syscalls are punchthroughs. We may want to have a dedicated provider for it?


jaybosamiya-ms:
> Currently all signal related syscalls are punchthroughs. We may want to have a dedicated provider for it?


Yeah, this has been in my mind for some time.  There is some cleanup needed.  I think it is out of scope of this PR though.  I'll open an issue to track it.

PR #348, opened 2025-09-13
Support nanosleep and clock_nanosleep

jstarks:
Add `nanosleep` and `clock_nanosleep` support for the existing supported clock IDs. Currently these sleeps cannot be interrupted.


litebox_common_linux/src/lib.rs
```diff
@@ -1772,6 +1781,12 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
         clockid: i32,
         res: Platform::RawMutPointer<Timespec>,
     },
+    ClockNanosleep {
+        clockid: i32,
+        flags: TimerFlags,
+        request: Platform::RawConstPointer<Timespec>,
+        remain: Option<Platform::RawMutPointer<Timespec>>,
+    },
```
CvvT, line 1789: For 32bit, the request is of type `TimeVal`: https://elixir.bootlin.com/linux/v6.16.4/source/kernel/time/posix-timers.c#L1404

jstarks: Good catch. I'll just support the 64-bit syscalls for now, as with the other clock_* syscalls.

PR #349, opened 2025-09-15
Use RefCell to protect TLS

jstarks:
Use a `RefCell` instead of manually implementing the same machinery. This moves the `borrowed` field out of the common TLS type, which closes a soundness hole. Move `self_ptr` out for the same reason.

Also, on Windows, use linker-provided TLS instead of runtime calls. This is more efficient and safer, and it works because Windows's TLS machinery uses gs as a base, which is not used by user code. Once fs is swapped across syscalls, litebox can use this approach for Linux and FreeBSD as well.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -1567,8 +1567,8 @@ impl litebox::platform::ThreadLocalStorageProvider for LinuxUserland {
             entry_number: self
                 .tls_entry_number
                 .load(core::sync::atomic::Ordering::Relaxed),
+            limit: u32::try_from(core::mem::size_of_val(tls.as_ref())).unwrap() - 1,
             base_addr: Box::into_raw(tls) as u32,
-            limit: u32::try_from(core::mem::size_of::<Self::ThreadLocalStorage>()).unwrap() - 1,
             flags,
         };
         let user_desc_ptr = litebox::platform::trivial_providers::TransparentMutPtr {


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1452,16 +1467,20 @@ impl LinuxUserland {
             return core::ptr::null_mut();
         }
 
-        let mut addr: usize;
+        let mut addr: *const TlsData<litebox_common_linux::ThreadLocalStorage<LinuxUserland>>;
         unsafe {
             core::arch::asm!(
-                "mov {0}, fs:{offset}",
+                "mov {0}, fs:0", // The first field of TlsData is a self-pointer
```
CvvT: When I implemented this, I also attempted to use `RefCell`. The only concern I have is how to get the offset to the self-pointer.

```
pub struct RefCell<T: ?Sized> {
    borrow: Cell<BorrowCounter>,
    // Stores the location of the earliest currently active borrow.
    // This gets updated whenever we go from having zero borrows
    // to having a single borrow. When a borrow occurs, this gets included
    // in the generated `BorrowError`/`BorrowMutError`
    #[cfg(feature = "debug_refcell")]
    borrowed_at: Cell<Option<&'static crate::panic::Location<'static>>>,
    value: UnsafeCell<T>,
}
```

From the source code, it doesn't look like the first field is a self-pointer or there is a self-pointer.

jstarks: Yes. For the Linux userland platform specifically, I don't use `RefCell` directly, I use it nested in this local `TlsData` type. That one has a self ptr (on 32-bit only) followed by the `RefCell`.

CvvT: ah, I missed the `TlsData`. Thanks for the effort on the improvement.

Is it better to use `offset = const core::mem::offset_of!(TlsData, self_ptr)` instead of hardcoding the offset `0`

jstarks: Yeah, good idea, I can restore the offset_of.

PR #351, opened 2025-09-15
Fix issues with accidental-multi-instance litebox

jaybosamiya-ms:
While the design of LiteBox supports having multiple instances of `LiteBox` object (we have carefully designed it this way), some shims and runners need to assume that there might not be multiple created (or at least any extra creations are safely gated, and not introduced accidentally).

This PR introduces a new `enforce_singleton_litebox_instance` cfg feature to the `litebox` crate, which allows us to detect any situations where we might be assuming a singleton instance of `LiteBox` object, but accidentally are making two.  We are not planning on making this the default, but we possible might want to consider making it a default in some runners (e.g., linux-on-linux should probably have it enabled).

Interestingly, we're accidentally making two instances in multiple places, which leads to some fairly hard to diagnose subtle bugs.  For example, I spent a non-trivial amount of time today chasing a bug that turned out to be entirely due to not following the singleton thing, despite _assuming_ a singleton in the Linux shim.  Interestingly, #24 would likely have already cleared this up, and it would be a lot safer to have multiple `LiteBox` instances (i.e., someone making two LiteBox instances would not be able to _accidentally_ cross subsystems across them; it is currently quite easy to do this mistake, which is what led to the hard to diagnose subtle bug, that was triggered by an _unrelated_ change I was making).

Note again that LiteBox doesn't _really_ need a singleton.  Otoh, due to how `litebox_shim_linux` needs to get the `LiteBox` object from _somewhere_ (and it cannot get it from the syscall), it picks up the object from a global.  Thus, it needs that global to be consistent, which esp when loading, could have accidentally gone out of sync due to creation of a LiteBox object before the rest of hte setup finishes up.  This can lead to some very surprising bugs.  This PR fixes up these minor loading ordering issues too.

As a drive-by, this PR also cleans up some test output to make it easier to read when it fails.

### DIFF
--- a/litebox/Cargo.toml
+++ b/litebox/Cargo.toml
@@ -20,7 +20,6 @@ buddy_system_allocator = { version = "0.11.0", default-features = false, feature
 slabmalloc = { git = "https://github.com/gz/rust-slabmalloc.git", rev = "19480b2e82704210abafe575fb9699184c1be110" }
 
 [features]
-default = ["enforce_singleton_litebox_instance"]
 lock_tracing = ["dep:arrayvec", "spin/mutex"]
 panic_on_unclosed_fd_drop = []
 enforce_singleton_litebox_instance = []


jstarks:
Should future changes to the Linux shim and the various userland platforms try to minimize the use of globals/singletons, in order to one day allow multiple instances in a process? Currently, I think the repo is moving in the opposite direction, e.g., some recent code added a global for tracking the current `umask`.

I think even Linux-on-Linux _could_ support multi instancing with some tweaks--all the hackery (fs handling, seccomp filters) seems like it's per-thread, and any globals can probably be replaced with an object passed on the stack or discovered via TLS.

jaybosamiya-ms:
Yes, imho, we should minimize the use of globals/singletons.  Indeed, I see that some code has introduced a few more globals recently, but I think we're going to want to clean up and remove as many of such instances as we can.  I would like to add a ratchet test (such as the one I added for `transmute`s in #154).  To prevent the repo from moving in the direction of more globals, I will soon be adding a ratchet test.  Thanks for your comment, I've bumped that ratchet implementation up in priority.

Yes, I agree Linux-on-Linux could support multi-instancing. It just isn't a direct priority at the moment, in which case, it is helpful to try to put some safeguards, thus the above comment.

PR #352, opened 2025-09-16
Fix the linux and windows runners to support relative path

wdcui:
This PR fixes the Linux and Windows runners to support using a relative path for the user program.

### DIFF
--- a/litebox_runner_linux_on_windows_userland/src/lib.rs
+++ b/litebox_runner_linux_on_windows_userland/src/lib.rs
@@ -9,7 +9,7 @@ use clap::Parser;
 use litebox::fs::FileSystem as _;
 use litebox_platform_multiplex::Platform;
 use std::os::windows::fs::MetadataExt;
-use std::path::{Component, PathBuf};
+use std::path::{Component, Path, PathBuf};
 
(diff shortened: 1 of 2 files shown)


litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -105,7 +105,7 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
     }
 
     let (ancestor_modes_and_users, prog_data): (Vec<(litebox::fs::Mode, u32)>, Vec<u8>) = {
-        let prog = PathBuf::from(&cli_args.program_and_arguments[0]);
+        let prog = std::fs::canonicalize(Path::new(&cli_args.program_and_arguments[0]))?;
```
jstarks: Do you want `canonicalize` (which resolves all symlinks) or do you just want `std::path::absolute` (which just makes the path absolute)? In my experience, the latter has fewer sharp edges if you don't actually need symlink resolution.

wdcui: Thank you for reviewing my code, John! Good point! I just replaced canonicalize with absolute.

PR #354, opened 2025-09-16
Support prctl

CvvT:
Add support for [prctl](https://linux.die.net/man/2/prctl).

Only support `PR_SET_NAME` and `PR_GET_NAME` for now.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1510,7 +1510,6 @@ pub enum FutexArgs<Platform: litebox::platform::RawPointerProvider> {
     },
 }
 
-#[non_exhaustive]
 #[repr(u32)]
 #[derive(Debug, IntEnum)]
 pub enum PrctlOption {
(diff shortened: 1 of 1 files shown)


CvvT:
Hi @jstarks You mentioned you have some tests that requires support for `prctl`. Do you know what were the options you encountered for `prctl`?

jstarks:
> Hi @jstarks You mentioned you have some tests that requires support for `prctl`. Do you know what were the options you encountered for `prctl`?

It was SET_NAME. So this PR fixes it. 👍

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -13,11 +13,49 @@ use litebox::platform::{
...
+        tls.current_task.comm[..comm.len()].copy_from_slice(comm);
+    });
+}
+
+/// Handle syscall `prctl`.
+pub(crate) fn sys_prctl(
+    arg: PrctlArg<litebox_platform_multiplex::Platform>,
+) -> Result<usize, Errno> {
+    match arg {
+        PrctlArg::GetName(name) => litebox_platform_multiplex::platform()
```
wdcui, line 34: Do we need to check the length of `name`?

CvvT: https://elixir.bootlin.com/linux/v6.16.4/source/kernel/sys.c#L2543

It needs to be at least 16 bytes long; If it is shorter, returns EFAULT.

PR #355, opened 2025-09-16
Support syscall rt_sigreturn

CvvT:
1. Add `rt_sigreturn` as a punchthrough.
2. Add a C test case for `rt_sigaction` and `rt_sigreturn` (which was mostly implemented by copilot).
3. Fix `SigSet`: it should be 8 bytes on both x64 and x86.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -881,24 +881,24 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
             }
             PunchthroughSyscall::RtSigreturn { stack } => {
                 // The stack pointer should point to a `ucontext` structure. Due to our syscall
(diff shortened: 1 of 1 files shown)


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -878,6 +879,41 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
                 })
                 .map_err(litebox::platform::PunchthroughError::Failure)
             }
+            PunchthroughSyscall::RtSigreturn { stack } => {
+                // The stack pointer should point to a `ucontext` structure. Due to our syscall
+                // interception mechanism (see syscall_callback), the actual stack pointer is
+                // 2 `usize`s below the provided pointer.
```
wdcui, line 885: Change the direction of the stack?

CvvT: Fixed, thanks!

PR #357, opened 2025-09-16
New `dev_tests` crate + ratcheting for globals

### LINKED ISSUE #356: Introduce a ratchet to prevent increasing number of globals
See https://github.com/microsoft/litebox/pull/351#issuecomment-3299553484 and https://github.com/microsoft/litebox/pull/351#issuecomment-3300209232

jaybosamiya-ms:
This PR does a couple of things:
- Moves the ratcheting from the CI yaml stuff to a crate where we can more easily manage such scripts, allowing more expressivity and nicer error messages and such.  It also makes it such that folks will notice this locally sooner than waiting on the CI.  It is triggered by running regular tests.
- Fixes #356 by adding the ratchet by looking for any `static` at the start of the line (so that it doesn't accidentally trigger on `'static` and such).   Note that I explicitly am not looking for `const` which is also a way to make globals, but such globals are fully read-only, so don't have the kind of downsides that come from `static` or `static mut` things.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,6 +250,9 @@ jobs:
           #   it is a helper binary that runs in userland to AOT "compile" ELFs.
           #
           # - `litebox_runner_snp` is `no_std` but requires custom target to build
+          #
+          # - `dev_tests` is meant to only be used for tests, and thus can
+          #   safely use std.
           find . -type f -name 'Cargo.toml' \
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
Oh I guess I should mention, for transmute, the reason it went from 5 -> 7 is simply because the new test introduces the word `transmute` in the name of the function and inside the string.  This is equivalent to the older count in the main LiteBox sources.

PR #358, opened 2025-09-16
Implement rmdir for layered fs

CvvT:
Note this PR was completely done by copilot except that I fixed some clippy warnings.

### DIFF
--- a/litebox/src/fs/layered.rs
+++ b/litebox/src/fs/layered.rs
@@ -1066,21 +1066,7 @@ impl<
         self.upper.mkdir(path, mode)
     }
 
-    #[expect(clippy::too_many_lines)]
     fn rmdir(&self, path: impl crate::path::Arg) -> Result<(), RmdirError> {
-        // Semantics:
-        // - If the directory exists only in upper: ensure empty -> remove.
-        // - If it exists only in lower: ensure empty in lower -> place tombstone.
-        // - If it exists in both: ensure BOTH are empty -> remove upper + tombstone lower.
-        // - Union emptiness = (upper empty) AND (lower empty (if it exists and upper empty)).
-        // - Root ("/") may never be removed -> Busy.
-        // - A non-directory in either layer -> NotADirectory.
-        // - Non-empty -> NotEmpty.
-        // - Path errors that lower cannot fix propagate immediately.
-        //
-        // Permissions: we mirror `unlink` semantics—tombstoning lower layer without mutating it.
-        // This keeps consistency with existing layered delete behavior.
-
         let path = self.absolute_path(path)?;
 
         // Prevent removing root explicitly (even if upper is empty).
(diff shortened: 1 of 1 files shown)


litebox/src/fs/layered.rs
```diff
@@ -1066,13 +1066,153 @@ impl<
...
+                .expect("upper.file_status succeeded but open failed");
+            let entries = self
+                .upper
+                .read_dir(&upper_fd)
+                .expect("read_dir upper failed");
+            self.upper
+                .close(upper_fd)
+                .expect("close upper dir fd failed");
+            // Convention: "." and ".." are always present; anything more => not empty.
+            if entries.len() > 2 {
```
wdcui: Is this because of "." and ".."?

litebox/src/fs/layered.rs
```diff
@@ -1066,13 +1066,153 @@ impl<
...
+                return Err(RmdirError::NotEmpty);
+            }
+        }
+
+        // Mutation phase.
+        if upper_exists {
+            // Remove from upper layer.
+            self.upper.rmdir(path.as_str())?;
+            if lower_exists {
```
wdcui: When !upper_exists and lower_exists, we are supposed to tombstone lower?

litebox/src/fs/layered.rs
CvvT: Yes, left a comment above.

litebox/src/fs/layered.rs
CvvT: lines 1209-1214 does that

wdcui: Ah, I missed that. I wouldn't write code this way. I will handle upper_exisits first, then lower_exisits. :-) Should I push a commit for it?

litebox/src/fs/layered.rs
```diff
@@ -1066,13 +1066,153 @@ impl<
...
+        if path == "/" {
+            return Err(RmdirError::Busy);
+        }
+
+        // Track existence and (later) emptiness.
+        let mut upper_exists = false;
+        let mut lower_exists = false;
+
+        // First, probe upper.
+        match self.upper.file_status(path.as_str()) {
```
jaybosamiya-ms: Seems like unnecessary mutability that just makes it harder to understand. Changing it to a `let upper_exists = match ...` and similar below would work.

litebox/src/fs/layered.rs
jaybosamiya-ms: Yeah, the current implementation is kind of redundant.

Just a
```rs
if upper {
  ...
}

if lower {
  ...
}

Ok(())
```
would do the correct thing, no?  This of course is if we're sticking to this implementation.  I believe there is a cleaner version that I've written in the other comment.

litebox/src/fs/layered.rs
CvvT: Thanks for the suggestion! I followed your idea to simplify the code. The only difference is that I used `open` with `OFlags::DIRECTORY` to check if the give path is a dir, because we need to open it anyway for subsequent `read_dir`.

PR #365, opened 2025-09-17
Fix TLS

CvvT:
On x86, LiteBox uses a fixed TLS entry across all threads (though each thread can still set its own TLS base address via `set_thread_area` using the same entry number).  See #320 for more details. Once it is initialized, we should never change it.

Temporarily disable the thread test due to some unknown bug.

### DIFF
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -257,7 +257,7 @@ impl WindowsUserland {
 
     fn is_aligned_to_granu(&self, x: usize) -> bool {
         let gran = self.sys_info.read().unwrap().dwAllocationGranularity as usize;
-        x % gran == 0
+        x.is_multiple_of(gran)
     }
 
     fn set_init_tls(&self) {


CvvT:
> Does this fix the flaky CI tests?

I believe both this PR and #366 fix the issue.

CvvT:
> > Does this fix the flaky CI tests?
> 
> I believe both this PR and #366 fix the issue.

I ran into another issue that I cannot reproduce locally. Looks like other PRs are blocked by the thread test: https://github.com/microsoft/litebox/actions/runs/17813503406/job/50642230000. While I'm debugging this, temporarily delete this test.

CvvT:
I was able to reproduce this issue in a VM with only two vCPUs. The initial main thread unmaps some stacks while other threads are still running using them. But it is unclear to me why that happened. It looks like very similar to what @jstarks described earlier in #364.  Do you have any idea or a concrete test I can run?

CvvT:
> I was able to reproduce this issue in a VM with only two vCPUs. The initial main thread unmaps some stacks while other threads are still running using them. But it is unclear to me why that happened. It looks like very similar to what @jstarks described earlier in #364. Do you have any idea or a concrete test I can run?

I believe I have a better understanding now.

The main thread is calling `pthread_join` to wait on a child thread. When the child thread `exit`, it wakes up the parent at line:

https://github.com/microsoft/litebox/blob/c4f6e653a947c321ae190c40cbab1ccab83313aa/litebox_shim_linux/src/syscalls/process.rs#L259

and then continue the execution to actually call the raw syscall `exit`. However, in between, the main thread wakes up and believe this child already terminates and thus release the memory used by the child, resulting in the child thread panics in `sys_exit` as it's still using the stack.

I believe #364 would fix this problem.

PR #366, opened 2025-09-17
Fix a race issue on thread creation

CvvT:
When creating a new thread, the program can pass `parent_tid` and `child_tid` with flags `CLONE_PARENT_SETTID` and `CLONE_CHILD_CLEARTID`.

CLONE_CHILD_CLEARTID (since Linux 2.5.49)
              Clear (zero) the child thread ID at the location pointed to
              by child_tid (clone()) or cl_args.child_tid (clone3()) in
              child memory when the child exits, and do a wakeup on the
              futex at that address.  The address involved may be changed
              by the [set_tid_address(2)](https://www.man7.org/linux/man-pages/man2/set_tid_address.2.html) system call.  This is used by
              threading libraries.

CLONE_PARENT_SETTID (since Linux 2.5.49)
              Store the child thread ID at the location pointed to by
              parent_tid (clone()) or cl_args.parent_tid (clone3()) in
              the parent's memory.  (In Linux 2.5.32-2.5.48 there was a
              flag CLONE_SETTID that did this.)  The store operation
              completes before the clone call returns control to user
              space.

In the test case we have, parent calls `pthread_join`, which would in turns call `futex(&child_tid, WAIT, tid)`.  Normally when the child exits, it would clear `child_tid` due to `CLONE_CHILD_CLEARTID` and then parent would be unblocked. However, I just realized that `parent_tid` and `child_tid` may point to the same address: https://codebrowser.dev/glibc/glibc/nptl/pthread_create.c.html#292. The problem with that was that in our implementation, parent thread may set `parent_tid` after `child_tid` is cleared; then `futex(&child_tid, tid)` got stuck forever as it expected the child would clear it. This is not a problem for Linux kernel, because it set `parent_tid` before waking up the child.

This PR addresses it by setting a gate for synchronization so that the child can only start running after parent has set `parent_tid` if `parent_tid` and `child_tid` point to the same address.

Alternatively, we may want to allocate our own tid so that parent can set `parent_tid` before spawning a new thread. However, its implication is unclear to me as it may affect how we implement signal related syscalls (or any syscalls that would take `tid/pid`). Also, I found having real `tid` helps debugging.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -380,7 +380,7 @@ pub(crate) fn sys_clone(
         && parent_tid_ptr.as_usize() == child_tid_ptr.as_usize()
     {
         // Lock the mutex before creating the thread, so that the new thread will block on it
-        // until this function returns (i.e., guard is dropped).
+        // until `guard` is dropped.
         Some(start_gate.lock())
     } else {
         None
@@ -416,6 +416,8 @@ pub(crate) fn sys_clone(
     if let Some(parent_tid_ptr) = set_parent_tid {
         let _ = unsafe { parent_tid_ptr.write_at_offset(0, child_tid.truncate()) };
     }
+    // After parent_tid is set, we can drop the guard to let the new thread proceed.
+    drop(guard);
 
     NR_THREADS.fetch_add(1, core::sync::atomic::Ordering::Relaxed);
     Ok(child_tid)


litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -265,10 +265,19 @@ fn new_thread_callback(
...
         set_child_tid,
+        start_gate,
         callback: _,
     } = args;
     let child_tid = task.tid;
 
+    // Wait for parent to write parent_tid if a start gate is present.
+    if let Some(gate) = start_gate {
+        // gate value 0 = locked, non-zero = go.
+        while gate.load(core::sync::atomic::Ordering::Acquire) == 0 {
```
jstarks: Can we use a RawMutex-based thing instead to avoid spinning? Most of the time we probably won't hit this condition, but still...

CvvT: Thanks, john!

Interestingly, copilot made the same point as yours:

`I used AtomicU32 as the gate with 0 meaning "wait" and 1 meaning "go". We spin with core::hint::spin_loop(). This is a very small, deterministic wait that is safe here because the parent will signal the gate immediately after writing the parent tid (no long blocking).
If you prefer a futex-based wait for lower CPU consumption, we can implement that later; for now spinning is simple and sufficient for the short window.
`

I feel it makes sense and thus didn't ask it to change. Also, as I mentioned in this PR, I may implement the tid allocation within LiteBox, then we don't have this issue anymore. Open to disscusion and any comments or suggestions are welcome.

jstarks: I was imagining a case where the child starts running immediately, interrupting the parent running on the same CPU. In this case, the child will have to spin until its time slice expires or until the parent gets scheduled elsewhere.

If you think litebox will take over tid allocation, then this seems reasonable in the short term. But if you expect to always pass through the tid from the OS then maybe it's better to just fix this now.

CvvT: Totally agree. I switched to use mutex as you suggested (which only requires a few lines of code change).

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -347,6 +353,39 @@ pub(crate) fn sys_clone(
...
+        parent_tid
+    } else {
+        None
+    };
+
+    // Create an optional `guard` when parent_tid and child_tid point to the same address.
+    // Clearing child_tid is always done after setting parent_tid, so this ensures these two
+    // operations occur in order if they are on the same address.
+    let start_gate = alloc::sync::Arc::new(crate::litebox().sync().new_mutex(()));
+    let start_gate_clone = start_gate.clone();
```
wdcui, line 377: Minor: should we clone it in line 399 when it's needed?

CvvT: `start_gate` is consumed by line 384, so I need to clone it beforehand.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -377,11 +413,10 @@ pub(crate) fn sys_clone(
             }),
         )
     }?;
-    if flags.contains(CloneFlags::PARENT_SETTID)
-        && let Some(parent_tid_ptr) = parent_tid
-    {
+    if let Some(parent_tid_ptr) = set_parent_tid {
         let _ = unsafe { parent_tid_ptr.write_at_offset(0, child_tid.truncate()) };
     }
```
jstarks, line 418: nit: should we explicitly `drop(guard)` here for clarity?

CvvT: Good point! Updated

PR #367, opened 2025-09-17
Support syscall alarm

CvvT:
To run UnixBench, `alarm` is required. We don't have a proper design/support for signal related syscalls yet. This PR adds `alarm` as a punchthrough. We will revisit it later once we have a better understanding of supporting signals on different platforms.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -935,6 +935,16 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
             PunchthroughSyscall::SetThreadArea { user_desc } => {
                 set_thread_area(user_desc).map_err(litebox::platform::PunchthroughError::Failure)
             }
+            PunchthroughSyscall::Alarm { seconds } => unsafe {
+                let remain = syscalls::syscall2(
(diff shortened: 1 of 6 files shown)


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -899,6 +899,16 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
             PunchthroughSyscall::SetThreadArea { user_desc } => {
                 set_thread_area(user_desc).map_err(litebox::platform::PunchthroughError::Failure)
             }
+            PunchthroughSyscall::Alarm { seconds } => unsafe {
+                let remain = syscalls::syscall2(
+                    syscalls::Sysno::alarm,
+                    seconds as usize,
```
wdcui, line 941: Why do we change from u32 to usize?

CvvT: Because `syscall2` requires `usize`:

```
pub unsafe fn syscall2(
    nr: Sysno,
    a1: usize,
    a2: usize,
) -> Result<usize, Errno> {
    Errno::from_ret(raw::syscall2(nr, a1, a2))
}
```

PR #369, opened 2025-09-18
LVBS: Revise per-CPU kernel variables

sangho2:
This PR revises the per-CPU kernel variables of the LVBS host/platform. Now we store per-CPU kernel variables in the kernel heap (except for one for the BSP), allowing us to support a large number of VCPUs. These per-CPU kernel variables can be accessed through the GS registers using `with_per_cpu_variables` and a closure along with `RefCell` to avoid multiple mutable borrows. In addition, this PR lets us build LiteBox for LVBS with the release profile.

### DIFF
--- a/litebox_platform_lvbs/src/host/per_cpu_variables.rs
+++ b/litebox_platform_lvbs/src/host/per_cpu_variables.rs
@@ -236,7 +236,7 @@ pub fn allocate_per_cpu_variables() {
         "# of possible CPUs ({num_cores}) exceeds MAX_CORES",
     );
 
-    // TODO: use `cpu_online_mask` to selectively allocate per-CPU variables
+    // TODO: use `cpu_online_mask` to selectively allocate per-CPU variables only for online CPUs.
     // Note. `PER_CPU_VARIABLE_ADDRESSES[0]` is expected to be already initialized to point to
     // `BSP_VARIABLES` before calling this function by `get_or_init_refcell_of_per_cpu_variables()`.
     #[allow(clippy::needless_range_loop)]


litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -133,24 +137,73 @@ static mut PER_CORE_KERNEL_CONTEXT: [KernelContext; MAX_CORES] = [KernelContext
...
-    apic_id as usize
+/// Store the addresses of per-CPU variables. The kernel threads are expected to access
+/// the corresponding per-CPU variables via the GS registers which will store the addresses later.
+/// Instead of maintaining this map, we might be able to use a hypercall to directly program each core's GS register.
+static mut PER_CPU_VARIABLE_ADDRESSES: [Option<u64>; MAX_CORES] = [None; MAX_CORES];
+
+/// Get the per-CPU variables for the current core.
+/// # Panics
+/// Panics if GSBASE is not set
+pub fn get_per_cpu_variables() -> &'static mut PerCpuVariables {
```
jstarks: This is inherently unsafe/unsound--you can call this twice to get two mutable references to the same value, which is undefined behavior in Rust.

sangho2: Thanks, John! I'll mark it as `unsafe` to use it with care.

Context: This is for per-CPU storage, and it will not be shared between threads. In fact, there is no threading/preemption here.

jaybosamiya-ms: Even without threading, if you ever do something like:
```rs
let x = get_per_cpu_variables();
let y = get_per_cpu_variables();
```
(which can easily happen accidentally, if a function with the x declaration calls a function with the y declaration before dropping the x declaration), then it is instant UB (even if you don't use x or y).

Simply placing an `unsafe` there is still making the reasoning quite non-local, which makes it quite scary.

I recommend instead returning a read-only ref to a `RefCell` (since you're planning on only using this within a single thread), so that it'll prevent you from accidentally doing a double-borrow like the above.  The runtime cost is miniscule, but will prevent a lot of potential headache.

sangho2: Instead of returning a reference or a pointer, now we do use `with_per_cpu_variables()` and closure for dealing with these per-CPU variables. In theory, one can still return them through a closure. We might be able to prevent this using `RefCell` if necessary (I tried but somewhat didn't work well. Still, we cannot prevent them from directly accessing the GS register or global variables).

jaybosamiya-ms: Thanks Sangho!  The closure is safer, but still not amazingly safe, since nothing prevents a recursive call `with_per_cpu_variables(|x| with_per_cpu_variables(|y| ...))` but that is a less likely to happen in practice, so _maybe_ is ok?

Directly accessing via GS register or such is hard to prevent, I agree.  The issue however is not that, but instead the creation of two Rust mutable-references to the same object.  Due to Rust's semantics, simply the _creation_ of two mutable references at the same program point (even if you don't use them) leads to instant UB.  That's why the main concern showed up.

I think if you updated `with_per_cpu_variables` to be `unsafe` and also added the `# Safety` constraint of "should not be invoked recursively", then it is probably good enough for now?

sangho2: Thanks, Jay! `unsafe` and no recursion might fine for now. Let me take a look at another approach. If it doesn't work, will do the `unsafe` stuff.

sangho2: Now this code uses both closure and `RefCell` which can catch many simple misusages.

litebox_platform_lvbs/src/arch/x86/instrs.rs
```diff
@@ -43,3 +43,19 @@ pub fn wrmsr(msr: u32, value: u64) {
         );
     }
 }
+
+#[expect(clippy::inline_always)]
```
wdcui: We have wrgsbase and rdgsbase implemented in litebox_common_linux.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -0,0 +1,275 @@
...
+        };
+        unsafe {
+            PER_CPU_VARIABLE_ADDRESSES[i].write(RefCell::new(Box::into_raw(per_cpu_variables)));
+        }
+    }
+}
+
+// A hypercall page is a shared read-only code page, so it's better not to use heap.
+unsafe extern "C" {
+    static _hypercall_page_start: u8;
```
wdcui: Should the hypercall page be per-core?

litebox_platform_lvbs/src/arch/x86/instrs.rs
sangho2: I'll use those ones.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: No.  it doesn't have to. This page is actually just a special call target which is never used as a page (more like per VTL). I'll move this to a different file to avoid misunderstanding.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
wdcui: @jstarks and @jaybosamiya-ms are you okay with Sangho's current approach? Thanks!

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -0,0 +1,254 @@
...
+        entries: [(0, 0); NUM_CONTROL_REGS],
+    },
+    gdt: const { None },
+};
+
+/// Store the addresses of per-CPU variables. The kernel threads are expected to access
+/// the corresponding per-CPU variables via the GS registers which will store the addresses later.
+/// Instead of maintaining this map, we might be able to use a hypercall to directly program each core's GS register.
+static mut PER_CPU_VARIABLE_ADDRESSES: [MaybeUninit<RefCell<*mut PerCpuVariables>>; MAX_CORES] =
+    [const { MaybeUninit::uninit() }; MAX_CORES];
```
jaybosamiya-ms: `RefCell::new()` is `const` initializable, so I am not sure why you have the `MaybeUninit` here.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
jaybosamiya-ms: The new signature seems largely reasonable.  However, a quick look at the types brought up a question regarding the use of `MaybeUninit` (https://github.com/microsoft/litebox/pull/369#discussion_r2383539924) that kinda stood out to me.  Also `PER_CPU_VARIABLE_ADDRESSES` is written to, but never read from?

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: Thanks, Jay! I tried `const { RefCell::new(core::ptr::null_mut()) }` but it crashed. Perhaps because `RefCell` doesn't implement `Copy`? not so sure.

jaybosamiya-ms: Huh, that's somewhat surprising.  I don't think `Copy` is the issue (if that is the issue, then it would prevent compilation; all you should need is `Clone` and `RefCell` implements `Clone`.

Please don't consider the `MaybeUninit` to be a blocker here to merging, but yeah, it is worth trying to see what is causing the crash to see if it can be fixed up, because generally speaking, we don't want to be increasing the number of `MaybeUninit`s around, and if such initialization is failing, that points at a somewhat fundamental problem.

`MaybeUninit`s are dangerous (_especially_ so when used in global contexts) and can cause some serious UB if there is ever an accidental `assume_init` before initialization, which can happen accidentally when moving code around and such.  I am not _too_ worried about that here, because all of them (well, all the ones up to the max cpus) get initialized as soon as you run `allocate_per_cpu_variables` so it is probably ok here.  But yeah, non-local reasoning needed.

sangho2: That's a good point. Thanks! Let me see whether I can get rid of `MaybeUninit`. If this isn't easy, perhaps it'll be added to `TODO`.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: > The new signature seems largely reasonable. However, a quick look at the types brought up a question regarding the use of `MaybeUninit` ([#369 (comment)](https://github.com/microsoft/litebox/pull/369#discussion_r2383539924)) that kinda stood out to me. Also `PER_CPU_VARIABLE_ADDRESSES` is written to, but never read from?

`read_or_populate_gsbase()` reads `PER_CPU_VARIABLE_ADDRESSES` to program `gsbase`.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: `litebox_platform_lvbs/src/host/mod.rs` now has `hv_hypercall_page_address()` which get or initialize a hypercall page.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: TODO for now. Looks like there are some issues with borrow flag initialization but need more investigation.

sangho2:
Turns out that the compiler optimization pass for release build doesn't like my last commit for using heap for GDT and TSS. Reverted this change. I will not introduce further changes to this PR unless someone finds other bugs or safety issues.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -0,0 +1,257 @@
...
+        usize::try_from(get_num_possible_cpus().expect("Failed to get number of possible CPUs"))
+            .unwrap();
+    assert!(
+        num_cores <= MAX_CORES,
+        "# of possible CPUs ({num_cores}) exceeds MAX_CORES",
+    );
+
+    // TODO: use `cpu_online_mask` to selectively allocate per-CPU variables
+    #[allow(clippy::needless_range_loop)]
+    for i in 1..num_cores {
```
jaybosamiya-ms, line 243: Minor: an explicit comment saying the 0 is handled by `BSP_VARIABLES` during `get_or_init_refcell_of_per_cpu_variables` would help.

sangho2: Thanks, Jay! Added a comment as suggested.

PR #370, opened 2025-09-19
Improved fs/net metadata handling

jaybosamiya-ms:
This PR cleans up metadata handling by migrating the handling from the net subsystem, and each fs that we support, into the fd subsystem.  This reduces the amount of redundant code we have and makes later migration in the shim easier.

This PR is primarily semantics-preserving refactoring, but _does_ change our north interfaces a bit.  Specifically, we no longer will have the parametric metadata directly accessible at the fs/net north interfaces, but instead are linked to the fd interface.  I've fixed up the relevant calls in the shim.

Interestingly, the net subsystem's metadata support was lacking behind the fs subsystem's metadata support (network FDs could previously not store metadata separately at the FD level and at entry level).  With this PR, it is up to full parity in terms of support.

---

This is yet another PR in the series of PRs towards the better FD design (connected to #31).

### DIFF
--- a/litebox/src/utilities/anymap.rs
+++ b/litebox/src/utilities/anymap.rs
@@ -7,9 +7,9 @@
 //! Whenever we want/need to make a new decision or add an interface, we are going to try our best
 //! to keep things largely consistent with the anymap crate.
 //!
-//! Due to how we're using it within LiteBox, we are doing something similar to `anymap::Map<dyn Any
-//! + Send + Sync>` rather than a direct `anymap::AnyMap` (which would just be equivalent to
-//! `anymap::Map<dyn Any>`.
+//! Due to how we're using it within LiteBox, what we are doing is something similar to
+//! `anymap::Map<dyn Any + Send + Sync>` rather than a direct `anymap::AnyMap` (which would just be
+//! equivalent to `anymap::Map<dyn Any>`).
 
 use alloc::boxed::Box;
 use core::any::{Any, TypeId};


litebox/src/fd/mod.rs
```diff
@@ -334,9 +431,37 @@ pub enum ErrRawIntFd {
...
+#[derive(Error, Debug)]
+pub enum MetadataError {
+    #[error("no such metadata available")]
+    NoSuchMetadata,
+}
+
+/// A module-internal fd-specific individual entry
+struct IndividualEntry<Platform: RawSyncPrimitivesProvider> {
+    x: Arc<RwLock<Platform, DescriptorEntry>>,
+    metadata: AnyMap,
```
wdcui, line 460: What if we want to store two values of the same type?

jaybosamiya-ms: Mentioned this IRL, but just briefly noting for anyone else who looks: introducing a runtime-zero-cost newtype in Rust is trivial: `struct Foo(Bar);` is a different type from `Bar`, despite otherwise being identical in terms of runtime.

jaybosamiya-ms:
The semver checks test caught an interesting point: the current approach makes `Descriptors` into non-`Send` and non-`Sync`, which is not ideal.  This is quite easily fixed by making `AnyMap` have the requirements of `Send` and `Sync` (this is fine because we plan to put only Send+Sync things in it anyways).

I've just pushed a new commit to fix this up, at which point the semver checks _should_ only point at things we expect to change.

PR #372, opened 2025-09-22
Support udp

CvvT:
Add send/receive/connect/bind/getsockname support for UDP and also fix some error handling for TCP.

For unit test, use `nc` instead of C code, which supports both TCP and UDP. Note that we don't support loopback yet, so we need to have an external client for testing.

### DIFF
--- a/litebox/src/net/errors.rs
+++ b/litebox/src/net/errors.rs
@@ -96,6 +96,8 @@ pub enum SendError {
     BufferFull,
     #[error("port allocation failed: {0}")]
     PortAllocationFailure(#[from] LocalPortAllocationError),
+    #[error("unnecessary destination address provided")]
+    UnnecessaryDestinationAddress,
 }
 
 /// Possible errors from [`Network::receive`]
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -886,7 +886,8 @@ where
 
     /// Send data over a socket, optionally specifying the destination address.
     ///
-    /// If the socket is connection-mode, destination address shall be ignored.
+    /// If the socket is connection-mode and the destination address is provided,
+    /// `Err(SendError::UnnecessaryDestinationAddress)` is returned.
     pub fn send(
         &mut self,
         fd: &SocketFd<Platform>,
@@ -903,11 +904,16 @@ where
         }
 
         let ret = match socket_handle.protocol() {
-            Protocol::Tcp => self
-                .socket_set
-                .get_mut::<tcp::Socket>(socket_handle.handle)
-                .send_slice(buf)
-                .map_err(|tcp::SendError::InvalidState| SendError::SocketInInvalidState),
+            Protocol::Tcp => {
+                if destination.is_some() {
+                    // TCP is connection-oriented, so no destination address should be provided
+                    return Err(SendError::UnnecessaryDestinationAddress);
+                }
+                self.socket_set
+                    .get_mut::<tcp::Socket>(socket_handle.handle)
+                    .send_slice(buf)
+                    .map_err(|tcp::SendError::InvalidState| SendError::SocketInInvalidState)
+            }
             Protocol::Udp => {
                 let destination = destination
                     .map(|s| match s {
--- a/litebox_common_linux/src/errno/mod.rs
+++ b/litebox_common_linux/src/errno/mod.rs
@@ -393,6 +393,7 @@ impl From<litebox::net::errors::SendError> for Errno {
             litebox::net::errors::SendError::Unaddressable => Errno::EDESTADDRREQ,
             litebox::net::errors::SendError::BufferFull => Errno::EAGAIN,
             litebox::net::errors::SendError::PortAllocationFailure(e) => e.into(),
+            litebox::net::errors::SendError::UnnecessaryDestinationAddress => Errno::EISCONN,
             _ => unimplemented!(),
         }
     }


wdcui:
The code looks good to me overall. Since @jaybosamiya-ms implemented the early part of the network code, I would defer to him for approving this PR.

litebox/src/net/mod.rs
```diff
@@ -627,6 +650,32 @@ where
...
+                        local_endpoint.port,
+                    )),
+                }
+            }
+            Protocol::Icmp => unimplemented!(),
+            Protocol::Raw { protocol: _ } => unimplemented!(),
+        }
+    }
+
     /// Bind a socket to a specific address and port.
```
jaybosamiya-ms: I think we should add to the doc-comment here stating that a port of 0 causes automatic allocation of ephemeral port

litebox/src/net/errors.rs
```diff
@@ -50,6 +52,8 @@ pub enum BindError {
     UnsupportedAddress(SocketAddr),
     #[error("Port {0} already in use")]
     PortAlreadyInUse(u16),
+    #[error("Already bound to an address")]
```
jaybosamiya-ms, line 55: Nit: "the address" rather than "an address"?

litebox/src/net/mod.rs
```diff
@@ -841,6 +890,17 @@ where
         fd: &SocketFd<Platform>,
         buf: &[u8],
         flags: SendFlags,
+    ) -> Result<usize, SendError> {
+        self.sendto(fd, buf, flags, None)
+    }
+
+    /// Send data over a socket, possibly specifying the destination address.
+    pub fn sendto(
```
jaybosamiya-ms: Few points:
- I don't think we need a separate `send` and `sendto`.  We can just add the `Option<SocketAddr>` argument to `send`.
- I suggest calling it `destination` rather than `sockaddr` (we already know it is a socket address, it is better to include the _why_ than the _what_).
- The connection oriented sockets (TCP) should check `if destination.is_some() { return Err(SendError::UnnecessaryDestinationSpecified); }`.

litebox/src/net/mod.rs
```diff
@@ -856,7 +916,40 @@ where
...
-            Protocol::Udp => unimplemented!(),
+            Protocol::Udp => {
+                let sockaddr = sockaddr
+                    .map(|s| match s {
+                        SocketAddr::V4(addr) => smoltcp::wire::IpEndpoint {
+                            addr: smoltcp::wire::IpAddress::Ipv4(*addr.ip()),
+                            port: addr.port(),
+                        },
+                        SocketAddr::V6(_) => unimplemented!(),
+                    })
```
jaybosamiya-ms, line 922: Rather than doing this manually, I believe you can just do a `smoltcp::wire::IpEndpoint::from` (see https://docs.rs/smoltcp/latest/smoltcp/wire/struct.IpEndpoint.html#impl-From%3CSocketAddr%3E-for-Endpoint)

litebox/src/net/mod.rs
```diff
@@ -869,12 +962,14 @@ where
...
     /// Receive data from a connected socket.
+    ///
+    /// On success, returns the number of bytes received and, if applicable, the source address.
     pub fn receive(
         &mut self,
         fd: &SocketFd<Platform>,
         buf: &mut [u8],
         flags: ReceiveFlags,
-    ) -> Result<usize, ReceiveError> {
+    ) -> Result<(usize, Option<SocketAddr>), ReceiveError> {
```
jaybosamiya-ms: Based on this documentation, it should return the source address for TCP sockets too, but the current implementation returns `None`.

My suggestion instead is to take an extra argument `Option<&mut SocketAddr>` so that the caller can request a source address if they want it.

The docstring would then be something like "If the `source_addr` is `Some` and the underlying protocol provides a source address, it will be updated."

litebox/src/net/mod.rs
```diff
@@ -892,11 +987,28 @@ where
                 .socket_set
                 .get_mut::<tcp::Socket>(socket_handle.handle)
                 .recv_slice(buf)
+                .map(|n| (n, None))
```
jaybosamiya-ms: This is essentially the one that seems surprising to me based on the docstring.  With the updated version, we could simply do a `if source_addr.is_some() { unimplemented!() }` and it'd be ok.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -330,6 +331,11 @@ impl litebox::platform::ExitProvider for LinuxUserland {
         // We don't need to explicitly drop this, but doing so clarifies our intent that we want to
         // close it out :). The type itself is re-specified here to make sure we look at this
         // particular function in case we decide to change up the types within `LinuxUserland`.
+        //
+        // Note that we only drop this if we are not using the `systrap_backend` feature,
+        // because in that case, all syscalls (including `close`) are intercepted and handled
```
jaybosamiya-ms: Nit: "in that case" -> "in the systrap-backend case"; the "that" is ambiguous whether it is referring to the systrap backend case or the "if we are not using the ..." case.

litebox/src/net/errors.rs
CvvT: This error means that this socket is already bound to some address (it does not need to be the same as the input). Basically, we cannot call `bind` more than once on the same socket.

litebox/src/net/mod.rs
CvvT: Updated

litebox/src/net/mod.rs
CvvT: The separation was only meant to provide a shortcut helper function. Since it only removes a single argument, it’s not particularly useful, so I removed it. Also updated the name of the argument as suggested.

According to the [man sendto](https://linux.die.net/man/2/sendto), "If sendto() is used on a connection-mode (SOCK_STREAM, SOCK_SEQPACKET) socket, the arguments dest_addr and addrlen are ignored (and the error EISCONN may be returned when they are not NULL and 0)". I just tested on Linux, it didn't return any error.

litebox/src/net/mod.rs
CvvT: Good point! The one you mentioned require enabling ipv6, so I used this one: https://docs.rs/smoltcp/latest/smoltcp/wire/struct.IpEndpoint.html#impl-From%3CSocketAddrV4%3E-for-Endpoint instead.

litebox/src/net/mod.rs
CvvT: Good point! I just tested it on Linux, `recvfrom` on TCP socket does not get source address. If we add an extra argument `Option<&mut SocketAddr>`, how does the caller know it is updated or not. I changed it to `source_addr: Option<&mut Option<SocketAddr>>`, so caller would receive `None` if it requests for it and the protocol does not provide it.

I added some comment: `UDP does provide the source address, while TCP does not (because it is connection-oriented, once it is established, both ends should already know each other's addresses)`

litebox/src/net/mod.rs
CvvT: See the comment above, now this line is removed.

litebox_platform_linux_userland/src/lib.rs
CvvT: Updated

litebox/src/net/errors.rs
jaybosamiya-ms: Good point, thanks!

litebox/src/net/mod.rs
jaybosamiya-ms: Yep, the manpage indeed says that (and I'd noticed it when I'd originally written my comment), but my reason for suggesting the stricter version is to prevent LiteBox internals from being overly Linux-specific.  If we find that some application in practice actually does depend on this ignoring behavior, then the shim can simply retry with `None` passed.  But yeah, I don't think the LiteBox crate itself should bake in this Linux-ism.

litebox/src/net/mod.rs
jaybosamiya-ms: That's a fair point, thanks for using the `Option<&mut Option<..>>` there.

litebox/src/net/mod.rs
CvvT: Good point! Updated to return an error instead.

PR #374, opened 2025-09-23
Support SETLK, SETLKW, and GETLK

CvvT:
This PR adds `SETLK`, `SETLKW`, and `GETLK` support for fcntl. After I finished implementing it, I noticed two things:

1. Since we don't support multi-processing yet and the same process can always set file lock succesfully, this PR could be simplified a lot by just having no-op or dummy implementation.
2. As I mentioned in #121 , `CLOSE_ON_EXEC` flag is per-fd, while file status flag is per-entry. A `dup`ed fd should inherit file status flag but not `CLOSE_ON_EXEC` flag. Now we have per-file lock. All entries that point to the same file should share same file locks. Our previous implementation had per-fd and per-file metadata, recent change #370 changed to per-fd and per-entry (but not per-file). If we decide to have dummy implementation for `SETLK`, `SETLKW`, and `GETLK`, then we don't need to change fs interface for now.

If we plan to support per-file metadata, I will keep this PR as a draft. Otherwise, I will simplify it to have dummy implementation.


wdcui:
Since @jaybosamiya-ms is working on the fs, I will let him chime in on his preference.

jaybosamiya-ms:
Good point, the file system actually requires a 3-layer hierarchy of metadata; fd, entry, and file.  I have not reviewed the current PR's implementation, but based on what you've written @CvvT, I would suggest that we change this one to draft, until we support the file-level metadata.

Interestingly, pre-#370 _looked_ like it was per-file, but actually was per-file _at the specific file system level_, so effectively (before #370) could lose metadata if you stored it onto a file, and then migrated the file up the layers.  A proper per-file metadata implementation would need to account for the layering migration.

I'll open an issue shortly to track the above. (EDIT: Opened #392)

As you've said, there is a "simple" fix for now (because we don't support multiprocessing yet), which is just dummy implementations for SETLK etc.

My suggestion is to change this PR to a draft and make a new PR with the dummy implementations.  We can revisit this draft PR after we resolve #392.

PS: I'm changing this PR to a draft.

CvvT:
Replaced by #409, will reopen once we have multi-processing support.

PR #377, opened 2025-09-23
Ignore MAP_SHARED

CvvT:
I was trying to run sqlite's [benchmark](https://sqlite.org/cpu.html) but ran into an unsupported feature: shared file mapping.

From strace's log on the test, once the file is mapped, there is no more open/read/write to that file. Thus, we can safely ignore this flag in this case. With this change, I can run the test and get some performance numbers but I'm not 100% sure if they are correct.

Supporting shared file mapping requires significant change; not sure how to do it yet and whether we are going to support it.


wdcui (approved):
LGTM. Thanks!

CvvT:
Replaced by #699

Issue #341, opened 2025-09-12
Improved fs support in Linux Userland Platform


jaybosamiya-ms:
We need to better handle FS restoration/resetting specifically for the Linux Userland Platform.

There are some benefits to this:
- ability to use `std`
- nicer panics
- easier portability of platforms

There are various steps towards the full migration (probably each could be their own PR?):
- [ ] handle fs store/restore at the correct chokepoints (syscall entry/exit)
- [ ] switch away from systrap backend
- [ ] migrate threading to properly handle fs + TLS setup
- [ ] switch more of the platform to std-specific bits
- [ ] coalesce commonalities between platforms (if we notice that there is sufficient commonality that we could make a `litebox_platform_userland` on its own)

Related important reading: https://github.com/microsoft/litebox/pull/339#issuecomment-3286916904

PR #380, opened 2025-09-23
Support swapping fs/gs

wdcui:
This PR adds the support for swapping fs/gs so that we can use std's tls implementation in the Linux userland platform.

This is the first step for addressing issue #341.

I temporarily disabled some tests and will add them back in future PRs.

### DIFF
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -476,6 +476,7 @@ pub fn optee_command_dispatcher(session_id: u32, is_sys_return: bool) -> ! {
 
         optee_command_completion_queue().push(session_id, stack.get_params_address());
 
+        // TODO: We will move the following line into the platform or the runner after the OPTEE shim is refactored.
         litebox_common_linux::swap_fsgs();
 
         unsafe {
--- a/litebox_shim_optee/src/loader/elf.rs
+++ b/litebox_shim_optee/src/loader/elf.rs
@@ -418,6 +418,9 @@ impl ElfLoader {
             load_trampoline(trampoline, base, fd);
         }
 
+        // TODO: the following memory page allocated for TLS in the TA
+        // will be freed when the TA is unloaded.
+
         // Since we don't use `ld` or `ldelf` for loading a TA, we should manually set up
         // the userspace thread local storage (TLS)
         let _ = crate::syscalls::mm::sys_mmap(


litebox/src/platform/mod.rs
```diff
@@ -675,6 +675,5 @@ pub trait ThreadLocalStorageProvider {
     ///
     /// This is used to help emulate certain syscalls (e.g., `execve`) that clear TLS,
     /// which is necessary for some platforms (e.g., x86 Linux).
```
CvvT: I guess this line can also be removed.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -918,25 +905,31 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
                     );
                 }
             }
+            // Since we use gs to store the TLS base for the guest, we read/write gsbase instead.
```
CvvT: More precisely, we swap gs and fs before and after a syscall so at this point guest's fs base is stored in gs.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1718,24 +1599,18 @@ mod tests {
...
-                tls.current_task.tid, tid,
-                "TLS should have the correct task ID"
-            );
             tls.current_task.tid = 0x1234; // Change the task ID
         });
+        let tid = platform.with_thread_local_storage_mut(|tls| {
+            let tid = tls.current_task.tid; //= 0x1234; // Change the task ID
+            assert!(tid == 0x1234, "TLS should have the modified task ID");
+            tid
+        });
```
CvvT: We can merge these two calls into one.

litebox_runner_linux_userland/tests/loader.rs
```diff
@@ -153,6 +153,20 @@ fn test_load_exec_common(executable_path: &str) {
...
+        litebox_common_linux::wrgsbase(litebox_common_linux::rdfsbase());
+        litebox_common_linux::wrfsbase(0);
+    }
+    #[cfg(target_arch = "x86")]
+    {
+        assert!(litebox_common_linux::rdfss() == 0);
+        litebox_common_linux::wrfss(litebox_common_linux::rdgss());
+        litebox_common_linux::wrgss(0);
+    };
+
```
CvvT: This is called when launching a new program or calling `execve`. In the former case, guest's TLS is not initialized, while in the latter case, we already called `clear_guest_thread_local_storage` in `execve`; so here we can just call `swap_fsgs`

litebox_shim_optee/src/lib.rs
```diff
@@ -476,6 +476,8 @@ pub fn optee_command_dispatcher(session_id: u32, is_sys_return: bool) -> ! {
 
         optee_command_completion_queue().push(session_id, stack.get_params_address());
 
+        litebox_common_linux::swap_fsgs();
```
CvvT, line 480: Do we also need to do it at the syscall entry?

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -418,6 +418,26 @@ impl ElfLoader {
...
+                | litebox_common_linux::ProtFlags::PROT_WRITE,
+            litebox_common_linux::MapFlags::MAP_PRIVATE
+                | litebox_common_linux::MapFlags::MAP_ANONYMOUS
+                | litebox_common_linux::MapFlags::MAP_FIXED
+                | litebox_common_linux::MapFlags::MAP_POPULATE,
+            -1,
+            0,
+        );
+        unsafe {
+            litebox_common_linux::wrgsbase(super::DEFAULT_FS_BASE);
```
CvvT, line 439: Do we need to release the memory at some point?

litebox_shim_optee/src/lib.rs
CvvT: why we need to do it at the shim layer? Looks like we directly jump to TA code. Does optee shim run in kernel mode or user mode?

litebox/src/platform/mod.rs
wdcui: Good point. Removed the last line about x86.

litebox_platform_linux_userland/src/lib.rs
wdcui: Good point. Updated.

litebox_platform_linux_userland/src/lib.rs
wdcui: Good point. Done.

litebox_runner_linux_userland/tests/loader.rs
wdcui: Good point. Fixed.

litebox_shim_optee/src/lib.rs
wdcui: @sangho2 can you help answer @CvvT 's questions here? Thanks!

litebox_shim_optee/src/loader/elf.rs
wdcui: @sangho2 can you help answer @CvvT 's questions here? Thanks!

sangho2: @CvvT Since OP-TEE doesn't support multi-threading, this page to support FS access should be maintained until we unload or terminate the loaded OP-TEE TA program.

litebox_shim_optee/src/lib.rs
sangho2: @CvvT For now, we maintain this here because OP-TEE requires a real or fake loop to repetitively enter a TA (either `jmp` or `iretq`) with a clean stack while maintaining other memory areas including code, data, and heap. Once we have such a loop within the runner, we can get rid of this from the OP-TEE shim.

wdcui: I added a TODO in the code for this one. @sangho2 plans to refactor the optee shim in the near future. At that time, this TODO will be resolved.

litebox_shim_optee/src/loader/elf.rs
wdcui: I added a TODO in the code for this one. @sangho2 plans to refactor the optee shim in the near future. At that time, this TODO will be resolved.

litebox_common_linux/src/lib.rs
```diff
@@ -998,6 +998,73 @@ pub unsafe fn wrgsbase(gs_base: usize) {
     }
 }
 
+/// Reads the current FS segment selector
+#[cfg(target_arch = "x86")]
+pub fn rdfss() -> u16 {
```
jaybosamiya-ms, line 1003: These functions should likely be `#[inline(always)]`, otherwise, cross-crate, they might not be inlinable.  `#[inline(always)]` does not really do much _within_ the same crate, but cross-crate, it has impact.

litebox_common_linux/src/lib.rs
```diff
@@ -998,6 +998,73 @@ pub unsafe fn wrgsbase(gs_base: usize) {
...
+            options(nostack, preserves_flags)
+        );
+    }
+    fs_selector
+}
+
+/// Writes the FS segment selector
+#[cfg(target_arch = "x86")]
+pub fn wrfss(fs_selector: u16) {
+    unsafe {
```
jaybosamiya-ms, line 1018: This function should have been marked `unsafe`, since it can break Rust's memory safety if used incorrectly.

litebox_common_linux/src/lib.rs
```diff
@@ -998,6 +998,73 @@ pub unsafe fn wrgsbase(gs_base: usize) {
...
+            out(reg) gs_selector,
+            options(nostack, preserves_flags)
+        );
+    }
+    gs_selector
+}
+
+/// Writes the GS segment selector
+#[cfg(target_arch = "x86")]
+pub fn wrgss(gs_selector: u16) {
```
jaybosamiya-ms, line 1043: same

litebox_common_linux/src/lib.rs
```diff
@@ -998,6 +998,73 @@ pub unsafe fn wrgsbase(gs_base: usize) {
...
+        core::arch::asm!(
+            "mov gs, {0:x}",
+            in(reg) gs_selector,
+            options(nostack, preserves_flags)
+        );
+    }
+}
+
+#[cfg(target_arch = "x86_64")]
+pub fn swap_fsgs() {
```
jaybosamiya-ms, line 1054: same

litebox_common_linux/src/lib.rs
```diff
@@ -998,6 +998,73 @@ pub unsafe fn wrgsbase(gs_base: usize) {
...
+    let fs_base = unsafe { rdfsbase() };
+    let gs_base = unsafe { rdgsbase() };
+    unsafe { wrgsbase(fs_base) };
+    unsafe { wrfsbase(gs_base) };
+}
+#[cfg(target_arch = "x86")]
+pub fn swap_fsgs() {
```
jaybosamiya-ms, line 1061: same

wdcui:
@jaybosamiya-ms Thank you very much for your comments! I will address them in my next PR that adds the support for std::thread::spawn and separate stacks.

PR #381, opened 2025-09-24
Make `fd_from_raw_integer` not maintain a borrow

jaybosamiya-ms:
This PR updates `fd_from_raw_integer` to return a `Weak` instead of maintaining a borrow, so that some upcoming changes can be managed better.  Specifically, this separates the borrow of the descriptor table from the actual descriptors themselves, instead replacing it with a fallible `upgrade` to make sure it is still alive.  This comes with the downside of such upgrades becoming fallible (+ now there is no _static_ guarantee that someone isn't storing the upgrade away; the documentation does mention that people should not do this though). However, it comes with the upside of much simplified lifetime management at the Linux shim layer for upcoming changes.  Otherwise, it can lead to a mess of recursive locks that can become quite problematic (via deadlocks) and would be very hard to untangle.

I believe that a component registry pattern (#24) might make this even nicer, but that particular issue has been open long enough, and I don't think we're solving that too soon, so for now, we live with the explicitly weak reference.

---

This is yet another PR in the series of PRs towards the better FD design (connected to https://github.com/microsoft/litebox/issues/31).

### DIFF
--- a/litebox/src/fd/mod.rs
+++ b/litebox/src/fd/mod.rs
@@ -5,7 +5,7 @@
     reason = "still under development, remove before merging PR"
 )]
 
-use alloc::sync::Arc;
+use alloc::sync::{Arc, Weak};
 use alloc::vec;
 use alloc::vec::Vec;
 use core::marker::PhantomData;
@@ -28,7 +28,7 @@ pub struct Descriptors<Platform: RawSyncPrimitivesProvider> {
     litebox: LiteBox<Platform>,
     entries: Vec<Option<IndividualEntry<Platform>>>,
     /// Stored FDs are used to provide raw integer values in a safer way.
-    stored_fds: Vec<Option<OwnedFd>>,
+    stored_fds: Vec<Option<Arc<OwnedFd>>>,
 }
 
 impl<Platform: RawSyncPrimitivesProvider> Descriptors<Platform> {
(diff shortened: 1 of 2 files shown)


litebox/src/fd/mod.rs
```diff
@@ -308,22 +311,38 @@ impl<Platform: RawSyncPrimitivesProvider> Descriptors<Platform> {
...
-        // SAFETY: Since `TypedFd` is `#[repr(transparent)]`, we can safety cast the type over (it
-        // is essentially the correct type, we simply want to expose a wrapped-type variant of it).
-        // We've just confirmed that it is the correct subsystem too.
-        let typed_fd: &TypedFd<Subsystem> = unsafe { &*core::ptr::from_ref(owned_fd).cast() };
-        Ok(typed_fd)
+
+        let typed_fd: Arc<TypedFd<Subsystem>> = {
+            let fd: Arc<OwnedFd> = Arc::clone(owned_fd);
+            let fd: *const OwnedFd = Arc::into_raw(fd);
+            // SAFETY: We effectively converting an `Arc<OwnedFd>` to an `Arc<TypedFd<Subsystem>>`.
```
wdcui: Minor: "We are" or "convert:?

jaybosamiya-ms: Good catch, updated, thanks!

PR #383, opened 2025-09-24
Support polling on socket

### LINKED ISSUE #371: Linux shim: support nonblocking sockets
LiteBox does not truely support nonblocking socket. For example, the following sequence causes LiteBox to block at `poll` forever:

    1. fd = socket(...)
    2. bind(fd, ...)
    3. listen(fd, ...)
    4. fcntl(fd, NONBLOCK)
    5. poll(fd, ...)

If a packet sent from a client arrives, it is not processed at all because the program is waiting in `poll`. In vsbox, we have a background worker, while in LiteBox we process packets manually before and/or after each call into network.

GPT proposed several solutions:

Option 1 — Minimal shim-driven slices (low-risk, localized)

+ Idea: sys_epoll_pwait calls a new small helper (e.g., Network::maybe_perform_platform_interaction) to run one step when Network is Manual, and uses short slice timeouts (100ms, tunable) to periodically re-run the helper while waiting.
+ Pros: minimal code change, small scope, easy to test, avoids touching core Pollee/condvar code.
+ Cons: periodic wake-ups (CPU + latency tradeoff), not as elegant; you must be careful about lock ordering (call network helper with the net mutex held briefly, but release before calling epoll_file.wait).
+ Deadlocks/borrows: ensure you do not hold the network lock across any blocking call that would try to acquire the same resource. The helper pattern avoids this.
+ When to use: immediate fix, safe to land quickly. 

Option 2 — Background network worker thread (best runtime performance)

+ Idea: spawn a dedicated thread that continuously calls perform_platform_interaction (or loops with small sleep), driving the smoltcp device regardless of callers.
+ Pros: no periodic wake-ups in epoll wait paths; low latency and efficient once implemented.
+ Cons: more invasive (thread lifetimes, shutdown), requires correct synchronization with existing mutexes (but the project already uses a Mutex for litebox_net), may complicate deterministic/manual test scenarios.
+ Deadlocks/borrows: still must respect the same mutexes. Worker must acquire litebox_net().lock() safely. Also ensure worker doesn't block while holding locks that epoll waits need.
+ When to use: when you want long-term stable low-latency behavior and threads are acceptable.

Option 3 — Device/driver notifies epoll directly (eventfd/wakeup)

+ Idea: integrate a platform wakeup primitive (eventfd/pipe) so the network device writes a byte or sets an eventfd when packets arrive; that write notifies epoll immediately (no manual drive).
+ Pros: very efficient and faithful to how OSes do it; true event-driven behaviour, minimal spinning.
+ Cons: needs integration between the platform's packet-receive path and the epoll-ready mechanism; may require changes in the device/phy layer and possibly a global eventfd per epoll instance or per device.
+ Deadlocks/borrows: less likely, but you must ensure the platform's receive callback can write the wakeup without requiring the same locks used by the blocked thread.
+ When to use: best for a production-quality event-driven design if you can refactor device receive path.

CvvT:
This PR fixes #371 by having a background worker that processes all packets and sends events. The background worker now uses a busy loop, which could be optimized by doing polling on the `tun_fd`.

Also, fixed a deadlock issue for `epoll`.

### DIFF
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -989,32 +989,6 @@ mod tests {
             .expect("epoll_wait failed")
     }
 
-    fn epoll_add(task: &crate::Task, epfd: i32, target_fd: i32, events: litebox::event::Events) {
-        let ev = litebox_common_linux::EpollEvent {
-            events: events.bits(),
-            data: u64::try_from(target_fd).unwrap(),
-        };
-        let ev_ptr = (&raw const ev).cast::<litebox_common_linux::EpollEvent>();
-        let ev_const = crate::ConstPtr::from_usize(ev_ptr as usize);
-        task.sys_epoll_ctl(
-            epfd,
-            litebox_common_linux::EpollOp::EpollCtlAdd,
-            target_fd,
-            ev_const,
-        )
-        .expect("epoll_ctl add server failed");
-    }
-
-    fn epoll_wait(
-        task: &crate::Task,
-        epfd: i32,
-        events: &mut [litebox_common_linux::EpollEvent],
-    ) -> usize {
-        let events_ptr = crate::MutPtr::from_usize(events.as_mut_ptr() as usize);
-        task.sys_epoll_pwait(epfd, events_ptr, events.len().truncate(), -1, None, 0)
-            .expect("epoll_wait failed")
-    }
-
     fn test_tcp_socket(
         task: &crate::Task,
         ip: [u8; 4],
@@ -1218,10 +1192,11 @@ mod tests {
             .expect("failed to bind server");
 
         // Create an epoll instance and register the server fd for EPOLLIN
-        let epfd = sys_epoll_create(litebox_common_linux::EpollCreateFlags::empty())
+        let epfd = task
+            .sys_epoll_create(litebox_common_linux::EpollCreateFlags::empty())
             .expect("failed to create epoll");
         let epfd = i32::try_from(epfd).unwrap();
-        epoll_add(task, epfd, server_fd, litebox::event::Events::IN);
+        epoll_add(&task, epfd, server_fd, litebox::event::Events::IN);
 
         let msg = "Hello from client";
         let mut child = std::process::Command::new("nc")
@@ -1258,7 +1233,7 @@ mod tests {
         }
         if is_nonblocking {
             let mut events = [litebox_common_linux::EpollEvent { events: 0, data: 0 }; 2];
-            let n = epoll_wait(task, epfd, &mut events);
+            let n = epoll_wait(&task, epfd, &mut events);
             assert_eq!(n, 1);
             for ev in &events[..n] {
                 assert!(ev.events & litebox::event::Events::IN.bits() != 0);


CvvT:
I got some numbers from `perf`:

```
Samples: 275K of event 'cycles:u', Event count (approx.): 130058124632
  Children      Self  Command          Shared Object                  Symbol
-   95.83%     0.03%  litebox_runner_  litebox_runner_linux_userland  [.] litebox::net::Network<Platform>::internal_perform_platform_interaction                                                                                       ◆
   - 95.80% litebox::net::Network<Platform>::internal_perform_platform_interaction                                                                                                                                                     ▒
      - 59.55% litebox::net::Network<Platform>::check_and_update_events                                                                                                                                                                ▒
         + 27.35% litebox::event::polling::Pollee<Platform>::notify_observers                                                                                                                                                          ▒
         + 14.58% <core::iter::adapters::filter_map::FilterMap<I,F> as core::iter::traits::iterator::Iterator>::next                                                                                                                   ▒
         + 14.46% litebox::net::Network<Platform>::check_socket_events                                                                                                                                                                 ▒
         + 1.78% core::ptr::drop_in_place<litebox::sync::rwlock::MappedRwLockReadGuard<litebox_platform_linux_userland::LinuxUserland,litebox::net::DescriptorEntry<litebox_platform_linux_userland::LinuxUserland>>>                  ▒
      + 29.75% smoltcp::iface::interface::Interface::poll_egress                                                                                                                                                                       ▒
      + 6.28% smoltcp::iface::interface::Interface::poll_ingress_single
```

My interpretation is that `check_and_update_events` generates too many events as it is called inside a busy loop.

Also, when I ran it against memcached benchmark, it had ~200 concurrent sockets at most.

wdcui:
@CvvT is this PR ready for review?

CvvT:
> @CvvT is this PR ready for review?

~Not yet, I made some changes but haven't pushed them yet. Will do soon.~

Added some change to this PR, which now only notifies when there are new events.

litebox/src/net/mod.rs
```diff
@@ -1055,43 +1071,115 @@ where
...
+                    .recv_queue
+                    .store(recv_queue_now, core::sync::atomic::Ordering::Relaxed);
+                // We have more data to receive than last time, so we should notify
+                events |= Events::IN;
+            }
+
+            let can_send_last = socket_handle
+                .can_send_last
+                .load(core::sync::atomic::Ordering::Relaxed);
+            if !can_send_last && can_send_now {
```
wdcui: Why do we require both `!can_send_last` and `can_send_now` here? Do we expect `Events::OUT` flag to be set in `events` if `can_send_last` is true?

CvvT: No, the idea is that we only send new events; so, we only need to send `Events::OUT` when the state changes from `cannot send last` (false) to `can send now` (true).

jaybosamiya-ms:
Will review it soon.

Sidenote: this _may_ have non-trivial interaction with #435 so we should coordinate on merging this and that one @CvvT, thanks!

CvvT:
> Will review it soon.
> 
> Sidenote: this _may_ have non-trivial interaction with #435 so we should coordinate on merging this and that one @CvvT, thanks!

Thanks Jay! Looks like it might be easier for me to rebase this one. Feel free to merge #416 and #435 first when they are done.

litebox/src/net/mod.rs
wdcui: I see. Thanks.

litebox/src/net/mod.rs
```diff
@@ -114,20 +115,29 @@ where
...
 /// from [`SocketFd`], _except_ the `Socket` itself which is stored in the [`Network::socket_set`].
-pub(crate) struct SocketHandle {
+pub(crate) struct SocketHandle<Platform: RawSyncPrimitivesProvider + TimeProvider> {
     /// The handle into the `socket_set`
     handle: smoltcp::iface::SocketHandle,
     // Protocol-specific data
     specific: ProtocolSpecific,
+    pollee: crate::event::polling::Pollee<Platform>,
+    /// Whether the socket was last known to be able to send
+    can_send_last: core::sync::atomic::AtomicBool,
```
jaybosamiya-ms: Nit: `previously_sendable`?  To me `can_send_last` seems to read more like "the 'last' is send-able"

litebox/src/net/mod.rs
```diff
@@ -151,6 +161,8 @@ pub(crate) struct TcpSpecific {
     local_port: Option<LocalPort>,
     /// Server socket specific data
     server_socket: Option<TcpServerSpecific>,
+    /// Whether the socket was last known to be able to accept
+    can_accept_last: core::sync::atomic::AtomicBool,
```
jaybosamiya-ms: Similar

litebox_shim_linux/src/syscalls/epoll.rs
```diff
@@ -353,15 +353,17 @@ impl ReadySet {
...
-        let mut nums = entries.len();
+        let mut nums = self.entries.lock().len();
         while nums > 0 {
             nums -= 1;
             if events.len() >= maxevents {
                 break;
             }
 
-            let Some(weak_entry) = entries.pop_front() else {
+            // Note the lock operation was moved into the loop to avoid holding the lock while calling `poll()`.
```
jaybosamiya-ms: Nit: "operation was moved into the loop" -> "operation is performed inside the loop".  No point referring to the pre-PR version in the source.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -907,6 +904,23 @@ mod tests {
     const SERVER_PORT: u16 = 8080;
     const CLIENT_PORT: u16 = 8081;
 
+    fn init_platform(tun_device_name: Option<&str>) {
+        crate::syscalls::tests::init_platform(tun_device_name);
+        crate::litebox_net()
+            .lock()
+            .set_platform_interaction(litebox::net::PlatformInteraction::Manual);
```
jaybosamiya-ms: After this PR, is it reasonable to say that the Linux shim will no longer be supported in `PlatformInteraction::Automatic` mode?  Or are there situations where `::Automatic` mode still make sense?

If we don't support `::Autoamatic`, we could probably (as part of `init_process` or something) perform this particular switch to `::Manual`.  That way, it is noisier if the user forgets to set up the `perform_platform_interaction` loop, because it will cause a _complete_ block, rather than only in the rarer polling-related case described in #371, which is a more subtle situation to be stuck in.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -907,6 +904,23 @@ mod tests {
...
+        std::thread::spawn(|| {
+            loop {
+                while crate::litebox_net()
+                    .lock()
+                    .perform_platform_interaction()
+                    .call_again_immediately()
+                {}
+                core::hint::spin_loop();
```
jaybosamiya-ms, line 960: Minor: I am mildly concerned we'll lose consistency of this test vs the version we'll have in the runner.  I think at minimum we probably want to put a comment here reminding us that this should be in sync with that?

litebox_shim_linux/src/lib.rs
```diff
@@ -169,8 +169,8 @@ pub fn litebox_page_manager<'a>() -> &'a PageManager<Platform, PAGE_SIZE> {
     VMEM.get_or_init(|| alloc::boxed::Box::new(PageManager::new(litebox())))
 }
 
-pub(crate) fn litebox_net<'a>()
--> &'a litebox::sync::Mutex<Platform, litebox::net::Network<Platform>> {
+/// Get the global network manager
+pub fn litebox_net<'a>() -> &'a litebox::sync::Mutex<Platform, litebox::net::Network<Platform>> {
```
jaybosamiya-ms: Rather than expose this, which (imho) is exposing a pretty large structure to the public interface, it might be better to expose something like:

```rs
/// Perform queued network interactions with the outside world.
///
/// This function should be invoked in a loop, based on the returned advice.
pub fn perform_network_interaction() -> PlatformInteractionReinvocationAdvice {
  litebox_net().lock().perform_platform_interaction()
}
```

Or possibly even the looped version itself:
```rs
/// Perform queued network interactions with the outside world.
///
/// This function runs in an infinite loop, invoking `f` whenever all queued interactions are finished, where `f` can decide to wait until more socket and/or device interactions occur.
///
/// `f` is normally expected to return `true`, but if it wants to stop the infinite loop, it can return `false` to terminate the loop.
pub fn continuous_network_interaction_with<R>(f: Fn() -> bool) {
  loop {
    while litebox_net().lock().perform_platform_interaction().call_again_immediately() {}
    if !f() { return; }
  }
}
```

litebox_shim_linux/src/syscalls/net.rs
jaybosamiya-ms: In the other comment, I describe a slightly different function like `continuous_network_interaction_with` but that doesn't quite solve this situation, since the passed `f` would still need to be kept in sync.

litebox/src/net/mod.rs
```diff
@@ -1111,43 +1128,115 @@ where
...
+    pub fn check_events(&self, fd: &SocketFd<Platform>) -> Events {
+        let descriptor_table = self.litebox.descriptor_table();
+        let table_entry = descriptor_table.get_entry(fd);
+        let socket_handle = &table_entry.entry;
+        self.check_socket_events(socket_handle, false)
+    }
+
+    /// Get the [`Events`] for a socket handle [`SocketHandle`].
+    ///
+    /// If `new_events_only` is true, only events that are new since the last check are returned.
```
jaybosamiya-ms: I am not sure if this documentation matches the implementation.  The documentation here seems to imply `new_events_only` is equivalent to holding on to a `prev_events` somewhere, and then always taking a diff of `curr_events - prev_events`, which is not what is happening.  For example, if there are more receivable bytes that from before, _even if_ previously `Events::IN` was returned, it is returned again.

The question is basically one of whether it is strictly-new events or practically-new events, roughly, right?

litebox/src/net/mod.rs
CvvT: Got it, updated now.

litebox/src/net/mod.rs
CvvT: Updated

litebox_shim_linux/src/syscalls/epoll.rs
CvvT: Good point! Fixed

litebox_shim_linux/src/syscalls/net.rs
CvvT: I don't think this has to be kept in sync with runner. This is a working but naive implementation for testing purposes only. In the future, I may only update runner for performance optimization (e.g., do polling on `tun` device instead of busy looping). I added some comment here. I'm not sure if we may have different or same implementation for different runners (windows runner does not support tun yet and we may merge runners in the future).

To test if the implementation is correct for runner, I plan to add an end-to-end integration test to measure the performance.

litebox_shim_linux/src/lib.rs
CvvT: Good point! I prefer the first option as it gives runners more flexibility.

litebox_shim_linux/src/syscalls/net.rs
CvvT: You're right. To support polling (e.g., `epoll`) on sockets, we cannot use `Automatic` mode (unless we change epoll implementation to use busy wait and each iteration calls `perform_platform_interaction`). Now `set_platform_interaction` is called in `litebox_net` as part of the initialization of network.

litebox/src/net/mod.rs
CvvT: I agree it is not clear. It is not diff between events. It supposes to indicate state change of a socket. I updated the comment:

```
/// If `new_events_only` is true, only events representing meaningful state changes
/// since the last check are returned (e.g., new data arrived, socket became writable,
/// new connections became acceptable).
```

wdcui:
#435 is merged. Is this one ready to be merged?

CvvT:
> #435 is merged. Is this one ready to be merged?

I just rebased this PR, resolving all conflicts. @jaybosamiya-ms Please take a look when you have time.

PR #385, opened 2025-09-25
Support getpeername

CvvT:
1. Add support for `getpeername`
2. Support `getsockname` for TCP
3. Make `accept` return the connected remote address

### DIFF
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -97,7 +97,7 @@ impl From<SocketAddrV4> for CSockInetAddr {
 /// Socket address structure for different address families.
 /// Currently only supports IPv4 (AF_INET).
 #[non_exhaustive]
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Debug)]
 pub(crate) enum SocketAddress {
     Inet(SocketAddr),
 }
@@ -1049,11 +1049,11 @@ mod tests {
             )
             .unwrap();
         let server = i32::try_from(server).unwrap();
-        let sockaddr = SocketAddress::Inet(SocketAddr::V4(core::net::SocketAddrV4::new(
-            core::net::Ipv4Addr::from(TUN_IP_ADDR),
+        let server_sockaddr = SocketAddress::Inet(SocketAddr::V4(core::net::SocketAddrV4::new(
+            core::net::Ipv4Addr::from(ip),
             port,
         )));
-        task.sys_bind(server, sockaddr)
+        task.sys_bind(server, server_sockaddr.clone())
             .expect("Failed to bind socket");
         task.sys_listen(server, 1)
             .expect("Failed to listen on socket");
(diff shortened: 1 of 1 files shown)


litebox/src/net/errors.rs
```diff
@@ -50,6 +50,16 @@ pub enum LocalAddrError {
     InvalidFd,
 }
 
+/// Possible errors from [`Network::get_remote_addr`]
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum RemoteAddrError {
+    #[error("Not a valid open file descriptor")]
```
wdcui, line 57: Minor: "Not a valid file descriptor" (since all file descriptors are for opened files)? Or should we say "Not a valid socket"?

litebox/src/net/mod.rs
```diff
@@ -868,6 +878,41 @@ where
...
+                .ok_or(RemoteAddrError::NotConnected)?,
+            Protocol::Udp => socket_handle
+                .udp()
+                .remote_endpoint
+                .ok_or(RemoteAddrError::NotConnected)?,
+            Protocol::Icmp => unimplemented!(),
+            Protocol::Raw { protocol: _ } => unimplemented!(),
+        };
+        match endpoint.addr {
+            smoltcp::wire::IpAddress::Ipv4(ipv4) => {
```
wdcui, line 910: Should we add an error message if it's not IPv4?

litebox/src/net/mod.rs
```diff
@@ -1093,7 +1144,14 @@ where
                     pollee: crate::event::polling::Pollee::new(&self.litebox),
                     previously_sendable: core::sync::atomic::AtomicBool::new(true),
                     recv_queue: core::sync::atomic::AtomicUsize::new(0),
-                }))
+                };
+                if let Some(peer) = peer {
```
wdcui, line 1148: Is there an else branch for this `if`?

litebox/src/net/tests.rs
```diff
@@ -33,7 +33,7 @@ fn bidi_tcp_comms(mut network: Network<MockPlatform>, comms: fn(&mut Network<Moc
 
     // Accept the connection on the listening socket
     let server_fd = network
-        .accept(&listener_fd)
+        .accept(&listener_fd, None)
```
wdcui, line 36: Why do we use `None` here? Do we have a test case to check the remote address/port is returned properly?

litebox/src/net/errors.rs
CvvT: This description is consistent with all other error messages. I can clean it up in a separate PR. "Note a valid socket` implies that it might be valid fd for file.

litebox/src/net/mod.rs
CvvT: We didn't config smoltcp to enable ipv6, so only ipv4 is available.

litebox/src/net/mod.rs
CvvT: No, this is optional. If `peer` is provided, it is filled with the remote address of the accepted connection.

litebox/src/net/tests.rs
CvvT: Yes, it is tested here:

https://github.com/microsoft/litebox/blob/81b512fae381cfe08c96774a1d637fc52fc2dca2/litebox_shim_linux/src/syscalls/net.rs#L1103-L1114

litebox/src/net/errors.rs
wdcui: Sure.

litebox/src/net/mod.rs
wdcui: Got it.

litebox/src/net/tests.rs
wdcui: Alright.

PR #387, opened 2025-09-25
Support MADV_FREE

CvvT:
Support one more option `MADV_FREE` for `madvise`. `MADV_FREE` is similar to `MADV_DONTNEED` in that kernel can release the corresponding physical pages but keep its mapping. The difference is that `MADV_FREE` allows kernel to free it lazily (at any point later). For simplicity, we always free it immediately. Also, `MADV_FREE` should only work on anonymous mapping.

### DIFF
--- a/litebox/src/mm/mod.rs
+++ b/litebox/src/mm/mod.rs
@@ -448,9 +448,8 @@ where
 
     /// Reset pages without removing its mapping.
     ///
-    /// `lazy_free` indicates whether to lazily free the pages (if `true`) or
-    /// immediately free them (if `false`). For simplicity, we always free the pages
-    /// immediately for now. Also, `lazy_free` should only work for anonymous mappings.
+    /// If `anonymous_only` is true and any part of the range is non‑anonymous (i.e., file‑backed),
+    /// returns `Err(VmemResetError::FileBacked)`.
     ///
     /// After calling this function, the memory region remains mapped, but its contents are invalidated.
     /// Subsequent accesses to the region will result in repopulating the memory contents, either from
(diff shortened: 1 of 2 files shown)


litebox_common_linux/src/lib.rs
```diff
@@ -1763,7 +1811,7 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
     Madvise {
         addr: Platform::RawMutPointer<u8>,
         length: usize,
-        behavior: MadviseBehavior,
+        behavior: i32,
```
wdcui: Why do we change this to `i32`?

litebox/src/mm/mod.rs
```diff
@@ -448,6 +448,10 @@ where
 
     /// Reset pages without removing its mapping.
     ///
+    /// `lazy_free` indicates whether to lazily free the pages (if `true`) or
+    /// immediately free them (if `false`). For simplicity, we always free the pages
+    /// immediately for now. Also, `lazy_free` should only work for anonymous mappings.
+    ///
```
jaybosamiya-ms: If we are _effectively_ ignoring `lazy_free`, I don't see why this is an argument added to LiteBox right now.  Shouldn't it just be at the shim layer that `MADV_FREE` -> `MADV_DONTNEED`.

Basically, I am concerned that this adds to the interface that we are promising, while at the same time telling people that we might not be implementing it.

litebox_common_linux/src/lib.rs
CvvT: I reverted it back. We could do the conversion either here or in the shim layer. I'm not sure which one is better.

litebox/src/mm/mod.rs
CvvT: I used it at the following line to return the correct error:

https://github.com/microsoft/litebox/blob/b149d48679489d32d1307629d90c45ae96d74af5/litebox/src/mm/linux.rs#L371-L374

jaybosamiya-ms: I think then the correct parameter should be `fail_if_file_backed: bool` or similar.  That way, you can still get the error you want (at the shim, change MADV_FREE -> MADV_DONTNEED, and set this flag to true), and it is clearer what the actual implications are.  Otherwise, the implications are kind of unclear, and confusing, given that the flag is called `lazy_free` but its _actual_ behavior is non-lazy for anonymous, and fail-for-file-backed.

jaybosamiya-ms: (Obviously, similar update to the other file too)

CvvT: Fair, changed it to `anonymous_only` and updated the doc accordingly. Thanks!

PR #388, opened 2025-09-25
Support setitimer

CvvT:
Similar to #346, another signal related syscall [`setitimer`](https://www.man7.org/linux/man-pages/man2/setitimer.2.html) is added as punchthrough.  We could potentially emulate it with a background thread and `tgkill`, but I'd like to defer it until we figure out how to handle signals. Adding this now so that we have a better understanding how many signal related syscalls we should consider.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1805,6 +1805,19 @@ pub enum PrctlArg<Platform: litebox::platform::RawPointerProvider> {
     CapBSetRead(usize),
 }
 
+#[repr(u32)]
+#[derive(Debug, IntEnum)]
+pub enum IntervalTimer {
+    /// This timer counts down in real (i.e., wall clock) time.  At each expiration, a SIGALRM signal is generated.
+    Real = 0,
+    /// This timer counts down against the user-mode CPU time consumed by the process. The measurement includes CPU time
+    /// consumed by all threads in the process. At each expiration, a SIGVTALRM signal is generated.
+    Virtual = 1,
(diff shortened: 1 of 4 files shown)


litebox_common_linux/src/lib.rs
```diff
@@ -2063,6 +2071,11 @@ pub enum SyscallRequest<'a, Platform: litebox::platform::RawPointerProvider> {
         tid: i32,
         sig: i32,
     },
+    SetITimer {
+        which: i32,
```
jaybosamiya-ms: Nit: for now I see why you're just using the i32, but can you add a TODO comment saying that after it moves away from the punchthrough, this should be an enum? Thanks!

litebox_common_linux/src/lib.rs
```diff
@@ -2631,6 +2645,11 @@ pub enum PunchthroughSyscall<Platform: litebox::platform::RawPointerProvider> {
         thread_id: i32,
         sig: Signal,
     },
+    SetItimer {
```
jaybosamiya-ms: Nit: capitalization in this enum is different from the other enum's capitalization of `SetITimer`.  I don't have a strong preference one way or another, but would prefer consistency :)

CvvT: Good catch! Fixed

litebox_common_linux/src/lib.rs
CvvT: You're right. I added an enum for it. We should use enum for this even if we remove the punchthrough.

PR #395, opened 2025-09-29
Support more socket options

CvvT:
This PR adds support for more socket options, including
1. `LINGER`: our current implementation of `close` does not support graceful close of socket yet, which may cause data loss. Plan to fix it after #383 is merged.
2. `KEEPINTVL`: set `KEEPALIVE` interval
3. `TCP_KEEPIDLE` & `KEEPCNT`: return `EOPNOTSUPP` as smoltcp does not seem to support them.

### DIFF
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -147,6 +147,7 @@ impl Socket {
                         let linger: crate::ConstPtr<litebox_common_linux::Linger> =
                             crate::ConstPtr::from_usize(optval.as_usize());
                         let linger = unsafe { linger.read_at_offset(0) }.ok_or(Errno::EFAULT)?;
+                        // TODO: our current implementation of `close` does not support graceful close yet.
                         if linger.onoff != 0 && linger.linger != 0 {
                             unimplemented!("SO_LINGER with non-zero timeout is not supported yet");
                         }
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
For the ones that are not supported, rather than no-op, does it work for you to return `EOPNOTSUPP`?  `man 7 tcp` says "Any errors defined for ip(7) or the generic socket layer may also be returned for TCP.", and for a whole bunch of options, says "This option should not be used in code intended to be portable." and `man 7 ip` allows for this error.  This means that it is possible that user-mode programs might gracefully handle us returning an error code instead of silently doing a no-op.

CvvT:
> Any errors defined

Good point! I can test if my testing program is okay with `EOPNOTSUPP`.

CvvT:
> For the ones that are not supported, rather than no-op, does it work for you to return `EOPNOTSUPP`? `man 7 tcp` says "Any errors defined for ip(7) or the generic socket layer may also be returned for TCP.", and for a whole bunch of options, says "This option should not be used in code intended to be portable." and `man 7 ip` allows for this error. This means that it is possible that user-mode programs might gracefully handle us returning an error code instead of silently doing a no-op.

Looks like `TCP_KEEPIDLE` and `KEEPCNT` do not need to be supported. `SO_LINGER` is tricky because it requires graceful close of a socket (which should be the default behavior, but we don't support yet). Will fix it in another PR.

PR #398, opened 2025-09-30
Eliminate StdioFile; manage via LiteBox raw FDs

jaybosamiya-ms:
This PR sets up the basis for directly using LiteBox raw FDs within the Linux shim.  While other components will be migrated in future PRs, as a way to demonstrate what this usage looks like, and to validate the migration, this PR also at the same time eliminates `StdioFile` from the Linux shim, switching all stdio behavior to using the raw FDs.

Given that this PR is setting up the basis for future PRs that will migrate components, this one is a little bit on the larger side in terms of code addition (since it kind of needs to place a bunch of `todo!`s and such that will be cleared out soon).  For future PRs that are cleaning up the shim, I expect some of the redundant code to go away.  For now, yes, there are some redundancies with (for example) regular files.

This PR depends on (and is blocked on #393).

---

 (connected to #31).

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -259,14 +259,16 @@ pub fn sys_mkdir(pathname: impl path::Arg, mode: u32) -> Result<(), Errno> {
 pub(crate) fn do_close(desc: Descriptor) -> Result<(), Errno> {
     match desc {
         Descriptor::LiteBoxRawFd(raw_fd) => {
-            // XXX(jayb): This is not ideal, but we should NEVER hit the `unreachable!` below in
-            // practice if all the places that use `fd_from_raw_integer` actually follow the
-            // constraints declared in its documentation. We need to run this for more than just
-            // once because it _can_ accidentally happen that the internals of `fd_from_raw_integer`
-            // itself, or in the short durations that something else is attempting to read/write to
-            // the FD at the same time that another thread is attempting to close that exact same
-            // FD. This should be incredibly rare in practice. To make this better, we probably need
-            // to update `fd_consume_raw_integer` interface on the LiteBox side itself.
+            // XXX(jayb): This is not ideal, but we should hopefully practically not hit the
+            // `unreachable!` below in practice if all the places that use `fd_from_raw_integer`
+            // actually follow the constraints declared in its documentation. We need to run this
+            // for more than just once because it _can_ accidentally happen that the internals of
+            // `fd_from_raw_integer` itself, or in the short durations that something else is
+            // attempting to read/write to the FD at the same time that another thread is attempting
+            // to close that exact same FD. This should be incredibly rare in practice. Care
+            // obviously must be taken for blocking operations to release things between attempted
+            // reads/writes. To make this better, we probably need to update
+            // `fd_consume_raw_integer` interface on the LiteBox side itself.
             for _fuel in 0..1000 {
                 let mut dt = crate::litebox().descriptor_table_mut();
                 return match dt.fd_consume_raw_integer(raw_fd) {


jaybosamiya-ms:
Hmm, seems like the nightlies that were being used on LVBS, SNP are now older than Rust stable, and we might need to bump them to satisfy the CI.  I'll do that separately, but yeah until then the CI might show a bunch of reds on them.

EDIT: #399

jaybosamiya-ms:
The PR itself is ready to review.  The CI will expectedly fail until the blockers #393 and #399 are merged, after which it should become green.

jaybosamiya-ms:
@CvvT I've added you as an additional reviewer to this PR because you'd built the `StdioFile` abstraction that this is removing, so might notice some things.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -253,11 +258,51 @@ pub fn sys_mkdir(pathname: impl path::Arg, mode: u32) -> Result<(), Errno> {
...
+        Descriptor::LiteBoxRawFd(raw_fd) => {
+            // XXX(jayb): This is not ideal, but we should NEVER hit the `unreachable!` below in
+            // practice if all the places that use `fd_from_raw_integer` actually follow the
+            // constraints declared in its documentation. We need to run this for more than just
+            // once because it _can_ accidentally happen that the internals of `fd_from_raw_integer`
+            // itself, or in the short durations that something else is attempting to read/write to
+            // the FD at the same time that another thread is attempting to close that exact same
+            // FD. This should be incredibly rare in practice. To make this better, we probably need
+            // to update `fd_consume_raw_integer` interface on the LiteBox side itself.
+            for _fuel in 0..1000 {
```
CvvT: What if a read/write to the FD is a blocking operation like `receive` for socket (socket is not included in this PR but I assume it is similar)? Even if we implement it in a way that acquires and drops FD in each iteration inside a loop, there is still a chance for race to happen (i.e., `fd_consume_raw_integer` always fails because FD is just acquired at some iteration).

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -827,9 +934,16 @@ pub fn sys_ioctl(
             .ok_or(Errno::EFAULT)?
             .into_owned();
         match desc {
-            Descriptor::File(file) => todo!(),
-            Descriptor::Stdio(file) => {
-                file.inner.set_status(OFlags::NONBLOCK, val != 0);
+            Descriptor::LiteBoxRawFd(raw_fd) => {
+                // TODO: stdio NONBLOCK?
+                todo!()
```
CvvT, line 941: Previous implementation defers the panic to `read`/`write`. I don't recall the detail. Maybe one of the test application sets NONBLOCK on stdio but never calls `read`/`write` on it.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -973,9 +1113,65 @@ pub fn sys_epoll_pwait(
...
-        Descriptor::Stdio(file) => Descriptor::Stdio(file.dup(flags.contains(OFlags::CLOEXEC))),
+        Descriptor::LiteBoxRawFd(raw_fd) => {
+            use alloc::sync::Arc;
+            use litebox::fd::ErrRawIntFd;
+            let mut dt = litebox().descriptor_table_mut();
+            match dt.fd_from_raw_integer(*raw_fd) {
+                Ok(fd) => {
+                    let fd: Arc<litebox::fd::TypedFd<crate::LinuxFS>> =
+                        fd.upgrade().ok_or(Errno::EBADF)?;
+                    let cloexec = dt
```
CvvT: `cloexec` should come from the argument not FD itself.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -973,9 +1113,65 @@ pub fn sys_epoll_pwait(
...
+                            // fs+net are the only subsystems at the moment
+                            unreachable!()
+                        }
+                    }
+                }
+            }
+        }
+        Descriptor::File(file) => {
+            let mut dt = litebox().descriptor_table_mut();
+            let cloexec = dt
```
CvvT: Similar

litebox_shim_linux/src/syscalls/file.rs
jaybosamiya-ms: Yes, things are not supposed to hold on to a strong reference for long.  Even blocking operations are indeed supposed to acquire and quickly drop (this is why the actual LiteBox crate network implementation is actually non-blocking).  The design for the sockets will become more obvious once I am migrating that over (explicitly not in this PR).

And yes, there is a small chance of a race where the timing of the short duration of pickups and the consume attempts turn out to _perfectly_ align up.  This is why I've placed a large number on the attempts (to reduce the chances of this happening), and also made it quite loud if even 1000 attempts later it fails (it will cause a panic).

I've also already included this detailed comment for this exact reason.  I am not happy with the design here, but it is what we need to do at the moment.  It is not ideal, and I don't like it, but trying to handle this change along with everything else in the same go is quite complicated _especially_ when other stuff keeps changing in the shim layer.

Fwiw, I _do_ have a design for a fix for this, but if I tried doing that in the same PR, then more changes to the shim keep piling on, and it makes the rebasing and interactions with subtle semantics hard.

Does the comment I've already included not already document your concern?

litebox_shim_linux/src/syscalls/file.rs
jaybosamiya-ms: I am surprised by a program that sets NONBLOCK and then never reads/writes.

In any case, the `todo!()` will complain loudly if we ever hit this in practice, and we can fix it up then.  Whether the panic is deferred to the read/write or handled sooner, this non-blocking stdio is something we should fix _as a whole_, not simply by "the application never reads/writes so it is ok to defer the panic".

I think fixing non-blocking stdio (especially since the previous implementation was just papering over things rather than actually handling, and it is something not covered by our CI and tests) is something that is out of scope of this PR.  Please let me know if you disagree.

litebox_shim_linux/src/syscalls/file.rs
jaybosamiya-ms: Good catch!  Fixed.

litebox_shim_linux/src/syscalls/file.rs
jaybosamiya-ms: Good catch!  Fixed.

litebox_shim_linux/src/syscalls/file.rs
CvvT: No, it's clear. I guess my thinking is any race could be rare in practice but could be triggered easily with carefully crafted input; thus "NEVER" sounds too strong to me. And I was not sure if you had considered blocking read/write. Since you already have a fix in mind, that's great and we can leave it for now.

litebox_shim_linux/src/syscalls/file.rs
CvvT: Sounds good, thanks!

litebox_shim_linux/src/syscalls/file.rs
jaybosamiya-ms: Fair point, the "NEVER" indeed seems stronger than I intended, I'll make the tiny tweak to tone it down a little bit before merging, thanks!

PR #400, opened 2025-09-30
Set file limit

CvvT:
Support setting file limit (i.e., `RLIMIT_NOFILE`) via prlimit/setrlimit and enforce it when adding a new fd to file descriptor table (return EMFILE when it exceeds the limit).

Resource limit is also per-process, so merge it with NR_THREAD into a single global variable.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -491,8 +491,6 @@ pub(crate) fn sys_gettid() -> i32 {
 // TODO: enforce the following limits:
 const RLIMIT_NOFILE_CUR: usize = 1024 * 1024;
 const RLIMIT_NOFILE_MAX: usize = 1024 * 1024;
-/// Number of resource limits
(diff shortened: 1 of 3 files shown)


litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -474,30 +491,110 @@ pub(crate) fn sys_gettid() -> i32 {
 // TODO: enforce the following limits:
 const RLIMIT_NOFILE_CUR: usize = 1024 * 1024;
 const RLIMIT_NOFILE_MAX: usize = 1024 * 1024;
+/// Number of resource limits
+pub const RLIMIT_COUNT: usize = 16;
```
wdcui: Is 16 sufficient?

CvvT: I moved this constant to:
```
impl RlimitResource {
    /// Maximum value for RlimitResource
    pub const RLIM_NLIMITS: usize = RlimitResource::RTTIME as usize + 1;
}
```

PR #403, opened 2025-10-03
Fix memory and stack alignment issues

### LINKED ISSUE #402: Fixing LiteBox LVBS crashes with the new version of Rust
Since #399 updates the Rust version, LiteBox for LVBS starts to crash.
I will try to fix this issue as soon as possible. Meanwhile, I recommend
others who are working on LVBS use older Rust (version 1.88).

sangho2:
This PR fixes #402. There were two issues.

First, it fixes the memory alignment issue of the memory container for
VTL0-VTL1 data exchange.
Second, it fixes the stack alignment issue which panics `memcmp` and
`memcpy`.

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -194,13 +194,7 @@ impl<Host: HostInterface> LinuxKernel<Host> {
             let src_ptr = page_addr.wrapping_add(page_offset).cast::<T>();
             assert!(src_ptr.is_aligned(), "src_ptr is not properly aligned");
 
-            let mut boxed = Box::<T>::new_uninit();
-            let dst_ptr = boxed.as_mut_ptr();
-
-            let boxed = unsafe {
-                core::ptr::copy_nonoverlapping(src_ptr, dst_ptr, 1);
-                boxed.assume_init()
-            };
+            let boxed = Box::<T>::new(unsafe { core::ptr::read_volatile(src_ptr) });
 
             assert!(
                 self.unmap_vtl0_pages(page_addr, length).is_ok(),
@@ -236,11 +230,8 @@ impl<Host: HostInterface> LinuxKernel<Host> {
             let dst_ptr = page_addr.wrapping_add(page_offset).cast::<T>();
             assert!(dst_ptr.is_aligned(), "dst_ptr is not properly aligned");
 
-            let src_ptr = core::ptr::from_ref::<T>(value);
+            unsafe { core::ptr::write_volatile(dst_ptr, *value) };
 
-            unsafe {
-                core::ptr::copy_nonoverlapping(src_ptr, dst_ptr, 1);
-            }
             assert!(
                 self.unmap_vtl0_pages(page_addr, length).is_ok(),
                 "Failed to unmap VTL0 pages"
@@ -276,11 +267,9 @@ impl<Host: HostInterface> LinuxKernel<Host> {
             let dst_ptr = page_addr.wrapping_add(page_offset).cast::<T>();
             assert!(dst_ptr.is_aligned(), "dst_ptr is not properly aligned");
 
-            let src_ptr = value.as_ptr();
+            let dst = unsafe { core::slice::from_raw_parts_mut(dst_ptr, value.len()) };
+            dst.copy_from_slice(value);
 
-            unsafe {
-                core::ptr::copy_nonoverlapping(src_ptr, dst_ptr, value.len());
-            }
             assert!(
                 self.unmap_vtl0_pages(page_addr, length).is_ok(),
                 "Failed to unmap VTL0 pages"
@@ -316,11 +305,8 @@ impl<Host: HostInterface> LinuxKernel<Host> {
             let src_ptr = page_addr.wrapping_add(page_offset).cast::<T>();
             assert!(src_ptr.is_aligned(), "src_ptr is not properly aligned");
 
-            let dst_ptr = buf.as_mut_ptr();
-
-            unsafe {
-                core::ptr::copy_nonoverlapping(src_ptr, dst_ptr, buf.len());
-            }
+            let src = unsafe { core::slice::from_raw_parts(src_ptr, buf.len()) };
+            buf.copy_from_slice(src);
 
             assert!(
                 self.unmap_vtl0_pages(page_addr, length).is_ok(),
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -447,8 +447,9 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
         return Err(Errno::EINVAL);
     }
 
-    let mut kinfo_buf = avec![0u8; kinfo_mem.len()];
-    let mut kdata_buf = avec![0u8; kdata_mem.len()];
+    let mut kinfo_buf = avec![[{ core::mem::align_of::<HekiKernelInfo>() }] | 0u8; kinfo_mem.len()];
+    let mut kdata_buf =
+        avec![[{ core::mem::align_of::<HekiKernelSymbol>() }] | 0u8; kdata_mem.len()];
 
     kinfo_mem
         .read_bytes(kinfo_mem.start().unwrap(), &mut kinfo_buf)
@@ -527,7 +528,7 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
         "Module ELF size exceeds the maximum allowed size"
     );
 
-    let mut original_elf_data = avec![0u8; elf_size];
+    let mut original_elf_data = vec![0u8; elf_size];
     module_as_elf
         .read_bytes(module_as_elf.start().unwrap(), &mut original_elf_data)
         .map_err(|_| Errno::EINVAL)?;


litebox_platform_lvbs/src/lib.rs
```diff
@@ -739,3 +743,63 @@ pub fn platform_low() -> &'static Platform {
         .get()
         .expect("set_platform_low should have already been called before this point")
 }
+
+/// This function behaves like C memcpy.
+/// # Safety
+/// `dest` and `src` must be valid for reads/writes of `count` bytes, unless
+/// `count == 0`, in which case no memory is touched and `dest` is returned.
+/// This function is only for debug builds to avoid erroneous safety checks of
```
jstarks: What erroneous safety checks?

sangho2: `compiler-builtins-mem` does use a bunch of pointer casting (to `*u8`--`*u128`) for optimized memory copy and comparison, which results in unexpected precondition check errors with debug build. The below is test code that panics (#402).

```rust
let data = vec![0u8; 4096]; // data.as_ptr() as u64 % 4096 == 0
if data.as_slice() != data.as_slice() {
    println!("123");
} else {
    println!("456");
}
```

CvvT: why u128 is problematic? In the example, data.as_ptr() is 4k-aligned.

sangho2: Aligning `data[0]` is not enough and this still panics. Don't know whether it is fully reliable, but GPT-5 says that core inserts an alignment check before doing word-size loads. `a.iter().eq(b.iter())` works (without this overriding), so this is not about invalid memory addresses or corruptions. Another hypothesis is this might be due to our `abort` panic strategy for LVBS, our allocator doesn't produce good enough provenance, or something else.

jstarks: I see, so this is a bug in `compiler_builtins`, that they are invalidly casting to and reading from unaligned pointers. Should we also open an issue there?

jstarks: (It seems that this isn't the first time `compiler_builtins` has had UB in their memcpy implementation, e.g. https://github.com/rust-lang/compiler-builtins/issues/559.)

jstarks: Actually, if you're seeing an undefined opcode exception, I wonder if it's not alignment but that `compiler_builtins` is depending on SSE2 being available. I noticed that your target specification disables SSE support.

sangho2: Thanks for sharing an interesting issue @jstarks ! 
I think we might need to open an issue in `compiler_builtins` (of course, once I double check it).
At least for now, what Iconfirm is the debug build of LiteBox-LVBS cannot use functions
relying on `compiler_builtins`'s `memcmp` and `memcpy`.

I checked the binary using `objdump`. RIP was at `ud2`.

`compiler-builtins` uses SSE2 for `c_string_length()` and some math functions.

jstarks: > I checked the binary using objdump. RIP was at ud2.

Interesting. Does `objdump -l -d` tell you the line numbers accurately enough to figure out what the exact check is?

This seems like a reasonable enough change to unblock, but I think it's probably worth figuring out what's really going on here, to make sure there's not a problem somewhere else.

sangho2: The below is manual backtrace based on `objdump` and stack dump for this vector slice comparison (we don't have unwinding and back tracing yet...)

```rust
83db5ff: ud2 // invalid opcode exception
824fa50: <x86_64::addr::VirtAddr as core::fmt::LowerHex>::fmt
83d7730: <core::fmt::num::<impl core::fmt::LowerHex for u64>>::fmt
83db330: core::ptr::const_ptr::<impl *const T>::read
83dbc5f: core::ptr::const_ptr::<impl *const T>::read_unaligned
83dd607: compiler_builtins::mem::memcmp
8247d47: <core::slice::cmp::<impl core::cmp::PartialEq<[U]> for [T]>>::ne
8257a36: <core::str::iter::Bytes as core::iter::traits::iterator::Iterator>::next
83cc16b: <core::result::Result<T, E> as core::ops::try_trait::Try>::branch
8292142: litebox_platform_lvbs::mshv::vsm::mshv_vsm_validate_guest_module
```

my guess is `core::ptr::read()` -> `ub_checks::assert_unsafe_precondition!()` -> `ub_checks::maybe_is_aligned_and_not_null()` -> `ud2` because this `ud2` is within `core::ptr::read_precondition_check()` which calls `core::ub_checks::maybe_is_aligned_and_not_null()`.

sangho2: Found the root cause eventually. Someone suffered from the same issue https://github.com/rust-lang/rust/issues/145668 and it turned out that all these were due to stack misalignment.
Everything seems fine now. Thank you!

litebox_platform_lvbs/src/lib.rs
```diff
@@ -191,25 +191,26 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
-                    core::mem::size_of::<T>(),
-                );
-            }
+            let mut boxed = Box::<T>::new_uninit();
+            let dst_ptr = boxed.as_mut_ptr();
+
+            let boxed = unsafe {
+                core::ptr::copy_nonoverlapping(src_ptr, dst_ptr, 1);
+                boxed.assume_init()
+            };
```
CvvT: I believe this can be simplified by:

```
let boxed = Box::<T>::new(unsafe { *src_ptr });
```

litebox_platform_lvbs/src/lib.rs
```diff
@@ -232,12 +233,13 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
+            let src_ptr = core::ptr::from_ref::<T>(value);
+
             unsafe {
-                core::ptr::copy_nonoverlapping(
-                    core::ptr::from_ref::<T>(value).cast::<u8>(),
-                    page_addr.wrapping_add(page_offset).cast::<u8>(),
-                    core::mem::size_of::<T>(),
-                );
+                core::ptr::copy_nonoverlapping(src_ptr, dst_ptr, 1);
             }
```
CvvT: Similarly, `unsafe { *dst_ptr = *value };`

litebox_platform_lvbs/src/lib.rs
```diff
@@ -271,12 +273,13 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
+            let src_ptr = value.as_ptr();
+
             unsafe {
-                core::ptr::copy_nonoverlapping(
-                    value.as_ptr().cast::<u8>(),
-                    page_addr.wrapping_add(page_offset).cast::<u8>(),
-                    core::mem::size_of_val(value),
-                );
+                core::ptr::copy_nonoverlapping(src_ptr, dst_ptr, value.len());
             }
```
CvvT: Similarly, 
```
let dst = unsafe { core::slice::from_raw_parts_mut(dst_ptr, value.len()) };
dst.copy_from_slice(value);
```

It's still two lines of code, but I feel it is clearer.

BTW, do we need to check if `length` >= `value.len() * size_of::<T>()` here?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -310,12 +313,13 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
+            let dst_ptr = buf.as_mut_ptr();
+
             unsafe {
-                core::ptr::copy_nonoverlapping(
-                    page_addr.wrapping_add(page_offset).cast::<u8>(),
-                    buf.as_mut_ptr().cast::<u8>(),
-                    core::mem::size_of_val(buf),
-                );
+                core::ptr::copy_nonoverlapping(src_ptr, dst_ptr, buf.len());
             }
```
CvvT: Similarly

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -437,8 +448,8 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
         return Err(Errno::EINVAL);
     }
 
-    let mut kinfo_buf = vec![0u8; kinfo_mem.len()];
-    let mut kdata_buf = vec![0u8; kdata_mem.len()];
+    let mut kinfo_buf = avec![0u8; kinfo_mem.len()];
+    let mut kdata_buf = avec![0u8; kdata_mem.len()];
```
CvvT: What is the alignment requirement for these two bufs? Do we need to specify it?

litebox_platform_lvbs/src/lib.rs
sangho2: Thanks! used to use `copy_nonoverlapping` because it didn't enforce alignment before and wanted to avoid optimization. Now, `read_volatile` and `write_volatile` should be enough.

litebox_platform_lvbs/src/lib.rs
sangho2: revised. Don't need to check `length` because `map_vtl0_phys_range` doesn't do partial mapping.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: revised. now we explicitly specify alignment (used to use `avec`'s default one because it cache-aligns buffers by default).

PR #404, opened 2025-10-03
Add support for std::spawn and separate stacks in Linux userland

wdcui:
This PR adds the support for std::spawn and separate stacks in Linux userland.

The key trick in enabling this support is to stitching the stack together between `thread_start_asm` and `syscall_callback`. Specifically, the stack/frame pointers we store in `thread_start_asm` right before calling `thread_start_internal` is the same as what's restored in `syscall_callback` right before calling `syscall_handler`. Furthermore, the stack layout of `thread_start_asm` is the same as `syscall_callback`, which allows the return to the caller of `thread_start_asm` when `sys_exit_group` or `sys_exit` is called on the main/child thread.

I had to disable seccomp tests because it cannot differentiate syscalls made by the platform's std from those made by the guest.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -1087,8 +1087,6 @@ pub(crate) fn sys_tgkill(
 
 #[cfg(test)]
 mod tests {
-    use litebox::mm::linux::PAGE_SIZE;
-
     #[cfg(target_arch = "x86_64")]
     #[test]
     fn test_arch_prctl() {


litebox_runner_linux_userland/src/lib.rs
```diff
@@ -309,46 +309,49 @@ fn load_program(
...
+        rdx: 0,
+        rsi: 0,
+        rdi: 0,
+        orig_rax: 0,
+        rip: 0,
+        cs: 0x33, // __USER_CS
+        eflags: 0,
+        rsp: 0,
+        ss: 0x2b, // __USER_DS
+    };
```
sangho2, line 323: OP-TEE TAs use a different set of registers for entry, possibly resulting in a compatibility issue. Other registers might be fine but `rcx` is concerning. We might need to 1) implement a custom `thread_start_internal` for OP-TEE TAs or 2) modify OP-TEE devkit/libutee to make it use different registers.

https://github.com/microsoft/litebox/blob/7466b065189ed8d90f2e8b6c35f2b438ed8701ac/litebox_runner_optee_on_linux_userland/src/dispatcher_impl/optee_ta.rs#L102-L119

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -409,141 +621,41 @@ impl litebox::platform::ThreadProvider for LinuxUserland {
...
-                _ => panic!("unexpected error {err}"),
-            });
+        {
+            ctx_copy.ebx = entry_point;
+            ctx_copy.ecx = stack.as_usize() + stack_size;
         }
-        Ok(ret)
+
+        // TODO: do we need to wait for the handle in the main thread?
+        let _handle = std::thread::spawn(move || thread_start(&thread_args, ctx_copy));
```
sangho2: In case of OP-TEE TAs, we might need to run `thread_start_asm` for each `invoke_command`, which might result in repetitive thread spawning and zombie threads. Do we plan to call `join()` at some point?

wdcui: Do we expect to receive multiple `invoke_command` requests simultaneously? If so, how would that be delivered from vtl0 to vtl1?

wdcui:
I tried to turn on a multi-thread test for Windows userland but found that the multi-threaded program makes signal related syscalls which are not supported on Windows userland yet. We will need to add a multi-thread test for Windows userland after the signal provider is implemented on it.

litebox_platform_linux_userland/src/lib.rs
sangho2: Don't worry about this comment. I misunderstood the code. `invoke_command` will be per thread.

litebox_shim_optee/src/lib.rs
```diff
@@ -477,8 +477,6 @@ pub fn optee_command_dispatcher(session_id: u32, is_sys_return: bool) -> ! {
         optee_command_completion_queue().push(session_id, stack.get_params_address());
 
         // TODO: We will move the following line into the platform or the runner after the OPTEE shim is refactored.
```
jstarks: Remove the TODO as well

litebox_common_linux/src/lib.rs
```diff
@@ -1257,6 +1222,12 @@ pub struct Task<Platform: litebox::platform::RawPointerProvider> {
     pub credentials: alloc::sync::Arc<Credentials>,
     /// Command name (usually the executable name, excluding the path)
     pub comm: [u8; TASK_COMM_LEN],
+    /// Stored stack pointer for the thread. Used for switching stacks between the guest and the platform.
+    pub stored_sp: usize,
+    /// Stored instruction pointer for the thread. Used for a platform to terminate a thread gracefully.
+    pub stored_bp: usize,
+    /// Indicate if it's about to terminate the current thread
+    pub to_terminate: usize,
```
CvvT: bool instead of usize?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1356,27 +1491,64 @@ syscall_callback:
...
+    call swap_fsgs
+
+    /* Retrieve platform rsp and rbp switch to them */
+    mov rdi, rbp
+    call swap_bp
+    mov rbp, rax
+
+    mov rdi, rsp
+    call swap_sp
+    mov rsp, rdx
```
CvvT: "mov rsp, rax"?

litebox_common_linux/src/lib.rs
```diff
@@ -1257,6 +1222,12 @@ pub struct Task<Platform: litebox::platform::RawPointerProvider> {
     pub credentials: alloc::sync::Arc<Credentials>,
     /// Command name (usually the executable name, excluding the path)
     pub comm: [u8; TASK_COMM_LEN],
+    /// Stored stack pointer for the thread. Used for switching stacks between the guest and the platform.
+    pub stored_sp: usize,
+    /// Stored instruction pointer for the thread. Used for a platform to terminate a thread gracefully.
+    pub stored_bp: usize,
```
CvvT: Is it possible to just store one? It looks like there is always (or at least it is okay to) `mov rsp, rbp` after `swap_sp` and `swap_bp`.

litebox_common_linux/src/lib.rs
```diff
@@ -1197,17 +1161,18 @@ pub struct NewThreadArgs<
...
-    /// the guest thread entry. This avoids races when the parent wants to write
-    /// its child-tid (parent_tid) into parent memory after spawn.
-    pub start_gate: Option<alloc::sync::Arc<litebox::sync::Mutex<Platform, ()>>>,
     /// Task struct that maintains all per-thread data
     pub task: alloc::boxed::Box<Task<Platform>>,
     /// A callback function that *MUST* be called when the thread is created.
     ///
     /// Note that `task.tid` must be set correctly before this function is called.
-    pub callback: fn(Self),
+    pub callback: fn(&Self),
```
CvvT: Why changed it to `&Self`?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -312,54 +318,254 @@ impl litebox::platform::ExitProvider for LinuxUserland {
...
-                (code as isize).reinterpret_as_unsigned(),
-                // Unused by the syscall but would be checked by Seccomp filter if enabled.
-                syscall_intercept::SYSCALL_ARG_MAGIC,
-            )
-        }
-        .expect("Failed to exit group");
-
-        unreachable!("exit_group should not return");
+    fn exit(&self, _code: Self::ExitCode) -> ! {
+        todo!("this function is not needed")
```
CvvT: This would be called by syscall `exit_group`; why we don't need it?

litebox/src/platform/mod.rs
```diff
@@ -652,14 +655,14 @@ pub trait ThreadLocalStorageProvider {
     /// # Panics
     ///
     /// Panics if TLS is set already.
-    fn set_thread_local_storage(&self, value: Self::ThreadLocalStorage);
+    fn set_thread_local_storage(value: Self::ThreadLocalStorage);
```
jaybosamiya-ms, line 632: I am not sure why the interfaces here (and related) were changed to not take `&self`

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -409,141 +621,41 @@ impl litebox::platform::ThreadProvider for LinuxUserland {
...
-                in("rsi") size_of::<litebox_common_linux::CloneArgs>(),
-                // Unused by the syscall but would be checked by Seccomp filter if enabled.
-                in("rdx") syscall_intercept::SYSCALL_ARG_MAGIC,
-                out("rcx") _, // rcx is used to store old rip
-                out("r11") _, // r11 is used to store old rflags
-                options(nostack, preserves_flags)
-            );
+        {
+            ctx_copy.r10 = entry_point;
+            ctx_copy.r11 = stack.as_usize() + stack_size;
```
CvvT: The original values are discarded here. Guest may expect all registers except rsp/rip/rax to remain the same. We can add additional arguments to `thread_args`.

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -309,46 +309,49 @@ fn load_program(
...
-    );
+    let pt_regs = litebox_common_linux::PtRegs {
+        r15: 0,
+        r14: 0,
+        r13: 0,
+        r12: 0,
+        rbp: 0,
+        rbx: 0,
+        r11: loaded_program.user_stack_top,
+        r10: loaded_program.entry_point,
```
CvvT: why not set rsp and rip?

litebox_common_linux/src/lib.rs
```diff
@@ -1197,17 +1161,18 @@ pub struct NewThreadArgs<
...
+}
+
+unsafe impl<Platform> Send for NewThreadArgs<Platform> where
+    Platform:
+        litebox::platform::RawPointerProvider + litebox::sync::RawSyncPrimitivesProvider + Send // if your platform object must be movable across threads
+{
```
jaybosamiya-ms, line 1216: I'm ok with this introduction of `Send` here I think, but I am not sure I understand the comment here.  Do you mean to say "`NewThreadArgs` is Send only if `Platform` is also `Send`", if so, then that matches the code written.  As the comment is currently written, I am somewhat confused how I should be reading it or what information it is telling me.

litebox_common_linux/src/lib.rs
jaybosamiya-ms: It says stored instruction pointer in the comment, but then uses `stored_bp` which implies `rbp` rather than `rip`.  Which one is it supposed to be?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -261,71 +261,73 @@ fn exit_code(
...
-            addr: clear_child_tid,
-            flags: litebox_common_linux::FutexFlags::PRIVATE,
-            count: 1,
-        });
-    }
-    if let Some(robust_list) = tls.current_task.robust_list.take() {
-        let _ = wake_robust_list(robust_list);
-    }
+pub(crate) fn sys_exit(_status: i32) {
+    litebox_platform_multiplex::Platform::with_thread_local_storage_mut(|tls| {
```
CvvT: Do we need to call `release_thread_local_storage` somewhere in `spawn_thread`?

litebox_common_linux/src/lib.rs
jaybosamiya-ms: Should this be a `bool` instead?  Alternatively, the comment should talk about what the different possible values here mean.

litebox/src/platform/mod.rs
```diff
@@ -99,6 +99,9 @@ pub trait ThreadProvider: RawPointerProvider + ExitProvider {
 
     /// Terminate the current thread with the given exit code.
     fn terminate_thread(&self, code: Self::ExitCode) -> !;
+
+    /// Get the next available thread ID.
+    fn next_thread_id(&self) -> i32;
```
jaybosamiya-ms: Does this reserve the thread ID?  What happens if it is invoked twice in a row?  What if two different threads invoke this at the same time?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -400,23 +402,16 @@ pub(crate) fn sys_clone(
...
-        None
-    };
+    let platform = litebox_platform_multiplex::platform();
+    let child_tid = platform.next_thread_id();
+    if let Some(parent_tid_ptr) = set_parent_tid {
+        let _ = unsafe { parent_tid_ptr.write_at_offset(0, child_tid) };
+    }
+    if let Some(child_tid_ptr) = set_child_tid {
+        let _ = unsafe { child_tid_ptr.write_at_offset(0, child_tid) };
+    }
```
CvvT: Why moved it from child thread to parent thread?

litebox_platform_freebsd_userland/src/lib.rs
```diff
@@ -49,7 +49,7 @@ impl FreeBSDUserland {
             reserved_pages: Self::read_proc_self_maps(),
         };
 
-        platform.set_init_tls();
+        Self::set_init_tls();
```
jaybosamiya-ms, line 53: I'm not entirely sure why these changes were made from `&self` -> `Self::`.  I'm not going to mark this everywhere this sort of change happens, but basically, this seems like unrelated changes that have been pushed into the same PR?

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -106,23 +106,26 @@ impl SnpLinuxKernel {
                 egid: boot_params.egid as usize,
             }),
             comm: [0; litebox_common_linux::TASK_COMM_LEN],
+            stored_sp: 0,
+            stored_bp: 0,
```
jaybosamiya-ms: Rather than using 0 to mean "nothing is stored", it is better to actually use `Option`s at the types.  If you are concerned about space use, you can actually use `Option<NonZeroUsize>` or similar but generally speaking, `Option<usize>` is prob good enough here.  Easier to ensure that there isn't a mistake being made, and errors will be a lot noisier with proper handling or panics rather than silently looking at the zero page.

litebox/src/platform/mod.rs
jstarks: Why does the platform need to provide this, instead of the shim?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -261,71 +261,73 @@ fn exit_code(
...
+            let _ = sys_futex(litebox_common_linux::FutexArgs::Wake {
+                addr: clear_child_tid,
+                flags: litebox_common_linux::FutexFlags::PRIVATE,
+                count: 1,
+            });
+        }
+        if let Some(robust_list) = tls.current_task.robust_list.take() {
+            let _ = wake_robust_list(robust_list);
+        }
+        tls.current_task.to_terminate = 1;
```
jstarks: I have been thinking about this problem as well. I think the better model is for `handle_syscall_request` to return some kind of `ContinueOperation` enum, something like:

```rust
enum ContinueOperation {
    ResumeGuest { return_value: usize },
    ExitThread(ExitCode),
    ExitProcess(ExitCode),
}
```

Rather than stash this in TLS.

litebox_platform_linux_userland/src/lib.rs
CvvT: Looks like `sys_exit_group` now returns instead of calling this function. Then effectively, `exit_group` now becomes `exit`?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1356,27 +1491,64 @@ syscall_callback:
...
+    /* Pass syscall number and pt_regs (saved on the guest stack) */
+    mov rdi, r15
+    mov rsi, r14
 
     /* Call syscall_handler */
     call syscall_handler
+    mov rbx, rax
+
+    /* Check if to terminate the thread */
+    call to_terminate_thread
```
jstarks: I think syscall_handler should just update the guest rax in pt_args rather than return it. Then you can have syscall_handler return whether the thread should be terminated, saving a step here.

litebox_common_linux/src/lib.rs
wdcui: I thought about this. We need the aligned `rsp` before handling syscall. We can store one if we do the alignment after swapping rbp. That actually can work. Will fix it together with other changes based on other discussions.

litebox_common_linux/src/lib.rs
wdcui: It's because now it passes a reference instead of a copy.

litebox_platform_linux_userland/src/lib.rs
wdcui: Yes, I didn't do the clean up to remove `exit` from the trait. Thought I would wait on that to see the review comments.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -186,7 +189,7 @@ extern "C" fn thread_start(
 
     // Set up thread-local storage for the new thread. This is done by
     // calling the actual thread callback with the unpacked arguments
-    (thread_start_args.thread_args.callback)(*(thread_start_args.thread_args));
+    (thread_start_args.thread_args.callback)(&thread_start_args.thread_args);
```
jaybosamiya-ms: Any reason this is non-consuming?

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: Do you mean `unreachable!()`?  It is not needed for a particular shim, or is it impossible for it to be reached?  What if some shim literally only depends on `ExitProvider` (does not have any notion of threads) and runs `exit`?  This would show up (say) in the Wasm shim.

litebox_platform_linux_userland/src/lib.rs
wdcui: We could do that; one of the reasons I didn't do it is because it was painful to make thread_args movable for spawn_thread.

litebox_runner_linux_userland/src/lib.rs
wdcui: Good point! Let me see if I can make that work.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -409,141 +621,41 @@ impl litebox::platform::ThreadProvider for LinuxUserland {
...
-            syscalls::syscall2(
-                syscalls::Sysno::exit,
-                (code as isize).reinterpret_as_unsigned(),
-                // Unused by the syscall but would be checked by Seccomp filter if enabled.
-                syscall_intercept::SYSCALL_ARG_MAGIC,
-            )
-        }
-        .expect("Failed to exit");
+    fn terminate_thread(&self, _code: Self::ExitCode) -> ! {
+        todo!("this function is not needed")
```

litebox_shim_linux/src/syscalls/process.rs
wdcui: I think our current thread local storage is managed by std, so we don't need to call `release_thread_local_storage`. Let me know if you think the `task` struct will not be dropped automatically.

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -266,15 +266,15 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
         envp
     };
 
-    load_program(prog_path, argv, envp).expect("failed to load program");
-    unreachable!("should have jumped to the program");
+    load_program(prog_path, argv, envp);
+    Ok(())
```
jaybosamiya-ms, line 258: Is this actually reachable now?  When is it reached?

litebox_shim_linux/src/syscalls/process.rs
wdcui: I don't understand your question here. I thought I didn't change the semantics of the original code in the new code.

litebox_shim_linux/src/lib.rs
```diff
@@ -389,8 +389,14 @@ fn pread_with_user_buf(
...
-        SyscallRequest::Exit { status } => syscalls::process::sys_exit(status),
-        SyscallRequest::ExitGroup { status } => syscalls::process::sys_exit_group(status),
+        SyscallRequest::Exit { status } => {
+            syscalls::process::sys_exit(status);
+            Ok(0)
+        }
+        SyscallRequest::ExitGroup { status } => {
+            syscalls::process::sys_exit_group(status);
+            Ok(0)
+        }
```
jaybosamiya-ms: Are the `Ok(0)` even reachable here?  What are the semantics here?

litebox/src/platform/mod.rs
wdcui: Good question! I think it can be done in the shim. I need to think about it to see which one makes more sense. I picked the platform because it has the ThreadProvider.

litebox_shim_linux/src/syscalls/process.rs
wdcui: I like this suggestion. Let me see if I can make it work.

litebox_platform_linux_userland/src/lib.rs
wdcui: Good suggestion! Let me improve the code based on it.

litebox_common_linux/src/lib.rs
wdcui: I don't know what I'm doing here; just tried to make it compile. :-). Please suggest changes. Thanks!

litebox_common_linux/src/lib.rs
wdcui: Inconsistent comment. Will fix it.

litebox/src/platform/mod.rs
wdcui: I don't remember why I did it. Let me try to recall.

litebox_shim_linux/src/syscalls/process.rs
CvvT: If it runs in kernel mode, there is no std. We may need to call `release_thread_local_storage` by ourselves

litebox_shim_linux/src/syscalls/process.rs
CvvT: Some subtle difference I am not sure if it matters. The previous version set child's tid in child's thread context. Now it the parent thread that set child's tid.

litebox/src/platform/mod.rs
jstarks: An advantage of having the shim fully manage the pid/tid space is that there are shim-specific semantics on top of this, right? Like pid 1 has a specific meaning, which only Linux cares about. And pids in Windows (hypothetical Windows shim) are multiples of 4.

litebox_platform_linux_userland/src/lib.rs
CvvT: You meant we don't need `terminate_thread`? `exit_group` supposes to terminate all threads. If we just return in `sys_exit_group`, std may only call `exit` to terminate itself eventually.

litebox_common_linux/src/lib.rs
jaybosamiya-ms: The code seems fine, you can just remove the comment :)

litebox_platform_linux_userland/src/lib.rs
jstarks: As I mentioned elsewhere, I think in all cases exit should result in a special return value from `handle_syscall_request`. Requiring the platform to actually terminate the backing thread or process (whether it be the initial thread or one of the spawned threads) is quite limiting. For process exit, what if the process wants to restart a new instance of the shim or go on to do something else? For thread exit, it's worse--there's no reliable, leak-free way to exit a system-created thread in most platforms without returning from the entry point.

So IMO the `ThreadProvider::terminate_thread` and `ExitProvider::exit` methods should be removed.

litebox_common_linux/src/lib.rs
jstarks: I suspect the real problem is that `Platform::RawMutPointer` is not `Send`, so you have to have this kind-of-risky explicit `Send` impl here.

This gets into what I consider a bug in Rust, that `*const T` is not `Send`. There is absolutely no soundness issue with making pointers `Send`, since all interesting options on pointers are already unsafe. The language designers just chose to make pointers `!Send` to encourage developers to _think_ about the semantics of the pointer they're wrapping in some more complex object. I think this is an annoying mistake, but one that's too late to fix even if the Rust maintainers agreed with me.

Anyway, I think the better fix here is to add `Send` and `Sync` impls for the user pointer types `UserMutPtr`, `TransparentMutPtr`, etc., plus a `Send` bound on `Platform::RawMutPointer`, etc. Then you won't need the harder-to-reason-about impl here.

litebox/src/platform/mod.rs
wdcui: Good point. Let me see how I can move it to the shim.

litebox_common_linux/src/lib.rs
wdcui: I was just trying to make the struct a multiple of usize. Based on John's suggestion, we won't need it anymore if I change the return value of `syscall_handler`.

litebox_platform_linux_userland/src/lib.rs
wdcui: No code calls these two functions right now. When `syscall_callback` returns from `sys_exit` or `sys_exit_group`, it will return to the caller of `thread_start_asm`.

I will see if I can remove `ThreadProvider::terminate_thread` and `ExitProvider::exit` as John suggested above.

litebox_platform_linux_userland/src/lib.rs
wdcui: See the discussion above.

litebox_runner_linux_userland/src/lib.rs
wdcui: When `syscall_callback` returns from `sys_exit` or `sys_exit_group`, it will return to the caller of `thread_start_asm`, which is `load_program`.

litebox_shim_linux/src/lib.rs
wdcui: `sys_exit_group` sets `to_terminate` to 1 and returns to `Ok(0)`.

litebox/src/platform/mod.rs
wdcui: @jaybosamiya-ms @CvvT do you have any suggestions where we can implement `next_thread_id` in the Linux shim? It's not clear to me where we should store the `thread_id`.

wdcui: Its implementation uses an AtomicI32. It increases by 1 every time it is invoked.

CvvT: Add a global variable in https://github.com/microsoft/litebox/blob/main/litebox_shim_linux/src/syscalls/process.rs? Or since there is only one process, add it to 
https://github.com/microsoft/litebox/blob/099330ae72126e35f26ec633e4fca1da14e57ba4/litebox_shim_linux/src/syscalls/process.rs#L29

litebox_shim_linux/src/syscalls/process.rs
wdcui: Per suggestions from @jaybosamiya-ms , I will duplicate this enum definition in all the platforms. We will work out a better interface for syscall handling in a future PR.

@jstarks and @CvvT , please let me know if you have any suggestions about this.

wdcui: Actually, I will put them in `litebox_common_linux` and `litebox_common_optee`.

wdcui: I coded this up and pushed my changes. Please take a look. Thanks!

litebox_platform_linux_userland/src/lib.rs
wdcui: It's done this way now.

litebox_platform_linux_userland/src/lib.rs
wdcui: I changed the code to use `rip` and `rsp` in the context to save the entry point and the stack top. Please take a look. Thanks.

litebox_runner_linux_userland/src/lib.rs
wdcui: I changed the code to set `rip` and `rsp`.

litebox_platform_linux_userland/src/lib.rs
wdcui: In `thread_start_internal`, I still use a couple of other registers to store the entry point and the stack top. I don't see a good way to avoid that.

litebox_shim_linux/src/syscalls/process.rs
wdcui: I think we did that before because we didn't know the child thread id at this point. Now we do. I think we can set it here. What do you think?

litebox/src/platform/mod.rs
wdcui: Fixed. Please take a look.

litebox_common_linux/src/lib.rs
wdcui: This one is because we cannot move a copy of NewThreadArgs.

wdcui: Resolved: changed to move the instance.

litebox_platform_linux_userland/src/lib.rs
wdcui: We decided to leave a TODO: we can modify syscall_callback to return if we need to exit the process. If so, we can exit the process inside `thread_start`.

litebox_platform_linux_userland/src/lib.rs
wdcui: Left a TODO. The code can be optimized to avoid overwriting r10 and r11.

litebox_shim_linux/src/syscalls/process.rs
wdcui: Fixed by calling `release_thread_local_storage` in `sys_exit`. For `sys_exit_group`, we left a TODO.

litebox/src/platform/mod.rs
wdcui: This was moved to the shim.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
wdcui: Fixed.

litebox_runner_linux_userland/src/lib.rs
wdcui: Left a TODO to return information from the guest to this place.

litebox_platform_linux_userland/src/lib.rs
wdcui: Removed.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
wdcui: Changed it to Option<usize>.

litebox_common_linux/src/lib.rs
wdcui: Removed the comment.

litebox/src/platform/mod.rs
wdcui: We access thread local storage inside the syscall callback. It's not an easy fix to get a reference to the platform in that code. I added an issue #410 to track this.

litebox_platform_linux_userland/src/lib.rs
wdcui: Tracking this with #411.

litebox_platform_linux_userland/src/lib.rs
wdcui: Tracking this with #412

jstarks:
Do the FreeBSD and Linux SNP kernel platforms handle exit/exit_group properly now? It looks to me like they don't self-terminate like they used to.

Are there instructions for testing these obscure platforms?

I wonder if we should kill the FreeBSD userland platform for now and just plan to merge it into the Linux userland platform, once we move away from the magic syscall mechanism for Linux. At that point, the platform code should mostly just work in generic POSIX environments, and we shouldn't need OS-specific ones.

If that's not acceptable, then I think FreeBSD should be updated to use the same separate stack model as the Linux platform, rather than self-termination.

For SNP, I guess there's already a separate kernel stack, by necessity. But it doesn't yet use your clever approach of returning back to the thread entry point at shim thread exit. I think it should (which probably means updating the `vmpl2_kernel_sp` field of `vsbox_task` as part of thread launch, so that the syscall handler always runs on top of the thread start function).

I have an idea on how to clean up TLS and thread management, improving safety and performance, but it depends on all the platforms use this thread unwind model that you've built for Linux and Windows userland.

CvvT:
> For SNP, I guess there's already a separate kernel stack, by necessity. But it doesn't yet use your clever approach of returning back to the thread entry point at shim thread exit. I think it should (which probably means updating the `vmpl2_kernel_sp` field of `vsbox_task` as part of thread launch, so that the syscall handler always runs on top of the thread start function).

I can take care of that in a separate PR. It's non-trivial to test SNP.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -299,108 +299,310 @@ impl LinuxUserland {
...
+    /// This function is the entry point for starting a guest thread. It performs the following:
+    ///
+    /// 1. Saves all general-purpose registers, flags, and other CPU state onto the stack
+    ///    in a layout that matches the `PtRegs` structure.
+    /// 2. Captures the current stack pointer (RSP/ESP) and frame pointer (RBP/EBP).
+    /// 3. Calls `thread_start_internal` with:
+    ///    - `ctx`: A reference to the provided `PtRegs` structure (passed in RDI/stack)
+    ///    - `frame_pointer`: The captured frame pointer value
+    /// 4. After `thread_start_internal` returns, restores all saved registers and returns
+    ///    to the caller.
```
CvvT: Remove this comment?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -1250,151 +1227,6 @@ mod tests {
...
-                limit: u32::try_from(core::mem::size_of::<
-                    litebox_common_linux::ThreadLocalStorage<litebox_platform_multiplex::Platform>,
-                >())
-                .unwrap()
-                    - 1,
-                flags,
-            }
-        };
-
-        make_aligned_entry!(main_wrapper, new_thread_main_test);
```
CvvT, line 1359: If you remove this test, `new_thread_main_test` and `make_aligned_entry` can also be removed

litebox_shim_linux/src/lib.rs
```diff
@@ -862,11 +887,6 @@ pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> usize {
             }
         }
         SyscallRequest::Futex { args } => syscalls::process::sys_futex(args),
-        SyscallRequest::Execve {
-            pathname,
-            argv,
-            envp,
-        } => syscalls::process::sys_execve(pathname, argv, envp).map(|()| unreachable!()),
```
CvvT, line 869: Why this is removed?

litebox_platform_linux_userland/src/lib.rs
wdcui: Done. Thanks!

litebox_platform_freebsd_userland/src/lib.rs
wdcui: This is tracked by #410 .

litebox_shim_linux/src/lib.rs
wdcui: It's moved to the top of this function.


```suggestion
pub fn handle_syscall_request(request: SyscallRequest<Platform>) -> ContinueOperation {
    if let SyscallRequest::Exit { status } = request {
        syscalls::process::sys_exit(status);
        return ContinueOperation::ExitThread(status);
    } else if let SyscallRequest::ExitGroup { status } = request {
        syscalls::process::sys_exit_group(status);
        return ContinueOperation::ExitProcess(status);
    } else if let SyscallRequest::Execve {
        pathname,
        argv,
        envp,
    } = request
    {
        match syscalls::process::sys_execve(pathname, argv, envp) {
            Ok(()) => return ContinueOperation::ExitProcess(0),
            Err(err) => {
                let e: i32 = err.as_neg();
                let Ok(e) = isize::try_from(e) else {
                    // On both 32-bit and 64-bit, this should never be triggered
                    unreachable!()
                };
                return ContinueOperation::ResumeGuest {
                    return_value: e.reinterpret_as_unsigned(),
                };
            }
        }
    }
    ```

litebox_shim_linux/src/syscalls/process.rs
CvvT: Tracked by #414

litebox_common_linux/src/lib.rs
jaybosamiya-ms: Just as a quick response to John's above comment: I mentioned this to Weidong IRL, but I'd rather keep things for the raw pointers closer to Rust's design (even though I largely agree with John above on it being a design bug in Rust for pointers to be `!Send`), just to keep things less "surprising" by keeping them closer to Rust.  We can revisit this in the future if it turns out to be an issue too often, but I suspect it rarely will be.  Thus, reasoning about a single impl here is good enough to me.

PR #407, opened 2025-10-04
Switch `devices::stdio` to `devices` + add /dev/null

CvvT:
1. Refactor the stdio device fs to incorporate /dev/null device
2. Add a helper function in `litebox_shim_linux` to simplify the construction of fs

### DIFF
--- a/litebox/src/fs/devices.rs
+++ b/litebox/src/fs/devices.rs
@@ -1,4 +1,4 @@
-//! Device provider for LiteBox inclduding:
+//! Device provider for LiteBox including:
 //! 1. Standard input/output devices.
 //! 2. /dev/null device.
 


CvvT:
> I'm generally ok with this PR, with a few thoughts that I'd like to hear from you @CvvT before we consider merging this in:
> 
> * I agree that there is quite some duplication.  Do we have any reason to believe someone might want the null device but not want stdio or vice-versa?  I am struggling to see a reason for this.  Back when I'd initially set up `devices::stdio` as a separate thing, it was partly because I was unsure of what sorts of devices we might need to provide support for.  However, I think combining things into just a single `devices` modules might be the right move?
> * If we decide _not_ to combine things, then naming it `devices::null` prevents us from adding full and zero into the same one; we probably could just name it `devices::trivial_sources_and_sinks` instead, and put in `/dev/full` and `/dev/zero` too.
> 
> Personally, I think we could just combine stdio with these (and if we want/need it in the future, we could potentially support some builder like `devices::FileSystem::with_stdio(false).with_dev_null(true).new()` if some future shim wants exactly some subset). Thoughts?

Both stdio and /dev/null are commonly used; it should be fine to bundle them together. In that case, we may need to change the name of this file from `stdio.rs` to something else. In the future, we may also put in `/dev/full` and `/dev/zero` too as you mentioned.

jaybosamiya-ms:
Seems like we're in agreement then.  We should just have it be part of `devices` directly (effectively it would be `litebox::fs::devices::FileSystem`).  Essentially, replace the entire `litebox/src/fs/devices/` directory with a single `litebox/src/devices.rs` and not have any submodules there.

Thanks for making the updates!

CvvT:
> Seems like we're in agreement then. We should just have it be part of `devices` directly (effectively it would be `litebox::fs::devices::FileSystem`). Essentially, replace the entire `litebox/src/fs/devices/` directory with a single `litebox/src/devices.rs` and not have any submodules there.
> 
> Thanks for making the updates!

Updated per our discussion, thanks!

PR #408, opened 2025-10-06
Clear out manually-handled `File` descriptors from shim

jaybosamiya-ms:
This PR clears out the manually handled `File` Descriptors from the shim.  It also updates the stdio initialization to do it eagerly rather than on-demand.  The moment we have an FS, we can initialize FDs 0, 1, and 2.

---

 (connected to #31).


jaybosamiya-ms:
There were some conflicts with recently merged PRs which required some updates, I have rebased up to the latest `main` again, fixed things, and marked as read-to-review again.

PR #413, opened 2025-10-13
Use jmp instead of call in syscall rewriter

jstarks:
On x86_64, store the return address in rcx instead of on the stack when branching to the syscall callback. This is necessary because the stack pointer may be invalid at the time of a syscall, such as during thread teardown in binaries using musl libc.

For now, just fix all the syscall callbacks to handle this new calling convention by pushing rcx to the stack on entry. Later changes will update platforms to take care to never touch the guest stack in the syscall path.


jstarks:
> One question: do we need to do the same thing for x86?

Probably we do. musl has the same behavior on x86, afaict. This seems trickier on x86, though, because you can't get eip without using the stack. So, the trampoline will need to compute it from some base address that has been relocated.

PR #416, opened 2025-10-13
Separated raw descriptor store + better `close`

jaybosamiya-ms:
This PR does a bunch of changes:

1. Separated raw descriptor store: this allows us to (eventually) have task/process-specific FD storage. This does require the shim to store the raw->typed descriptor table, which technically (as it currently stands) leads to an additional descriptor table at the shim, but the plan is that the old descriptor table will go away.
2. Removes the usage of `Weak` and instead just gives strong `Arc`s.  This is more of a drastic change, since it now has the potential downside of something within LiteBox holding on to an `Arc` for "too long" and leading (essentially) to a resource leak. However, I suspect we'll be able to set up some watchers if we notice something like this happen, and clean that up if needed. However, this simplifies some reasoning regarding what happens around closed raw FDs (see next point).
3. Closing of raw FDs (i.e., `fd_consume_raw_integer`) no longer can return a `CurrentlyUnconsumable` but instead always returns an `Arc`. This means that we no longer need the fueled-loop and cannot get blocked by other operations. This does need some changes to LiteBox itself (see next point) but basically, this allows us to emulate Linux semantics better.
4. Updates the `close` operations in FS/Net in LiteBox to not be FD-consuming but instead to take a ref. This does have the downside of us no longer being able to use the FD as a liveness-token, which means that all operations can now return a "operating on closed FD" error, but has the upside of allowing for an unblocked `close` operation. For the FS, this is basically largely straightforward, since the close operation doesn't _really_ need to do much cleanup. However, for the Net subsystem, we need more work (see next point).
5. Updates the Net subsystem to support a queued-close, to account for ongoing operations.  There are fast-paths to close out early if we can, but if something cannot be closed and cleaned up immediately, then it gets queued into a system that attempts a close on every network-platform interaction (again with fast-paths to prevent much slowdown) but this effectively gives us the GC-like behavior we'd like.

---

To review the PR, it is likely easiest to read each commit on its own.

---

### DIFF
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -1056,8 +1056,10 @@ where
             return Err(ReceiveError::InvalidFd);
         }
 
-        if flags.intersects(ReceiveFlags::DONTWAIT.complement()) {
-            unimplemented!()
+        if flags.intersects(
+            (ReceiveFlags::DONTWAIT | ReceiveFlags::TRUNC | ReceiveFlags::DISCARD).complement(),
+        ) {
+            unimplemented!("flags: {:?}", flags);
         }
 
         let ret = match socket_handle.protocol() {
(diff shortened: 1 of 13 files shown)


litebox_shim_linux/src/lib.rs
```diff
@@ -277,6 +278,20 @@ enum Descriptor {
     },
 }
 
+// XXX: This should likely be made task-specific, since each task has its own list of FDs
```
jstarks, line 330: #421 should make this easier.

jaybosamiya-ms:
Blocked on #425 (I just changed the base to reflect this and make it easier to read, but yeah this PR is ready to review even though it is not marked as `main`-based)

litebox/src/net/mod.rs
```diff
@@ -558,13 +567,51 @@ where
...
+        if let Some(socket_handle) = dt.remove(fd) {
+            // Can immediately close it out.
+            drop(dt);
+            self.close_handle(socket_handle.entry);
+            Ok(())
+        } else {
+            // It seems like there might be other duplicates around (e.g., due to `dup`), so we
+            // can't immediately close it out (or this FD has already been closed out). We
+            // attempt to queue it for future closure and then just return.
+            let mut dt = self.litebox.descriptor_table_mut();
```
wdcui: Can we use the `dt` from line 572?

litebox/src/net/mod.rs
```diff
@@ -558,13 +567,51 @@ where
...
+            return false;
+        }
+        let mut dt = self.litebox.descriptor_table_mut();
+        let entries = dt.drain_entries_full_covered_by(&mut self.queued_for_closure);
+        drop(dt);
+        if entries.is_empty() {
+            return false;
+        }
+        for entry in entries {
+            self.close_handle(entry.entry);
```
wdcui, line 606: Why do we know it's okay to close it now here?

litebox_shim_linux/src/lib.rs
```diff
@@ -276,6 +277,20 @@ enum Descriptor {
...
```
wdcui, line 330: `task` here is a process or a thread?

litebox/src/fd/mod.rs
```diff
@@ -530,14 +444,165 @@ impl<Platform: RawSyncPrimitivesProvider> Descriptors<Platform> {
...
+    #[expect(
+        clippy::missing_panics_doc,
+        reason = "not guaranteed as an API-level guarantee, but instead as a defensive panic to re-consider implementation if we hit it"
+    )]
+    pub fn fd_into_specific_raw_integer<Subsystem: FdEnabledSubsystem>(
+        &mut self,
+        fd: TypedFd<Subsystem>,
+        raw_fd: usize,
+    ) -> bool {
+        // TODO(jayb): Should we be storing things via a HashMap to make sure this operation cannot
```
wdcui, line 529: I think we should use a HashMap; maybe add an issue to track this.

jaybosamiya-ms: This TODO existed from before (this function just moved from earlier in the file to later).  It was introduced in #373.

I am also _fairly_ sure we should use a HashMap here but changing that is out of scope of this PR and should be separate.  The assertion that I have here is to make it loud enough if we see it in practice, so I don't think we need a separate issue to track it.

litebox/src/net/mod.rs
jaybosamiya-ms: Good catch!   Updated.

litebox/src/net/mod.rs
jaybosamiya-ms: The guarantee of the `drain_entries_full_covered_by` introduced in this PR (see `fd/mod.rs` for the docs) is that it will only return entries which are fully covered by the FDs it was passed.  It consumes those FDs, and then returns a fully owned entry (not just a reference to it).  There clearly (by type guarantees) cannot be any more references to the entry anymore.  This ownership is _also_ confirmed by `close_handle` that consumes the `SocketHandle` rather than taking a reference to it.

litebox_shim_linux/src/lib.rs
jaybosamiya-ms: It will likely be an `Arc` that is shared across threads within a process but separate across different processes.  Since we don't have multi-process support yet, I'm just marking it as a concern to be handled when we do that.

litebox/src/fd/mod.rs
jstarks: Linux and probably all unices use a flat, resizable array, and they limit its size via `RLIMIT_NOFILE`. It's considered a misconfiguration if an application can cause problems by setting a huge fd. The corollary is that if `RLIMIT_NOFILE` is large, an application is entitled to use a large fd without failure (assuming there's enough memory available), even if it's wasteful.

This approach simple and fast. So, unless there's some other consideration, I suggest sticking with a flat vector, ensuring that RLIMIT_NOFILE is respected, and removing the assertion.

jaybosamiya-ms: Thanks John, that's a valid point, I'm going to leave the assertion and TODO here for now, and then in a separate PR update it to have support for something like `RLIMIT_NOFILE`.  We already have some stuff for `EMFILE` at the shim level, but as part of the migration, we'll need to have something of that form with this particular type.

litebox/src/net/mod.rs
wdcui: Got it. Thanks.

litebox/src/fd/mod.rs
wdcui: I will resolve this conversation so Jay can merge this PR.

litebox_shim_linux/src/lib.rs
wdcui: I see, so we can leverage the TLS primitives added by John for this (as he suggested).

PR #420, opened 2025-10-15
Push syscall parsing and handling into the shims

jstarks:
Call into the shims with a mutable thread register context and let the shims determine what to do to parse syscalls and update the context after the syscall completes. This centralizes the syscall parsing logic and makes it easier for OPTEE and Linux to have different behaviors and ABIs.

### DIFF
--- a/litebox_runner_snp/src/main.rs
+++ b/litebox_runner_snp/src/main.rs
@@ -169,16 +169,13 @@ pub extern "C" fn sandbox_task_exit() {
 }
 
 #[unsafe(no_mangle)]
-pub extern "C" fn do_syscall_64(nr: u64, pt_regs: &mut litebox_common_linux::PtRegs) {
-    pt_regs.rax = match litebox_common_linux::SyscallRequest::try_from_raw(nr.truncate(), pt_regs) {
-        Ok(req) => match litebox_shim_linux::handle_syscall_request(req) {
-            ContinueOperation::ResumeGuest { return_value } => return_value,
-            ContinueOperation::ExitThread(status) | ContinueOperation::ExitProcess(status) => {
-                status.cast_unsigned() as usize
-            }
-        },
-        Err(err) => (err.as_neg() as isize).reinterpret_as_unsigned(),
-    };
+pub extern "C" fn do_syscall_64(pt_regs: &mut litebox_common_linux::PtRegs) {
+    match litebox_shim_linux::handle_syscall_request(pt_regs) {
+        ContinueOperation::ResumeGuest => {}
+        ContinueOperation::ExitThread(status) | ContinueOperation::ExitProcess(status) => {
+            pt_regs.rax = status.cast_unsigned() as usize;
+        }
+    }
 }
 
 /// This function is called on panic.
--- a/litebox_runner_snp/src/entry.S
+++ b/litebox_runner_snp/src/entry.S
@@ -89,8 +89,7 @@ entry_SYSCALL_64:
     push    r14         /* pt_regs->r14 */
     push    r15         /* pt_regs->r15 */
 
-    mov     rdi, rax
-    mov     rsi, rsp
+    mov     rdi, rsp
     call    do_syscall_64
 
 swapgs_restore_regs_and_return_to_usermode:


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1652,64 +1652,31 @@ unsafe extern "C" {
 #[allow(clippy::cast_sign_loss)]
 #[unsafe(no_mangle)]
 unsafe extern "C" fn syscall_handler(
-    syscall_number: usize,
-    ctx: *mut litebox_common_linux::PtRegs,
+    _syscall_number: usize,
```
jstarks, line 1655: Of course I could remove this parameter by changing the asm further, but I have some other asm refactoring coming, so I chose to just leave it alone.

jstarks: (Ditto elsewhere.)

litebox_runner_snp/src/main.rs
```diff
@@ -169,16 +169,13 @@ pub extern "C" fn sandbox_task_exit() {
...
-            ContinueOperation::ExitThread(status) | ContinueOperation::ExitProcess(status) => {
-                status.cast_unsigned() as usize
-            }
-        },
-        Err(err) => (err.as_neg() as isize).reinterpret_as_unsigned(),
-    };
+pub extern "C" fn do_syscall_64(pt_regs: &mut litebox_common_linux::PtRegs) {
+    match litebox_shim_linux::handle_syscall_request(pt_regs) {
+        ContinueOperation::ResumeGuest => {}
+        ContinueOperation::ExitThread(status) | ContinueOperation::ExitProcess(status) => {
```
wdcui, line 175: Will the platform or the runner need to do anything for thread/process exit?

litebox_shim_linux/src/lib.rs
```diff
@@ -384,35 +384,47 @@ fn pread_with_user_buf(
...
-                let e: i32 = err.as_neg();
-                let Ok(e) = isize::try_from(e) else {
-                    // On both 32-bit and 64-bit, this should never be triggered
-                    unreachable!()
-                };
-                return ContinueOperation::ResumeGuest {
-                    return_value: e.reinterpret_as_unsigned(),
-                };
-            }
+pub fn handle_syscall_request(ctx: &mut litebox_common_linux::PtRegs) -> ContinueOperation {
```
wdcui, line 387: Do you anticipate that we just define `PtRegs` for ARM64 when we add support for it?

jstarks: Yes, I think so. Although probably we don't need to stick with `PtRegs`--we should probably define a generic (but arch-specific) litebox context structure. There's no need to exactly match what Linux defined.

litebox_runner_snp/src/main.rs
jstarks: Maybe. I think the exit path needs to be refactored a bit more. It's a little tricky because the platform needs to know when _all_ threads have terminated so that it can exit the process... and this might happen _after_ the initial thread returns.

Probably `ExitProcess` should go away, and the platform/runner should be responsible for this refcounting and waiting. Then the shim can expose a function to get the exit status of the process. So we'll be left with `ResumeGuest` and `ExitThread`, both with no fields.

litebox_shim_linux/src/lib.rs
wdcui: I agree. We probably should use a different name since PtRegs implies that it matches with Linux.

litebox_runner_snp/src/main.rs
wdcui: I remember there is some subtle difference between `sys_exit_group` and `sys_exit`. After `sys_exit_group` is called, LiteBox should just terminate all the threads without worrying about the ref counting. Is this consistent with your understanding? If so, we will still need to pass `ExitPrcess` to the the platform/runner.

jstarks: `sys_exit_group` means that all the threads in the user process should be killed and the current thread should exit. `sys_exit` just exits the current thread.

We _could_ just let the runner/platform exit the process immediately on `sys_exit_group`, like we did before. This works for userland platforms because the kernel will clean everything up when the host process exits. But if the platform is something like coconut and the shim is running in kernel mode, then the shim needs to clean up all the threads explicitly.

So, I think it's probably better to just model the full clean up path, for the fully general case. This is slightly wasteful for the userland case, so we could later consider adding a fast path for runners that want it.

PR #421, opened 2025-10-15
Make shim TLS type opaque to platform

### LINKED ISSUE #410: Better handling of TLS
There are two issues with TLS right now.

1. The shim and the platform share one struct inside TLS.
2. TLS is accessed as a global (no reference to the platform).

jstarks:
Simplify the platform TLS trait to just operate on a single pointer, moving the complexity of managing this pointer into litebox.

This simplifies the code the platform needs to implement. But more importantly, this allows the shim to define its TLS type locally, without making it visible outside the crate. This will be important as we expand the complexity of the per-task state in the Linux shim.

Currently, only a single TLS value can be set on a thread at a time. This is sufficient for the needs of the existing shims.

Fixes #410.

### DIFF
--- a/litebox/src/tls.rs
+++ b/litebox/src/tls.rs
@@ -22,8 +22,14 @@ use core::sync::atomic::AtomicU8;
 /// # Example
 ///
 /// ```no_run
+/// # struct MyPlatform;
+/// # unsafe impl litebox::platform::ThreadLocalStorageProvider for MyPlatform {
+/// #     fn get_thread_local_storage() -> *mut () { todo!() }
+/// #     unsafe fn replace_thread_local_storage(value: *mut ()) -> *mut () { todo!() }
+/// #     fn clear_guest_thread_local_storage() { todo!() }
+/// # }
 /// litebox::shim_thread_local! {
-///     #[platform = litebox::platform::mock::MockPlatform]
+///     #[platform = MyPlatform]
 ///     static MY_TLS: core::cell::Cell<u32>;
 /// }
 ///


litebox/src/platform/tls.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// A key used to access thread local storage of type `T` for platform `P`.
+pub struct TlsKey<T, P> {
+    // Used to ensure there's a unique address for this.
+    _dummy: AtomicU8,
+    _phantom: PhantomData<fn(T, &P) -> T>,
+}
+
+#[repr(C)] // needed so that the offset of `key` is predictable across different `T` and `P`.
+struct Tls<T: 'static, P: 'static> {
```
jaybosamiya-ms: Minor: we consistently use `Platform` as the generic name for platform, rather than the single letter `P`.

litebox/src/platform/mod.rs
```diff
@@ -621,35 +622,38 @@ pub trait SystemInfoProvider {
...
-
-    /// Release the thread-local storage value for the current thread
+///
+/// Currently, this provides just a single thread-local storage pointer. Shims
+/// should use [`shim_thread_local!`](crate::shim_thread_local) macro for a safe
+/// and ergonomic interface to TLS.
+///
+/// # Safety
+/// The implementation must ensure that the TLS pointer that is set for the
+/// thread is the one that is returned.
```
jaybosamiya-ms: There is no explicit `set_...` so this text can be a bit confusing.  My understanding is that this is stating that whatever is set by `replace` must always be returned by the `get`?

litebox/src/platform/mod.rs
```diff
@@ -621,35 +622,38 @@ pub trait SystemInfoProvider {
...
+pub unsafe trait ThreadLocalStorageProvider {
+    /// Gets the current thread-local storage pointer.
+    //
+    // DEVNOTE: note that this does not take `&self`. So far, this has not been
+    // a problem for platform implementations, and allowing this does improve
+    // performance by avoiding a platform lookup on every TLS access. But we
+    // could consider changing this in the future if needed.
+    fn get_thread_local_storage() -> *mut ();
```
jaybosamiya-ms: We should probably document the initial value of null, rather than leaving it implicit.  Similarly for the replace.

litebox/src/tls.rs
jaybosamiya-ms, line 1: I think this should be moved to `litebox/src/tls.rs` rather than `litebox/src/platform/tls.rs`, since it is just functionality provided by LiteBox itself, rather than some platform-specific stuff.

litebox/src/platform/tls.rs
```diff
@@ -0,0 +1,242 @@
...
+///
+/// # Example
+///
+/// ```no_run
+/// # struct MyPlatform;
+/// # unsafe impl litebox::platform::ThreadLocalStorageProvider for MyPlatform {
+/// #     fn get_thread_local_storage() -> *mut () { todo!() }
+/// #     unsafe fn replace_thread_local_storage(value: *mut ()) -> *mut () { todo!() }
+/// #     fn clear_guest_thread_local_storage() { todo!() }
+/// # }
```
jaybosamiya-ms: Hmm, apparently, we hadn't added `ThreadLocalStorageProvider` to `litebox::platform::mock::MockPlatform`; this set of `todo!()` initializations probably go there.

litebox/src/platform/tls.rs
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        unsafe fn replace_thread_local_storage(value: *mut ()) -> *mut () {
+            TLS.replace(value)
+        }
+
+        fn clear_guest_thread_local_storage() {
+            unimplemented!()
+        }
+    }
```
jaybosamiya-ms: Yep, this one should just be referencing the mock platform directly :)

litebox/src/tls.rs
```diff
@@ -0,0 +1,242 @@
+//! Thread-local storage support for shims.
+//!
+//! Shims can use [`shim_thread_local!`](crate::shim_thread_local) to define
+//! thread-local static variables.
```
jaybosamiya-ms, line 4: I'm now thinking that we might want to have a `dev_tests/` test that tells people using `thread_local` in any shim to use `shim_thread_local` instead.  Not crucial to do in this PR, just making a note for myself.

litebox_shim_linux/src/lib.rs
```diff
@@ -59,8 +61,25 @@ pub(crate) fn boot_time() -> &'static <Platform as litebox::platform::TimeProvid
         .expect("litebox() should have already been called before this point")
 }
 
+/// Initialize the shim to run a task with the given parameters.
+///
+/// Returns the global litebox object.
+pub fn init_process<'a>(task: litebox_common_linux::Task<Platform>) -> &'a LiteBox<Platform> {
+    // TODO: ensure this gets torn down even if the thread never runs. Consider
```
jaybosamiya-ms: Minor: "never runs." -> "never runs `sys_exit`."?

litebox/src/platform/tls.rs
```diff
@@ -0,0 +1,242 @@
...
+///
+/// MY_TLS.init(42.into());
+/// MY_TLS.with(|tls| {
+///     assert_eq!(tls.get(), 42);
+/// });
+/// let v = MY_TLS.deinit();
+/// ```
+#[macro_export]
+macro_rules! shim_thread_local {
+    (static $name:ident: $ty:ty: $platform:ty;) => {
```
jaybosamiya-ms: Musing: It's mildly awkward to use an `static FOO: Bar: Baz` syntax; we might want to consider slightly different syntax?

Most natural looking would probably be like
```rs
shim_thread_local! {
  #[platform = Baz]
  static FOO: Bar;
}
```
but if we're going the style I've ended up using in other places, it'd look a bit more like
```rs
shim_thread_local! {
  static FOO: Bar @ Baz;
}
```
or
```rs
shim_thread_local! { @ Baz
  static FOO: Bar;
}
```
neither of which are much better than the `a : b : c`.

Anyways, this is just me musing about syntax; no actual changes needed here to merge this in.

litebox_shim_optee/src/lib.rs
```diff
@@ -829,3 +829,11 @@ pub(crate) fn tee_cryp_state_map() -> &'static TeeCrypStateMap {
...
     TEE_CRYPT_STATE_MAP.get_or_init(|| alloc::boxed::Box::new(TeeCrypStateMap::new()))
 }
+
+#[cfg(feature = "platform_linux_userland")]
+pub fn set_session_id(session_id: u32) {
+    SESSION_ID.store(session_id, core::sync::atomic::Ordering::Relaxed);
+}
+
+#[cfg(feature = "platform_linux_userland")]
+static SESSION_ID: core::sync::atomic::AtomicU32 = core::sync::atomic::AtomicU32::new(0);
```
wdcui, line 839: Should `SESSION_ID` be a thread local?

litebox/src/platform/tls.rs
jstarks: Yeah, I didn't really like the syntax I came up with. I like the `#[platform = Baz]` idea, personally.

litebox_shim_optee/src/lib.rs
jstarks: To be honest I don't really understand what this session_id thing is about. It seemed like a temporary thing, so I was just trying to preserve the existing behavior while removing the dependency on the Linux TLS from within the OPTEE shim.

litebox/src/platform/tls.rs
jstarks: This one I fixed.

litebox/src/platform/tls.rs
jstarks: I wasn't able to switch to `MockPlatform` here since that's not `pub` and so is not available to doc tests. For now I've kept the custom impl here... we could make `MockPlatform` `pub` in the future (perhaps with `#[doc(hidden)]` if we want to avoid these types of one-offs in doctests.

PR #423, opened 2025-10-15
Centralize constructing initial context for new threads

jstarks:
The shim should fully construct the thread's initial context (including stack pointer, instruction pointer, and return value) and not leave it up to the individual platforms.

As part of this, fix `ThreadProvider` so that the thread arguments do not leak internal shim details. This is straightforward now that the thread ID is managed by the shim and not the platform.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -327,7 +327,7 @@ impl litebox::platform::InitThread for NewThreadArgs<litebox_platform_multiplex:
             set_child_tid,
         } = *self;
 
-        let new_task = Box::new(litebox_common_linux::Task {
+        let new_task = litebox_common_linux::Task {
             pid: task.pid,
             tid: task.tid,
             ppid: task.ppid,
@@ -335,7 +335,7 @@ impl litebox::platform::InitThread for NewThreadArgs<litebox_platform_multiplex:
             robust_list: task.robust_list,
             credentials: task.credentials.clone(),
             comm: task.comm,
-        });
+        };
         let child_tid = task.tid;
 
         // Set the TLS for the platform itself


litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -289,48 +289,79 @@ pub(crate) fn sys_exit(_status: i32) {
...
+    /// Where to store child TID in child's memory
+    set_child_tid: Option<Platform::RawMutPointer<i32>>,
+    /// Task struct that maintains all per-thread data
+    task: alloc::boxed::Box<litebox_common_linux::Task<Platform>>,
+}
 
-    // Set the TLS for the platform itself
-    let litebox_tls = litebox_common_linux::ThreadLocalStorage::new(new_task);
-    litebox_platform_multiplex::Platform::set_thread_local_storage(litebox_tls);
+// TODO: impl Send for RawMutPointer and family so that this manual impl is not needed.
```
jaybosamiya-ms: Even though I largely agree with you on the Rust design "bug", I'd (at least currently) prefer _not_ implementing `Send` for `RawMutPointer` and family (see https://github.com/microsoft/litebox/pull/404#discussion_r2430580663).  Thus I recommend removing this particular TODO comment from this PR.

We can open an issue to track this if you like.

litebox/src/platform/mod.rs
```diff
@@ -49,33 +49,33 @@ pub trait Provider:
...
-    /// `entry_point` is the address of the first instruction in the guest program that will be executed
-    /// in the new thread.
-    ///
-    /// `thread_args` is the additional platform-specific argument needed for the new thread to start
+    /// `init_thread` provides an object used to initialize the shim on the new thread.
     ///
     /// # Safety
     ///
-    /// The `entry_point` must be a valid entry point for the new thread.
+    /// The context must be valid.
```
jaybosamiya-ms, line 63: We had this issue before the PR too, so does not need to be handled by the PR itself, but just making a note:

I am not particularly jazzed about the fact that we are just depending on an implicit constructability of the ctx object on the shim side.  It makes the "context must be valid" difficult to actually reason about on its own.  

I don't have a concrete suggestion within this PR, but will shortly convert this comment to an issue so that we track it.

litebox/src/platform/mod.rs
```diff
@@ -49,33 +49,33 @@ pub trait Provider:
...
-        entry_point: usize,
-        thread_args: alloc::boxed::Box<Self::ThreadArgs>,
-    ) -> Result<Self::ThreadId, Self::ThreadSpawnError>;
+        init_thread: alloc::boxed::Box<dyn InitThread>,
+    ) -> Result<(), Self::ThreadSpawnError>;
+}
+
+/// An object to initialize a newly spawned platform thread for use with the
+/// shim that spawned it.
+pub trait InitThread: Send {
```
jaybosamiya-ms, line 75: Minor: since this is one of the few traits in the platform module that is expected to be implemented by the shim, we might want to call that out in the doc string?

litebox/src/platform/mod.rs
jaybosamiya-ms: #429 opened. Marking this as resolved since this won't be done in this PR.

litebox_shim_linux/src/syscalls/process.rs
jstarks: Happy to defer to your judgement. But I see two potential issues with the status quo in this PR:

One, when you impl `Send` on an aggregate type like this, it's easy for another developer to come along and add a non-`Send` field without revisiting the `Send` impl. So, I find it safer to impl `Send` on the primitive type and let the compiler infer `Send` on the aggregate. Of course, this could be done with an explicit newtype around the ptrs, but...

Two, if we don't require that the pointer types are mechanically `Send`, we at least need to require that they are _logically_ `Send` in the trait impl. E.g., with this code as is, it would be unsound to `impl RawConstPtr<T> for Rc<T>`. Now, that's already ruled out because of the `Copy` bound on `RawConstPtr`, but it seems very possible to have a `Copy + 'static` type that is also not "logically" `Send`, e.g., something representing an index into some array stored in TLS.

To address this, it seems like we at least need to document this requirement on `RawConstPtr` and friends, and we need to make the trait `unsafe` in order to enforce this requirement soundly.

But if we're going to do that, well, should we just make the raw pointer types `Send`?

Counter argument: no, because in general user pointers _shouldn't_ be sent between threads without additional thinking--what if the target thread is in a different address space (assuming we eventually implement fork for some platforms). So maybe forcing this code to use a newtype (which litebox could provide as a convenience) to document why this is OK would be more clear.

But either way, we need to do something to keep the code sound, I think. So, we should probably open an issue.

jstarks: Opened #431.

Issue #171, opened 2025-06-23
Add tracing support


wdcui:

jaybosamiya-ms:
Brief notes: I tried to get a version of this working, but apparently, something seems to break in particular with the rtld_audit stuff (leading to a segfault due to a null-pointer deref).  I suspect that the issue is not in rtld audit itself, but possibly due to the fs/gs stuff we are having to do (and the fact that setting up a tracing-subscriber by default sets up some thread locals).  Will need to debug further, or might need to implement our own tracing subscriber (which is not ideal, since we are missing out on a lot of nice funcionality that is provided by the community).  Ideally, we figure out where we need to hook things so that we can make sure that the thread locals stay reasonable.

PR #424, opened 2025-10-15
Silently ignore rt_sigprocmask on Windows

jstarks:
Until signal support is further along, ignore this punchthrough syscall on Windows. This is necessary to unblock pthread testing on Windows.

### DIFF
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -745,6 +745,7 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
             }
             PunchthroughSyscall::RtSigprocmask { .. } => {
                 // Ignored for now.
+                eprintln!("WARNING: ignoring RtSigprocmask punchthrough");
                 Ok(0)
             }
             p => {


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -743,10 +743,12 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
                 )?;
                 Ok(0)
             }
-            _ => {
-                unimplemented!(
-                    "PunchthroughToken for WindowsUserland is not fully implemented yet"
-                );
+            PunchthroughSyscall::RtSigprocmask { .. } => {
+                // Ignored for now.
```
jaybosamiya-ms, line 747: Minor comment: we may wish to consider debug logging a warning here.  Not a blocker to merging though.

jstarks:
> Would it make sense to add the multi-thread testd in this PR?

We have a pthread test in the Linux runner, but the Linux tests don't seem to be shared/sharable with the Windows runner. Is that something we should fix more comprehensively?

litebox_platform_windows_userland/src/lib.rs
jstarks: Good idea. What's the recommended way to log from a platform? Just use `eprintln!` since we can use `std` here? I'd use `litebox::log_println!`, but to use that I'd have to plumb the WindowsUserland platform ref through to the punchthrough token, and that's would require using GATs in the `PunchthroughProvider` trait and...

jaybosamiya-ms: I think an `eprintln` is good enough here; it is meant to be temporary anyways, and yes, we already have `std` here.

Some context on `litebox::log_println!`: it is a hack that exists until I finish #171.  That was blocked in the past, but now I think has been unblocked, so I should (in theory) be able to get that working, and then things would be nicer.  Some more higher-priority items to finish first though, but yeah, maybe something I should bring back into my active TODO set.

PR #426, opened 2025-10-15
Handle execve fully within the shim

jstarks:
Instead of calling out to the platform to load and run the next program, handle the execve syscall mostly within the Linux shim (a callback has been preserved to allow the Linux userland shim to update auxv and envp.)

This requires fixing the platforms to properly restore rsp and rip if they are changed in the context structure as part of the syscall. This has been done for the 64-bit userland platforms.

Note that execve still does not work on Windows userland, apparently due to some failure when unmapping memory.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -1005,9 +1005,6 @@ pub(crate) fn sys_execve(
     // Close CLOEXEC descriptors
     crate::file_descriptors().write().close_on_exec();
 
-    // Clear TLS for the new program.
-    clear_tls(ctx);
-
     // unmmap all memory mappings and reset brk
     with_current_task(|task| {
         if let Some(robust_list) = task.robust_list.take() {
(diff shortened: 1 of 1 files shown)


jstarks:
This isn't really correct with the existing userland platform impls, since the context structures are still sitting on the guest stacks. It works, but only because the stack doesn't seem to get unmapped (or maybe it got remapped into the same place?). Holding off on this change until I rewrite the platform code to put the contexts in the right place.

I need to split out the TLS clearing bits of this code, though, to make the platform changes easier. I'll open a separate PR for that.

CvvT:
> This isn't really correct with the existing userland platform impls, since the context structures are still sitting on the guest stacks. It works, but only because the stack doesn't seem to get unmapped (or maybe it got remapped into the same place?). Holding off on this change until I rewrite the platform code to put the contexts in the right place.
> 
> I need to split out the TLS clearing bits of this code, though, to make the platform changes easier. I'll open a separate PR for that.

Haven't reviewed the PR yet, but I do remember I had a hacky way to avoid unmapping stack while still using it (which is not ideal and could be improved once we stop using guest's stack in LiteBox).

https://github.com/microsoft/litebox/blob/bf472de4cdfa74edd3119c978b270de11e19dcce/litebox_shim_linux/src/syscalls/process.rs#L1011-L1042

PR #427, opened 2025-10-15
Support MSG_TRUNC for tcp and udp

CvvT:
This PR adds support for `MSG_TRUNC`. `MSG_TRUNC` behaves differently for TCP vs datagram (UDP) sockets.

[TCP](https://www.man7.org/linux/man-pages/man7/tcp.7.html):
    Since Linux 2.4, Linux supports the use of MSG_TRUNC in the flags
       argument of [recv(2)](https://www.man7.org/linux/man-pages/man2/recv.2.html) (and [recvmsg(2)](https://www.man7.org/linux/man-pages/man2/recvmsg.2.html)).  This flag causes the
       received bytes of data to be discarded, rather than passed back in
       a caller-supplied buffer.

[UDP](https://www.man7.org/linux/man-pages/man2/recv.2.html):
    For raw (AF_PACKET), Internet datagram (since Linux
    2.4.27/2.6.8), netlink (since Linux 2.6.22), and UNIX
    datagram as well as sequenced-packet (since Linux 3.4)
    sockets: return the real size of the packet or datagram,
    even when it was longer than the passed buffer.

### DIFF
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -978,8 +978,9 @@ where
         let mut table_entry = descriptor_table.get_entry_mut(fd);
         let socket_handle = &mut table_entry.entry;
 
-        let flags = flags & ReceiveFlags::ALL; // ignore irrelevant flags (e.g., NOSIGNAL)
-        if flags.intersects((ReceiveFlags::DONTWAIT | ReceiveFlags::TRUNC).complement()) {
+        if flags.intersects(
+            (ReceiveFlags::DONTWAIT | ReceiveFlags::TRUNC | ReceiveFlags::DISCARD).complement(),
+        ) {
             unimplemented!("flags: {:?}", flags);
         }
 
@@ -991,7 +992,9 @@ where
                 }
                 let tcp_socket = self.socket_set.get_mut::<tcp::Socket>(socket_handle.handle);
                 if flags.contains(ReceiveFlags::TRUNC) {
-                    // This flag causes the received bytes of data to be discarded
+                    unimplemented!("TRUNC flag for tcp");
+                }
+                if flags.contains(ReceiveFlags::DISCARD) {
                     let discard_slice =
                         |tcp_socket: &mut tcp::Socket<'_>| -> Result<usize, tcp::RecvError> {
                             // See [`tcp::Socket::recv_slice`] and [`tcp::Socket::recv`] for why we do two `recv` calls.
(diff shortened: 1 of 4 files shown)


jaybosamiya-ms:
Top-level comment: this is another one of those Linux-isms that I think we should not be baking into LiteBox itself.  We should instead get rid of `TRUNC` from `ReceiveFlags` and have separate flags for "discard data" and "truncate data but return real size".  I'll add some review comments shortly.

litebox/src/net/mod.rs
```diff
@@ -1124,6 +1147,14 @@ bitflags! {
         const WAITALL = 0x100;
         /// <https://docs.rs/bitflags/*/bitflags/#externally-defined-flags>
         const _ = !0;
+
+        const ALL = Self::CMSG_CLOEXEC.bits()
```
jaybosamiya-ms: Rather than calling this `ALL` (which is in conflict with the above `externally-defined-flags`), we could call it something like `HANDLED_BY_LITEBOX` or similar?

litebox/src/net/mod.rs
```diff
@@ -988,34 +989,56 @@ where
...
+                let tcp_socket = self.socket_set.get_mut::<tcp::Socket>(socket_handle.handle);
+                if flags.contains(ReceiveFlags::TRUNC) {
+                    // This flag causes the received bytes of data to be discarded
+                    let discard_slice =
+                        |tcp_socket: &mut tcp::Socket<'_>| -> Result<usize, tcp::RecvError> {
+                            // See [`tcp::Socket::recv_slice`] and [`tcp::Socket::recv`] for why we do two `recv` calls.
+                            // Basically, the socket buffer is implemented as a ring buffer, and if the data to be read
+                            // wraps around, a single `recv` call will not be able to read all the data.
+                            let size1 = tcp_socket.recv(|data| (data.len(), data.len()))?;
+                            let size2 = tcp_socket.recv(|data| (data.len(), data.len()))?;
```
jaybosamiya-ms, line 1004: What happens if it is _not_ wrapped around?  Don't you need to check `recv_queue`, or are you sure that the second one is guaranteed to return `Ok(0)` if there is no wrap-around?

CvvT:
> Top-level comment: this is another one of those Linux-isms that I think we should not be baking into LiteBox itself. We should instead get rid of `TRUNC` from `ReceiveFlags` and have separate flags for "discard data" and "truncate data but return real size". I'll add some review comments shortly.

Good point! The shim layer and LiteBox should have different flags. I made a copy of `ReceiveFlags` and `SendFlags` to `litebox_common_linux` and modified LiteBox's `ReceiveFlags` to have flag `DISCARD`.

litebox/src/net/mod.rs
CvvT: I made a copy of `ReceiveFlags` and `SendFlags` to `litebox_common_linux` so we don't need this anymore.

litebox/src/net/mod.rs
CvvT: My understanding is calling `recv` twice is equivalent to calling `recv_slice` once (except you can handle the data on your own). `recv_slice` calls `dequeue_slice` which in turn calls `dequeue_many_with` twice (see https://github.com/smoltcp-rs/smoltcp/blob/39cd44e585aae109f37a2bc9f31376414968456d/src/storage/ring_buffer.rs#L273), while `recv` only call `dequeue_many_with` once (see https://github.com/smoltcp-rs/smoltcp/blob/39cd44e585aae109f37a2bc9f31376414968456d/src/socket/tcp.rs#L1339).

jaybosamiya-ms: Ah, makes sense, thanks for double-checking!

PR #432, opened 2025-10-16
Store context in platform memory, not on guest stack

jstarks:
Instead of pushing the guest context onto the guest stack, store it in pre-allocated context memory, on the platform stack. This fixes issues with musl on x86_64 where sometimes it issues syscalls without a valid stack pointer.

This change also ensures that the guest context is accurately restored in all cases that are possible. One exception is Windows, where we return to the guest with rcx = rip; this matches the semantics of sysret, so that's OK for returning from syscalls, but it doesn't work for signal return. In that case, we use `NtContinue` to atomically set the full context.

Another benefit is that the stack can now be unwound during a syscall, so you can see the full stack if a syscall handler panics. This is enabled by including unwind metadata in the asm routines, and by merging the thread entry point with the syscall entry point, so that the unwinder can see that they share a stack frame.

### DIFF
--- a/dev_tests/src/ratchet.rs
+++ b/dev_tests/src/ratchet.rs
@@ -30,7 +30,7 @@ fn ratchet_globals() -> Result<()> {
             ("litebox_platform_linux_userland/", 3),
             ("litebox_platform_lvbs/", 17),
             ("litebox_platform_multiplex/", 1),
-            ("litebox_platform_windows_userland/", 5),
+            ("litebox_platform_windows_userland/", 7),
             ("litebox_runner_linux_userland/", 1),
             ("litebox_shim_linux/", 13),
             ("litebox_shim_optee/", 10),


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -304,234 +304,311 @@ impl LinuxUserland {
...
+    mov gs:guest_fsbase@tpoff, rdx
+    rdgsbase rdx
+    wrfsbase rdx
 
+    // Handle the syscall and resume the guest.
+    call {syscall_handler}
+
+    // This thread is done. Return.
+    lea  rsp, [rbp - 5*8]
+    pop  r15
```
wdcui, line 459: Just to confirm these are the only non-volatile registers for Linux?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -304,234 +304,311 @@ impl LinuxUserland {
...
+    mov ebp, fs:host_bp@ntpoff
+
+    // Align the stack to 16 bytes
     and esp, -16
-    sub esp, 8
 
-    push ebp      /* frame_pointer */
-    push eax      /* ctx */
-    call thread_start_internal
+    // Switch to host gs
```
wdcui, line 565: Why do we need to switch to gs?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1409,254 +1489,35 @@ impl litebox::mm::allocator::MemoryProvider for LinuxUserland {
...
 ///
+/// Returns only if the guest thread is exiting. Otherwise, resumes the guest
+/// without returning.
+///
 /// # Safety
 ///
-/// - The `ctx` pointer must be valid pointer to a `litebox_common_linux::PtRegs` structure.
-/// - If any syscall argument is a pointer, it must be valid.
+/// - The `ctx` pointer must be valid pointer to a
+///   `litebox_common_linux::PtRegs` structure.
```
wdcui: extra empty spaces in the comment?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -285,137 +285,250 @@ impl WindowsUserland {
...
     push r13
+    .seh_pushreg r13
     push r14
+    .seh_pushreg r14
     push r15
+    .seh_pushreg r15
+    sub rsp, 168 // align + space for xmm6-xmm15
+    .seh_stackalloc 168
+    movdqa [rsp + 0*16], xmm6
+    .seh_savexmm xmm6, 0*16
```
wdcui, line 359: Why do we save xmm registers on Windows but not on Linux

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -285,137 +285,250 @@ impl WindowsUserland {
...
+    .seh_savexmm xmm13, 7*16
+    movdqa [rsp + 8*16], xmm14
+    .seh_savexmm xmm14, 8*16
+    movdqa [rsp + 9*16], xmm15
+    .seh_savexmm xmm15, 9*16
+    .seh_endprologue
+
+    // Save the host rsp and rbp and guest context
+    mov     r11d, DWORD PTR [rip + _tls_index]
+    mov     r8, QWORD PTR gs:[88]
```
wdcui: It would be hard to track/update these constant offsets. Any better way than just use a number here?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -285,137 +285,250 @@ impl WindowsUserland {
...
+    movdqa [rsp + 7*16], xmm13
+    .seh_savexmm xmm13, 7*16
+    movdqa [rsp + 8*16], xmm14
+    .seh_savexmm xmm14, 8*16
+    movdqa [rsp + 9*16], xmm15
+    .seh_savexmm xmm15, 9*16
+    .seh_endprologue
+
+    // Save the host rsp and rbp and guest context
+    mov     r11d, DWORD PTR [rip + _tls_index]
```
wdcui: Why do we need this low-level direct access to tls?

litebox_platform_windows_userland/src/lib.rs
jstarks: xmm6-15 are non-volatile on Windows but volatile on Linux

litebox_platform_linux_userland/src/lib.rs
jstarks: Yes, that's right.

litebox_platform_linux_userland/src/lib.rs
jstarks: We are switching from guest TLS to host TLS before calling to Rust code.

litebox_platform_windows_userland/src/lib.rs
jstarks: Before we can touch any guest registers (other than r11, which is available after a syscall) or the stack (which is the guest stack and may be invalid), we have to save register state somewhere and switch to the host stack. We can't call into Rust code (e.g., to access TLS via high-level mechanisms) before that. And in Windows in particular, where we don't control gsbase, the only place to store the guest context pointer and the host stack pointer is TLS.

On Linux, we control gsbase (and fs on x86), so we could use non-standard TLS, settings gsbase/fs to any address we want. But I figured it would be simpler to use the same TLS-based model for everything (except it's simpler on Linux because there is less indirection--this makes it possible to fully restore the context when returning to the guest without restoring to making a syscall).

litebox_platform_windows_userland/src/lib.rs
jstarks: This 88 is a magic number, part of the Windows TLS ABI for x86-64. It's the offset into the TEB. Let me see if there's a structure defined in windows_sys or something so that we can use offset_of!. If not, I'll define a constant and use it so that it's self-documenting.

litebox_platform_windows_userland/src/lib.rs
jstarks: I thought about this some more and decided it's actually better to use "classic" Windows TLS slots instead of this loader-supported TLS. This simplifies the logic for finding the TLS data, which allows us to avoid clobbering rflags on syscall entry (probably doesn't matter, but better safe than sorry).

litebox_platform_linux_userland/src/lib.rs
wdcui: Ah that's right. This is x86 so gs is used (instead of fs).

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -285,137 +285,270 @@ impl WindowsUserland {
...
+    movdqa [rsp + 7*16], xmm13
+    .seh_savexmm xmm13, 7*16
+    movdqa [rsp + 8*16], xmm14
+    .seh_savexmm xmm14, 8*16
+    movdqa [rsp + 9*16], xmm15
+    .seh_savexmm xmm15, 9*16
+    .seh_endprologue
+
+    // Offset into the TEB (gs segment) where TLS slots are stored.
+    .equ TEB_TLS_SLOTS_OFFSET, 5248
```
wdcui, line 381: How stable is this offset? Do we need to worry about Windows ever changing it?

litebox_platform_windows_userland/src/lib.rs
wdcui: The new approach looks good to me. Thank you, John.

litebox_platform_windows_userland/src/lib.rs
jstarks: This is part of the public ABI.

wdcui: Got it. Thanks!

PR #434, opened 2025-10-17
Support pselect

CvvT:
1. Add support for syscall pselect based on `poll` functionality (they are the same except that pselect has a different format of input).
2. Fix a deadlock issue in `poll`.

### DIFF
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -1796,66 +1796,6 @@ bitflags::bitflags! {
     }
 }
 
-pub const FD_SETSIZE: usize = 1024;
-#[derive(Debug, Default, Clone, Copy)]
-#[repr(C)]
-pub struct FdSet {
-    fds_bits: [usize; FD_SETSIZE / Self::USIZE_BITS],
-}
-
-impl FdSet {
-    const USIZE_BITS: usize = core::mem::size_of::<usize>() * 8;
-    /// Check if a file descriptor is set in the set
-    ///
-    /// # Panics
-    ///
-    /// Panics if `fd` is out of range ( `fd` >= [`FD_SETSIZE`]).
-    pub fn is_set(&self, fd: u32) -> bool {
-        let idx = fd as usize / Self::USIZE_BITS;
-        let bit = fd as usize % Self::USIZE_BITS;
-        (self.fds_bits[idx] & (1 << bit)) != 0
-    }
-
-    /// Clear all file descriptors in the set
-    pub fn clear(&mut self) {
-        self.fds_bits.fill(0);
-    }
-
-    /// Set a file descriptor in the set
-    ///
-    /// # Panics
-    ///
-    /// Panics if `fd` is out of range ( `fd` >= [`FD_SETSIZE`]).
-    pub fn set(&mut self, fd: u32) {
-        let idx = fd as usize / Self::USIZE_BITS;
-        let bit = fd as usize % Self::USIZE_BITS;
-        self.fds_bits[idx] |= 1 << bit;
-    }
-
-    /// Iterate over all set file descriptors in the set
-    pub fn iter(&self) -> impl Iterator<Item = u32> {
-        self.fds_bits.iter().enumerate().flat_map(|(idx, bits)| {
-            let mut bits = *bits;
-            core::iter::from_fn(move || {
-                if bits == 0 {
-                    None
-                } else {
-                    let bit = bits.trailing_zeros();
-                    bits &= !(1 << bit);
-                    Some((idx * Self::USIZE_BITS + bit as usize).truncate())
-                }
-            })
-        })
-    }
-}
-
-/// Packaged sigset with its size, used by `pselect` syscall
-#[derive(Clone, Copy)]
-pub struct SigSetPack {
-    pub sigset: SigSet,
-    pub size: usize,
-}
-
 /// Packaged sigset with its size, used by `pselect` syscall
 #[derive(Clone, Copy)]
 pub struct SigSetPack {
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -1358,9 +1358,11 @@ fn select_common(
 ) -> Result<usize, Errno> {
     if nfds >= i32::MAX as u32
         || nfds as usize
-            > super::process::LITEBOX_PROCESS
-                .limits
-                .get_rlimit_cur(litebox_common_linux::RlimitResource::NOFILE)
+            > with_current_task(|task| {
+                task.process
+                    .limits
+                    .get_rlimit_cur(litebox_common_linux::RlimitResource::NOFILE)
+            })
     {
         return Err(Errno::EINVAL);
     }


litebox_common_linux/src/lib.rs
```diff
@@ -1757,6 +1757,59 @@ pub enum IntervalTimer {
     Prof = 2,
 }
 
+pub const FD_SETSIZE: usize = 1024;
```
jstarks: I think this is a C library convention, not a kernel requirement. We need to only read the number of bytes (longs?) that are required by the # of fds user mode specifies in the call, which is allowed to be as large or small as they want.

jstarks: (I think the fdset size is limited by RLIMIT_NOFILE, so there _is_ a limit, but it's dynamic.)

litebox_common_linux/src/lib.rs
```diff
@@ -2519,6 +2580,28 @@ impl<Platform: litebox::platform::RawPointerProvider> SyscallRequest<Platform> {
...
             }
+            #[cfg(target_arch = "x86_64")]
+            Sysno::pselect6 => {
+                sys_req!(Pselect {
+                    nfds,
+                    readfds:*,
+                    writefds:*,
+                    exceptfds:*,
+                    timeout:*,
+                    sigmask:*,
```
jstarks: This isn't quite right--pselect6's last argument is a pointer to a `[usize; 2]`, where `v[0]` is the sigset pointer and `v[1]` is the size of the sigset.

litebox_common_linux/src/lib.rs
```diff
@@ -2519,6 +2580,28 @@ impl<Platform: litebox::platform::RawPointerProvider> SyscallRequest<Platform> {
             Sysno::poll => {
                 sys_req!(Poll { fds:*, nfds, timeout })
             }
+            #[cfg(target_arch = "x86_64")]
+            Sysno::pselect6 => {
```
jstarks, line 2619: Might as well support `select` while you're at it.

litebox_shim_linux/src/syscalls/epoll.rs
```diff
@@ -493,6 +493,7 @@ impl PollSet {
                     register = false;
                 }
             }
+            drop(fds);
```
jstarks, line 496: Thanks for the fix.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1266,6 +1267,120 @@ fn do_ppoll(
...
 }
 
+pub(crate) fn do_pselect(
+    nfds: u32,
+    readfds: Option<&mut litebox_common_linux::FdSet>,
+    writefds: Option<&mut litebox_common_linux::FdSet>,
+    exceptfds: Option<&mut litebox_common_linux::FdSet>,
+    timeout: Option<core::time::Duration>,
+) -> Result<usize, Errno> {
+    if nfds as usize >= litebox_common_linux::FD_SETSIZE {
```
wdcui: Should we return `EINVAL` if `nfds` is greater than `max_fd`?

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1266,6 +1267,120 @@ fn do_ppoll(
...
+
+    set.wait_or_timeout(|| file_descriptors().read(), timeout);
+
+    let mut ready_count = 0;
+    if let Some(fds) = readfds {
+        fds.clear();
+        for (i, revents) in set.revents_with_fds() {
+            if revents.intersects(Events::IN | Events::ALWAYS_POLLED) {
+                // no negative fds added to the set
+                fds.set(i.reinterpret_as_unsigned());
```
wdcui: Does this `fds` leave this `if let Some(fds) .. {}` closure?

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1266,6 +1267,120 @@ fn do_ppoll(
...
+        unimplemented!("no sigmask support yet");
+    }
+    let timeout = timeout
+        .map(super::process::get_timeout)
+        .transpose()?
+        .map(Into::into);
+
+    let mut kreadfds = readfds
+        .map(|fds| unsafe { fds.read_at_offset(0) }.ok_or(Errno::EFAULT))
+        .transpose()?
```
wdcui, line 1372: I remember @jaybosamiya-ms wanted to avoid `transpose` as much as we can. Is this one avoidable?

litebox_common_linux/src/lib.rs
CvvT: You're right. It should point to [`struct [sigset_argpack](https://elixir.bootlin.com/linux/v5.19.17/C/ident/sigset_argpack)`](https://elixir.bootlin.com/linux/v5.19.17/source/fs/select.c#L773). Fixed now.

litebox_shim_linux/src/syscalls/file.rs
CvvT: I just tested it on Linux. Actually, it is okay to have a large `nfds`, but we should return `EBADF` if there is any invalid fd, which I didn't handled. It is now fixed.

litebox_shim_linux/src/syscalls/file.rs
CvvT: It should be local to this closure.

litebox_common_linux/src/lib.rs
CvvT: Yeah, it almost identical to `pselect`. It's added now.

litebox_common_linux/src/lib.rs
CvvT: Good catch. I was confused by the [manual]( https://www.man7.org/linux/man-pages/man2/select.2.html) and `fd_set` defined in Linux kernel: https://elixir.bootlin.com/linux/v5.19.17/source/include/uapi/linux/posix_types.h#L27. 

```
WARNING: select() can monitor only file descriptors numbers that
       are less than FD_SETSIZE (1024)—an unreasonably low limit for many
       modern applications—and this limitation will not change.  All
       modern applications should instead use [poll(2)](https://www.man7.org/linux/man-pages/man2/poll.2.html) or [epoll(7)](https://www.man7.org/linux/man-pages/man7/epoll.7.html), which
       do not suffer this limitation.
```

However, the `ERRORS` section also mentions:
```
EINVAL nfds is negative or exceeds the RLIMIT_NOFILE resource
              limit (see [getrlimit(2)](https://www.man7.org/linux/man-pages/man2/getrlimit.2.html)).
```

I feel it might be fine to use a fixed-sized array in practice, but let me try to switch to a dynamic one.

litebox_shim_linux/src/syscalls/file.rs
CvvT: Did you mean `transmute` #154 ?

jaybosamiya-ms: Yeah `transpose` is fine, `transmute` is the one to avoid :)

litebox_common_linux/src/lib.rs
CvvT: Switched to `bitvec` now

PR #435, opened 2025-10-17
Clear out manually-handled `Socket`s from the shim

jaybosamiya-ms:
This PR switches out the manually handled `Socket`s in the shim, and instead uses LiteBox raw FDs for them.

---

### DIFF
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -267,20 +267,9 @@ fn setsockopt(
                         // CORK is the opposite of NODELAY
                         val == 0
                     };
-                    if let Err(err) = litebox_net()
+                    litebox_net()
                         .lock()
-                        .set_tcp_option(fd, litebox::net::TcpOptionData::NODELAY(on))
-                    {
-                        match err {
-                            litebox::net::errors::SetTcpOptionError::InvalidFd => {
-                                return Err(Errno::EBADF);
-                            }
-                            litebox::net::errors::SetTcpOptionError::NotTcpSocket => {
-                                return Err(Errno::EOPNOTSUPP);
-                            }
-                            _ => unimplemented!(),
-                        }
-                    }
+                        .set_tcp_option(fd, litebox::net::TcpOptionData::NODELAY(on))?;
                     Ok(())
                 }
                 TcpOption::KEEPINTVL => {
@@ -787,6 +776,7 @@ pub(crate) fn sys_sendto(
     match socket {
         Descriptor::LiteBoxRawFd(raw_fd) => with_socket_fd(*raw_fd, |fd| {
             let sockaddr = sockaddr.map(|SocketAddress::Inet(addr)| addr);
+            drop(file_table); // Drop before possibly-blocking `sendto`
             sendto(fd, &buf, flags, sockaddr)
         }),
         _ => Err(Errno::ENOTSOCK),
@@ -870,12 +860,9 @@ pub(crate) fn sys_getsockopt(
         .get_fd(sockfd)
         .ok_or(Errno::EBADF)?
     {
-        Descriptor::LiteBoxRawFd(raw_fd) => crate::run_on_raw_fd(
-            *raw_fd,
-            |fd| Err(Errno::ENOTSOCK),
-            |fd| getsockopt(fd, optname, optval, optlen),
-        )
-        .flatten(),
+        Descriptor::LiteBoxRawFd(raw_fd) => {
+            with_socket_fd(*raw_fd, |fd| getsockopt(fd, optname, optval, optlen))
+        }
         _ => Err(Errno::ENOTSOCK),
     }
 }


jaybosamiya-ms:
Currently blocked on #416.  

I've had to rebase this a couple of times while working on it, so one of the bits (`KEEPINVTL`) apparently didn't make it over (probably got lost in one of the conflict fixes).  I'll fix that up soon.

Nonetheless, at least the overall core design for the migration should be fairly clear at this point, so I'm saying "ready for review".

Oh, also, more cleanup is definitely possible in a future PR. I've mostly tried to keep things _roughly_ one-to-one-mapped to keep the reviews easier to do.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -68,340 +84,304 @@ pub(crate) enum SocketAddress {
...
+                                return Err(Errno::EBADF);
                             }
+                            litebox::net::errors::SetTcpOptionError::NotTcpSocket => {
+                                return Err(Errno::EOPNOTSUPP);
+                            }
+                            _ => unimplemented!(),
                         }
-                        Ok(())
-                    }
-                    TcpOption::KEEPINTVL => {
```
jaybosamiya-ms, line 265: Hmm, this case seems to have got dropped somewhere in the rebase conflict fixups, will fix soon.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -613,12 +591,13 @@ pub(crate) fn sys_sendto(
...
-        Descriptor::Socket(socket) => {
-            let socket = socket.clone();
-            // drop file table as `sendto` may block
-            drop(file_table);
-            socket.sendto(&buf, flags, sockaddr.map(|SocketAddress::Inet(addr)| addr))
-        }
+        Descriptor::LiteBoxRawFd(raw_fd) => with_socket_fd(*raw_fd, |fd| {
+            if get_status(fd).contains(OFlags::NONBLOCK) {
+                flags.insert(SendFlags::DONTWAIT);
+            }
```
CvvT: This is unnecessary because it is done inside `sendto`.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -527,20 +507,18 @@ pub(crate) fn sys_accept(
...
-            let fd = socket.accept()?;
-            Descriptor::Socket(alloc::sync::Arc::new(Socket::new(
-                fd,
-                flags,
-                crate::litebox(),
-                Events::empty(),
-            )))
-        }
+        Descriptor::LiteBoxRawFd(raw_fd) => with_socket_fd(*raw_fd, |fd| {
+            let fd = accept(fd)?;
```
CvvT: `accept` might be blocking so we don't want hold lock on `file_table` while calling `accept`?

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -636,15 +615,12 @@ pub(crate) fn sys_recvfrom(
...
-        Descriptor::Socket(socket) => {
-            let socket = socket.clone();
-            // drop file table as `receive` may block
-            drop(file_table);
+    match file_table.get_fd(fd).ok_or(Errno::EBADF)? {
+        Descriptor::LiteBoxRawFd(raw_fd) => with_socket_fd(*raw_fd, |fd| {
             let mut buffer: [u8; 4096] = [0; 4096];
             let mut addr = None;
-            let size = socket.receive(
+            let size = receive(
```
CvvT, line 804: Similarly, holding lock on `file_table` while calling blocking `receive`

litebox_shim_linux/src/syscalls/net.rs
jaybosamiya-ms: Ah, apparently a leftover from when I was rebasing things.  Seems like you removed these bits in #372 that you merged two days ago.  Removed, thanks!

litebox_shim_linux/src/syscalls/net.rs
jaybosamiya-ms: By the end of the set of changes for shared-fd, there won't be a `file_table` anymore, so it won't be a problem then, that might be why when thinking about the blocking behavior here, I might have missed this.

There is a fairly simple fix here (the drop below can just be moved up since all we need from the `file_table` is the `usize`), updated. 

Good catch, thanks!

litebox_shim_linux/src/syscalls/net.rs
jaybosamiya-ms: Thanks!  Similar fix as earlier, we only need the `usize` from the `file_table` and don't need to hold the lock any longer.  Fixed.  I also aim to remove `file_table` entirely in upcoming PRs, so basically, I am not _too_ concerned in general but indeed, we should make sure not to introduce regressions in the meanwhile.

litebox_shim_linux/src/syscalls/net.rs
jaybosamiya-ms: Indeed, was introduced by #395 which was merged two days ago, and the rebase missed that. Fixed it up.

PR #439, opened 2025-10-22
fix epoll for LiteBoxRawFd

CvvT:
1. Due to recent change #435 on LiteBox's API, `epoll` needs to be updated accordingly. Polling on socket is going to be done in a separate PR #383
2.  Add `IoctlArg::FIONBIO` and `FcntlArg::GETFL` support for socket

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -1174,8 +1174,7 @@ pub(crate) fn sys_epoll_ctl(
     };
 
     let file = locked_file_descriptors.get_fd(fd).ok_or(Errno::EBADF)?;
-    let file_descriptor = super::epoll::EpollDescriptor::try_from(file)
-        .expect("file should still be alive under lock");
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1146,6 +1174,8 @@ pub fn sys_epoll_ctl(
     };
 
     let file = locked_file_descriptors.get_fd(fd).ok_or(Errno::EBADF)?;
+    let file_descriptor = super::epoll::EpollDescriptor::try_from(file)
+        .expect("file should still be alive under lock");
```
jaybosamiya-ms: If it is a raw FD, it might have gone away even though you have a lock on the shim-level file descriptors, no?

CvvT: If the fd is valid in the shim-level file table, it should also be valid in LiteBox's file table unless somewhere in the shim layer we close it without removing it from the shim-level file table. To be safe and easy to reason about it, I changed to return an error instead.

PR #442, opened 2025-10-23
Support SETFL for socket

CvvT:


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -831,11 +831,33 @@ pub(crate) fn sys_fcntl(
...
+                                f.toggle(diff);
+                            })
+                            .map_err(|err| match err {
+                                MetadataError::ClosedFd => Errno::EBADF,
+                                MetadataError::NoSuchMetadata => {
+                                    unreachable!("all sockets have SocketOFlags when created")
+                                }
+                            })
+                    },
+                )??,
```
wdcui, line 860: The previous code has a single "?". Why do we need two "?" now?

CvvT: Previously the closure returns `()`, so `run_on_raw_fd` returns `Result<(), Errno>`. I changed the closure to return `Result<(), Errno>`, so now `run_on_raw_fd` returns `Result<Result<(), Errno>, Errno>`.

PR #443, opened 2025-10-23
Add dup support for others

CvvT:
Add `dup` support for all other types of files (i.e., pipe, epoll, and eventfd).


jaybosamiya-ms:
Fwiw, the raw-fd cleanups would automatically add support for dup to things, but doing this right now is helpful in the meanwhile, thanks!

PR #446, opened 2025-10-23
Set up pipes as an FD-enabled subsystem

jaybosamiya-ms:
This PR makes `litebox::pipes` an FD-enabled subsystem.

This specific PR is concerned entirely with introducing a new public interface to be used; a future PR will actually use this from the shim.

This shim updates the tests _within_ the LiteBox crate itself to use this new interface, so that a future PR can eliminate the direct dependence on internal details more cleanly.

---

### DIFF
--- a/litebox/src/pipes.rs
+++ b/litebox/src/pipes.rs
@@ -76,7 +76,7 @@ impl<Platform: RawSyncPrimitivesProvider + TimeProvider> Pipes<Platform> {
         (sender, receiver)
     }
 
-    /// Close the pipe at at `fd`.
+    /// Close the pipe at `fd`.
     ///
     /// Future operations on the `fd` will start to return `ClosedFd` errors.
     pub fn close(&self, fd: &PipeFd<Platform>) -> Result<(), errors::CloseError> {
@@ -168,9 +168,8 @@ bitflags::bitflags! {
     #[repr(transparent)]
     #[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
     pub struct Flags: u32 {
-        /// `NON_BLOCKING` impacts what happens when a full channel is written or, or an empty
-        /// channel is read from. If set, the operations returns immediately with a `WouldBlock`
-        /// error.
+        /// `NON_BLOCKING` impacts what happens when a full channel is written, or an empty channel
+        /// is read from. If set, the operations returns immediately with a `WouldBlock` error.
         const NON_BLOCKING = 0x1;
     }
 }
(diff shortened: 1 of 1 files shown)


litebox/src/pipes.rs
```diff
@@ -27,6 +27,215 @@ use crate::{
...
+        );
+        let sender = PipeEnd::Sender(sender);
+        let receiver = PipeEnd::Receiver(receiver);
+        let mut dt = self.litebox.descriptor_table_mut();
+        let sender = dt.insert(sender);
+        let receiver = dt.insert(receiver);
+        (sender, receiver)
+    }
+
+    /// Close the pipe at at `fd`.
```
wdcui: "at at"

litebox/src/pipes.rs
```diff
@@ -27,6 +27,215 @@ use crate::{
...
+    ///
+    /// Future operations on the `fd` will start to return `ClosedFd` errors.
+    pub fn close(&self, fd: &PipeFd<Platform>) -> Result<(), errors::CloseError> {
+        self.litebox.descriptor_table_mut().remove(fd);
```
wdcui, line 83: We close `sender` and `receiver` separately? Any implications from doing that?

litebox/src/pipes.rs
```diff
@@ -27,6 +27,215 @@ use crate::{
...
+    Receiver(Arc<ReadEnd<Platform, u8>>),
+    Sender(Arc<WriteEnd<Platform, u8>>),
+}
+
+bitflags::bitflags! {
+    /// Flags for controlling the pipe behaviors.
+    #[repr(transparent)]
+    #[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+    pub struct Flags: u32 {
+        /// `NON_BLOCKING` impacts what happens when a full channel is written or, or an empty
```
wdcui: "or, or"

litebox/src/pipes.rs
```diff
@@ -27,6 +27,215 @@ use crate::{
...
+        WouldBlock,
+    }
+
+    /// Possible errors from [`Pipes::write`]
+    #[non_exhaustive]
+    #[derive(Error, Debug)]
+    pub enum WriteError {
+        #[error("not an open file descriptor")]
+        ClosedFd,
+        #[error("not open for writeing")]
```
wdcui: "writeing"

litebox/src/pipes.rs
jaybosamiya-ms: Both the sender and receiver have `Drop` implementations within them that trigger their peer to shutdown.  The removal from the descriptor table will automatically take care of shutdowns (that's why I have the comment right after this line).  The Drop implementations are existing behavior elsewhere in the file, unchanged by this PR.

wdcui: Got it. Thanks!

PR #447, opened 2025-10-24
Add exception handler infrastructure

jstarks:
Add an interface for the platform to call the shim when the guest code causes an unresolved hardware exception (page fault, invalid opcode, divide by zero, etc.). Implement support for catching exceptions in the Linux and Windows userland platforms.

For now, the Linux and OPTEE shims panic when an exception occurs. In future changes, the Linux shim will inject the appropriate signal into the current task in response to a signal. The correct OPTEE behavior is less clear.

This includes changing the shim interface from a single function to the `EnterShim` trait. For now, shims are implementing this on ZSTs and are still being registered globally with the platforms. In the future, we should probably make this registration per-thread and move global shim state (if there is any) into the shim object.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -1901,6 +1901,9 @@ unsafe extern "C" fn exception_signal_handler(
     }
     *orig_eax = *eax;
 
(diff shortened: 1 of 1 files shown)


wdcui:
John, I forgot to check yesterday. Do we currently have unit tests that trigger the signal handling paths? Thanks!

PR #449, opened 2025-10-24
Add an integration test for python

CvvT:
Add a simple end-to-end test for python3.

As I commented in the code, pre-compiled .pyc files are not used because of the lack of support for timestamp in fs.

Also, updated the function `run_target_program` so that it can 1) enable network (i.e., tun), 2) pass environment variables, and 3) create tar file using `ustar` format.

### DIFF
--- a/litebox_runner_linux_userland/tests/run.rs
+++ b/litebox_runner_linux_userland/tests/run.rs
@@ -381,29 +381,50 @@ fn test_runner_with_python() {
                 let python_lib_src = Path::new(each);
                 if python_lib_src.is_dir() {
                     let python_lib_dst = out_dir.join(&each[1..]); // remove leading '/'
-                    if python_lib_dst.exists() {
-                        continue;
-                    }
-                    std::fs::create_dir_all(&python_lib_dst).unwrap();
-                    println!(
-                        "Copying python3 lib from {} to {}",
-                        python_lib_src.to_str().unwrap(),
-                        python_lib_dst.to_str().unwrap()
-                    );
-                    // TODO: we may also need to rewrite all .so files under the python lib directory
-                    let output = std::process::Command::new("cp")
-                        .args([
-                            "-a",
+                    if !python_lib_dst.exists() {
+                        std::fs::create_dir_all(&python_lib_dst).unwrap();
+                        println!(
+                            "Copying python3 lib from {} to {}",
                             python_lib_src.to_str().unwrap(),
-                            python_lib_dst.parent().unwrap().to_str().unwrap(),
-                        ])
-                        .output()
-                        .expect("Failed to copy python3 lib");
-                    assert!(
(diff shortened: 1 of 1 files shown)


litebox_runner_linux_userland/tests/run.rs
```diff
@@ -310,3 +358,72 @@ fn test_runner_with_ls() {
...
+                    if python_lib_dst.exists() {
+                        continue;
+                    }
+                    std::fs::create_dir_all(&python_lib_dst).unwrap();
+                    println!(
+                        "Copying python3 lib from {} to {}",
+                        python_lib_src.to_str().unwrap(),
+                        python_lib_dst.to_str().unwrap()
+                    );
+                    // TODO: we may also need to rewrite all .so files under the python lib directory
```
jaybosamiya-ms: I think we probably wish to have functionality that rewrites all ELFs.  Just attempting the rewriter on everything (glob `**/*`) should probably be enough, because the rewriter will just quickly exit out of anything that is not an ELF.

CvvT: Good point! Now it tries to rewrite everything except files with certain extensions like `py` and `pyc` to speed up.

jaybosamiya-ms:
Thanks @CvvT for adding Python to our test suite!

Unfortunately, this PR implicitly takes a version of Python from the environment, and at least my local version of python3 (3.12.2) fails the test with
```
Error: failed to generate object file: Unaligned sh_offset value 0x201000 for section '.text', expected 0x400000
```

Do you mind mentioning which version of python3 works for you?  I'm guessing whatever is on the CI probably also works, which is why the CI didn't complain.

We should probably fix this issue in the rewriter, but at least until then, I can set up a check within the test and also set up the necessary stuff to ensure that we're testing on the python3 version that works.  I'm happy to make the PR (and open an issue to track the rewriting thing) if you can let me know the working Python version, thanks!

CvvT:
> Thanks @CvvT for adding Python to our test suite!
> 
> Unfortunately, this PR implicitly takes a version of Python from the environment, and at least my local version of python3 (3.12.2) fails the test with
> 
> ```
> Error: failed to generate object file: Unaligned sh_offset value 0x201000 for section '.text', expected 0x400000
> ```
> 
> Do you mind mentioning which version of python3 works for you? I'm guessing whatever is on the CI probably also works, which is why the CI didn't complain.
> 
> We should probably fix this issue in the rewriter, but at least until then, I can set up a check within the test and also set up the necessary stuff to ensure that we're testing on the python3 version that works. I'm happy to make the PR (and open an issue to track the rewriting thing) if you can let me know the working Python version, thanks!

The error message sounds familiar, but I don't recall what's the root cause. Anyway, I'm using `Python 3.10.12`. Thanks!

PR #451, opened 2025-10-28
Clear out manually-handled pipes from the shim

jaybosamiya-ms:
This PR switches out the manually handled pipes in the shim and instead uses LiteBox raw FDs for them.

It additionally switches away the old internal details from `litebox::pipes` to be private, so that we can use the more consistent-with-other-subsystems interface introduced by #446, which will then allow future cleanups to happen inside `litebox::pipes`.

---

### DIFF
--- a/litebox_common_linux/src/errno/mod.rs
+++ b/litebox_common_linux/src/errno/mod.rs
@@ -498,6 +498,7 @@ impl From<litebox::pipes::errors::WriteError> for Errno {
     fn from(value: litebox::pipes::errors::WriteError) -> Self {
         match value {
             litebox::pipes::errors::WriteError::ClosedFd => Errno::EBADF,
+            litebox::pipes::errors::WriteError::ReadEndClosed => Errno::EPIPE,
             litebox::pipes::errors::WriteError::NotForWriting => Errno::EINVAL,
             litebox::pipes::errors::WriteError::WouldBlock => Errno::EWOULDBLOCK,
             _ => todo!(),
--- a/litebox/src/pipes.rs
+++ b/litebox/src/pipes.rs
@@ -88,6 +88,9 @@ impl<Platform: RawSyncPrimitivesProvider + TimeProvider> Pipes<Platform> {
     /// Read values in the pipe into `buf`, returning the number of elements read.
     ///
     /// See [`Self::create_pipe`] for details on blocking behavior.
+    ///
+    /// Note: currently, this function returns `Ok(0)` if the peer end has been shut down, this may
+    /// change in the future to an explicit "peer has shut down" error.
     pub fn read(&self, fd: &PipeFd<Platform>, buf: &mut [u8]) -> Result<usize, errors::ReadError> {
         let dt = self.litebox.descriptor_table();
         let p = match &dt.get_entry(fd).ok_or(errors::ReadError::ClosedFd)?.entry {
@@ -234,6 +237,8 @@ pub mod errors {
     pub enum WriteError {
         #[error("not an open file descriptor")]
         ClosedFd,
+        #[error("the reading end of this pipe is closed")]
+        ReadEndClosed,
         #[error("not open for writing")]
         NotForWriting,
         #[error("write would block")]
@@ -323,8 +328,10 @@ struct WriteEnd<Platform: RawSyncPrimitivesProvider + TimeProvider, T> {
 #[derive(Error, Debug)]
 #[non_exhaustive]
 enum PipeError {
-    #[error("this pipe has been closed down")]
-    Closed,
+    #[error("this end has been shut down")]
+    ThisEndShutdown,
+    #[error("peer has been shut down")]
+    PeerShutdown,
     #[error("this operation would block")]
     WouldBlock,
 }
@@ -332,7 +339,10 @@ enum PipeError {
 impl From<PipeError> for errors::ReadError {
     fn from(err: PipeError) -> Self {
         match err {
-            PipeError::Closed => errors::ReadError::ClosedFd,
+            PipeError::ThisEndShutdown => errors::ReadError::ClosedFd,
+            PipeError::PeerShutdown => {
+                unreachable!("unreachable for now; see documentation of `read`")
+            }
             PipeError::WouldBlock => errors::ReadError::WouldBlock,
         }
     }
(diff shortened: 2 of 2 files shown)


litebox_common_linux/src/errno/mod.rs
```diff
@@ -483,12 +483,38 @@ impl From<litebox::sync::futex::FutexError> for Errno {
...
+            _ => todo!(),
+        }
+    }
+}
+
+impl From<litebox::pipes::errors::CloseError> for Errno {
+    fn from(_value: litebox::pipes::errors::CloseError) -> Self {
+        todo!()
+    }
+}
```
wdcui, line 513: Is this one same as the `impl` below?

litebox_shim_linux/src/lib.rs
```diff
@@ -170,16 +171,24 @@ pub fn litebox_page_manager<'a>() -> &'a PageManager<Platform, PAGE_SIZE> {
...
     NET.get_or_init(|| {
-        let net = litebox::net::Network::new(litebox());
+        let net = Network::new(litebox());
         alloc::boxed::Box::new(litebox().sync().new_mutex(net))
     })
 }
 
+pub(crate) fn litebox_pipes<'a>()
+-> &'a litebox::sync::RwLock<Platform, litebox::pipes::Pipes<Platform>> {
+    static PIPES: OnceBox<litebox::sync::RwLock<Platform, litebox::pipes::Pipes<Platform>>> =
```
wdcui: Minor: to be consistent with the code change in `litebox_net`, should we skip `litebox::pipes::` here and below?

litebox_common_linux/src/errno/mod.rs
```diff
@@ -483,12 +483,38 @@ impl From<litebox::sync::futex::FutexError> for Errno {
...
+            litebox::pipes::errors::ReadError::WouldBlock => Errno::EWOULDBLOCK,
+            _ => todo!(),
+        }
+    }
+}
+
+impl From<litebox::pipes::errors::WriteError> for Errno {
+    fn from(value: litebox::pipes::errors::WriteError) -> Self {
+        match value {
+            litebox::pipes::errors::WriteError::ClosedFd => Errno::EBADF,
```
CvvT, line 500: Similarly, `PipeError::Closed` is different from`errors::WriteError::ClosedFd` and is corresponding to EPIPE as shown in the following example:

```
    if (pipe(fd1) == -1) {
        fprintf(stderr, "Pipe Failed");
        return 1;
    }
    close(fd1[0]); // Close reading end of first pipe

    signal(SIGPIPE, SIG_IGN);
    int n = write(fd1[1], input_str, strlen(input_str) + 1);
    if (n == -1) {
        perror("write");
        return 1;
    }
```
`write` should return `EPIPE`.

We need to add `Closed` to `errors::WriteError`.

litebox_common_linux/src/errno/mod.rs
```diff
@@ -483,12 +483,38 @@ impl From<litebox::sync::futex::FutexError> for Errno {
...
 
-impl From<litebox::pipes::PipeError> for Errno {
-    fn from(value: litebox::pipes::PipeError) -> Self {
+impl From<litebox::pipes::errors::ReadError> for Errno {
+    fn from(value: litebox::pipes::errors::ReadError) -> Self {
         match value {
-            litebox::pipes::PipeError::Closed => Errno::EPIPE,
-            litebox::pipes::PipeError::WouldBlock => Errno::EAGAIN,
-            _ => unimplemented!(),
+            litebox::pipes::errors::ReadError::ClosedFd => Errno::EBADFD,
```
CvvT, line 489: `PipeError::Closed` is different from`errors::ReadError::ClosedFd` and should correspond to EPIPE. We need to add `Closed` to `errors::ReadError`.

litebox_common_linux/src/errno/mod.rs
jaybosamiya-ms: Nope, this one is errors from `close` operation, while the other one is a shared thing for a bunch of related operations whose only potential error is "closed fd".  The `todo!()` is because it is a `#[non_exhaustive]` enum.  I plan to clear out some of the non-exhaustives in the future, which should make things more obvious, but for now we're leaving it as non-exhaustive since we haven't yet finalized what the potential errors can/should be.

litebox_common_linux/src/errno/mod.rs
jaybosamiya-ms: Good catch!  Updated.

litebox_common_linux/src/errno/mod.rs
jaybosamiya-ms: Hmm interestingly, `man 2 read` does not mention EPIPE, so I wrote a quick test:
```c
#include <stdio.h>
#include <unistd.h>

int main(void) {
    int r1, r2;
    int fds[2];
    char buffer[128];

    pipe(fds);
    write(fds[1], "TEST", 4);
    close(fds[1]);

    r1 = read(fds[0], buffer, sizeof(buffer));
    r2 = read(fds[0], buffer, sizeof(buffer));
    printf("%d %d\n", r1, r2);
}
```

This returns `4 0`, so the closed write end does not trigger an EPIPE on the write end.  I am updating to follow this behavior, with a doc-comment to note that we might wish to expose this a bit more nicely at the LiteBox crate level (because implicit `Ok(0)` is not ideal imho).

litebox_shim_linux/src/lib.rs
jaybosamiya-ms: Good catch, updated!

litebox_common_linux/src/errno/mod.rs
jaybosamiya-ms: See 62eb4f17daa7d707b4ee6fc5b12823570503dc97

CvvT: You're right. Read from a closed pipe should be able to drain all the remaining data in the buffer and return zero eventually, but we still need to distinguish between `no data for now` and `the pipe is closed` because blocking read need to retry if it is just `no data for now`. That's why I would like to return a specific error like you did and let the shim to handle it properly. I plan to have another PR to handle EPIPE error.

PR #452, opened 2025-10-28
Fix page management provider for Windows

CvvT:
This PR fixes multiple issues in the page management provider for Windows:

1. `VirtualProtect` only works on a single region but the provided range may span multiple regions. Fix `update_permissions` by iterating all regions within the given range.
2. When allocating pages, separate `MEM_RESERVE` and `MEM_COMMIT` because `MEM_RESERVE` requires the base address to be system allocation granularity (e.g., 64KB on my machine) aligned while `MEM_COMMIT` only asks for page aligned base address. Call single `VirtualAlloc2` with both `MEM_RESERVE` and `MEM_COMMIT` may waste a lot of memory.
3. Another long-standing issue in `allocate_pages` was that when `suggested_range` is given but `fixed_address` is false, we need to first check if there are any committed pages within the given range. If not, we can then operate on each subregion. Otherwise, we should ask OS to assign a new available range for us.

### DIFF
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -222,12 +222,6 @@ impl WindowsUserland {
         x & !(gran - 1)
     }
 
-    #[expect(dead_code)]
-    fn is_aligned_to_granu(&self, x: usize) -> bool {
-        let gran = self.sys_info.read().unwrap().dwAllocationGranularity as usize;
-        x.is_multiple_of(gran)
-    }
-
     pub fn init_task(&self) -> litebox_common_linux::TaskParams {
         // TODO: Currently we are using a static thread ID and credentials (faked).
         // This is a placeholder for future implementation to use passthrough.
@@ -943,6 +937,13 @@ where
     Ok(())
 }
 
+macro_rules! debug_assert_alignment {
+    ($r:ident, $page_size:expr) => {
+        debug_assert!($r.start.is_multiple_of($page_size));
+        debug_assert!($r.end.is_multiple_of($page_size));
+    };
+}
+
 impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for WindowsUserland {
     // TODO(chuqi): These are currently "magic numbers" grabbed from my Windows 11 SystemInformation.
     // The actual values should be determined by `GetSystemInfo()`.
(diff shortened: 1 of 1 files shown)


CvvT:
#291 was also trying to address the first and third problems mentioned above and another issue unresolved by this PR. Because the page manager in LiteBox is not aware of the memory allocated to global allocator, #291 also tries to detect whether the given `suggested_range` collides with memory assigned to global allocator when `fixed_address` is true (e.g., guest application calls `mmap` with `MAP_FIXED` set). 

For `mmap`, we probably don't need to worry about it, because the common pattern for `MAP_FIXED` is the following:

```
addr = mmap(NULL, len, ...);
mmap(addr, len, MAP_FIXED, ...)
```

The application should set `MAP_FIXED` only when the memory is allocated to it.

However, it might cause issue for `remap` when `remap` tries to grow an existing mapping. I believe it can be addressed without the complicated tracking implemented in #291. Will do it in a separate PR.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -222,6 +222,7 @@ impl WindowsUserland {
         x & !(gran - 1)
     }
 
+    #[expect(dead_code)]
```
wdcui: Why do we keep this function if it's dead code?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -919,181 +958,169 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
...
+            if ptr.is_null() {
+                core::ptr::null_mut()
+            } else {
+                unsafe {
+                    VirtualAlloc2(
+                        GetCurrentProcess(),
+                        if r.start == 0 {
+                            ptr
+                        } else {
+                            r.start as *mut c_void
```
wdcui, line 997: Is `r.start` guaranteed to be page aligned?

CvvT: Yes, all the ranges are `ALIGN`-byte aligned, guaranteed by LiteBox's page manager. I added some debug_assert to ensure `ALIGN` is a multiple of the page size, and all ranges are aligned with `ALIGN`.

litebox_platform_windows_userland/src/lib.rs
CvvT: I thought it might be useful in the future. Removed for now.

PR #454, opened 2025-10-28
OP-TEE Shim and Runner refactoring

sangho2:
This PR revises OP-TEE Shim and Linux Runner to let them use `run_thread`. It also re-enables OP-TEE TA tests.

### DIFF
--- a/dev_tests/src/ratchet.rs
+++ b/dev_tests/src/ratchet.rs
@@ -62,7 +62,7 @@ fn ratchet_maybe_uninit() -> Result<()> {
             ("dev_tests/", 1),
             ("litebox_platform_linux_kernel/", 1),
             ("litebox_platform_linux_userland/", 3),
-            ("litebox_platform_lvbs/", 5),
+            ("litebox_platform_lvbs/", 6),
             ("litebox_shim_linux/", 5),
         ],
         |file| {


litebox_platform_lvbs/src/ptr.rs
```diff
@@ -17,12 +20,22 @@ impl<T: Clone> core::fmt::Debug for UserConstPtr<T> {
 
 impl<T: Clone> Copy for UserConstPtr<T> {}
 impl<T: Clone> RawConstPointer<T> for UserConstPtr<T> {
-    unsafe fn read_at_offset<'a>(self, _count: isize) -> Option<alloc::borrow::Cow<'a, T>> {
-        todo!()
+    unsafe fn read_at_offset<'a>(self, count: isize) -> Option<alloc::borrow::Cow<'a, T>> {
+        if self.inner.is_null() || !self.inner.is_aligned() {
```
wdcui: Why does it need to be aligned?

litebox_platform_lvbs/src/ptr.rs
```diff
@@ -81,16 +104,46 @@ impl<T: Clone> RawConstPointer<T> for UserMutPtr<T> {
...
+            core::ops::Bound::Unbounded => {
+                return None;
+            }
+        };
+        let len = if start <= end {
+            start.abs_diff(end)
+        } else {
+            return None;
+        };
+        let _ = start.checked_mul(size_of::<T>().try_into().ok()?)?;
```
wdcui, line 187: Why do we do `checked_mul` on `start`?

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -418,13 +412,13 @@ impl ElfLoader {
...
         }
 
-        // TODO: the following memory page allocated for TLS in the TA
-        // will be freed when the TA is unloaded.
-
-        // Since we don't use `ld` or `ldelf` for loading a TA, we should manually set up
-        // the userspace thread local storage (TLS)
-        let _ = crate::syscalls::mm::sys_mmap(
-            super::DEFAULT_FS_BASE,
+        // Normally, loader or libc allocates memory for TLS and sets up the FS base.
```
wdcui: Can we handle tls in the optee shim in a way similar to the linux shim?

litebox_runner_optee_on_linux_userland/src/lib.rs
```diff
@@ -97,27 +98,193 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
...
+        unsafe { litebox_platform_linux_userland::run_thread(&mut pt_regs) };
+
+        if func_id == UteeEntryFunc::CloseSession {
+            litebox_shim_optee::deinit_session();
+        }
+    }
+}
+
+/// Run the loaded TA with a sequence of test commands
+#[cfg(test)]
```
wdcui: Is it possible to move all of this into a test? I understand the test code will need to do some setup before it can run the KMPP test commands. But I thought I should at least check with you about it.

litebox_runner_optee_on_linux_userland/src/lib.rs
```diff
@@ -162,23 +331,28 @@ enum TaCommandParamsBase64 {
     },
 }
 
+#[cfg(test)]
```
wdcui: Why is all of the code here only for test?

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -418,13 +412,19 @@ impl ElfLoader {
...
-        let _ = crate::syscalls::mm::sys_mmap(
-            super::DEFAULT_FS_BASE,
+        // Initialize the guest TLS for an OP-TEE TA.
+        // Typically, the loader (e.g., OP-TEE's ldelf) or libc initializes the guest TLS before
+        // calling the main function. However, currently, LiteBox does not use ldelf or libc for
+        // OP-TEE TAs such that no guest code attempts to initialize the guest TLS.
+        // To avoid this problem, `litebox_shim_optee` initializes the guest TLS by
+        // allocating a guest page and programming the FS base to point to the allocated page.
+        // Note that we use `PunchthroughSyscall::SetFsBase` for this purpose not to deal with
+        // the underlying platform behaviors (i.e., swapping FS and GS).
```
wdcui: Minor: remove the "(i.e., ....)" since the swapping behavior may be removed.

litebox_runner_optee_on_linux_userland/src/lib.rs
sangho2: I've moved all test stuffs to a separate file.

litebox_platform_lvbs/src/ptr.rs
sangho2: `alloc::borrow::Cow()` requires an aligned address. I'll revise this function and related functions to copy data if it isn't aligned.

litebox_platform_lvbs/src/ptr.rs
sangho2: to calculate the lower address (because `start` is a range index value)?

PR #455, opened 2025-10-29
LVBS: Support for x86_64 extended states

sangho2:
This PR adds support for x86_64 extended states. It enables extended state instructions and registers. It also provides functions to save and restore these extended registers. Currently, it saves/restores the VTL0 kernel's extended states on every VTL switch and the VTL1 kernel's extended states before/after running OP-TEE TAs.

### DIFF
--- a/litebox_runner_lvbs/x86_64_vtl1.json
+++ b/litebox_runner_lvbs/x86_64_vtl1.json
@@ -23,5 +23,6 @@
 	},
 	"panic-strategy": "abort",
 	"disable-redzone": true,
+	"features": "-avx,-avx2,-avx512f",
 	"dynamic-linking": false
 }


litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -33,6 +38,7 @@ pub struct PerCpuVariables {
     pub vtl1_state: VtlState,
     pub vtl0_locked_regs: ControlRegMap,
     pub gdt: Option<&'static gdt::GdtWrapper>,
+    xsave_area_addr: VirtAddr,
```
wdcui: This is used for VTL0-VTL switch or user-kernel switch?

sangho2: This is for VTL1 user-kernel switch. Linux kernel uses extended registers only within certain subsystems, so this PR doesn't consider them. If needed, we can create another buffer to save and restore VTL0 extended states for every VTL switch.

sangho2: As suggested, this PR now saves/restores VTL0 extended states on every VTL switch.

wdcui: Got it. Thanks! I saw that you added the save/restore between VTL0 and VTL1.

PR #458, opened 2025-10-30
Mostly remove use of TLS within Linux shim

jstarks:
Instead of using `with_current_task` through the shim, make most shim functions methods on `Task`. This clarifies the data flow with a minimum of extra syntax (but with some extra indentation almost everywhere).

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -191,23 +191,6 @@ impl Task {
             _ => unimplemented!("Unsupported punchthrough error {:?}", e),
         })
     }
-
-    /// Handle syscall `rt_sigreturn`.
-    pub(crate) fn sys_rt_sigreturn(&self, stack: usize) -> ! {
-        let punchthrough = litebox_common_linux::PunchthroughSyscall::RtSigreturn { stack };
-        let token = litebox_platform_multiplex::platform()
-            .get_punchthrough_token_for(punchthrough)
-            .expect("Failed to get punchthrough token for RT_SIGRETURN");
-        token
-            .execute()
-            .map(|_| ())
-            .map_err(|e| match e {
-                litebox::platform::PunchthroughError::Failure(errno) => errno,
-                _ => unimplemented!("Unsupported punchthrough error {:?}", e),
-            })
-            .expect("rt_sigreturn failed");
-        unreachable!("rt_sigreturn should not return");
-    }
 }
 
 const ROBUST_LIST_LIMIT: isize = 2048;
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -2225,6 +2225,7 @@ pub enum ContinueOperation {
     ResumeGuest,
     ExitThread(i32),
     ExitProcess(i32),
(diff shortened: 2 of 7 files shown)


jstarks:
I recommend reviewing this PR with whitespace changes hidden: <https://github.com/microsoft/litebox/pull/458/files?w=1>

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -340,75 +347,76 @@ impl litebox::shim::InitThread for NewThreadArgs {
...
+                task.sys_arch_prctl(ArchPrctlArg::SetFs(tls.as_usize()));
             }
         }
 
         if let Some(child_tid_ptr) = set_child_tid {
             // Set the child TID if requested
             let _ = unsafe { child_tid_ptr.write_at_offset(0, child_tid) };
         }
+
+        // Set the TLS for the platform itself
```
wdcui: Minor: "for the *shim* itself"?

wdcui:
@jaybosamiya-ms or @CvvT  should take a look before this PR is merged.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -278,31 +282,34 @@ fn wake_robust_list(
...
     }
 
-    task.process
-        .nr_threads
-        .fetch_sub(1, core::sync::atomic::Ordering::Relaxed);
-}
+    pub(crate) fn sys_exit(&self, _status: i32) {}
 
-pub(crate) fn sys_exit_group(_status: i32) {}
+    pub(crate) fn sys_exit_group(&self, _status: i32) {}
```
jaybosamiya-ms: Minor: might want to add a comment stating that the teardown is handled by the `Drop` implementation, which will trigger `prepare_for_exit`.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -340,75 +347,76 @@ impl litebox::shim::InitThread for NewThreadArgs {
 
         let child_tid = task.tid;
 
-        // Set the TLS for the platform itself
-        crate::SHIM_TLS.init(crate::LinuxShimTls { current_task: task });
-
```
jaybosamiya-ms, line 345: I am not fully sure why this bit moved down; nothing immediately stands out as being an issue to have moved it but just pointing out this change.

jaybosamiya-ms:
PS: difftastic (https://difftastic.wilfred.me.uk/) does a great job for making it easier to review PRs like this (I generally find it especially helpful whenever a PR would need `git diff -b` or `git diff -w`)

litebox_shim_linux/src/syscalls/process.rs
jstarks: Ah. I think I just didn't want to have to use `with_current_task` here--by moving this down, I could call the methods directly on `task` before moving ownership to TLS.

It's a bit subtle and could introduce bugs if someone expects TLS to be set. OTOH, every instance of `with_current_task` is another place to fix later, arguably, since I think we're now on a path to eliminate TLS from this shim. I'll just add a comment for now so that there's a breadcrumb if something breaks.

PR #461, opened 2025-10-31
Switch to guest fs/gs before calling rt_sigreturn

CvvT:
Now that we have signal support, we could add back the signal test removed by #380. This PR also fixes the syscall handler for `rt_sigreturn`. Because syscall `rt_sigreturn` does not return, we need to retore guest's fsbase or gs before calling it.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -582,8 +582,7 @@ syscall_callback:
     mov esp, fs:host_sp@ntpoff
     mov ebp, fs:host_bp@ntpoff
 
-    // Save guest gs and switch to host gs
-    mov fs:guest_gs@tpoff, gs
+    // Switch to host gs
     mov ax, fs
     mov gs, ax
 
@@ -686,8 +685,6 @@ host_sp:
 host_bp:
     .long 0
 guest_context_top:
-    .long 0
-guest_gs:
     .long 0
 in_guest:
     .long 0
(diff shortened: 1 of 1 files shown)


wdcui:
@jstarks can you take a look at this PR when you get a chance? Thanks!

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -582,7 +582,8 @@ syscall_callback:
     mov esp, fs:host_sp@ntpoff
     mov ebp, fs:host_bp@ntpoff
 
-    // Switch to host gs
+    // Save guest gs and switch to host gs
+    mov fs:guest_gs@tpoff, gs
```
jstarks: `gs` is already saved in the pt_regs structure itself, so don't save it again.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1076,19 +1079,28 @@ impl litebox::platform::PunchthroughToken for PunchthroughToken {
...
+                        SYSCALL_NUM = const syscalls::Sysno::rt_sigreturn as usize,
                         options(noreturn)
                     );
                 }
                 #[cfg(target_arch = "x86")]
                 unsafe {
                     core::arch::asm!(
                         "mov esp, {0}",
+                        "mov BYTE PTR gs:in_guest@ntpoff, 1",
+                        // Switch to guest gs
```
jstarks: Don't need to do anything special here--rt_sigreturn will restore gs for us.

jstarks:
Conceptually this is fine. I have a native implementation of rt_sigreturn coming soon, but this change is fine for now (and adds tests, which is good).

But the 32-bit path needs to be reverted first.

litebox_platform_linux_userland/src/lib.rs
CvvT: Good point! It's removed.

litebox_platform_linux_userland/src/lib.rs
CvvT: Removed as well.

PR #462, opened 2025-10-31
Store per-session data in OpteeShimTls

sangho2:
OP-TEE has per-session data like `tee_cryp_state` and `tee_obj` to maintain private information like crypto keys and cipher info within the kernel and let a TA reference them using `handles`. This PR stores them in `OpteeShimTls` instead of globally accessible memory. In addition, this PR introduces a session ID pool to allocate/maintain unique session IDs and fixes some bugs in test code.

### DIFF
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -642,15 +642,15 @@ impl SessionIdPool {
         }
     }
 
+    /// # Panics
+    /// Panics if session IDs are exhausted.
     pub fn allocate(&self) -> u32 {
         let mut inner = self.inner.lock();
         if let Some(session_id) = inner.pop_front() {
             session_id
         } else {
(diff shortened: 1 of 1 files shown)


litebox_shim_optee/src/lib.rs
```diff
@@ -625,3 +626,52 @@ litebox::shim_thread_local! {
...
+        }
+    }
+
+    pub fn allocate(&self) -> u32 {
+        let mut inner = self.inner.lock();
+        if let Some(session_id) = inner.pop_front() {
+            session_id
+        } else {
+            let mut session_id = self.next_session_id.fetch_add(1, SeqCst);
+            if session_id == Self::PTA_SESSION_ID {
```
wdcui: Can we really handle this case? Should we just panic when the session id reaches `PTA_SESSION_ID`?

sangho2: Agree with you. Revised.

PR #463, opened 2025-11-03
Support fallible memcpy for user mode

CvvT:
We need fallible `memcpy` in user space even for benign applications, because on the `exit` syscall path, we still need to access some user provided memory (e.g., in `wake_robust_list`) but all user memory might have been unmapped at that moment. To support this fallible `read`, platform needs to handle page faults/SIGSEGV properly.

### DIFF
--- a/litebox/src/platform/common_providers/mod.rs
+++ b/litebox/src/platform/common_providers/mod.rs
@@ -1,3 +1,7 @@
-//! Common platform provider
+//! Common Platform Provider Implementations
+//!
+//! This module contains reusable implementations of platform provider traits that
+//! can be shared across multiple platform implementations. These providers handle
+//! common functionality that doesn't require platform-specific customization.
 
 pub mod userspace_pointers;


jaybosamiya-ms:
I haven't yet done a full review of the code (since it is in draft) but the intention of `read_at_offset` is to be fallible already.  That's why it is set up as an `Option<CoW<..>>`.  Platforms that don't care about fallibility would take the borrowed approach (they can do trivial checks if they like, but basically they're returning borrowed values), whereas ones that care about fallibility would return the owned version after doing whatever checks and memcpy.  Basically, I don't think we should be adding a new `fallible_read_at_offset`; instead _every_ `read_at_offset` should be protected this way.

Ref: dbafc76

CvvT:
> I haven't yet done a full review of the code (since it is in draft) but the intention of `read_at_offset` is to be fallible already. That's why it is set up as an `Option<CoW<..>>`. Platforms that don't care about fallibility would take the borrowed approach (they can do trivial checks if they like, but basically they're returning borrowed values), whereas ones that care about fallibility would return the owned version after doing whatever checks and memcpy. Basically, I don't think we should be adding a new `fallible_read_at_offset`; instead _every_ `read_at_offset` should be protected this way.
> 
> Ref: [dbafc76](https://github.com/microsoft/litebox/commit/dbafc7633a653fc7319aefa3e50f710627c92874)

I was concerned about the overhead if every read performs `memcpy`, but I guess it's not that critical. I added new `UserConstPtr` and `UserMutPtr` to `LinuxUserland`, which is identical to the ones in `LinuxKernel` (for now; For kernel mode, we might need to do more stuffs there, e.g., enable/disable SMEP).

The current implementation relies on signal handling. An alternative solution is: 1) hold a lock on `Vmem`, 2) check if the given range is valid; 3) call given callback function on the slice, and 4) release the lock of `Vmem`. However, in the platform, we cannot access `PageManager` which is defined in the shim layer. Also, this solution only works for user mode, because in kernel mode step 3) may trigger page faults and page fault handler also requires lock on `Vmem`.

CvvT:
This fixes the crash we encountered while running our multi-threading test. However, I also found another unknow race issue while running it with i386 target.

It might crash at libc's function `__ctype__init` or the following assertion (somehow `mprotect` returned `ENOMEM`):

https://github.com/microsoft/litebox/blob/0cc46d55473c8d3034ab564531d535f8f534012c/litebox_platform_linux_userland/src/lib.rs#L1699-L1704

litebox/src/mm/exception_table.rs
```diff
@@ -0,0 +1,89 @@
...
+    .global __ex_table_end
+    __ex_table_end:
+"#
+);
+
+unsafe extern "C" {
+    /// Copies `size` bytes from `src` to `dst`. This function works with exception handling
+    /// and can recover from page fault.
+    /// Returns number of bytes that failed to copy.
+    pub fn __memcpy_fallible(dst: *mut u8, src: *const u8, size: usize) -> usize;
```
jaybosamiya-ms: Minor: rather than "works with exception handling", maybe something like "requires exception handling to be set up to correctly recover from exceptions inside it" or similar, and then link to `search_exception_tables` and tell users how they should set things up?

jaybosamiya-ms: Actually I think the better documentation is probably to tell users that the requirements are documented on `litebox::platform::common_providers::userspace_pointers` and over there tell users to set things up with the `search_exception_tables`?

litebox_platform_linux_userland/src/ptr.rs
```diff
@@ -0,0 +1,153 @@
+//! Userspace Pointer Abstraction
```
jaybosamiya-ms: We can probably move this file over to `litebox/src/platform/common_providers/userspace_pointers.rs` (in a separate PR, we can move the `trivial_providers` over to `common_providers`, but in this one, we can just introduce a new `common_providers` module that has the `userspace_pointers` module inside it).

I think it is generically helpful (kinda like how `Transparent*Pointer` has been helpful), and all of its dependencies exist inside `litebox` already _except_ the setup needed for exception handling.  We can handle that by adding a documentation comment to this module telling the users that they should catch any failed reads, look things up via `search_exception_tables` and fixup the program counter.  We can also note what happens if the user has not set this up (i.e., it then just behaves like a slightly more expensive `Transparent*Pointer`).

CvvT:
> This fixes the crash we encountered while running our multi-threading test. However, I also found another unknow race issue while running it with i386 target.
> 
> It might crash at libc's function `__ctype__init` or the following assertion (somehow `mprotect` returned `ENOMEM`):
> 
> https://github.com/microsoft/litebox/blob/0cc46d55473c8d3034ab564531d535f8f534012c/litebox_platform_linux_userland/src/lib.rs#L1699-L1704

This should be fixed by #470

litebox_platform_linux_userland/src/ptr.rs
CvvT: Make sense, thanks for the suggestion. I moved the implementation to litebox with doc comments.

litebox/src/mm/exception_table.rs
CvvT: Fixed the comment as suggested, thanks!

PR #464, opened 2025-11-04
litebox_shim_linux: move file descriptor tables into `Task`

jstarks:
Eliminate `file_descriptors()` and `raw_descriptor_store()`, moving their state into an `Arc` on `Task`. This eliminates some globals and sets us up for supporting `!CLONE_FILES` in the future when desired.

Also, move the file system instance into an `Arc` that's referenced by the `Task`. Update the pattern used to construct a Linux shim instance to ensure the file system is set before the `Task`.

There are a few other globals to eliminate, but cutting this PR off here to avoid it getting too big.

### DIFF
--- a/litebox_runner_linux_on_windows_userland/tests/common/mod.rs
+++ b/litebox_runner_linux_on_windows_userland/tests/common/mod.rs
@@ -46,7 +46,7 @@ impl TestLauncher {
             this.install_file(data, each);
         }
 
-        platform.register_shim(shim.entrypoints());
+        platform.register_shim(this.shim.entrypoints());
         this
     }
 


litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -144,7 +144,8 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
 
     let platform = Platform::new();
     litebox_platform_multiplex::set_platform(platform);
-    let litebox = litebox_shim_linux::init_process(platform.init_task());
+    let mut builder = litebox_shim_linux::ShimLauncher::new();
```
wdcui: Minor: it's a little confusing to see a `Launcher` returns a `builder`? Would it make sense to call it `shim` or `linux_shim` instead of `builder`?

litebox_runner_linux_on_windows_userland/tests/common/mod.rs
```diff
@@ -3,65 +3,87 @@
...
-    set_fs(litebox_shim_linux::default_fs(in_mem_fs, tar_ro_fs));
+impl TestLauncher {
+    pub fn init_platform(
+        tar_data: &'static [u8],
+        initial_dirs: &[&str],
+        initial_files: &[&str],
+    ) -> Self {
+        let platform = Platform::new();
+        litebox_platform_multiplex::set_platform(platform);
+        let launcher = litebox_shim_linux::ShimLauncher::new();
```
wdcui: See my comment above about `builder`. We should use a consistent name. :-)

litebox_runner_linux_on_windows_userland/tests/common/mod.rs
```diff
@@ -3,65 +3,87 @@
...
+            let data = std::fs::read(each).unwrap();
+            this.install_file(data, each);
+        }
 
-pub fn install_dir(path: &str) {
-    litebox_fs()
-        .mkdir(path, Mode::RWXU | Mode::RWXG | Mode::RWXO)
-        .expect("Failed to create directory");
-}
+        platform.register_shim(&litebox_shim_linux::LinuxShim);
```
wdcui: Why do we use `LinuxShim` instead of the `launcher` here?

litebox_shim_linux/src/lib.rs
```diff
@@ -1177,8 +1232,10 @@ struct Task {
     credentials: Arc<syscalls::process::Credentials>,
     /// Command name (usually the executable name, excluding the path)
     comm: Cell<[u8; litebox_common_linux::TASK_COMM_LEN]>,
-    /// Filesystem state.
+    /// Filesystem state. `RefCell` to support `unshare` in the future.
     fs: RefCell<Arc<syscalls::file::FsState>>,
+    /// File descriptors. `RefCell` to support `unshare` in the future.
+    files: RefCell<Arc<syscalls::file::FilesState>>,
```
wdcui, line 1248: What's the reason that `files` is not in `GlobalState`?

litebox_runner_linux_on_windows_userland/tests/common/mod.rs
jstarks: Updated the naming of everything.

litebox_runner_linux_on_windows_userland/tests/common/mod.rs
jstarks: Yeah, good idea, we can have a method to return this from the launcher.

litebox_shim_linux/src/lib.rs
jstarks: `files` is the set of file descriptors, which can differ for different tasks depending on the state of the `CLONE_FILES` flag passed to `clone` or `unshare`.

We don't support this yet (we need to implement `Clone` for the file descriptor tables first), but I wanted the data structures to reflect the eventual goal even they don't reflect the current capabilities.

wdcui: I see. I wasn't aware of the subtlties of `CLONE_FILES`. Thank you for the explanation, John.

litebox_runner_linux_on_windows_userland/tests/common/mod.rs
wdcui: Thank you for changing the variable names, John.

jaybosamiya-ms:
I re-triggered the CI on the latest commit just in case it is that spurious failure again, please feel free to merge when you're ready to merge.

PR #468, opened 2025-11-04
Change enablement of VTL1 for secondary processors

athvu:
This PR moves enabling VTL to the mshv_vsm_boot_aps() function and enable VTL based on the cpu_present_mask instead of num_possible_cpus. This also changes init_vtl_aps() to init_vtl_ap(), which only enables a single processor at a time.

### DIFF
--- a/dev_tests/src/ratchet.rs
+++ b/dev_tests/src/ratchet.rs
@@ -29,7 +29,7 @@ fn ratchet_globals() -> Result<()> {
             ("litebox/", 7),
             ("litebox_platform_linux_kernel/", 5),
             ("litebox_platform_linux_userland/", 5),
-            ("litebox_platform_lvbs/", 19),
+            ("litebox_platform_lvbs/", 20),
             ("litebox_platform_multiplex/", 1),
             ("litebox_platform_windows_userland/", 7),
             ("litebox_runner_linux_userland/", 1),


litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -113,25 +113,33 @@ pub fn mshv_vsm_enable_aps(cpu_present_mask_pfn: u64) -> Result<i64, Errno> {
...
-    }
 
-    // TODO: cpu_present_mask vs num_possible_cpus in kernel command line. which one should we use?
-    if let Ok(num_cores) = get_num_possible_cpus() {
-        debug_serial_println!("the number of possible cores: {num_cores}");
-        init_vtl_aps(num_cores).map_err(|_| Errno::EINVAL)?;
+        let present_cpus: Vec<u32> = decoded_mask
+            .iter()
+            .enumerate()
+            .filter_map(|(i, &present)| if present { Some(i as u32) } else { None })
```
praveen-pk: Just curious, do you really need the filter here? Seems like the cpumask is being copied from vtl0. Wouldn't it only have online cpus in the mask?

athvu:

litebox_platform_lvbs/src/mshv/vsm.rs
athvu: I think the filter was necessary because from VTL0 the cpumask is defined as
`typedef struct cpumask { DECLARE_BITMAP(bits, NR_CPUS); } cpumask_t;`
with each bit being 0 or 1 depending on if the CPU is present. Though I haven't ran into a case in which any of the bits are 0, it is still something I think we should account for

PR #469, opened 2025-11-04
No vdso on rewriter backend

CvvT:
Now that we have enough support for syscalls, the Linux runner now can run without `vdso`, thus we don't need the patch on `aux` vec to pass CI.

Another issue with `vdso` I found recently was that `vdso` on x86 has this function `kernel_vsyscall` that allow guest to call any syscalls, but our rewriter backend does not instrument the `vdso` section. To address it, we don't provide `vdso` address to guest if it runs with the rewriter backend. This ensures that all syscalls would go through LiteBox.

### DIFF
--- a/litebox_runner_linux_userland/src/lib.rs
+++ b/litebox_runner_linux_userland/src/lib.rs
@@ -227,7 +227,7 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
         });
     }
 
-    shim.set_load_filter(fixup_env_aux);
+    shim.set_load_filter(fixup_env);
     platform.register_shim(shim.entrypoints());
     match cli_args.interception_backend {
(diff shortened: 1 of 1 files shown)


litebox_runner_linux_userland/src/lib.rs
```diff
@@ -257,33 +256,8 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
 
 fn fixup_env_aux(
     envp: &mut Vec<alloc::ffi::CString>,
-    aux: &mut litebox_shim_linux::loader::auxv::AuxVec,
+    _aux: &mut litebox_shim_linux::loader::auxv::AuxVec,
```
wdcui: Why not remove `aux` completely?

CvvT: I thought it does not hurt to keep it, but I guess removing it is clearer. It's removed completely now.

PR #470, opened 2025-11-05
Fix setting new TLS

CvvT:
Finally figure out the root cause of the issue I mentioned in the [comment](https://github.com/microsoft/litebox/pull/463#issuecomment-3487385084), which is caused by incorrect handling of TLS when calling `sys_clone` with TLS (`*mut UserDesc`). The problem was that we set TLS for the new thread in the new thread's context instead of its parent's (in Linux, kernel can set TLS for another thread while user code cannot), which means after `sys_clone` returns and before the child runs, the parent may reset `*mut UserDesc`, resulting in the child setting the wrong TLS. This only affects x86, because on x64 we only need to set `fsbase` with the raw pointer.

To address it, we save a copy of `UserDesc` instead of the original pointer passed from user space. One caveat is that if we need to allocate a new entry for TLS (i.e., `UserDesc.entry_number == -1` for the new thread, we don't write the new entry number back to user space because the original pointer to `UserDesc` might be invalid as mentioned above. This shouldn't be a problem in practice. The entry number for a new thread is usually the same as the main thread, so no need for allocating a new one.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -1211,11 +1211,24 @@ impl Task {
             #[cfg(target_arch = "x86_64")]
             let desc = MutPtr::from_usize(addr);
             #[cfg(target_arch = "x86")]
-            let desc = unsafe {
-                MutPtr::<litebox_common_linux::UserDesc>::from_usize(addr).read_at_offset(0)
-            }
-            .ok_or(Errno::EFAULT)?
-            .into_owned();
+            let desc = {
+                let desc = unsafe {
+                    MutPtr::<litebox_common_linux::UserDesc>::from_usize(addr).read_at_offset(0)
+                }
+                .ok_or(Errno::EFAULT)?
+                .into_owned();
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/lib.rs
```diff
@@ -1186,7 +1207,16 @@ impl Task {
             ))
         };
         let tls = if clone_args.tls != 0 {
-            Some(MutPtr::from_usize(usize::try_from(clone_args.tls).unwrap()))
+            let addr = usize::try_from(clone_args.tls).unwrap();
+            #[cfg(target_arch = "x86_64")]
+            let desc = MutPtr::from_usize(addr);
+            #[cfg(target_arch = "x86")]
+            let desc = unsafe {
```
wdcui: Please add some comment as we discussed.

CvvT: Added, thanks!

wdcui: Looks good! Thank you, Weiteng.

PR #473, opened 2025-11-07
Use async functions for operations that can block

jstarks:
To allow the shim to abort indefinite waits (e.g., due to signals, thread termination, and time out), model abortable waits using async functions.

Add a simple executor that uses `RawMutex` to suspend and resume the current thread. Update most abortable waits to use this new infrastructure. The futex manager requires a bigger refactor to adopt this model.

### DIFF
--- a/litebox/src/pipes.rs
+++ b/litebox/src/pipes.rs
@@ -649,7 +649,7 @@ mod tests {
                     }
                     pipes.close(&prod).unwrap();
                     assert_eq!(i, data.len());
-                })
+                });
             });
 
             litebox.sync().new_executor().run(async {
@@ -697,7 +697,7 @@ mod tests {
                     }
                     pipes.close(&prod).unwrap();
                     assert_eq!(i, data.len());
-                })
+                });
             });
 
             litebox.sync().new_executor().run(async {


litebox/src/sync/mod.rs
```diff
@@ -10,6 +10,7 @@
 use crate::platform;
 
 mod condvar;
+mod executor;
```
jstarks, line 13: Not really sure this is the right place to put this module.

litebox_shim_linux/src/lib.rs
```diff
@@ -297,11 +301,11 @@ pub fn perform_network_interaction() -> litebox::net::PlatformInteractionReinvoc
     litebox_net().lock().perform_platform_interaction()
 }
 
-pub(crate) fn litebox_pipes<'a>() -> &'a litebox::sync::RwLock<Platform, Pipes<Platform>> {
-    static PIPES: OnceBox<litebox::sync::RwLock<Platform, Pipes<Platform>>> = OnceBox::new();
+pub(crate) fn litebox_pipes<'a>() -> &'a Pipes<Platform> {
```
jstarks, line 304: Interruptible waits (via async or not) are not really compatible with synchronous locks. Luckily it seems like `Pipes` doesn't need to be wrapped in a lock at all, so I was able to just remove it.

jstarks:
Going a different way.

PR #474, opened 2025-11-07
Refactor exit handling

jstarks:
Don't expose to the platform the distinction between thread and process exit. Instead, have the shim manage tearing down all threads for a process in response to the `exit_group` syscall. Update the userland runners to wait for all threads to exit before exiting the program.

Note that this will cause hangs at process exit until we implement code to interrupt threads running in user mode and waiting in syscalls.


jstarks:
Closing for now--this needs more infrastructure to avoid regressing the node tests.

PR #476, opened 2025-11-07
Add default implementation for remap_pages

CvvT:
Add a default implementation for `remap_pages` that emulates it via `allocate_pages` and `deallocate_pages`. This can be used by Windows as it does have native support for it.

### DIFF
--- a/litebox/src/platform/page_mgmt.rs
+++ b/litebox/src/platform/page_mgmt.rs
@@ -116,11 +116,21 @@ pub trait PageManagementProvider<const ALIGN: usize>: RawPointerProvider {
             })
             .expect("failed to update permissions on old range for copying");
         }
-        let old_ptr = <Self as RawPointerProvider>::RawConstPointer::from_usize(old_range.start);
-        unsafe {
-            new_ptr.write_slice_at_offset(0, &old_ptr.to_cow_slice(old_range.len()).unwrap())
+        // Copy in chunks of ALIGN bytes to handle very large memory regions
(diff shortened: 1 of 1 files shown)


litebox/src/platform/page_mgmt.rs
```diff
@@ -79,11 +81,66 @@ pub trait PageManagementProvider<const ALIGN: usize>: RawPointerProvider {
...
+            let old_ptr =
+                <Self as RawPointerProvider>::RawConstPointer::from_usize(old_range.start + offset);
+            unsafe {
+                new_ptr.write_slice_at_offset(
+                    isize::try_from(offset).unwrap(),
+                    &old_ptr.to_cow_slice(old_range.len()).unwrap(),
+                )
+            }
+            .unwrap();
+            offset += ALIGN;
```
wdcui, line 132: It's not clear to me where the number of bytes to be copied is defined?

jstarks:
This copying strategy is indeed what we do for WSL1, since even in the Windows kernel we don't have the functionality to move mappings.

litebox/src/platform/page_mgmt.rs
CvvT: `ALIGN` is a constant parameter that is specified in shim, e.g., https://github.com/microsoft/litebox/blob/837cfc3ac617b3870c9b3b24ad1f4fa064dca931/litebox_common_linux/src/mm.rs#L141-L145

PR #482, opened 2025-11-09
Add interrupt support to userland platforms

jstarks:
Add a mechanism to interrupt a thread that's running in the guest so that the shim can inject a signal or terminate the thread. On the Linux platform, use a signal to interrupt the target thread. On Windows, use `GetThreadContext`/`SetThreadContext`.

This is not actually hooked up to any shims yet in this PR, but I have tested this code somewhat via some private changes.

### DIFF
--- a/litebox/src/utils/mod.rs
+++ b/litebox/src/utils/mod.rs
@@ -120,10 +120,30 @@ impl<F: FnOnce()> Drop for Defer<F> {
 
 /// Returns an object that will run `f` when it goes out of scope.
 ///
+/// This is useful for cleaning up custom resources or for restoring invariants
+/// while calling user-provided closures that may panic.
+///
 /// Caution: the returned object must be bound to a variable to ensure the
 /// closure runs at scope end. In particular, binding to `_` will not work; use
 /// a named variable or a variable with a name starting with `_` (e.g.,
 /// `_defer`) instead.
+///
+/// # Example
+/// ```rust
+/// fn nest<R>(cell: &core::cell::Cell<usize>, f: impl FnOnce() -> R) -> R {
+///     cell.set(cell.get() + 1);
+///     let _defer = litebox::utils::defer(|| cell.set(cell.get() - 1));
+///     f()
+/// }
+/// let n = 0.into();
+/// nest(&n, || {
+///    assert_eq!(n.get(), 1);
+///    nest(&n, || {
+///       assert_eq!(n.get(), 2);
+///    });
+///    assert_eq!(n.get(), 1);
+/// });
+/// ```
 #[must_use = "Must be bound to a variable to defer until scope end; variable name cannot be just '_' (but '_foo' is fine)."]
 pub fn defer(f: impl FnOnce()) -> Defer<impl FnOnce()> {
     Defer(Some(f))


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1966,6 +2193,59 @@ unsafe fn next_signal_handler(
...
+    _signum: libc::c_int,
+    _info: &mut libc::siginfo_t,
+    context: &mut libc::ucontext_t,
+) {
+    // The interrupt signal can arrive in different contexts:
+    // 1. The thread is running in the host at the beginning of the syscall
+    //    handler. Do nothing--the syscall handler will handle the interrupt.
+    // 2. The thread is running in the host, with in_guest = 0. Just record that
+    //    an interrupt is pending; it will be checked next time we switch to the
+    //    guest.
```
CvvT, line 2245: why in the first case we don't need to record that an interrupt is pending like in the second case? I didn't follow how syscall handler handles the interrupt if it is not recorded.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -464,6 +465,9 @@ syscall_callback:
     mov     rsp, fs:host_sp@tpoff
     mov     rbp, fs:host_bp@tpoff
 
+    // Clear any pending interrupt.
+    mov     BYTE PTR fs:interrupt@tpoff, 0
+
```
CvvT: Why we clear `interrupt` here? Does it mean we ignore any interruptions occurred between `syscall_callback` and this line?

jstarks: The contract for `interrupt_thread` as written is that _one_ of the `EnterShim` methods will be called. The expectation is that if an interrupt happens at the same time as a syscall or an exception, it's OK to just call `syscall()` or `exception()`.

We can change that, but at least for the Linux shim there's not really a point.

litebox_platform_linux_userland/src/lib.rs
jstarks: Same explanation for this as in the other comment.

litebox/src/platform/mod.rs
```diff
@@ -66,6 +67,24 @@ pub trait ThreadProvider: RawPointerProvider {
         ctx: &Self::ExecutionContext,
         init_thread: alloc::boxed::Box<dyn crate::shim::InitThread>,
     ) -> Result<(), Self::ThreadSpawnError>;
+
+    /// Returns a handle to the current thread, which can be used to interrupt
+    /// it later.
+    ///
+    /// # Panics
+    /// Panics if called from a non-platform thread.
```
wdcui: In what contexts do we expect to see a non-platform thread? By "non-platform thread", I assume you meant threads that are not created/managed by LiteBox?

litebox/src/shim.rs
```diff
@@ -24,6 +24,10 @@ pub trait EnterShim: Send + Sync {
     /// FUTURE: use a single per-LiteBox type for all shims and platforms.
     type ContinueOperation;
 
+    /// Initialize a new thread. Must be called by the platform exactly once
+    /// before running the thread in the guest for the first time.
+    fn init(&self, ctx: &mut Self::ExecutionContext) -> Self::ContinueOperation;
```
wdcui, line 44: It might help to extend the comment here to explain how this function is different from `InitThread::init()`.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -207,6 +208,14 @@ impl litebox::platform::ThreadProvider for SnpLinuxKernel {
         })?;
         Ok(())
     }
+
+    fn current_thread(&self) -> Self::ThreadHandle {
```
wdcui, line 212: If we need to add these functions here, do we also need to add them in `litebox_platform_lvbs`?

litebox_runner_snp/src/main.rs
```diff
@@ -149,6 +151,10 @@ pub extern "C" fn sandbox_process_init(
             );
         }
     };
+    assert!(matches!(
```
wdcui, line 155: This change is added in this PR because it was missed previously?

litebox_shim_linux/src/lib.rs
```diff
@@ -65,6 +65,10 @@ impl litebox::shim::EnterShim for LinuxShimEntrypoints {
     type ExecutionContext = litebox_common_linux::PtRegs;
     type ContinueOperation = ContinueOperation;
 
+    fn init(&self, _ctx: &mut Self::ExecutionContext) -> Self::ContinueOperation {
```
wdcui, line 68: Is this a placeholder? Do we expect more code in this function in the future? Please see my question about this function and `InitThread::init()` above.

litebox_platform_linux_userland/src/lib.rs
wdcui: John, you commented for the PR that "Add a mechanism to interrupt a thread that's running in the guest so that the shim can inject a signal or terminate the thread". If we ignore the interrupt here and ask the shim to handle a syscall, how can the shim know if it's time to inject a signal or terminate the thread? Do we expect the syscall handler in the shim will check it?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -474,8 +478,20 @@ exception_callback:
     // Restore the stack and frame pointer.
     mov     rsp, fs:host_sp@tpoff
     mov     rbp, fs:host_bp@tpoff
+    // Clear any pending interrupt.
```
wdcui: Do we need to update `in_guest` here?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -639,39 +678,43 @@ unsafe extern "fastcall-unwind" fn syscall_handler_fast(ctx: &mut litebox_common
...
-            "popfq",
-            "pop rsp",
-            "jmp gs:scratch@tpoff", // jump to the guest
-            ctx = in(reg) ctx,
-            options(noreturn, nostack),
-        );
-    }
+    core::arch::naked_asm!(
+        "switch_to_guest_start:",
+        "mov BYTE PTR fs:in_guest@tpoff, 1",
```
wdcui, line 679: It will help understand why we set `in_guest` to 1 this early by adding some comment here. If we invoke the interrupt handler after this, we will be in the shim not the guest?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -774,25 +880,28 @@ impl RawMutex {
...
-            /* ignored */ None,
-        ) {
-            Ok(0) => Ok(UnblockedOrTimedOut::Unblocked),
-            Err(syscalls::Errno::EAGAIN) => Err(ImmediatelyWokenUp),
-            Err(syscalls::Errno::ETIMEDOUT) => Ok(UnblockedOrTimedOut::TimedOut),
-            Err(e) => {
-                panic!("Unexpected errno={e} for FUTEX_WAIT")
-            }
-            _ => unreachable!(),
+        loop {
```
wdcui, line 885: Is this a drive-by fix?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1725,31 +1900,23 @@ fn with_signal_alt_stack<R>(f: impl FnOnce() -> R) -> R {
...
 }
 
 #[cfg(target_arch = "x86_64")]
-unsafe extern "C" fn exception_signal_handler(
-    signum: libc::c_int,
-    info: &mut libc::siginfo_t,
-    context: &mut libc::ucontext_t,
-) {
-    let regs = unsafe {
+fn signal_handler_exit_guest(
```
wdcui, line 1937: Can we add a comment for this function? I can't tell based on its name what it's for and when it should be called. :)

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1824,55 +2040,15 @@ unsafe extern "C" fn exception_signal_handler(
...
     ] {
-        *reg = sigctx.gregs[sig_reg.reinterpret_as_unsigned() as usize]
+        *reg = context.uc_mcontext.gregs[sig_reg.reinterpret_as_unsigned() as usize]
             .reinterpret_as_unsigned()
             .truncate();
     }
     *orig_rax = *rax;
-
-    // Ensure that `run_thread` is linked in so that `exception_callback` is visible.
-    let _ = run_thread as usize;
```
wdcui, line 1834: We don't need this line to ensure `run_thread` is linked in?

litebox_platform_linux_userland/src/lib.rs
wdcui: It would help too if some comments are added for other new functions created below so it's clear what each function is for.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1966,6 +2193,59 @@ unsafe fn next_signal_handler(
...
+
+    let Some(regs) = signal_handler_exit_guest(context, true) else {
+        // Case 2: not in guest.
+        return;
+    };
+
+    // If the interrupt happened while returning to the guest, don't overwrite
+    // the saved context.
+    let in_switch_to_guest =
+        (switch_to_guest_start as usize..switch_to_guest_end as usize).contains(&ip);
```
wdcui, line 2282: Is it safe to assume `switch_to_guest_start < switch_to_guest_end`?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1966,6 +2193,59 @@ unsafe fn next_signal_handler(
...
+    if in_switch_to_guest {
+        // Case 3: in the middle of restoring guest context. Don't overwrite it.
```
wdcui, line 2284: This code doesn't use `in_guest`. Then why do we set `in_guest` very early in `switch_to_guest` before handling interrupt?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -450,10 +474,8 @@ unsafe extern "C-unwind" fn run_thread_inner(
     mov     QWORD PTR [rdx + {HOST_SP}], rsp
     mov     QWORD PTR [rdx + {HOST_BP}], rbp
 
-    // Switch to the guest context. When the guest issues a syscall, it will
-    // jump back into the middle of this function, at `syscall_callback`.
-    call {switch_to_guest}
-    ud2
+    call {init_handler}
```
wdcui, line 477: It's called `init_thread` in the Linux userland code. Maybe we should keep them consistent?

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
jstarks: Currently, the LVBS platform does not implement `ThreadProvider`.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -562,36 +590,53 @@ exception_callback:
...
-            };
-            #[link(name = "ntdll")]
-            unsafe extern "system" {
-                fn NtContinue(
-                    ctx: *const CONTEXT,
-                    raise_alert: u8,
-                ) -> windows_sys::Win32::Foundation::NTSTATUS;
-            }
-            let win_ctx = CONTEXT {
+            win_ctx.write(CONTEXT {
```
wdcui, line 639: Sangho added xsave/xrstor in LVBS. Do we need to worry about it in our userland platforms?

litebox/src/platform/mod.rs
jstarks: The platform could spawn its own threads, not requested by the shim, e.g., to run networking tasks or something like that. And the shim could allow calls from such threads--for example, you might call the shim on a separate thread to send a SIGTERM to the initial process.

I'll clarify the comment.

litebox_runner_snp/src/main.rs
jstarks: No, the `init` entry point is new. For now, I am asserting that it always succeeds because I don't know enough about the SNP code to do the right thing in case it fails--I'll add a comment.

litebox_shim_linux/src/lib.rs
jstarks: Yes, this is a placeholder for now. In the future, this will attach the thread to the process, which might fail if the process was terminated or an `execve` occurred while the thread was in the middle of spawning.

(The other model would be to wait in the `clone` path for the new thread to start up so that the process attach cannot fail. But I like this model better, it's more scalable.)

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -678,6 +722,258 @@ impl litebox::platform::ThreadProvider for WindowsUserland {
...
+        unsafe {
+            windows_sys::Win32::System::Threading::SuspendThread(inner.handle.as_raw_handle());
+        }
+        let _resume_guard = litebox::utils::defer(|| unsafe {
+            windows_sys::Win32::System::Threading::ResumeThread(inner.handle.as_raw_handle());
+        });
+
+        // SAFETY: The target TLS state is accessible while the thread is
+        // suspended.
+        let tls = unsafe { &*inner.tls.0 };
```
wdcui: Minor: it might help if we rename it to `target_tls`.

litebox_platform_linux_userland/src/lib.rs
jstarks: Sort of. The interrupt signal can potentially now arrive while we're in this path, depending on shim behavior, so we need to ignore the EINTR and continue. Although I need to take a closer look at the timeout behavior in this case...

litebox_platform_linux_userland/src/lib.rs
jstarks: I think this just moved to a different function.

litebox_platform_linux_userland/src/lib.rs
jstarks: Yes. We can add an assertion, but I don't think it will really buy us anything.

litebox_platform_linux_userland/src/lib.rs
jstarks: Yes--the shim will need to check for pending exits/signals just before returning to the guest from all entry points.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -1414,9 +1720,28 @@ unsafe extern "C-unwind" fn exception_handler(
...
+    let &shim = SHIM.get().expect("Should have run `register_shim` by now");
+    let op = if clear_interrupt() {
+        shim.interrupt(ctx)
+    } else {
+        ContinueOperation::ResumeGuest
+    };
+    continue_operation(op, ctx);
+}
+
+fn clear_interrupt() -> bool {
```
wdcui: This `clear_interrupt` function is not in the Linux userland platform. Would it make sense to create a similar function there?

litebox_platform_windows_userland/src/lib.rs
jstarks: We do. But we don't have a place to store this state right now. We need to replace `PtRegs` with a more comprehensive context structure that includes xsave state.

I'll open an issue.

litebox_platform_linux_userland/src/lib.rs
jstarks: No, this was done by the signal handler.

litebox_platform_linux_userland/src/lib.rs
wdcui: Just to confirm: the shim doesn't have such code yet, right?

litebox/src/platform/mod.rs
wdcui: If the thread is spawned by the platform, we probably shouldn't call it a non-platform thread?

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
wdcui: Got it.

litebox_runner_snp/src/main.rs
wdcui: Got it. Thanks.

litebox_shim_linux/src/lib.rs
wdcui: When this function is fully implemented, do we still need `InitThread::init()` for the clone path?

litebox_platform_windows_userland/src/lib.rs
jstarks: Good idea. I'm going to clean this up even further and match Windows and Linux.

litebox_shim_linux/src/lib.rs
jstarks: We do, in order to construct the `Task` object and put it in TLS. I'll clarify this in comments.

litebox_platform_linux_userland/src/lib.rs
jstarks: Correct, not in this PR.

litebox/src/platform/mod.rs
jstarks: I clarified the contract here.

litebox/src/shim.rs
jstarks: Done.

litebox_platform_linux_userland/src/lib.rs
jstarks: `in_guest` is examined and cleared above by the call to `signal_handler_exit_guest`. Added comments to clarify this.

litebox_platform_windows_userland/src/lib.rs
jstarks: Renamed to `init_handler` for consistency with Windows and with the other `*_handler` functions.

litebox/src/platform/mod.rs
wdcui: Thanks!

litebox/src/shim.rs
wdcui: Thanks!

litebox_platform_linux_userland/src/lib.rs
wdcui: Thanks!

litebox_platform_windows_userland/src/lib.rs
wdcui: Thank you, John!

litebox/src/utils/mod.rs
```diff
@@ -104,3 +106,20 @@ impl_reinterpret! { u64, i64 }
...
+            f();
+        }
+    }
+}
+
+/// Returns an object that will run `f` when it goes out of scope.
+#[must_use]
+pub fn defer(f: impl FnOnce()) -> Defer<impl FnOnce()> {
+    Defer(Some(f))
+}
```
jaybosamiya-ms, line 150: It is probably also helpful to explicitly tell people to _not_ use `let _ = defer(...);` because this is one of the _extremely few_ handful of cases where names of variables matter in Rust.  This is due to drop order.

Specifically, a concrete example of where this counter-intuitive behavior shows up:
```rs
fn main() {
    let _ = defer(|| println!("a"));
    let _ = defer(|| println!("b"));
    let _c = defer(|| println!("c"));
    let _d = defer(|| println!("d"));
    println!("---------------------");
}
```
outputs:
```
a
b
---------------------
d
c
```

Something like
```
#[must_use = "Must be bound to a variable to defer until scope end; variable name cannot be just '_' (but '_foo' is fine)."]
```
and adding similar to the doc-comment would likely be fine.

jaybosamiya-ms: Sidenote: this is why in projects where I've introduced something like `defer` I usually do it via a macro with a `doc(hidden)` fn that the macro invokes. That way, we don't even need to actually worry about usage.

```rs
#[macro_export]
macro_rules! defer {
    ($e:expr) => {
        let _named = $crate::utils::defer(|| $e);
    };
}
```

would just make the above code into:
```rs
fn main() {
    defer!(println!("a"));
    defer!(println!("b"));
    defer!(println!("c"));
    defer!(println!("d"));
    println!("---------------------");
}
```
such that the user does not need to remember the var naming thing.

The above has the intuitive output of:
```
---------------------
d
c
b
a
```

jstarks: Good point!

I think I prefer your former suggestion. The macro approach has its appeal, but it's also strictly less powerful since it enforces lexical scoping--I can't `drop` the defer guard early, initialize it late, or pass it to another function, and I can't add something like `Defer::cancel` in the future.

PR #487, opened 2025-11-11
Use polling for socket operations

CvvT:
1. Refactor `wait_or_timeout` so that it calls `try_op` right after `register_observer`. This is useful when `check` is unnecessary and always returns `True`. Also, it now takes a custom `register_observer` function as its argument.
2. `sendto`/`receive`/`accept` now use polling instead of busy loop
3. make `connect` blocking if it is a blocking socket


wdcui:
@jstarks This PR probably causes merge conflict with your async PR. Can you please take a look at this PR and let Weiteng know if it's okay to merge? Thanks!

jstarks:
This is fine, we can clean this up later once we are in agreement about the interruptible wait approach (async vs `WaitState`).

PR #488, opened 2025-11-12
Fix fixed address memory map race conditions

jstarks:
When the page manager is asked to perform an allocation at a fixed address, it first unmaps any old allocations before putting the new one in place. If the platform allocates memory on another thread between these two steps, then the desired address may be reused, causing memory corruption.

Fix this by telling the platform whether to expect an existing mapping to be present when allocating pages, and rely on the platform to overwrite those ranges atomically--don't unmap (and thereby release the VA ranges of) the existing mappings first.

### DIFF
--- a/litebox/src/mm/linux.rs
+++ b/litebox/src/mm/linux.rs
@@ -590,32 +590,20 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
                 unimplemented!("file-backed mapping expansion is not supported yet");
             }
             let range = PageRange::new(range.end, new_end).unwrap();
-            // Although we have checked the overlap above, there is still a small chance that this range
-            // collides with memory allocated to global allocator when LiteBox is used in user mode. This
-            // is because page manager is not aware of the memory allocated to global allocator.
-            // Try to insert the mapping with `fixed_address = false` to see if it is free first.
-            let ptr = match unsafe {
-                self.insert_mapping(range, *cur_vma, false, FixedAddressBehavior::Hint)
+            // Try to extend the mapping. Although we checked that there are no
+            // litebox mappings in this range, this may fail if there are
+            // platform mappings in the way.
+            match unsafe {
+                self.insert_mapping(range, *cur_vma, false, FixedAddressBehavior::NoReplace)
             } {
-                Ok(p) => p,
+                Ok(_) => {}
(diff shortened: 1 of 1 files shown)


litebox/src/mm/linux.rs
```diff
@@ -566,10 +594,18 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
             // collides with memory allocated to global allocator when LiteBox is used in user mode. This
             // is because page manager is not aware of the memory allocated to global allocator.
             // Try to insert the mapping with `fixed_address = false` to see if it is free first.
```
CvvT: The comment needs to be updated.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -1427,17 +1131,42 @@ impl<const ALIGN: usize> litebox::platform::PageManagementProvider<ALIGN> for Wi
...
+            {
+                return Err(AllocationError::AddressInUse);
             } else {
                 process_memory_range_by_regions(
                     suggested_range,
                     |r, state| -> Result<bool, std::convert::Infallible> {
                         let ok = match state {
                             // In case the region is already reserved, we just need to commit it.
-                            Win32_Memory::MEM_RESERVE => {
+                            // In case the region is already committed, decommit and recommit it.
```
CvvT, line 1148: Does this `decommit` and then `recommit` could potentially result in another race issue? Current global allocator used by Windows userland performs `MEM_RESERVE | MEM_COMMIT` so it won't be a problem. Not sure if we switch to std's allocator, do we need to worry about it?

jstarks: I don't think it's a problem--those reserved VAs are owned by us.

There are other problems, though, because even while we're looping over the mappings, the mappings could change. I think ultimately we need to change Windows to use placeholders <https://devblogs.microsoft.com/oldnewthing/20240201-00/?p=109346>--this will allow us to separate VA reservation from mapping/allocation.

litebox/src/mm/linux.rs
jstarks: Ah, and that makes me realize that we can request "no replace" behavior here.

litebox_platform_windows_userland/src/lib.rs
CvvT: Got it, thanks! Tracked the issue in #489

PR #491, opened 2025-11-13
LiteBox kernel platform and machine runner

sangho2:
This PR covers an initial effort to have a minimal LiteBox kernel platform and a test runner to use the kernel platform on top of a virtual machine (i.e., QEMU, KVM) to run a hello-world OP-TEE TA.


wdcui:
@jstarks @CvvT I think we should merge this minimal kernel platform into LiteBox. Can you please help take a look at this PR? Thanks!

wdcui:
@sangho2 would it make sense to move the changes to the lvbs platform to a separate PR?

CvvT:
> @jstarks @CvvT I think we should merge this minimal kernel platform into LiteBox. Can you please help take a look at this PR? Thanks!

Why we need to merge it into LiteBox? I thought it should be a separate crate for common kernel functionality like `litebox_common_linux`.

wdcui:
> > @jstarks @CvvT I think we should merge this minimal kernel platform into LiteBox. Can you please help take a look at this PR? Thanks!
> 
> 
> 
> Why we need to merge it into LiteBox? I thought it should be a separate crate for common kernel functionality like `litebox_common_linux`.

Sorry. I meant to merge it into the LiteBox repo. :-)

sangho2:
> @sangho2 would it make sense to move the changes to the lvbs platform to a separate PR?

The are simple bug fixes but I agree that a separate PR makes more sense.

litebox_platform_kernel/src/arch/x86/interrupts.rs
```diff
@@ -0,0 +1,84 @@
...
+fn idt() -> &'static InterruptDescriptorTable {
+    static IDT_ONCE: Once<InterruptDescriptorTable> = Once::new();
+    IDT_ONCE.call_once(|| {
+        let mut idt = InterruptDescriptorTable::new();
+        idt.divide_error.set_handler_fn(divide_error_handler);
+        idt.breakpoint.set_handler_fn(breakpoint_handler);
+        unsafe {
+            // Rust no longer allows a function with the custom ABI to have a return type.
+            // Unfortunately, the `x86_64` crate has not caught up this change.
+            // the below is a workaround mentioned in [link](https://github.com/rust-lang/rust/issues/143072).
```
wdcui: Minor: "the below" => "below"?

litebox_platform_kernel/src/arch/x86/ioport.rs
```diff
@@ -0,0 +1,178 @@
+//! I/O Port-based serial communication
+
+use core::{arch::asm, fmt};
+use spin::{Mutex, Once};
+
+const COM_PORT_1: u16 = 0x3F8;
+// const COM_PORT_2: u16 = 0x2F8;
```
wdcui: Remove the commented out code?

litebox_platform_kernel/src/lib.rs
```diff
@@ -0,0 +1,588 @@
...
+        }))
+    }
+
+    pub fn init(&self, cpu_mhz: u64) {
+        CPU_MHZ.store(cpu_mhz, core::sync::atomic::Ordering::Relaxed);
+    }
+
+    /// Create a new page table for user space. Currently, it maps the entire kernel memory for
+    /// proper operations (e.g., syscall handling). We should consider implementing
+    /// partial mapping to mitigate side-channel attacks and shallow copying to get rid of redudant
```
wdcui: typo" redudant"

litebox_platform_kernel/src/lib.rs
```diff
@@ -0,0 +1,588 @@
...
+    /// page table data structures for kernel space.
+    #[allow(clippy::unused_self)]
+    pub(crate) fn new_user_page_table(&self) -> mm::PageTable<PAGE_SIZE> {
+        // let pt = unsafe { mm::PageTable::new_top_level() };
```
wdcui: Remove code that's commented out?

litebox_platform_kernel/src/lib.rs
```diff
@@ -0,0 +1,588 @@
...
+    }
+}
+
+impl RawMutex {
+    fn block_or_maybe_timeout(
+        &self,
+        _val: u32,
+        _timeout: Option<core::time::Duration>,
+    ) -> Result<UnblockedOrTimedOut, ImmediatelyWokenUp> {
+        unimplemented!()
```
wdcui, line 201: The kernel works without a RawMutex implementation?

litebox_platform_kernel/src/lib.rs
```diff
@@ -0,0 +1,588 @@
...
+    ) -> Result<(), litebox::platform::page_mgmt::PermissionUpdateError> {
+        let range = PageRange::new(range.start, range.end)
+            .ok_or(litebox::platform::page_mgmt::PermissionUpdateError::Unaligned)?;
+        let new_flags =
+            litebox::mm::linux::VmFlags::from_bits(new_permissions.bits().into()).unwrap();
+        unsafe { self.page_table.mprotect_pages(range, new_flags) }
+    }
+
+    fn reserved_pages(&self) -> impl Iterator<Item = &core::ops::Range<usize>> {
+        core::iter::empty()
```
wdcui, line 365: Do we need to add a TODO here?

litebox_platform_kernel/src/per_cpu_variables.rs
```diff
@@ -0,0 +1,253 @@
...
+        }
+    }
+}
+
+// Allocate per-CPU variables in heap for all possible cores. We expect that the BSP will call
+// this function to allocate per-CPU variables for other APs because our per-CPU variables are
+// huge such that each AP without a proper stack cannot allocate its own per-CPU variables.
+// # Panics
+// Panics if the number of possible CPUs exceeds `MAX_CORES`
+// pub fn allocate_per_cpu_variables() {
```
wdcui, line 211: If we don't need this function, maybe we should remove it?

litebox_platform_kernel/src/ptr.rs
```diff
@@ -0,0 +1,210 @@
...
+    count: isize,
+) -> Option<alloc::borrow::Cow<'a, T>> {
+    if ptr.is_null() {
+        return None;
+    }
+    if ptr.is_aligned() {
+        Some(alloc::borrow::Cow::Borrowed(unsafe { &*ptr.offset(count) }))
+    } else {
+        // TODO: consider whether we should use `litebox_platform_linux_kernel`'s `memcpy_fallible`.
+        // `litebox_platform` currently preallocates all memory, so there would be no page fault.
```
wdcui: Do you mean `litebox_platform_kernel`?

litebox_shim_optee/Cargo.toml
```diff
@@ -33,6 +33,7 @@ elf_loader = { version = "0.12.0", default-features = false, features = ["rel"]
 default = ["platform_lvbs"]
 platform_linux_userland = ["litebox_platform_multiplex/platform_linux_userland_with_optee_syscall"]
 platform_lvbs = ["litebox_platform_multiplex/platform_lvbs_with_optee_syscall"]
+platform_kernel = ["litebox_platform_multiplex/platform_kernel_with_optee_syscall"]
```
wdcui: It doesn't look like that we need this feature definition?

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -147,7 +147,8 @@ impl ElfLoaderMmap {
             litebox_common_linux::MapFlags::from_bits(
                 flags.bits() | MapFlags::MAP_ANONYMOUS.bits(),
             )
-            .expect("unsupported flags"),
+            .expect("unsupported flags")
```
wdcui: maybe you can combine the bug fixes in `litebox_shim_optee` with `litebox_platform_lvbs` into a single separate PR?

wdcui (commented):
Thank you for creating this minimal kernel platform, Sangho. I left some minor comments. Please wait for comments from John and Weiteng.

litebox_platform_kernel/src/lib.rs
sangho2: done.

litebox_platform_kernel/src/lib.rs
sangho2: For now, the kernel (LVBS one specifically) uses spin locks if synchronization is needed. We would be able to implement simple RawMutex based on it.

litebox_shim_optee/src/loader/elf.rs
sangho2: some bug fixes are separated.

CvvT (commented):
~~A lot of the code seem to be duplicated from other platforms, which make this PR difficult to review. I can quickly draft a PR to move some common code into a separate crate.~~ Will try to clean up the code later.

litebox_platform_kernel/src/syscall_entry.rs
```diff
@@ -0,0 +1,241 @@
...
+        "push r8",
+        "push r10",
+        "push rdx",
+        "push rsi",
+        "push rdi",
+        "mov rdi, rax",
+        "mov rsi, rsp",
+        "and rsp, {stack_alignment}",
+        "call {syscall_entry}",
+        "add rsp, {register_space}",
```
CvvT: why we skip the first `NUM_REGISTERS_TO_POP` registers?

CvvT: `rsp` may not point to `SyscallContextRaw` after the alignment. You may need to save and restore it after `call`.

litebox_platform_kernel/src/syscall_entry.rs
```diff
@@ -0,0 +1,241 @@
...
+                );
+            }
+        }
+    }
+}
+
+#[unsafe(naked)]
+unsafe extern "C" fn syscall_entry_wrapper() {
+    naked_asm!(
+        "swapgs",
```
CvvT: We may also need to swap stack?

litebox_platform_kernel/src/lib.rs
```diff
@@ -0,0 +1,603 @@
...
+    .quad 0
+.globl guest_fsbase
+guest_fsbase:
+    .quad 0
+in_guest:
+    .byte 0
+.globl interrupt
+interrupt:
+    .byte 0
+    "
```
CvvT: For kernel, these may not be useful?

litebox_platform_kernel/src/lib.rs
```diff
@@ -0,0 +1,603 @@
...
+            "mov cr3, rax",
+            // Switch to user mode
+            "mov rax, {user_ds}",
+            "push rax",
+            "push gs:scratch2@tpoff", // user rsp
+            "mov rax, {rflags}",
+            "push rax",
+            "mov rax, {user_cs}",
+            "push rax",
+            "push gs:scratch@tpoff", // jump to the guest
```
CvvT: Why we need to do it again? `PtRegs` should have these already.

```
struct PtRegs {
    ...
    /* Return frame for iretq */
    pub rip: usize,
    pub cs: usize,
    pub eflags: usize,
    pub rsp: usize,
    pub ss: usize,
    ...
}
```

litebox_platform_kernel/src/syscall_entry.rs
jstarks: Yes, definitely.

`syscall` is pretty tricky to get right in the general case (e.g., what happens if you get an NMI before you swap GS? Does your NMI handler look at GS?). A simpler approach, which we'll take in Coconut initially, is to not enable the syscall instruction, so that a `syscall` causes a `#UD` exception. Then, in the `#UD` handler, you can read two bytes at RIP, and if it's a syscall instruction, you can emulate it. This is slower than `syscall` but is easier to reason about.

litebox_platform_kernel/src/syscall_entry.rs
```diff
@@ -0,0 +1,241 @@
...
+        "mov rdi, rax",
+        "mov rsi, rsp",
+        "and rsp, {stack_alignment}",
+        "call {syscall_entry}",
+        "add rsp, {register_space}",
+        "pop rcx",
+        "pop r11",
+        "pop rbp",
+        "swapgs",
+        "sysretq",
```
jstarks: Similar to the syscall side of things, I suggest we start with `iret` here. I like the way Linux handles this--they treat `sysret` as an optimization for switching to user mode when `rcx` == `rip` and `r11` == `rflags`. It doesn't matter if they entered from an interrupt or exception or `syscall`, just that the context matches what would happen if they entered via a `syscall`.

litebox_platform_kernel/src/ptr.rs
```diff
@@ -0,0 +1,210 @@
+//! Userspace Pointer Abstraction
+//! additional sanity checks and extra mode switches (e.g., SMAP/SMEP)
+
+use litebox::platform::{RawConstPointer, RawMutPointer};
+
+/// Represent a user space pointer to a read-only object
+#[repr(C)]
+#[derive(Clone)]
+pub struct UserConstPtr<T> {
```
jstarks, line 9: Can we use the existing `common_providers` implementation right away instead of cloning another copy of this?

sangho2:
Draft for now. I'll enhance it while transferring portions of it to the LVBS platform.

litebox_platform_kernel/src/syscall_entry.rs
sangho2: fixed.

sangho2: fixed.

litebox_platform_kernel/src/lib.rs
sangho2: removed

litebox_platform_kernel/src/lib.rs
sangho2: removed and now using stuffs in PtRegs.

litebox_platform_kernel/src/syscall_entry.rs
sangho2: Now we are using the kernel stack. Disabling `syscall` sounds interesting.

litebox_platform_kernel/src/syscall_entry.rs
sangho2: now we are using `iretq` instead of `sysretq`.

sangho2:
Overall Progress/Plan:
- I've updated this PR to catch up some recent changes related to `run_thread`.
- I'll port these changes to the LVBS platform.
- Once the LVBS platform is ready (OP-TEE and syscall supports), I'll work on this PR again to clean it up (and hopefully support multi cores and linux shim).

PR #492, opened 2025-11-14
Support secure kernel relocation

athvu:
Add support for loading the secure kernel at any memory location.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,7 @@ jobs:
       - run: |
           ./.github/tools/github_actions_run_cargo build
           ./.github/tools/github_actions_run_cargo build -Z build-std-features=compiler-builtins-mem -Z build-std=core,alloc --manifest-path=litebox_runner_lvbs/Cargo.toml --target litebox_runner_lvbs/x86_64_vtl1.json
-      - run: ./.github/tools/github_actions_run_cargo nextest
+      - run: ./.github/tools/github_actions_run_cargo nextest -p litebox_platform_lvbs
       - run: |
           ./.github/tools/github_actions_run_cargo test --doc
           # We need to run `cargo test --doc` separately because doc tests


litebox_runner_lvbs/x86_64_vtl1.json
```diff
@@ -13,14 +13,17 @@
...
 		],
 		"ld.lld": [
 			"--script=litebox_runner_lvbs/x86_64_vtl1.ld",
 			"--gc-sections",
-			"-no-pie"
+			"--static",
+			"--pie"
 		]
 	},
+	"relocation-model": "pic",
```
jstarks: I wonder if you'd get better codegen with "pie" instead of "pic" here, or if rustc is smart enough to know you're building an executable.

litebox_runner_lvbs/src/main.rs
```diff
@@ -6,12 +6,118 @@ use core::arch::asm;
...
+    }
+
+    // Get link-time address (where linker EXPECTED us to be)
+    // SAFETY: This is also position-independent - it's just the address
+    // the linker embedded in the relocation entries themselves
+
+    const EXPECTED_BASE: u64 = 0x0;
+    let expected_base = EXPECTED_BASE;
+
+    let offset = actual_base.wrapping_sub(expected_base);
```
sangho2: Do we need this `EXPECTED_BASE`? Is it a placeholder?

litebox_runner_lvbs/src/main.rs
```diff
@@ -6,12 +6,118 @@ use core::arch::asm;
...
+        let r_type = rela.info & 0xffffffff;
+
+        // Only handle R_X86_64_RELATIVE relocations
+        if r_type == R_X86_64_RELATIVE {
+            // Calculate target address: original offset + load offset
+            // SAFETY: Target address is valid after offset adjustment
+            let target = (rela.offset.wrapping_add(offset)) as *mut u64;
+            unsafe {
+                target.write_volatile((rela.addend as u64).wrapping_add(offset));
+            }
```
sangho2: Minor: `offset.wrapping_add(...)` could be more intuitive.

Do we need to consider negative `addend`?

litebox_runner_lvbs/x86_64_vtl1.ld
```diff
@@ -31,11 +31,25 @@ SECTIONS
         _bss_end = .;
     }
 
+    /* Relocation section for position-independent execution */
+    . = ALIGN(0x200000);
```
sangho2: this might be redundant.

litebox_runner_lvbs/src/main.rs
```diff
@@ -6,12 +6,118 @@ use core::arch::asm;
...
+}
+
+const R_X86_64_RELATIVE: u64 = 8;
+
+/// Apply ELF relocations to support position-independent execution.
+/// This code has NO dependency on absolute addresses - uses only RIP-relative addressing.
+///
+/// # Safety
+/// - Must be called before any absolute addresses are accessed
+/// - Must be called exactly once at boot
```
sangho2, line 29: We might want to strongly clarify or emphasize this (using something like `Once` here is tricky due to relocation...)

jaybosamiya-ms:
@athvu you should respond to https://github.com/microsoft/litebox/pull/492#issuecomment-3531056907 or https://github.com/microsoft/litebox/pull/468#issuecomment-3487872236 (you only need to do this once, and then it will stop pinging you on each PR).   Please see https://github.com/microsoft/litebox/pull/211#issuecomment-3085786586 for an example response.

litebox_runner_lvbs/x86_64_vtl1.json
athvu: Changed to pie. Thank you!

PR #493, opened 2025-11-14
Add infrastructure for interruptible waits

jstarks:
To support interrupted and terminated threads that are waiting on guest- or externally-triggered events, add infrastucture for performing interruptible waits. Each thread stores a `WaitState`, which can be used to generate `WaitContext` objects to wait on a condition, plus `Waker` objects that can be used to cause a thread to unblock to reevaluate the condition.

Also add a `ThreadHandle` object to allow interrupting a thread whether it is waiting or running in the guest.

Update all IO-related code to use this new wait infrastructure. Fix the networking code to provide a `check` function when waiting to avoid spinning in the new model.

Update `Instant` to support more operations (comparisons, copy/clone, adding durations). This makes it easier to work with deadlines/absolute timeouts.

Futex has yet to be updated since it must be rewritten to work with a wait-on-condition style of waiting, and that should be reviewed separately.

### DIFF
--- a/litebox/src/event/polling.rs
+++ b/litebox/src/event/polling.rs
@@ -29,7 +29,7 @@ pub enum TryOpError<E> {
     #[error("operation should be retried")]
     TryAgain,
     #[error("wait error")]
-    WaitError(WaitError),
+    WaitError(#[source] WaitError),
     #[error(transparent)]
     Other(E),
 }
@@ -40,9 +40,9 @@ impl<Platform: RawSyncPrimitivesProvider + TimeProvider> WaitContext<'_, Platfor
     ///
     /// If `nonblock` is true, returns `TryAgain` instead of waiting.
     ///
-    /// If `try_op` returns `TryAgain`, the thread will be woken try again when
-    /// the observer, registered via the call to `register_observer`, is called
-    /// with events that match the given `events` filter (or an event in
+    /// If `try_op` returns `TryAgain`, the thread will be woken to try again
+    /// when the observer, registered via the call to `register_observer`, is
+    /// called with events that match the given `events` filter (or an event in
     /// `Events::ALWAYS_POLLED`).
     pub fn wait_on_events<R, E>(
         &self,
@@ -95,9 +95,10 @@ impl<Platform: RawSyncPrimitivesProvider + TimeProvider> Pollee<Platform> {
     ///
     /// If `nonblock` is true, returns `TryAgain` instead of waiting.
     ///
-    /// If `try_op` returns `TryAgain`, the thread will be woken try again when
-    /// [`notify_observers`](Self::notify_observers) is called with events that
-    /// match the given `events` filter (or an event in `Events::ALWAYS_POLLED`).
+    /// If `try_op` returns `TryAgain`, the thread will be woken to try again
+    /// when [`notify_observers`](Self::notify_observers) is called with events
+    /// that match the given `events` filter (or an event in
+    /// `Events::ALWAYS_POLLED`).
     pub fn wait<R, E>(
         &self,
         cx: &WaitContext<'_, Platform>,
--- a/litebox/src/event/wait.rs
+++ b/litebox/src/event/wait.rs
@@ -324,27 +324,16 @@ impl<'a, Platform: RawSyncPrimitivesProvider + TimeProvider> WaitContext<'a, Pla
 
     /// Returns a new context that has a deadline after the given duration.
     ///
-    /// If the existing context already has an earlier deadline, then this just
-    /// clones the context.
-    #[must_use]
-    pub fn with_timeout(&self, timeout: core::time::Duration) -> Self {
-        // If this overflows, treat that as no deadline.
-        if let Some(deadline) = self.waker.0.platform.now().checked_add(timeout) {
-            self.with_deadline(deadline)
-        } else {
-            Self { ..*self }
-        }
-    }
-
-    /// Returns a new context that has a deadline after the given duration, if
-    /// one is provided.
-    ///
     /// If the existing context already has an earlier deadline or if no timeout
     /// is provided, then this just clones the context.
     #[must_use]
-    pub fn with_maybe_timeout(&self, timeout: Option<core::time::Duration>) -> Self {
-        if let Some(dur) = timeout {
-            self.with_timeout(dur)
+    pub fn with_timeout(&self, timeout: impl Into<Option<core::time::Duration>>) -> Self {
+        // If this overflows, treat that as no deadline.
+        if let Some(deadline) = timeout
+            .into()
+            .and_then(|timeout| self.waker.0.platform.now().checked_add(timeout))
+        {
+            self.with_deadline(deadline)
         } else {
             Self { ..*self }
         }
@@ -355,25 +344,14 @@ impl<'a, Platform: RawSyncPrimitivesProvider + TimeProvider> WaitContext<'a, Pla
     /// If the existing context already has an earlier deadline, then this just
     /// clones the context.
     #[must_use]
-    pub fn with_deadline(&self, deadline: Platform::Instant) -> Self {
+    pub fn with_deadline(&self, deadline: impl Into<Option<Platform::Instant>>) -> Self {
         let mut this = Self { ..*self };
-        if self.deadline.is_none_or(|d| deadline < d) {
+        if let Some(deadline) = deadline.into()
+            && self.deadline.is_none_or(|d| deadline < d)
+        {
             this.deadline = Some(deadline);
         }
         this
-    }
-
-    /// Returns a new context that has the given deadline, if one is provided.
-    ///
-    /// If the existing context already has an earlier deadline or if no deadline
-    /// is provided, then this just clones the context.
-    #[must_use]
-    pub fn with_maybe_deadline(&self, deadline: Option<Platform::Instant>) -> Self {
-        if let Some(d) = deadline {
-            self.with_deadline(d)
-        } else {
-            Self { ..*self }
-        }
     }
 
     /// Returns the deadline for this wait context, if any.
@@ -451,9 +429,9 @@ impl<'a, Platform: RawSyncPrimitivesProvider + TimeProvider> WaitContext<'a, Pla
         self.wait_until(|| false).unwrap_err()
     }
 
-    /// Waits until `f` returns `true`.
+    /// Waits until `ready` returns `true`.
     ///
-    /// `f` is called once before the thread sleeps and then again each time the
+    /// `ready` is called once before the thread sleeps and then again each time the
     /// thread is woken up. The caller must arrange for wakeups at the
     /// appropriate time. This can be done by a call to [`Waker::wake`] or
     /// [`ThreadHandle::interrupt`].
@@ -464,11 +442,11 @@ impl<'a, Platform: RawSyncPrimitivesProvider + TimeProvider> WaitContext<'a, Pla
     ///
     /// # Panics
     /// Panics if the thread is not currently in the running state, either
-    /// because `f` calls `wait_until` recursively, or  because
+    /// because `ready` calls `wait_until` recursively, or  because
     /// [`prepare_to_run_guest`](WaitState::prepare_to_run_guest) was called
     /// without a subsequent call to
     /// [`finish_running_guest`](WaitState::finish_running_guest).
-    pub fn wait_until(&self, mut f: impl FnMut() -> bool) -> Result<(), WaitError> {
+    pub fn wait_until(&self, mut ready: impl FnMut() -> bool) -> Result<(), WaitError> {
         assert_eq!(
             self.waker.0.state_for_assert(),
             ThreadState::RUNNING_IN_HOST
@@ -476,7 +454,7 @@ impl<'a, Platform: RawSyncPrimitivesProvider + TimeProvider> WaitContext<'a, Pla
         let _end_wait = crate::utils::defer(|| self.end_wait());
         loop {
             self.start_wait();
-            if f() {
+            if ready() {
                 break Ok(());
             }
             self.commit_wait()?;
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -1389,7 +1389,7 @@ impl Task {
                 _ => return Err(Errno::EBADF),
             }
         };
-        match epoll_file.wait(&self.wait_cx().with_maybe_timeout(timeout), maxevents) {
+        match epoll_file.wait(&self.wait_cx().with_timeout(timeout), maxevents) {
             Ok(epoll_events) => {
                 if !epoll_events.is_empty() {
                     events
@@ -1434,10 +1434,7 @@ impl Task {
             set.add_fd(fd.fd, events);
         }
 
-        match set.wait(
-            &self.wait_cx().with_maybe_timeout(timeout),
-            &self.files.borrow(),
-        ) {
+        match set.wait(&self.wait_cx().with_timeout(timeout), &self.files.borrow()) {
             Ok(()) => {}
             Err(WaitError::Interrupted) => {
                 // TODO: update the remaining time.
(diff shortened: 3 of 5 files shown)


litebox/src/event/polling.rs
```diff
@@ -32,152 +26,97 @@ struct PolleeInner<Platform: RawSyncPrimitivesProvider + TimeProvider> {
...
+        &self,
+        nonblock: bool,
+        mut try_op: impl FnMut() -> Result<R, TryOpError<E>>,
+        check: impl Fn() -> bool,
+        register_observer: impl FnOnce(Weak<dyn Observer<Events>>, Events),
+    ) -> Result<R, TryOpError<E>>
+    where
+        Platform: RawSyncPrimitivesProvider + TimeProvider,
+    {
+        let mut register_observer = Some(register_observer);
```
wdcui: I'm not familiar with this coding pattern. Why does the code wrap `register_observer` with `Some()`?

litebox_shim_linux/src/syscalls/epoll.rs
```diff
@@ -476,81 +482,70 @@ impl PollSet {
...
+            entry.revents = if entry.fd < 0 {
+                continue;
+            } else if let Some(file) = fds.get_fd(entry.fd.reinterpret_as_unsigned())
+                && let Ok(poll_descriptor) = EpollDescriptor::try_from(files, file)
+            {
+                let observer = if !is_ready && let Some(waker) = waker {
+                    // TODO: a separate allocation is necessary here
+                    // because registering an observer twice with two
+                    // different event masks results in the last one
+                    // replacing the first. If this is changed to
```
wdcui, line 485: Minor: the comment is not clear: "instead OR"?

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1479,7 +1501,20 @@ impl Task {
...
         }
 
-        set.wait_or_timeout(&self.files.borrow(), timeout);
+        match set.wait(
+            &self.wait_cx().with_maybe_timeout(timeout),
+            &self.files.borrow(),
+        ) {
+            Ok(()) => {}
+            Err(WaitError::Interrupted) => {
+                // TODO: update the remaining time.
```
wdcui, line 1504: Do we need to resolve this TODO in this PR? What happens when the remaining time is not updated?

litebox/src/event/wait.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// [`finish_running_guest`](Self::finish_running_guest) before using the
+    /// wait state again.
+    ///
+    /// # Panics
+    /// Panics if called while not in the running state.
+    #[must_use]
+    pub fn prepare_to_run_guest(&self, f: impl FnOnce() -> bool) -> bool {
+        assert_eq!(self.waker.0.state_for_assert(), State::RUNNING);
+        self.waker.0.set_state(State::IN_GUEST, Ordering::SeqCst);
+        let ready = f();
```
wdcui: `ready` means `ready_to_run_guest` or `ready_to_finish_guest`?

litebox/src/event/wait.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// # Panics
+    /// Panics if there was not a prior successful call to
+    /// [`prepare_to_run_guest`](Self::prepare_to_run_guest).
+    pub fn finish_running_guest(&self) {
+        let state = self.waker.0.state_for_assert();
+        assert!(
+            state == State::IN_GUEST || state == State::INTERRUPTED_GUEST,
+            "{state:?}"
+        );
+        self.waker.0.set_state(State::RUNNING, Ordering::Relaxed);
```
wdcui: Why do we set the state to be `RUNNING` when we try to `finish_running_guest`?

litebox/src/event/wait.rs
```diff
@@ -0,0 +1,484 @@
...
+        let v = condvar.underlying_atomic().fetch_update(
+            Ordering::Release,
+            Ordering::Relaxed,
+            |state| match State(state) {
+                State::RUNNING | State::WOKEN | State::INTERRUPTED_GUEST | State::IN_GUEST => None,
+                State::WAITING => Some(State::WOKEN.0),
+                state => unreachable!("{state:?}"),
+            },
+        );
+        match v.map(State) {
```
wdcui: I'm not familiar with this pattern. Isn't `State` a type?

litebox/src/event/wait.rs
```diff
@@ -0,0 +1,484 @@
...
+                // Provide a consistent release fence even if we didn't wake up
+                // the thread.
+                core::sync::atomic::fence(Ordering::Release);
+            }
+        }
+    }
+}
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+struct State(u32);
```
wdcui: `State` is a very general name. Can we change its name to make it more specific?

litebox/src/event/wait.rs
```diff
@@ -0,0 +1,484 @@
...
+                Ok(UnblockedOrTimedOut::Unblocked) | Err(ImmediatelyWokenUp) => Ok(()),
+                Ok(UnblockedOrTimedOut::TimedOut) => Err(WaitError::TimedOut),
+            }
+        } else {
+            let _ = self.waker.0.condvar.block(State::WAITING.0);
+            Ok(())
+        }
+    }
+
+    /// Sleep until the wait is interrupted or times out.
```
wdcui, line 420: The `sleep` function doesn't take a timeout argument?

litebox/src/event/polling.rs
jstarks: We only want to call it once, and indeed by the function signature we _can_ only call it once. So we put it in an `Option` and `.take()` it the first time through the loop--subsequent times will get `None`.

litebox_shim_linux/src/syscalls/file.rs
jstarks: This is just a behavior mismatch with Linux. In practice it doesn't matter, because this is a POSIX violation to update the time, and glibc hides this behavior from the caller. It might matter once we support restarting syscalls via `SA_RESTART`.

litebox/src/event/wait.rs
jstarks: `RUNNING` means "running in litebox/shim", as opposed to "running in guest". I'll rename these a bit to make it clearer.

litebox/src/event/wait.rs
jstarks: Yes, but since it's a tuple type (with one field), it can also be used as a function. I.e., this is equivalent to `v.map(|v| State(v))`.

litebox/src/event/wait.rs
jstarks: No, the timeout is provided by a separate call to `with_timeout` or similar. This pattern is useful because it allows the caller to set the timeout separately from the wait operation. I'll add a comment here and in `wait_until` so that it's easy to find.

litebox/src/event/polling.rs
wdcui: Got it. Thanks.

litebox/src/event/wait.rs
wdcui: I see. Thanks.

litebox_shim_linux/src/syscalls/file.rs
wdcui: I see. Thanks.

litebox/src/event/wait.rs
wdcui: Okay, thanks.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -442,35 +455,22 @@ fn connect(fd: &SocketFd, sockaddr: SocketAddr) -> Result<(), Errno> {
...
+    cx.wait_on_events(
+        get_status(fd).contains(OFlags::NONBLOCK),
+        || {
+            if is_ready(fd).map_err(TryOpError::Other)? {
+                Ok(())
+            } else {
+                Err(TryOpError::TryAgain)
+            }
+        },
+        || check_ready(fd, Events::IN | Events::OUT),
```
CvvT: In #497, I fixed the event for a socket so that it can correctly return `Events::HUP` to indicate close of the socket (e.g., upon receipt of RST packet). Here it should be `Events::IN | Events::OUT | Events::ALWAYS_POLLED`. I will fix it later when doing a rebase for #497.

I wonder if it is possible to make this `check` optional. In the previous model, I used `|| true` and assume compiler would optimize `check` away. For network operations, `check` is probably unnecessary (when some event happens, `check` is most likely to return true and call `try_op` anyway). Also, in this case, `check` is almost identical to `try_op`, resulting in some duplication.

litebox/src/event/polling.rs
```diff
@@ -32,152 +26,97 @@ struct PolleeInner<Platform: RawSyncPrimitivesProvider + TimeProvider> {
...
+        loop {
             match try_op() {
                 Err(TryOpError::TryAgain) => {}
                 ret => return ret,
             }
+            if nonblock {
+                return Err(TryOpError::TryAgain);
+            }
+            if let Some(register_observer) = register_observer.take() {
+                register_observer(self.waker().observer(), Events::all());
```
CvvT: Should we unregister the observer once it is done? In the previously implementation, once `poller` is out of scope, it would be automatically unregistered by `pollee` upon its next operation `notify_observers`.

jstarks: Oh, good point. We do need rework this a bit--ideally we would not allocate here. But for now, I'll restore `poller`. I'll also remove `check` altogether, since if we're allocating for `poller` then we can just remember whether we were explicitly woken that way--`check` makes more sense if there are spurious wakeups.

litebox/src/event/polling.rs
```diff
@@ -32,152 +28,138 @@ struct PolleeInner<Platform: RawSyncPrimitivesProvider + TimeProvider> {
 pub enum TryOpError<E> {
     #[error("operation should be retried")]
     TryAgain,
-    #[error("operation timed out")]
-    TimedOut,
+    #[error("wait error")]
+    WaitError(WaitError),
```
jaybosamiya-ms: Minor: I think this may benefit from a `#[from]` (i.e., `WaitError(#[from] WaitError),`).  Not crucial to add (we can always add it later, just something I noticed).

litebox/src/event/polling.rs
```diff
@@ -32,152 +28,138 @@ struct PolleeInner<Platform: RawSyncPrimitivesProvider + TimeProvider> {
...
+impl<Platform: RawSyncPrimitivesProvider + TimeProvider> WaitContext<'_, Platform> {
+    /// Run `try_op` until it returns a non-`TryAgain` result, waiting after
+    /// each `TryAgain`.
+    ///
+    /// If `nonblock` is true, returns `TryAgain` instead of waiting.
+    ///
+    /// If `try_op` returns `TryAgain`, the thread will be woken try again when
+    /// the observer, registered via the call to `register_observer`, is called
+    /// with events that match the given `events` filter (or an event in
+    /// `Events::ALWAYS_POLLED`).
```
jaybosamiya-ms: Something about the English here seems off "will be woken try again".  Maybe "woken to try again" or something?

litebox/src/event/polling.rs
```diff
@@ -32,152 +28,138 @@ struct PolleeInner<Platform: RawSyncPrimitivesProvider + TimeProvider> {
...
+    /// If `nonblock` is true, returns `TryAgain` instead of waiting.
     ///
-    /// The `check` function is used to reduce the number of times `try_op` is run. Specifically, if
-    /// `check` returns `false`, then it means that it is not worth attempting to run `try_op` and
-    /// that we can wait (for a notification to arrive, via [`Self::notify_observers`]) until
-    /// `check` returns `true` again. Note that `try_op` can be (sometimes) run even if `check` is
-    /// false; it is not to be taken as a pre-requisite to running `try_op` but merely used as a way
-    /// to reduce the number of times `try_op` is invoked.
-    pub fn wait_or_timeout<R, E>(
+    /// If `try_op` returns `TryAgain`, the thread will be woken try again when
```
jaybosamiya-ms: Same

litebox/src/event/wait.rs
```diff
@@ -0,0 +1,499 @@
...
+//! Ordinary waits in litebox, via the [`RawMutex`] trait, are not easily
+//! interrupted--they can only be woken up by another thread explicitly
+//! signaling the raw mutex. This is fine for ordinary uses of mutexes, locks,
+//! and condition variables, cases where waits can be guaranteed to eventually
+//! complete.
+//!
+//! However, waits on guest- or externally-controlled conditions (e.g., futexes,
+//! eventfds, IO events) must be interruptible due to process termination or
+//! asynchronous signals, since the waited-on event may never occur. This module
+//! provides infrastructure for handling such cases.
```
jaybosamiya-ms, line 12: We might wish to put a 
```
FUTURE: Update to use the [`litebox::event::wait`] infrastructure
``` 
on `litebox::sync::futex`?  Might be ok to just open an issue to track it too without adding it as a comment in the code.

litebox/src/event/wait.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// If the existing context already has an earlier deadline or if no timeout
+    /// is provided, then this just clones the context.
+    #[must_use]
+    pub fn with_maybe_timeout(&self, timeout: Option<core::time::Duration>) -> Self {
+        if let Some(dur) = timeout {
+            self.with_timeout(dur)
+        } else {
+            Self { ..*self }
+        }
+    }
```
jaybosamiya-ms, line 340: These two could technically be combined with a `timeout: impl Into<Option<Duration>>`

litebox/src/event/wait.rs
```diff
@@ -0,0 +1,499 @@
...
+        }
+        this
+    }
+
+    /// Returns a new context that has the given deadline, if one is provided.
+    ///
+    /// If the existing context already has an earlier deadline or if no deadline
+    /// is provided, then this just clones the context.
+    #[must_use]
+    pub fn with_maybe_deadline(&self, deadline: Option<Platform::Instant>) -> Self {
```
jaybosamiya-ms: Similar

litebox/src/event/wait.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// [`with_deadline`](Self::with_deadline) or
+    /// [`with_timeout`](Self::with_timeout).
+    ///
+    /// # Panics
+    /// Panics if the thread is not currently in the running state, either
+    /// because `f` calls `wait_until` recursively, or  because
+    /// [`prepare_to_run_guest`](WaitState::prepare_to_run_guest) was called
+    /// without a subsequent call to
+    /// [`finish_running_guest`](WaitState::finish_running_guest).
+    pub fn wait_until(&self, mut f: impl FnMut() -> bool) -> Result<(), WaitError> {
```
jaybosamiya-ms: Consider renaming `f` to `ready` or similar?  I can see an argument for keeping it `f` too, so not crucial, but generally I prefer functions that return `bool` to have something in their name that immediately tells you the direction of true/false.

A different approach that might actually be even nicer is to leave it as `f` but make it `f: impl FnMut() -> Option<T>` with the `wait_until` returning a `Result<T, WaitError>`, at which point the function signature already tells you when things can/will be returned (since, at a type level itself, it is impossible to construct the `T` unless `f` had been run and it had returned `Some`).  I'm ok with it if you think this is unnecessary and would prefer it just being a `bool`.

litebox/src/event/wait.rs
jstarks: I'll just open a PR :)

litebox/src/event/wait.rs
jstarks: Interesting. I'll try that.

litebox/src/event/wait.rs
jstarks: Happy to rename `f` to `ready`.

I originally had `f` return `Option<T>` for the reasons you give, but `bool` ended up being more ergonomic. We can reconsider later if needed, or even add a bare `wait` method with `wait_until` as a special case.

litebox/src/event/polling.rs
jstarks: `TryOpError` is a kind of weird error type, in that you aren't really expected to construct the `WaitError` variant--the wait error is provided by the `wait_on_events` method. So, it's not clear to me that we want `#[from]`--I don't think it should be any easier for callers to construct this variant. I'll add `#[source]`, though.

Maybe we should split this type in two...

jaybosamiya-ms: `#[source]` is good enough, thanks!

PR #495, opened 2025-11-15
Fix soundness issues with Mutex and RwLock

jstarks:
As an owning wrapper of `T`, `Mutex<T>` can only be soundly `Send`/`Sync` if `T` is `Send` (`Sync` is not required because the mutex ensures the `T` can only be concurrently accessed by a single thread). `RwLock<T>` is even stricter since it does not guarantee exclusive shared access and must have a matching `Send`/`Sync` bound. These bounds were missing from the litebox impls.

Fix this by adding the appropriate bounds, propagating them to other traits where necessary. Luckily, it doesn't appear that there was any real harm here--there was no undefined behavior to fix anywhere.

### DIFF
--- a/litebox/src/fd/mod.rs
+++ b/litebox/src/fd/mod.rs
@@ -745,7 +745,8 @@ impl DescriptorEntry {
 /// A file descriptor that refers to entries by the `Subsystem`.
 #[repr(transparent)] // this allows us to cast safely
 pub struct TypedFd<Subsystem: FdEnabledSubsystem> {
-    _phantom: PhantomData<fn(Subsystem)>,
+    // Invariant in `Subsystem`: <https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns>
+    _phantom: PhantomData<fn(Subsystem) -> Subsystem>,
     x: OwnedFd,
 }
 
--- a/litebox/src/sync/rwlock.rs
+++ b/litebox/src/sync/rwlock.rs
@@ -644,8 +644,9 @@ impl<Platform: RawSyncPrimitivesProvider, T> RwLock<Platform, T> {
 }
 
 // SAFETY: `RwLock<T>` inherits `Send` from `T`.
-unsafe impl<Platform: RawSyncPrimitivesProvider, T: Sync> Sync for RwLock<Platform, T> {}
-// SAFETY: `RwLock<T>` inherits `Sync` from `T`. Note that this is a different
-// bound from `Mutex<T>`, because `RwLock` does not guarantee mutual
-// exclusion--there could be multiple readers at the same time.
 unsafe impl<Platform: RawSyncPrimitivesProvider, T: Send> Send for RwLock<Platform, T> {}
+// SAFETY: `RwLock<T>` is `Sync` when `T` is `Send+Sync`. Note that this is a
+// different bound from `Mutex<T>`--the `Send` bound is still necessary since a
+// writer can transfer `T` between threads, but the `Sync` bound is necessary,
+// too, since readers on multiple threads can share `T` simultaneously.
+unsafe impl<Platform: RawSyncPrimitivesProvider, T: Send + Sync> Sync for RwLock<Platform, T> {}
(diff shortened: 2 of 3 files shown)


wdcui:
I requested @jaybosamiya-ms to review this PR.

litebox/src/sync/rwlock.rs
```diff
@@ -643,5 +643,9 @@ impl<Platform: RawSyncPrimitivesProvider, T> RwLock<Platform, T> {
...
 }
 
-unsafe impl<Platform: RawSyncPrimitivesProvider, T> Sync for RwLock<Platform, T> {}
-unsafe impl<Platform: RawSyncPrimitivesProvider, T> Send for RwLock<Platform, T> {}
+// SAFETY: `RwLock<T>` inherits `Send` from `T`.
+unsafe impl<Platform: RawSyncPrimitivesProvider, T: Sync> Sync for RwLock<Platform, T> {}
+// SAFETY: `RwLock<T>` inherits `Sync` from `T`. Note that this is a different
+// bound from `Mutex<T>`, because `RwLock` does not guarantee mutual
+// exclusion--there could be multiple readers at the same time.
+unsafe impl<Platform: RawSyncPrimitivesProvider, T: Send> Send for RwLock<Platform, T> {}
```
jaybosamiya-ms, line 647: The comments here are mildly confusing, possibly due to swapped order of send/sync.

Also, it should be
- `T:Send` -> `RwLock<T>:Send` and 
- `T:Send+Sync` -> `RwLock<T>:Sync` 

respectively, right?

The second one is currently at "`T:Sync` -> `RwLock<T>:Sync`" which may not be sufficient?

litebox/src/fd/mod.rs
```diff
@@ -745,7 +745,7 @@ impl DescriptorEntry {
 /// A file descriptor that refers to entries by the `Subsystem`.
 #[repr(transparent)] // this allows us to cast safely
 pub struct TypedFd<Subsystem: FdEnabledSubsystem> {
-    _phantom: PhantomData<Subsystem>,
+    _phantom: PhantomData<fn(Subsystem)>,
```
jaybosamiya-ms: Given that we're updating this, we might wish to actually make it invariant, no?  We don't actually want any `TypedFd<A>` between `TypedFd<B>` coercions to happen even if one is a subtype of another (this might actually be a moot point in practice, but nonetheless, might as well update it while we're updating this).  Specifically, we could make it invariant by making this a `PhantomData<fn(Subsystem) -> Subsystem>`.  We probably also want to add a doc comment pointer to https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -34,7 +34,7 @@ pub fn update_cpu_mhz(freq: u64) {
 /// This is the platform for running LiteBox in kernel mode.
 /// It requires a host that implements the [`HostInterface`] trait.
 pub struct LinuxKernel<Host: HostInterface> {
-    host_and_task: core::marker::PhantomData<Host>,
+    host_and_task: core::marker::PhantomData<fn(Host)>,
```
jaybosamiya-ms: I am not fully sure whether we want covariant or contravariant in `Host` here.  Might be safer to just make it invariant right now?

Also similar link to https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns would help.

litebox/src/fd/mod.rs
jstarks: Good call. I hope they stabilize `PhantomInvariant` soon.

I seem to recall that I ran into some surprising problems with `Sized` when I tried this the first time, though. I'll try again.

litebox/src/sync/rwlock.rs
jstarks: Oh, yeah, good catch. You're right, you need `Send` for the `&mut T` to be safe.

(Yeah, the comments got flipped.)

litebox/src/fd/mod.rs
jaybosamiya-ms: Yep would reduce the need to look that table up each time I see a `PhantomData`.  I can walk through the implications from first principles, but that table is much faster, but just literally having it encoded directly in the name is even nicer.

PR #496, opened 2025-11-17
Manually load ELF binaries

jstarks:
Parse and load ELF binaries manually instead of using the `elf_loader` crate. This is done for a few reasons:

* `elf_loader` is not safe for our use--it assumes it can safely access mapped memory directly via Rust memory access primitives--this is not always true for LiteBox since that memory is user-mode memory. Even when we know there are no other guest threads, so that concurrent mutation should be impossible, we still need to handle that memory differently from Rust memory.

* `elf_loader` creates a dependency on shim TLS to get the task object, since the various mmap-related traits do not take `&self`. This prevents some future refactoring to eliminate the use of TLS within the shims. Fixing this upstream seems like a big change, one that would break semver compatibility.

* `elf_loader` seems to combine parsing and mapping operations into one step, preventing us from returning errors from `execve` when the target binary is not a valid ELF executable.

* `elf_loader` has an awful lot of code marked unsafe with safety contracts that are unclear (to me).

* Loading the binary is a pretty important part of the Linux and OPTEE shims--we want to have a good understanding and control over exactly what loader ABI we are actually implementing.

For now, just update the Linux shim to use this new loader--this work has uncovered some design problems with OPTEE that need to be addressed before we can remove the dependency on `elf_loader`.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -277,11 +277,11 @@ impl LinuxShim {
         let entrypoints = crate::LinuxShimEntrypoints {
             _not_send: core::marker::PhantomData,
         };
-        let process = with_current_task(|task| {
+        with_current_task(|task| {
             task.load_program(loader::elf::ElfLoader::new(task, path)?, argv, envp)?;
             Ok(LinuxShimProcess(task.process().clone()))
-        })?;
-        Ok(LoadedProgram {
+        })
+        .map(|process| LoadedProgram {
             entrypoints,
             process,
         })


litebox_shim_linux/src/loader/elf.rs
```diff
@@ -1,207 +1,147 @@
...
 
-/// [`elf_loader::mmap::Mmap`] implementation for ELF loader
-struct ElfLoaderMmap<'a>(&'a Task);
+impl litebox_common_linux::loader::MapMemory for ElfFile<'_> {
+    type Error = Errno;
+
+    fn reserve(&mut self, len: usize, align: usize) -> Result<usize, Self::Error> {
+        // Allocate a mapping larger than needed to satisfy alignment, then
+        // unmap the excess.
+        let mapping_len = len + (align - PAGE_SIZE);
```
wdcui: I don't follow this line. Should `mapping_len` be aligned with `align` or `PAGE_SIZE`?

litebox_common_linux/src/loader.rs
```diff
@@ -0,0 +1,763 @@
...
+        // Read the program headers.
+        let phent_size = if cfg!(target_pointer_width = "64") {
+            size_of::<elf::segment::Elf64_Phdr>()
+        } else {
+            size_of::<elf::segment::Elf32_Phdr>()
+        };
+        if usize::from(header.e_phentsize) != phent_size {
+            return Err(ElfParseError::BadFormat);
+        }
+        // Limit to 64KB of program headers.
```
wdcui, line 187: I don't see how the code limits the size based on this comment.

litebox_common_linux/src/loader.rs
```diff
@@ -0,0 +1,763 @@
...
+            {
+                return Err(ElfLoadError::InvalidProgramHeader);
+            }
+            let adjusted_vaddr = base_addr + p_vaddr;
+            if p_filesz > 0 {
+                // Map the file-backed portion.
+                let map_start = page_align_down(adjusted_vaddr);
+                let map_end = page_align_up(adjusted_vaddr + p_filesz);
+                mapper
+                    .map_file(
```
wdcui: Is there any weird case that `ph.p_offset` is not aligned intentionally?

litebox_common_linux/src/loader.rs
jstarks: Yeah, it's subtle because it relies on the types of the arguments (u16). I'll make it clearer.

litebox_shim_linux/src/loader/elf.rs
jstarks: `mapping_len` should be big enough that we can find an aligned subrange of `len` bytes inside the mapping.

jstarks: (`align` is at least `PAGE_SIZE`, but that's not clear or guaranteed here. I'll add a `.max` to be safe.)

litebox_common_linux/src/loader.rs
jstarks: I wrote the code to expect it to be co-aligned with p_vaddr. But I see that `musl` just rounds it down to the nearest page and doesn't worry about this. I'll do the same--ELF has a lot of unwritten rules, AFAICT.

jstarks: Actually, I take that back. Linux does what I'm doing, so I'll leave it as is. `map_file` will fail with `EINVAL` if `p_vaddr` and `p_offset` are not co-aligned, which is what we want.

litebox_common_linux/src/loader.rs
```diff
@@ -0,0 +1,773 @@
+//! ELF loader and mapper.
```
CvvT, line 1: Do we want to make it a separate crate?

litebox_common_linux/src/loader.rs
wdcui: I see. Thanks.

litebox_common_linux/src/loader.rs
wdcui: Ah I see.

litebox_shim_linux/src/loader/elf.rs
wdcui: Got it.

litebox_common_linux/src/loader.rs
jstarks: I don't have a strong preference, but this was easier for now because of the various `From<X> for Errno` impls, as well as `Protection::flags()` method, both of which rely on types in `litebox_common_linux`.

If OPTEE will eventually stop using `litebox_common_linux` then we could split this up.

sangho2:
It seems that this new ELF loader has a problem with the KMPP TA (other three TAs work well). I'll take a look at it more.

`cargo run -p litebox_runner_optee_on_linux_userland litebox_runner_optee_on_linux_userland/tests/kmpp-ta.elf.hooked litebox_runner_optee_on_linux_userland/tests/kmpp-ta-cmds.json`

jstarks:
> It seems that this new ELF loader has a problem with the KMPP TA (other three TAs work well). I'll take a look at it more.
> 
> `cargo run -p litebox_runner_optee_on_linux_userland litebox_runner_optee_on_linux_userland/tests/kmpp-ta.elf.hooked litebox_runner_optee_on_linux_userland/tests/kmpp-ta-cmds.json`

Is KMPP TA not part of the CI tests?

jstarks:
Hmm, I see, KMPP-TA fails to run but the test succeeds anyway...

wdcui:
@sangho2 @CvvT, if you think this PR is ready to be merged, please approve it. Thanks!

litebox_common_linux/src/loader.rs
```diff
@@ -0,0 +1,604 @@
...
+    mm::linux::PAGE_SIZE,
+    platform::{RawConstPointer as _, RawMutPointer as _, RawPointerProvider},
+    utils::{ReinterpretSignedExt as _, TruncateExt as _},
+};
+use thiserror::Error;
+use zerocopy::{FromBytes, FromZeros as _, IntoBytes};
+
+use crate::errno::Errno;
+
+type Endian = elf::endian::LittleEndian;
```
CvvT, line 20: Should we get the endianness from the binary or check it to ensure it's little endian?

jstarks: `parse_ident` will verify the endian matches for us.

CvvT:
The test `test_tun_tcp_socket_as_client` seems to be flaky; I will take a look.

PR #497, opened 2025-11-17
Support graceful close & fix connect

CvvT:
When close a socket, we shouldn't remove it from the `socket set` directly. Instead, we need to keep it around for a while so that it can finish the `FIN` handshake. The behavior of socket close can be configured via `setsockopt` with `SO_LINGER`:

```
Enable(l_onoff!=0):
    Timeout (l_linger > 0): The close() call will block until the socket's unsent data is transmitted and acknowledged, or until the timeout is reached. If the timeout expires, the connection is closed forcefully.
    Timeout (l_linger = 0): The close() call will immediately send a TCP RST (reset) packet, terminating the connection abruptly without waiting for data to be sent. This can cause data loss.
Disable (l_onoff=0): The socket is closed gracefully using the standard TCP handshake (FIN sequence) without any specific timeout delay beyond the normal TCP process. 
```

To the end, `close` now accepts additional argument `CloseBehavior`, and when a socket is closed, we save it to another queue instead of removing it from `socket_set`.

Following the same pattern, `connect` now can be called multiple times with `check_progress = true` to check whether a TCP connection is established, instead of relying on events on a socket.

Fix closing a server socket in `close_handle` where we need to remove its all listening sockets as well.

This PR also fixes an event issue for socket. A closed socket should have `Events::HUP`, which is useful when a connection fails (e.g., no server is listening or receiving invalid packets) and needs to notify any waiters.

### DIFF
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -1216,7 +1216,7 @@ where
                     socket.state() != tcp::State::Closed
                 });
                 // Find a socket that has progressed further in its TCP state machine, by finding a
-                // socket in a established state
+                // socket in an established state
                 let Some(position) = server_socket.socket_set_handles.iter().position(|&h| {
                     let socket: &tcp::Socket = self.socket_set.get(h);
                     socket.state() == tcp::State::Established


litebox/src/event/polling.rs
```diff
@@ -112,6 +112,7 @@ impl<Platform: RawSyncPrimitivesProvider + TimeProvider> Pollee<Platform> {
     ) -> Result<R, TryOpError<E>> {
         wait_or_timeout(&self.litebox, timeout, try_op, check, |observer, filter| {
             self.register_observer(observer, filter);
+            Ok(())
```
wdcui: Does the return value need to be an instance of `R`?

litebox/src/fd/mod.rs
```diff
@@ -124,23 +124,39 @@ impl<Platform: RawSyncPrimitivesProvider> Descriptors<Platform> {
...
+                    .unwrap();
+                Some(CloseResult::Closed(entry))
+            } else {
+                // Put it back
+                let old = self.entries[idx].replace(old);
+                assert!(old.is_none());
+                Some(CloseResult::Deferred)
+            }
         } else {
+            fd.x.mark_as_closed();
```
wdcui, line 159: This is a bug fix of previous code?

litebox/src/net/tests.rs
```diff
@@ -25,9 +25,7 @@ fn bidi_tcp_comms(mut network: Network<MockPlatform>, comms: fn(&mut Network<Moc
     let client_fd = network
         .socket(Protocol::Tcp)
         .expect("Failed to create TCP socket");
-    network
-        .connect(&client_fd, &listen_addr)
-        .expect("Failed to connect TCP socket");
+    let _ = network.connect(&client_fd, &listen_addr, false);
```
wdcui: Why do we ignore the return value here?

litebox/src/net/mod.rs
```diff
@@ -759,32 +824,66 @@ where
...
+                    }
+                }
+                if let Some(local_port) = tcp_specific.local_port.take() {
                     self.local_port_allocator.deallocate(local_port);
                 }
-                // TODO: Should we `.close()` or should we `.abort()`?
-                socket.abort();
-                socket_handle.pollee.notify_observers(Events::HUP);
+                let tcp_socket: &mut tcp::Socket = self.socket_set.get_mut(handle);
+                if tcp_specific.immediate_close.load(Ordering::Relaxed) {
```
wdcui, line 861: Does `abort()` mean close immediately regardless of the state?

litebox/src/net/mod.rs
```diff
@@ -1111,18 +1216,16 @@ where
                     socket.state() != tcp::State::Closed
                 });
                 // Find a socket that has progressed further in its TCP state machine, by finding a
-                // socket in a may-send-or-recv state
+                // socket in a established state
```
wdcui: Minor: "an established state"

litebox/src/event/polling.rs
CvvT: This is a callback for registering an observer, which is just one part of the process for `wait_or_timeout`. If it succeeds, `wait_or_timeout` can continue (thus no return value on this case). Otherwise, propagate the error back to the caller.

litebox/src/fd/mod.rs
CvvT: No, this is different from #490. #490 tries to do `close` and `dup` together, which is still the case in this PR. This PR is to prevent `close` if the callback function returns false (that is why `fd.x.mark_as_closed();` only needs to be called on all the other cases).

litebox/src/net/tests.rs
CvvT: Right, we shouldn't ignore it. Add a check on the return value.

litebox/src/net/mod.rs
CvvT: Good catch! Fixed.

litebox/src/net/mod.rs
CvvT: It means changing its state to `Closed`, which would then trigger sending out RST packet instead of FIN packet.

litebox/src/event/polling.rs
wdcui: Ah, I missed that.

PR #498, opened 2025-11-18
Add cryptographic random data provider

jstarks:
Add a provider for `getrandom` to use to get cryptographically-random bytes. Implement it in the userland platforms, passing through to the `getrandom` crate (which just forwards to the platform's native CRNG API).

It's not really possible to implement this correctly generically; it really has to be done on a per-platform basis. Future changes could provide infrastructure for kernel platforms to implement a CRNG, though, if desired.

Related: #41

### DIFF
--- a/litebox/src/platform/mod.rs
+++ b/litebox/src/platform/mod.rs
@@ -681,25 +681,31 @@ pub unsafe trait ThreadLocalStorageProvider {
 
 /// A provider of cryptographically-secure random data.
 ///
-/// The precise behavior and implementation is platform specific, and in general
-/// these methods should pass through to the platform's native cryptographic RNG
-/// API when one exists.
+/// The purpose of this provider is to allow LiteBox code to efficiently
+/// generate cryptographically-secure random bytes. This must be an infallible
+/// operation, with no possibility of failure, blocking, or returning
+/// low-quality randomness. The implementation must ensure that the CRNG is
+/// appropriately initialized and seeded by the time this method can be called.
 ///
-/// **Caution**: it may be tempting to write a more efficient provider that runs
-/// fully in the litebox runtime, seeding the RNG from the platform's kernel
-/// CRNG or other trusted sources. Don't do this! Implementing this correctly as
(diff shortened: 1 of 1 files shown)


litebox/src/platform/mod.rs
```diff
@@ -678,3 +678,37 @@ pub unsafe trait ThreadLocalStorageProvider {
...
 }
+
+/// A provider of cryptographically-secure random data.
+///
+/// The precise behavior and implementation is platform specific, and in general
+/// these methods should pass through to the platform's native cryptographic RNG
+/// API when one exists.
+///
+/// **Caution**: it may be tempting to write a more efficient provider that runs
+/// fully in the litebox runtime, seeding the RNG from the platform's kernel
```
jaybosamiya-ms: Minor: "provider" -> "implementation".  The precise definition of "provider" is not yet nailed down.  Changing this line to say "implementation of this method" makes it a bit less confusing.  It is also prob a little confusing due to the "fully in the litebox runtime" (which can accidentally be confused by "within the `litebox` crate", which is also a valid concern here).

Also like the other spot for consistency, `litebox` (in backticks if referring to the crate) or LiteBox (if referring to the project).

litebox/src/platform/mod.rs
```diff
@@ -678,3 +678,37 @@ pub unsafe trait ThreadLocalStorageProvider {
...
+pub trait CrngProvider {
+    /// Fill `buf` with cryptographically secure random bytes.
+    ///
+    /// This may take a long time for large buffers. Consider calling this
+    /// multiple times, checking for interrupts between calls, if you need to
+    /// fill a very large buffer.
+    ///
+    /// # Panics
+    /// Panics if unable to fill the buffer with random bytes. Litebox code is
+    /// not expected to handle such failures.
```
jaybosamiya-ms: Minor capitalization Litebox -> LiteBox _or_ if you are explicitly intending to say that the `litebox` crate does not handle such failures, then putting it into backticks is prob a better move.

PR #499, opened 2025-11-18
Support TCP_CONGESTION

CvvT:
This PR supports TCP option `TCP_CONGESTION`. 

Note that smoltcp supports two control algorithms (Reno and Cubic), but we didn't enable any of them. We can decide which we want to enable later. According to smoltcp's doc, "`CongestionControl::Cubic` represents a modern congestion control algorithm. It is the default choice for Linux, Windows, and macOS." But "`CongestionControl::Cubic` relies on double precision (`f64`) floating point operations, which may cause issues in some contexts". See https://github.com/smoltcp-rs/smoltcp/blob/39cd44e585aae109f37a2bc9f31376414968456d/src/socket/tcp.rs#L629-L649

### DIFF
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -1118,7 +1118,10 @@ mod tests {
     use alloc::string::ToString as _;
     use litebox::platform::RawConstPointer as _;
     use litebox::utils::TruncateExt as _;
-    use litebox_common_linux::{AddressFamily, ReceiveFlags, SendFlags, SockFlags, SockType};
+    use litebox_common_linux::{
+        AddressFamily, ReceiveFlags, SendFlags, SockFlags, SockType, SocketOptionName, TcpOption,
+        errno::Errno,
+    };
 
     use super::SocketAddress;
     use crate::{ConstPtr, MutPtr};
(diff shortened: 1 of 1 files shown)


litebox/src/net/mod.rs
```diff
@@ -1375,6 +1375,12 @@ where
                     TcpOptionData::KEEPALIVE(keepalive) => {
                         tcp_socket.set_keep_alive(keepalive.map(smoltcp::time::Duration::from));
                     }
+                    TcpOptionData::CONGESTION(congestion) => match congestion {
+                        CongestionControl::None => {
```
wdcui, line 1379: What does `CongestionControl::None` mean? To use the default congestion algorithm?

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -248,45 +248,67 @@ fn setsockopt(
...
-                    litebox_net()
-                        .lock()
-                        .set_tcp_option(fd, litebox::net::TcpOptionData::NODELAY(on))?;
-                    Ok(())
-                }
-                TcpOption::KEEPINTVL => {
-                    const MAX_TCP_KEEPINTVL: u32 = 32767;
-                    if !(1..=MAX_TCP_KEEPINTVL).contains(&val) {
-                        return Err(Errno::EINVAL);
+            if let TcpOption::CONGESTION = to {
```
wdcui, line 251: Minor: why do we use `if else` instead of `match` for different socket options?

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -354,38 +376,58 @@ fn getsockopt(
...
-            unsafe { optval.write_slice_at_offset(0, data) }.ok_or(Errno::EFAULT)?;
-            size_of::<u32>()
+            if let TcpOption::CONGESTION = tcpopt {
+                let TcpOptionData::CONGESTION(congestion) = litebox_net()
+                    .lock()
+                    .get_tcp_option(fd, litebox::net::TcpOptionName::CONGESTION)?
+                else {
+                    unreachable!()
+                };
+                let name = match congestion {
```
wdcui, line 386: Should we add some tests for setting/getting the tcp congestion option?

litebox/src/net/mod.rs
CvvT: `none` means no congestion control at all. On Linux, the default one is `Cubic` and there is no `none` option, but the application I tested seems to be fine with unknown congestion control algo.

litebox_shim_linux/src/syscalls/net.rs
CvvT: All other cases can be grouped together because they all read `u32` from user space. `TcpOption::CONGESTION` needs to read a str.

litebox_shim_linux/src/syscalls/net.rs
CvvT: Sure, I added one for `TCP_CONGESTION`. Will try to add more for other options in a separate PR.

wdcui: Thank you for adding the tests.

PR #501, opened 2025-11-19
Reimplement futex manager to use interruptible waits

jstarks:
Update futex manager to use interruptible waits. As part of this, implement an intrusive linked list to avoid heap
allocations in the futex path.

The futex and linked list tests have been run against miri to check for uses of undefined behavior.

### DIFF
--- a/litebox/src/sync/futex.rs
+++ b/litebox/src/sync/futex.rs
@@ -2,7 +2,7 @@
 
 // Implementation note: other submodules of `crate::sync` should NOT depend on
 // this module directly, because this module itself depends on some of the other
-// modules (specifically, this module depends on on `LoanList`, which depends on
+// modules (specifically, this module depends on `LoanList`, which depends on
 // `Mutex`). A refactoring could clean this up and prevent this dependency, but
 // at the moment, it has been decided that this ordering of dependency is more
 // fruitful.


litebox/src/sync/futex.rs
```diff
@@ -1,55 +1,50 @@
...
-// depends on on `RwLock`). A refactoring could clean this up and prevent this dependency, but at
-// the moment, it has been decided that this ordering of dependency is more fruitful.
-
-use super::{RawSyncPrimitivesProvider, RwLock};
-use crate::platform::{
-    ImmediatelyWokenUp, Instant, RawConstPointer as _, RawMutex, TimeProvider, UnblockedOrTimedOut,
-};
+// Implementation note: other submodules of `crate::sync` should NOT depend on
+// this module directly, because this module itself depends on some of the other
+// modules (specifically, this module depends on on `LoanList`, which depends on
```
wdcui: Minor: "on on"

litebox/src/sync/futex.rs
```diff
@@ -83,137 +83,68 @@ impl<Platform: RawSyncPrimitivesProvider + RawPointerProvider + TimeProvider>
...
+        ));
+
+        // Insert into the bucket's list. It will be removed when woken or the
+        // entry goes out of scope.
+        entry.as_mut().insert(bucket);
+
+        // Check the value once. Do this only after inserting into the list so
+        // that we don't miss a wakeup.
+        if futex_addr.load(Ordering::Relaxed) != expected_value {
+            return Err(FutexError::ImmediatelyWokenBecauseValueMismatch);
```
wdcui, line 127: Why do we return this case as an error?

jstarks:
> LGTM! With this new design of `WaitContext`, things got simplifed a lot and `futex` now can support bitset mask. I only have a high level question: why we switched from hashmap to hashtable with linked list? You mentioned it avoids heap allocation. I wonder it is purely for optimization or has any other reasons.

As LiteBox becomes more general-purpose, I think we'll have to confront the reality that "panic on OOM" is not a viable strategy for some scenarios. So, if we can design some of our core data structures to avoid needing allocations in the first place, then the eventual conversion will be that much easier.

It's probably OK to terminate the current process on OOM in cases where the contract doesn't allow us to return `ENOMEM`, so this is not strictly necessary. But still, better to just avoid allocations when it's straightforward to do so.

In this case, I had to change the data structure for the futex manager anyway--the old hash table was based on a single entry per futex, but with the new wait model (and to correctly support bitsets) we need an entry per waiter. Or at least a list of waiters per entry. So, I figured that we might as well switch to a data structure that's much more efficient for the use case, allocations or no--an external chaining hash table gives us good concurrency (lock per bucket instead of a global lock) while not requiring resizing under high load.

It might be interesting to use this `LoanList` for event wait queues as well. I have a prototype to do this.

litebox/src/sync/futex.rs
jstarks: Linux requires that we return `EAGAIN` in this case, although I don't really know why this distinction is important for callers in practice.

Arguably, it makes more sense to model this case as a bool return value indicating whether the wait actually happened. But I left this as-is since it's orthogonal to the refactoring I'm performing here.

PR #502, opened 2025-11-19
Clean up `ContinueOperation`

jstarks:
Some clean up in preparation for more task/signal work:

* Define one `ContinueOperation` for all shims, with just `ContinueGuest` and `ExitThread`.

* Remove exit status from `ExitThread`--this, plus the distinction between thread and process exit, will be communicated separately in future PRs, on a per-shim basis.

* Unify Linux shim entry/exit code.

* Record the `is_exiting` state of a Linux thread in `Task` and use it to determine whether to return `ExitThread` from a shim entrypoint. This is a more natural fit for Linux signal/termination handling.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -1348,7 +1348,7 @@ struct Task {
     fs: RefCell<Arc<syscalls::file::FsState>>,
     /// File descriptors. `RefCell` to support `unshare` in the future.
     files: RefCell<Arc<syscalls::file::FilesState>>,
-    /// If true, call the sigreturn punchline instead of returning to user mode.
+    /// If true, call the sigreturn punchthrough instead of returning to user mode.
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/lib.rs
```diff
@@ -1345,6 +1348,11 @@ struct Task {
     fs: RefCell<Arc<syscalls::file::FsState>>,
     /// File descriptors. `RefCell` to support `unshare` in the future.
     files: RefCell<Arc<syscalls::file::FilesState>>,
+    /// If true, call the sigreturn punchline instead of returning to user mode.
```
wdcui: Minor: "punchline" => "punchthrough"?

jstarks: Haha, whoops.

jstarks:
```suggestion
    /// If true, call the sigreturn punchthrough instead of returning to user mode.
```

PR #503, opened 2025-11-19
Use lazy memory container

femiatl:
Copying data from VTL0 involves a page by page copy into a memory container, followed by a copy into one contiguous buffer for parsing. For large data like KernelData, the two copies of data may exhaust the allocator free list. Instead, only copy VTL0 data at the time it needs to be parsed.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -56,12 +56,6 @@ use x509_cert::{Certificate, der::Decode};
 #[repr(align(4096))]
 struct AlignedPage([u8; PAGE_SIZE]);
 
-impl AlignedPage {
-    pub fn new() -> Self {
-        AlignedPage([0; PAGE_SIZE])
-    }
-}
-
 // For now, we do not validate large kernel modules due to the VTL1's memory size limitation.
 const MODULE_VALIDATION_MAX_SIZE: usize = 64 * 1024 * 1024;
 
@@ -364,7 +358,7 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
                 match heki_range.heki_kdata_type() {
                     HekiKdataType::SystemCerts => {
                         system_certs_mem
-                            .write_bytes_from_heki_range(heki_range)
+                            .add_heki_address_range(heki_range)
                             .map_err(|_| Errno::EINVAL)?;
                     }
                     HekiKdataType::KexecTrampoline => {
@@ -372,17 +366,17 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
                     }
                     HekiKdataType::PatchInfo => {
                         patch_info_mem
-                            .write_bytes_from_heki_range(heki_range)
+                            .add_heki_address_range(heki_range)
                             .map_err(|_| Errno::EINVAL)?;
                     }
                     HekiKdataType::KernelInfo => {
                         kinfo_mem
-                            .write_bytes_from_heki_range(heki_range)
+                            .add_heki_address_range(heki_range)
                             .map_err(|_| Errno::EINVAL)?;
                     }
                     HekiKdataType::KernelData => {
                         kdata_mem
-                            .write_bytes_from_heki_range(heki_range)
+                            .add_heki_address_range(heki_range)
                             .map_err(|_| Errno::EINVAL)?;
                     }
                     HekiKdataType::Unknown => {
@@ -597,12 +591,12 @@ fn prepare_data_for_module_validation(
                 }
                 ModMemType::ElfBuffer => {
                     module_as_elf
-                        .write_bytes_from_heki_range(heki_range)
+                        .add_heki_address_range(heki_range)
                         .map_err(|_| Errno::EINVAL)?;
                 }
                 ModMemType::Patch => {
                     patch_info_for_module
-                        .write_bytes_from_heki_range(heki_range)
+                        .add_heki_address_range(heki_range)
                         .map_err(|_| Errno::EINVAL)?;
                 }
                 _ => {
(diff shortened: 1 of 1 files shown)


femiatl:
- Tested by reducing memory reserved for secure kernel to 48MB, the minimum for an 8CPU machine.
- Tested with empty secure kernel boot param, together with https://github.com/microsoft/litebox/pull/492

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1511,121 +1505,62 @@ impl MemoryContainer {
...
-                PAGE_SIZE
-            } else {
-                usize::try_from(phys_end - phys_cur).unwrap()
-            };
-            self.write_bytes(addr + (phys_cur - phys_start), &page.0[..to_write])?;
-            phys_cur += u64::try_from(to_write).unwrap();
+            self.pages.insert(
+                addr_cur,
+                PhysFrame::from_start_address(phys_cur)
+                    .map_err(|_| MemoryContainerError::WriteFailed)?,
```
sangho2, line 1547: `WriteFailed` -> `InvalidAddress`?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1511,121 +1505,62 @@ impl MemoryContainer {
...
+    pub(crate) fn add_vtl0_address_range(
         &mut self,
         addr: VirtAddr,
         phys_start: PhysAddr,
         phys_end: PhysAddr,
     ) -> Result<(), MemoryContainerError> {
+        let mut addr_cur = addr;
         let mut phys_cur = phys_start;
+        if addr_cur - addr_cur.align_down(Size4KiB::SIZE)
+            != phys_cur - phys_cur.align_down(Size4KiB::SIZE)
```
sangho2, line 1530: `addr_cur.page_offset() != phys_cur.page_offset()` might be better.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1640,11 +1575,17 @@ impl MemoryContainer {
...
 
-            buf[buf_offset..buf_offset + len]
-                .copy_from_slice(&page.0[page_offset..page_offset + len]);
+            if !unsafe {
+                crate::platform_low().copy_slice_from_vtl0_phys(
+                    phys.start_address() + u64::try_from(page_offset).unwrap(),
+                    &mut buf[buf_offset..buf_offset + len],
+                )
+            } {
+                return Err(MemoryContainerError::CopyFromVtl0Failed);
```
sangho2, line 1587: One difference between the old pre-copy approach and this new lazy-copy approach is that the old approach checks whether the physical addresses belong to VTL0 at the registration whereas the new approach checks them when reading the data. I don't think there is significant security difference, but there might be some delay in detecting misbehaviors.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1640,11 +1575,17 @@ impl MemoryContainer {
...
             }
 
-            let page_offset = copy_start.page_offset().into();
+            let page_offset: usize = copy_start.page_offset().into();
             let buf_offset = usize::try_from(copy_start - start).expect("buffer offset error");
 
-            buf[buf_offset..buf_offset + len]
-                .copy_from_slice(&page.0[page_offset..page_offset + len]);
+            if !unsafe {
+                crate::platform_low().copy_slice_from_vtl0_phys(
```
sangho2, line 1582: We might want to mention potential data race issues here.

litebox_platform_lvbs/src/mshv/vsm.rs
tgopinath-microsoft: Yes.. I think this is acceptable from a security point of view. Atleast for now

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1413,14 +1407,14 @@ impl ModuleMemory {
     /// Write physical memory bytes from VTL0 specified in `HekiRange` at the specified virtual address of
```
praveen-pk, line 1407: Please update this comment to correspond to your change.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1640,11 +1575,17 @@ impl MemoryContainer {
                 break;
             }
 
-            let page_offset = copy_start.page_offset().into();
+            let page_offset: usize = copy_start.page_offset().into();
             let buf_offset = usize::try_from(copy_start - start).expect("buffer offset error");
 
-            buf[buf_offset..buf_offset + len]
-                .copy_from_slice(&page.0[page_offset..page_offset + len]);
+            if !unsafe {
```
praveen-pk, line 1581: what is the purpose of `!` here? `unsafe` is a keyword to mark an operation as unsafe.

Do you mean to check if the copy fails?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1511,121 +1505,62 @@ impl MemoryContainer {
...
     ) -> Result<(), MemoryContainerError> {
+        let mut addr_cur = addr;
         let mut phys_cur = phys_start;
+        if addr_cur - addr_cur.align_down(Size4KiB::SIZE)
+            != phys_cur - phys_cur.align_down(Size4KiB::SIZE)
+        {
+            return Err(MemoryContainerError::InvalidAddress);
+        }
+
         if !phys_cur.is_aligned(Size4KiB::SIZE) {
```
praveen-pk, line 1535: nit: Instead of a fixed size here, define a `PAGE_SIZE` variable and use it.

femiatl: It implements the PAGE_SIZE trait, maybe it could be named better

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Closing this PR to rework it

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Yes, its for the copy failing

PR #509, opened 2025-11-20
More thread startup cleanup

jstarks:
Make Linux shim entrypoint objects to be per-thread and not global, and update the platforms to handle this. This eliminates the global shim registration, which in theory would allow multiple shims to be used at the same time (on different threads, of course).

For now, there is still a bit of a TLS dance, which should go away once we stop using TLS in the Linux shim.

Also, stop returning contexts from the shim during program load and instead initialize the context during the first call into the shim. This will make switching to platform-specific contexts easier in the future.

The OPTEE shim keeps its current model for now.

### DIFF
--- a/litebox_runner_optee_on_linux_userland/tests/loader.rs
+++ b/litebox_runner_optee_on_linux_userland/tests/loader.rs
@@ -3,7 +3,6 @@ use litebox_platform_multiplex::{Platform, set_platform};
 fn init_platform() {
     let platform = Platform::new(None);
     set_platform(platform);
-    platform.register_shim(&litebox_shim_optee::OpteeShim);
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


litebox/src/shim.rs
```diff
@@ -6,15 +6,20 @@
...
     ///
-    /// After calling this, the caller must run the thread in the shim until it
-    /// exits, or there may be hangs or leaks.
-    fn init(self: alloc::boxed::Box<Self>);
+    /// FUTURE: use a single per-architecture type for all shims and platforms.
+    type ExecutionContext;
+
+    /// Initializes the thread, returning the shim interface for the new thread.
+    #[must_use]
+    fn init(
```
CvvT, line 16: I found it is very confusing that we have many `init` functions for thread initialization, including this one, `EnterShim::init`, and `Task::init_thread_state`. I don't have a better idea though.

jstarks: Yeah, it's a bit much! I am hoping that we can collapse some these after yet some more cleanup, but I'll think about renames along the way to clarify.

jstarks: I'll leave it as is for now and clean up during the next phases.

PR #510, opened 2025-11-20
Fix LVBS swapgs and shim_tls

sangho2:
This PR fixes LVBS platform's `swapgs` and `shim_tls` bugs and potential information leak.

### DIFF
--- a/litebox_platform_lvbs/src/user_context.rs
+++ b/litebox_platform_lvbs/src/user_context.rs
@@ -163,7 +163,8 @@ impl<Host: HostInterface> UserSpaceManagement for LinuxKernel<Host> {
         // Currently, `litebox_platform_lvbs` uses `swapgs` to efficiently switch between
         // kernel and user GS base values during kernel-user mode transitions.
         // This `swapgs` usage can pontetially leak a kernel address to the user, so
-        // we clear the `KernelGsBase` MSR before running the user thread.
+        // we clear the `KernelGsBase` MSR before running the user thread (only for
+        // the first entry).
         crate::arch::write_kernel_gsbase_msr(VirtAddr::zero());
 
         unsafe {


litebox_platform_lvbs/src/user_context.rs
```diff
@@ -166,7 +166,6 @@ impl<Host: HostInterface> UserSpaceManagement for LinuxKernel<Host> {
                 "push r12",
                 "push r13",
                 "push r14",
-                "swapgs",
```
jstarks, line 169: Why isn't this necessary now?

jstarks:
Side note, Coconut avoids use of gsbase in kernel mode altogether and instead has a per-CPU region of the kernel-mode address space (and per-CPU kernel-mode page tables). Interesting design choice...

litebox_platform_lvbs/src/user_context.rs
sangho2: Thank you for pointing this out. It needs to be something like `mov rax, 0x0; wrgsbase rax;` to do not leak kernel GSBASE.

sangho2:
> Side note, Coconut avoids use of gsbase in kernel mode altogether and instead has a per-CPU region of the kernel-mode address space (and per-CPU kernel-mode page tables). Interesting design choice...

Interesting. Our LiteBox kernel platform is still WiP, so we can adopt coconut's idea if it makes sense. Using `gsbase` and `swapgs` is not a strict requirement (might have a minor performance impact, though).

litebox_platform_lvbs/src/user_context.rs
jstarks: I don't understand--why isn't it just `swapgs`? We still do a `swapgs` on syscall entry, I think.

jstarks:
> > Side note, Coconut avoids use of gsbase in kernel mode altogether and instead has a per-CPU region of the kernel-mode address space (and per-CPU kernel-mode page tables). Interesting design choice...
> 
> Interesting. Our LiteBox kernel platform is still WiP, so we can adopt coconut's idea if it makes sense. Using `gsbase` and `swapgs` is not a strict requirement (might have a minor performance impact, though).

I'm trying to get more information about why they made this design choice. I think with their current design it's less efficient than using `gsbase`, but I don't think that has to be the case...

litebox_platform_lvbs/src/user_context.rs
sangho2: A problem is that the `KernelGsBase` MSR can have a stale GS base value which might be a kernel address or the address of another user process, and this will be written to the GS base by `swapgs`. To avoid this potential info leak, we now write zero to the `KernelGsBase` MSR before using `swapgs` and entering the userspace. We don't assume the guest uses GS, so this would be enough for now. This `iretq` is for initiating a user process which differs from syscall handling (no `swapgs` symmetry).

litebox_platform_lvbs/src/user_context.rs
```diff
@@ -159,13 +159,22 @@ impl<Host: HostInterface> UserSpaceManagement for LinuxKernel<Host> {
...
+        crate::arch::write_kernel_gsbase_msr(VirtAddr::zero());
+
         unsafe {
             asm!(
                 "push r10",
                 "push r11",
                 "push r12",
                 "push r13",
                 "push r14",
+                // clear the GS base register (as the `KernelGsBase` MSR contains 0)
```
CvvT, line 177: When this function is called? If we do `swapgs` at the entry and exit of every user-kernel transition, then we don't need to clear guest gsbase (also, does it mean we always overwrite guest gsbase to zero?)

sangho2: This is for jumping to the entry point. The syscall handling function is in a different file.

CvvT: I see. Then it makes sense.

PR #512, opened 2025-11-20
Add /dev/urandom

CvvT:
Add device `/dev/urandom` to LiteBox by using the recently introduced `CrngProvider`.

For now, just assume `CrngProvider` and `StdioProvider` are always implemented. In the future, we could decouple these by using registry pattern.

### DIFF
--- a/litebox/src/platform/mock.rs
+++ b/litebox/src/platform/mock.rs
@@ -10,8 +10,8 @@ extern crate std;
 
 use core::sync::atomic::AtomicU32;
 use std::collections::VecDeque;
-use std::sync::RwLock;
 use std::sync::atomic::{AtomicU64, Ordering};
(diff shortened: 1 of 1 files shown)


litebox/src/platform/mock.rs
```diff
@@ -31,6 +31,7 @@ use super::*;
 pub(crate) struct MockPlatform {
     current_time: AtomicU64,
     ip_packets: RwLock<VecDeque<Vec<u8>>>,
+    random: RwLock<crate::utils::rng::FastRng>,
```
jstarks: nit: might as well use `Mutex` here since we never acquire shared.

CvvT: Thanks! Fixed now.

PR #513, opened 2025-11-20
Use poll on tun

CvvT:
Use `poll` instead of busy waiting while waiting for incoming data on tun. It may cause delay for outgoing data, but overall it helps improve the throughput when tested against `iperf3`.


CvvT:
We are close to support `iperf3`. Besides #490 and #497, the only remaining bit is support of `CLOCK_PROCESS_CPUTIME_ID` and `getrusage`. It seems okay to just return errors for them though (at least `iperf3` won't complain about it).

PR #515, opened 2025-11-21
Remove remaining globals from Linux shim

jstarks:
Store all the global shim state in `GlobalState`.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -17,7 +17,7 @@ use litebox::platform::ThreadProvider;
 use litebox::platform::{Instant as _, SystemTime as _, TimeProvider};
 use litebox::platform::{
     PunchthroughProvider as _, PunchthroughToken as _, RawConstPointer as _, RawMutex as _,
-    RawMutexProvider as _, ThreadLocalStorageProvider as _,
+    ThreadLocalStorageProvider as _,
 };
 use litebox::sync::Mutex;
 use litebox::utils::TruncateExt as _;
@@ -146,7 +146,7 @@ pub(crate) enum ExitStatus {
 impl Process {
     /// Creates a new process with the given initial thread.
     fn new(global: &GlobalState, pid: i32, remote: Arc<ThreadRemote>) -> Self {
-        let nr_threads = global.platform.new_raw_mutex();
+        let nr_threads = <Platform as litebox::platform::RawMutexProvider>::RawMutex::INIT;
         nr_threads.underlying_atomic().store(1, Ordering::Relaxed);
         Self {
             nr_threads,
--- a/dev_tests/src/ratchet.rs
+++ b/dev_tests/src/ratchet.rs
@@ -25,7 +25,7 @@ fn ratchet_transmutes() -> Result<()> {
 fn ratchet_globals() -> Result<()> {
     ratchet(
         &[
-            ("litebox/", 7),
+            ("litebox/", 8),
             ("litebox_platform_linux_kernel/", 5),
             ("litebox_platform_linux_userland/", 5),
             ("litebox_platform_lvbs/", 20),
--- a/litebox/src/event/wait.rs
+++ b/litebox/src/event/wait.rs
@@ -80,7 +80,7 @@ impl<Platform: RawSyncPrimitivesProvider> WaitState<Platform> {
         Self {
             waker: Waker(Arc::new(WaitStateInner {
                 platform,
-                condvar: platform.new_raw_mutex(),
+                condvar: <Platform::RawMutex as RawMutex>::INIT,
             })),
             _phantom: PhantomData,
         }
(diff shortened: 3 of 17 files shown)


litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -213,6 +213,7 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
         shim.default_fs(in_mem, tar_ro)
     };
     shim.set_fs(initial_file_system);
+    let shim = shim.build();
```
wdcui: Minor: it might be helpful to call an instance of `ShimBuilder` `shim_builder` instead of `shim`?

litebox_shim_linux/src/lib.rs
```diff
@@ -203,41 +191,53 @@ impl LinuxShim {
...
+            load_filter: self.load_filter,
+            next_thread_id: 2.into(), // start from 2, as 1 is used by the main thread
+            litebox: self.litebox,
+        });
+        LinuxShim(global)
     }
+}
+
+#[derive(Clone)]
+pub struct LinuxShim(Arc<GlobalState>);
```
wdcui, line 206: Why do we need to store a reference to the `GlobalState` in both the shim and each task?

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -49,17 +49,20 @@ impl FsState {
 
 /// Task state shared by `CLONE_FILES`.
 pub(crate) struct FilesState {
+    pub global: alloc::sync::Arc<GlobalState>,
```
wdcui: If we initialize a task with a reference to the `GlobalState`, can we always access the global state from that path?

litebox_shim_linux/src/lib.rs
jstarks: The shim object here is what the runner uses to instantiate tasks (well, the first task). So it's the thing that creates the global state, which then gets passed to each task.

To put a different way, the `LinuxShim` is might even be gone by the time the tasks are running.

litebox_shim_linux/src/syscalls/file.rs
jstarks: We can. But given how `FilesState` is used, it's rather convenient to have another reference to `GlobalState` here rather than having to pass `&Task` to all the `FilesState` methods. And conceptually, I think it makes sense that `FilesState` can operate on files without necessarily having a `Task`.

Since the cost of this extra reference is quite low, I think this is probably OK.

wdcui: Can we use `with_current_task` to access the reference to the global state from `Task`? This way we can avoid adding a reference to `GlobalState` in many data structures?

litebox_shim_linux/src/lib.rs
wdcui: Got it. Thanks.

litebox_shim_linux/src/syscalls/file.rs
jstarks: Yes, we could. But we are very close to getting rid of `with_current_task`! And I'd like to do so, because that will make it much clearer which code must be called from a shim-created thread, and which can be called from an arbitrary platform thread. In the future, I expect the platform will need to call into the shim from non-shim threads for some IO-related use cases (we already have something like this for the networking code), and we don't want to accidentally call something that needs a `Task`.

jstarks: I removed this. Seems better your way, especially if we can remove a few more `LiteBox` references in the future via #523.

litebox_shim_linux/src/syscalls/epoll.rs
```diff
@@ -133,11 +138,12 @@ impl EpollFile {
     pub(crate) fn wait(
         &self,
         cx: &WaitContext<'_, Platform>,
+        global: &GlobalState,
```
wdcui, line 140: Should we make it consistent that `global` is the first argument after `self`?

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -114,576 +114,605 @@ struct SocketOptions {
 
 pub(crate) struct SocketOFlags(pub OFlags);
 
-fn initialize_socket(fd: &SocketFd, sock_type: SockType, flags: SockFlags) {
-    let mut status = OFlags::RDWR;
-    status.set(OFlags::NONBLOCK, flags.contains(SockFlags::NONBLOCK));
+impl GlobalState {
```
wdcui, line 121: It might be help to add a comment here explaining that the approach in network handling is different from other parts? Here we use `GlobalState` directly (because we don't have a struct for net) and we pass `global` to functions of other structs such as epoll.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -56,11 +52,11 @@ pub(crate) struct ThreadState {
 unsafe impl Send for ThreadState {}
 
 impl ThreadState {
-    pub fn new_process(pid: i32) -> Self {
+    pub fn new_process(globals: &GlobalState, pid: i32) -> Self {
```
wdcui: In other places, we call it `global` instead of `globals`?

litebox_shim_linux/src/syscalls/file.rs
wdcui: Thank you, John!

PR #516, opened 2025-11-21
Unify and improve userspace pointers

jstarks:
* Use a single userspace pointer type for all the platforms.
* Add a trait to allow the platforms to perform pointer validation (i.e. to ensure that the pointer is not in the kernel half of the address space), but do not implement any validation yet.
* Extend the exception table functionality to support Windows userspace, borrowing from the `trycopy` crate from OpenVMM.
* Use inline asm instead of a separate file to improve code generation.
* Use ordinary `mov`s for small accesses, guaranteeing that small, aligned accesses are atomic. Use this to fix the undefined behavior in the futex manager.

Future directions:
* Implement the appropriate validation for the kernel platforms.
* Add mechanisms to support SMAP.
* Improve memcpy performance--`rep movsb` is very slow for small accesses on many processors, even some new ones.


litebox/src/platform/common_providers/userspace_pointers.rs
```diff
@@ -1,8 +1,9 @@
 //! Userspace Pointer Abstraction with Fallible Memory Access
 //!
 //! This module implements fallible userspace pointers that can safely handle invalid
-//! memory accesses from userspace. The pointers use [`__memcpy_fallible`] internally,
-//! which relies on an exception table mechanism to recover from memory faults.
+//! memory accesses from userspace. The pointers use failable memory access routines
```
jaybosamiya-ms: Nit: `fallible`

litebox/src/platform/common_providers/userspace_pointers.rs
```diff
@@ -47,58 +98,71 @@ impl<T: Clone> core::fmt::Debug for UserConstPtr<T> {
 ///
 /// Note that this is fallible only if recovering from exceptions (e.g., page fault or SIGSEGV)
 /// is supported.
-unsafe fn read_at_offset<'a, T: Clone>(
+unsafe fn read_at_offset<'a, V: ValidateAccess, T: Clone>(
     ptr: *const T,
     count: isize,
 ) -> Option<alloc::borrow::Cow<'a, T>> {
+    let ptr = V::validate(ptr.cast_mut())?.cast_const();
     let src = unsafe { ptr.add(usize::try_from(count).ok()?) };
```
jaybosamiya-ms: The validation here is done before the offset; should we be doing it after too?  The fallible memcpy does help for part of it, but yeah, I think we may need the post-addition validation too, just in case user+off==kernel

jaybosamiya-ms: Actually, do we need the pre-validation?  Only the post-addition validation is necessary since that is where the read _actually_ occurs.

litebox/src/platform/common_providers/userspace_pointers.rs
```diff
@@ -47,58 +98,71 @@ impl<T: Clone> core::fmt::Debug for UserConstPtr<T> {
...
-    let mut data = core::mem::MaybeUninit::<T>::uninit();
-    let failed_bytes = unsafe {
-        __memcpy_fallible(
-            data.as_mut_ptr().cast(),
-            src.cast(),
-            core::mem::size_of::<T>(),
-        )
+    let val = unsafe {
+        match size_of::<T>() {
+            1 => core::mem::transmute_copy(
```
jaybosamiya-ms, line 113: Might be good to mention somewhere here that this is a perf optimization to split across these as 1,2,4,8,...

litebox/src/platform/trivial_providers.rs
```diff
@@ -120,9 +120,12 @@ impl<T: Clone> RawConstPointer<T> for TransparentConstPtr<T> {
...
         }
-        Some(alloc::borrow::Cow::Borrowed(unsafe {
-            &*self.inner.offset(count)
-        }))
+        Some(match size_of::<T>() {
+            // Try to ensure a single access for primitive types. The use of
+            // volatile here is dubious--this should really use inline asm.
+            1 | 2 | 4 | 8 => alloc::borrow::Cow::Owned(unsafe { self.inner.read_volatile() }),
+            _ => alloc::borrow::Cow::Borrowed(unsafe { &*self.inner.offset(count) }),
+        })
```
jaybosamiya-ms, line 129: Hmm, the scenario might be even more dubious (I believe the prev version had this issue too, not just new to the update, it is just that I also noticed that now): if the pointer is unaligned there are issues.

Also, fwiw, we don't need inline asm, we can use something like `AtomicU8::from_ptr(self.inner).load(...)`.  We also should probably wrap it with a `if p.is_aligned() { Atomic...load } else { read_unaligned(p) }` or similar?

litebox/src/platform/common_providers/userspace_pointers.rs
```diff
@@ -27,17 +28,67 @@
...
+/// access, just that it is in the userspace address range and won't be used to
+/// access kernel memory.
+pub trait ValidateAccess {
+    /// Validate that the given pointer is a valid userspace pointer.
+    ///
+    /// Returns `Some(ptr)` if valid. If the pointer is not valid, returns
+    /// `None` or `Some(invalid)` where `invalid` is adjusted to a valid
+    /// userspace address but will deterministically cause a fault on
+    /// access.
+    fn validate<T>(ptr: *mut T) -> Option<*mut T>;
```
jaybosamiya-ms, line 46: [No changes needed here, just discussion]

For this and for `validate_slice`, are we giving up perf / wasting some space by not having it be `Option<NonNull<T>>` which is the same size as `*mut T`?  I would need to give more thought towards variance though (`NonNull` essentially implies additional variance constraints) before suggesting we move to that.

For now, keeping it `Option<*mut T>` is good enough for this PR, but yeah, something that came to mind, just in case you have thoughts on it.

litebox/src/platform/common_providers/userspace_pointers.rs
jstarks: Yes, good catch.

litebox/src/platform/common_providers/userspace_pointers.rs
jstarks: I think in practice the validation function will be inlined (at least in the kernel case, where validation is simple), so this return value will never be in a register.

Interesting question for other places in the codebase. I especially wonder about some of the `Option<MutPtr<T>>` types in the various the syscall handling methods. Makes me wonder if we should have `NonNullRawPointer<T>` for such use cases (but I think this is probably impractical unless we move away from trait-based raw pointers).

litebox/src/platform/trivial_providers.rs
jstarks: Yeah, good call.

Somewhat related, I discovered that miri with stacked borrows (the default) gets mad if you use `AtomicU8` to read from a `u8`. miri with tree borrows thinks it's OK. I think the problem is that with stacked borrows, the type of borrow is evaluated when the reference is created--so when you transmute &u8 to &AtomicU8, it's a problem, because &AtomicU8 is a read-write borrow due to AtomicU8's interior mutability. But with tree borrows, the borrow is evaluated when the reference is used, not created, and an AtomicU8 load is a read, so that's OK.

Or something like that.

jstarks: Actually, looks like we already enforce that the value is aligned for the transparent pointer provider.

It's a bit laborious to fix this up and to switch to the atomic methods, so I'm going to skip it for now--nothing but the MockPlatform is actually using this code right now. Arguably, we should just use the userspace providers for the MockPlatform, too, but I wanted to leave it in at least until we close on whether the `Cow` stuff is worth keeping.

PR #518, opened 2025-11-22
Track threads per-process and kill them on exit

jstarks:
Now that we can reliably wake up threads that are waiting or are running in the guest, let's leverage that capability when terminating a process via `exit_group`. Track the list of threads in a process, and when a group exit is requested, mark all threads as exiting and interrupt them.

Also, provide a mechanism for a runner to wait on all threads for a process to exit and to retrieve the process exit code. Propagate this exit code to the userland runner's process, and update our tests to validate a successful exit.

Finally, update `execve` so that it terminates all other threads before loading the new program.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -205,9 +205,9 @@ impl Process {
 
             let nr_threads = self.nr_threads.underlying_atomic();
             let n = nr_threads.load(Ordering::Relaxed);
-            assert!(n > 0, "decrementing from zero threads");
-            nr_threads.store(n - 1, Ordering::Release);
-            if n == 1 {
+            let new_count = n.checked_sub(1).expect("decrementing from zero threads");
+            nr_threads.store(new_count, Ordering::Release);
+            if new_count == 0 {
                 assert!(inner.threads.is_empty());
                 // The last thread exited. Prevent new threads.
                 inner.group_exit = true;
@@ -216,7 +216,7 @@ impl Process {
             // Notify waiters if this is the last thread of the process
             // (`wait_for_exit`) or if this is the last thread being killed
             // during an exec (`kill_other_threads`).
-            n == 1 || (n == 2 && inner.is_killing_other_threads)
+            new_count == 0 || (new_count == 1 && inner.is_killing_other_threads)
         };
         if notify {
             self.nr_threads.wake_all();


litebox_shim_linux/src/wait.rs
```diff
@@ -11,18 +11,36 @@ impl WaitState {
...
+    pub(crate) fn prepare_to_run_guest(&self) -> bool {
+        self.wait_state
+            .0
+            .prepare_to_run_guest(|| !self.is_exiting())
+    }
 }
 
 impl litebox::event::wait::CheckForInterrupt for Task {
     fn check_for_interrupt(&self) -> bool {
-        // TODO: interrupt on signals, process exit, that kind of thing.
```
CvvT, line 25: Should we keep this TODO but remove the `process exit` part?

jstarks: Sure, good idea.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -3,67 +3,294 @@
...
+            }
+            let _ = self.nr_threads.block(n);
+        }
+        self.inner.lock().exit_code
+    }
+
+    /// Attaches a new thread to this process, returning a new remote state for
+    /// the thread.
+    fn attach_thread(&self, tid: i32) -> Option<Arc<ThreadRemote>> {
+        let remote = Arc::new(ThreadRemote::new());
```
wdcui, line 183: Minor: we can create a `remote` after checking the `inner`?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -3,67 +3,294 @@
...
+            assert!(n > 0, "decrementing from zero threads");
+            nr_threads.store(n - 1, Ordering::Release);
+            if n == 1 {
+                assert!(inner.threads.is_empty());
+                // The last thread exited. Prevent new threads.
+                inner.group_exit = true;
+            }
+
+            // Notify waiters if this is the last thread (`wait_for_exit`) or if
+            // this is the last thread from before an exec
```
wdcui: Minor: "from before an exec"?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -3,67 +3,294 @@
...
+            // (`kill_other_threads`).
+            n == 1 || (n == 2 && inner.is_killing_other_threads)
```
wdcui: In what case would we have `n==2` and `is_killing_other_threads`?

litebox_shim_linux/src/wait.rs
```diff
@@ -11,18 +11,36 @@ impl WaitState {
...
-        false
+        self.is_exiting()
```
wdcui, line 45: `check_for_interrupt` equals to `is_exiting` now because we don't handle other cases yet?

litebox_shim_linux/src/syscalls/process.rs
jstarks: We can, but this way we do the allocation outside the lock, which seems like the better tradeoff--the `None` case is rare.

litebox_shim_linux/src/syscalls/process.rs
jstarks: This happens while handling `execve`--we have to kill all the threads except the one calling `execve` before we can proceed.

litebox_shim_linux/src/wait.rs
jstarks: Yes. Restored the comment.

litebox_shim_linux/src/syscalls/process.rs
jstarks: Clarified.

litebox_shim_linux/src/syscalls/process.rs
jstarks: Added a comment.

litebox_shim_linux/src/syscalls/process.rs
wdcui: If only the thread calling execve is left, why would n be 2? Is it because we initialize n to be 1 when a process is created (before the first thread is attached)?

jstarks: 2 is the value before we decremented 1. So there's still 1 thread left.

I'll rewrite the code to make this clearer, so that we're looking at the new count (and thus checking for 0 or 1).

PR #523, opened 2025-11-25
Make `RawMutex`, `Mutex`, and `RwLock` const constructible

jstarks:
Currently, constructing a `RawMutex` required a `Platform` object. Similarly, constructing a `Mutex` or `RwLock` requires a `Synchronization` object. This is inconvenient for some use cases:

* It's not straightforward to initialize a mutex of any kind in a const context. Doing so would be helpful for avoiding runtime initialization overhead for some globals, and for defining a LiteBox version of `OnceLock` for lazy initialization of other globals. This would allow us to eliminate some uses of `spin::Once` and `once_cell::race`.
* It's hard to use different mutex types in the same project. This would be useful, for example, to allow using `LoanList<SpinRawMutex>` or `FutexManager<SpinRawMutex>` in a platform implementation of a wait-based `RawMutex`.
* It's hard to implement `Clone` for types containing mutexes--you have to preserve the `&'static Platform` pointer to make this possible, wasting space, or you have to implement `clone_with_platform`-style functions. The latter can't be used with `Arc::make_mut()`, etc.

These problems go away if the mutex types can be constructed out of thin air and in a const context.

Toward that end, this change updates the `RawMutex` trait to contain an `const INIT: Self` member that contains the const-initialized state of a `RawMutex`. It updates `Mutex` and `RwLock` to have `pub const fn new()` methods that take advantage of this.

To facilitate this, this changes updates the lock tracker to be a singleton, initialized at runtime with the platform to use for timekeeping and log output. This singleton is used for all `Mutex`/`RwLock` instances, regardless of whether they use the platform's `RawMutex`. This is slightly weird in the presence of multiple platform instances, but I don't think that's a real problem in practice.

This approach puts a small burden on platforms to ensure their `RawMutex` types are const-constructible without having access to the underlying `Platform` instance. In practice, I have not found any implementations where this is a problem, nor do I anticipate any such implementations. Intra-process synchronization primitives are a place where a capabilities-based approach doesn't seem to add much value.

So, although there may be other ways to solve the problems enumerated above, this seems like the right tradeoff.

To keep this PR scoped, I have not yet updated any callers to leverage these changes. A future change can eliminate the use of `litebox.sync()` across the codebase, eliminating a bunch of references to the `LiteBox` object.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -21,7 +21,7 @@ use litebox::platform::ThreadProvider;
 use litebox::platform::{Instant as _, SystemTime as _, TimeProvider};
 use litebox::platform::{
     PunchthroughProvider as _, PunchthroughToken as _, RawConstPointer as _, RawMutex as _,
-    RawMutexProvider as _, ThreadLocalStorageProvider as _,
+    ThreadLocalStorageProvider as _,
 };
 use litebox::sync::Mutex;
 use litebox::utils::TruncateExt as _;
@@ -150,8 +150,7 @@ pub(crate) enum ExitStatus {
 impl Process {
     /// Creates a new process with the given initial thread.
     fn new(pid: i32, remote: Arc<ThreadRemote>) -> Self {
-        let platform = litebox_platform_multiplex::platform();
-        let nr_threads = platform.new_raw_mutex();
+        let nr_threads = <Platform as litebox::platform::RawMutexProvider>::RawMutex::INIT;
         nr_threads.underlying_atomic().store(1, Ordering::Relaxed);
         Self {
             nr_threads,


litebox/src/platform/mock.rs
```diff
@@ -67,6 +67,16 @@ struct MockRawMutexInternalState {
 }
 
 impl MockRawMutex {
+    const fn new() -> Self {
+        Self {
+            inner: AtomicU32::new(0),
+            internal_state: std::sync::RwLock::new(MockRawMutexInternalState {
```
wdcui, line 73: Can we use `std` inside the litebox crate?

jstarks: Yes, for the mock platform only.

jstarks:
I'd like to get @jaybosamiya-ms's perspective on this before merging.

litebox/src/platform/mod.rs
```diff
@@ -216,12 +216,13 @@ where
...
     type RawMutex: RawMutex;
-    /// Allocate a new [`RawMutex`].
-    fn new_raw_mutex(&self) -> Self::RawMutex;
 }
 
 /// A raw mutex/lock API; expected to roughly match (or even be implemented using) a Linux futex.
-pub trait RawMutex: Send + Sync {
+pub trait RawMutex: Send + Sync + 'static {
+    /// A const initializer for this raw mutex.
+    const INIT: Self;
```
jaybosamiya-ms: I had considered this originally, but there are some _weird_ implications that show up if someone defines a weird initializer here, since `const` is effectively Rust's way of saying "inline", which means that the compiler actually could optimize the places back together from differing initializations.  This was one of the key reasons I had set it up as the previous thing.  In hindsight, I am not sure if this is a particularly important concern though, since I have not been able to cause such an optimization to be triggered in any scenario that contains a real `AtomicU32` within the initializer (it is easy to trigger if you're just storing a `u8` or something).  If someone actively is trying to break things by setting up a weird platform, they are to blame anyways.

So yeah, TL;DR: this is a reasonable initializer, but I wanted to document my thoughts on why I had not gone this initializer route in the past.

---

Sidenote: the doc-comment's "this" is a tiny bit confusing, it implies a specific raw mutex.  Perhaps "A const initializer for raw mutexes"?  It _is_ weird for it to not be thunk-like, but I couldn't figure out a way to phrase things better than that, but yeah, the "this" is def confusing.

litebox/src/sync/lock_tracing.rs
```diff
@@ -121,38 +123,64 @@ impl Locked {
...
 }
 
-pub(super) struct LockTracker<Platform: RawSyncPrimitivesProvider> {
-    x: alloc::sync::Arc<spin::Mutex<LockTrackerX<Platform>>>,
+struct LockTrackerPlatform<Platform: RawSyncPrimitivesProvider> {
+    platform: &'static Platform,
+    start_time: Platform::Instant,
+}
+
+trait DynLockTrackerProvider: Send + Sync {
```
jaybosamiya-ms, line 150: I'm not entirely sure why this particular trait was introduced, is it basically the instant->duration conversion?  Even so, I am still kinda hazy on why this is introduced as a trait.

litebox/src/sync/lock_tracing.rs
```diff
@@ -121,38 +123,64 @@ impl Locked {
...
+        LOCK_TRACKER.call_once(|| Self::new(platform));
+    }
+}
+
+static LOCK_TRACKER: spin::Once<LockTracker> = spin::Once::new();
+
+/// The main tracker, which manages both tracking and (if necessary) panicking upon invariant
+/// failure. Can/should only be accessed from the singleton that is automatically used upon usage of
+/// any of the `pub(crate)` functions available via the tracker.
+struct LockTrackerX<T: ?Sized + DynLockTrackerProvider = dyn DynLockTrackerProvider> {
```
jaybosamiya-ms: Nit: I'm mildly surprised by this doc comment move from `LockTracker` to here.  Also, I am ok with the `&self` being dropped from the methods on `LockTracker` such as `begin_lock_attempt` but I _mildly_ prefer the grouping that provides. Not crucial one way or the other tbqh, happy to leave things the way you've updated them.

litebox/src/platform/mod.rs
jstarks: Yeah, I see what you're getting at. I think since the type will have interior mutability (structurally due to the `AtomicU32`, but also in general how can you implement the contract without it?), the compiler can't merge instances of it.

Also, this is the same approach taken by the [`lock_api` crate](https://docs.rs/lock_api/0.4.14/lock_api/trait.RawMutex.html). That's probably where I got the idea.

litebox/src/sync/lock_tracing.rs
jstarks: I'll add a comment. The motivation here is to support a singleton lock tracker. `static`s can't be generic, and `Platform` must be a generic in the `litebox` crate. So, in order to have a singleton lock tracker, we have to register a trait object, so the underlying trait must be dyn compatible. This is not true of the platform traits here, so I defined a dyn-compatible trait with just what we need.

litebox/src/sync/lock_tracing.rs
jstarks: Yeah, this is probably the product of multiple revisions before I landed on something that worked. I'll fix this up.

PR #524, opened 2025-11-25
Implement thread signal support

jstarks:
Add a first layer of support for Linux signal handling. This includes support for:
* Signal delivery
* Signal handler configuration (`rt_sigaction`)
* Signal masks (`rt_sigprocmask`)
* Alternate signal stacks (`sigaltstack`)
* Returning from a signal handler (`rt_sigreturn`)
* Sending signals to self (`tkill`, `tgkill`)
* Delivering signals on hardware exceptions

Known to be missing:
* Cross-thread signaling
* Process-wide signals
* FPU/xsave state in signal contexts
* `SIGALRM` support (`alarm`, `setitimer`)
* Misc. signal syscalls (e.g., `rt_sigtimedwait`, `rt_sigpending`)
* `SIGSTOP` support

### DIFF
--- a/litebox_shim_linux/src/syscalls/signal/mod.rs
+++ b/litebox_shim_linux/src/syscalls/signal/mod.rs
@@ -188,12 +188,6 @@ impl PendingSignals {
         }
     }
 
-    fn find(&self, signal: Signal) -> Option<usize> {
-        self.queue
-            .iter()
-            .position(|info| info.signo == signal.as_i32())
-    }
-
     fn next(&self, blocked: SigSet) -> Option<Signal> {
         const EXCEPTION_SIGNALS: SigSet = SigSet::empty()
             .with(Signal::SIGSEGV)
@@ -299,7 +293,7 @@ impl SignalState {
     fn set_sigaltstack(&self, ss: SigAltStack) -> Result<(), Errno> {
         if !ss
             .flags
-            .difference(SsFlags::DISABLE | SsFlags::AUTODISARM)
+            .difference(SsFlags::DISABLE | SsFlags::ONSTACK | SsFlags::AUTODISARM)
             .is_empty()
         {
             Err(Errno::EINVAL)
@@ -311,7 +305,11 @@ impl SignalState {
         } else if ss.size < MINSIGSTKSZ {
             Err(Errno::ENOMEM)
         } else {
-            self.altstack.set(ss);
+            self.altstack.set(SigAltStack {
+                sp: ss.sp,
+                flags: ss.flags & SsFlags::AUTODISARM,
+                size: ss.size,
+            });
             Ok(())
         }
     }
@@ -327,6 +325,7 @@ impl SignalState {
 
     fn deliver_signal(
         &self,
+        signal: Signal,
         siginfo: &Siginfo,
         action: &SigAction,
         ctx: &mut PtRegs,
@@ -345,15 +344,19 @@ impl SignalState {
 
         let frame_addr = arch::get_signal_frame(sp, action);
 
-        self.set_signal_mask(action.mask);
-
         if (switch_stacks || on_alt_stack) && !is_on_stack(&altstack, frame_addr) {
             return Err(DeliverFault);
         }
 
         self.write_signal_frame(frame_addr, siginfo, action, ctx)?;
 
-        if switch_stacks && altstack.flags.contains(SsFlags::AUTODISARM) {
+        let mut mask = self.blocked.get() | action.mask;
+        if !action.flags.contains(SaFlags::NODEFER) {
+            mask.add(signal);
+        }
+        self.set_signal_mask(mask);
+
+        if altstack.flags.contains(SsFlags::AUTODISARM) {
             self.clear_sigaltstack();
         }
         Ok(())
@@ -570,7 +573,9 @@ impl Task {
                 }
                 SIG_IGN => {}
                 _ => {
-                    if let Err(DeliverFault) = self.signals.deliver_signal(&siginfo, &action, ctx) {
+                    if let Err(DeliverFault) =
+                        self.signals.deliver_signal(signal, &siginfo, &action, ctx)
+                    {
                         // Failed to deliver signal. Inject a SIGSEGV
                         // (terminating the process if we were trying to deliver
                         // a SIGSEGV).
--- a/litebox_shim_linux/src/syscalls/signal/x86.rs
+++ b/litebox_shim_linux/src/syscalls/signal/x86.rs
@@ -174,6 +174,7 @@ impl SignalState {
             ctx.edx = 0;
             ctx.ecx = 0;
         }
+        ctx.eflags &= !litebox_common_linux::EFLAGS_DF;
         Ok(())
     }
 }
--- a/litebox_shim_linux/src/syscalls/signal/x86_64.rs
+++ b/litebox_shim_linux/src/syscalls/signal/x86_64.rs
@@ -105,6 +105,7 @@ impl SignalState {
         ctx.rsi = frame_addr.wrapping_add(offset_of!(SignalFrame, siginfo));
         ctx.rdx = frame_addr.wrapping_add(offset_of!(SignalFrame, ucontext));
         ctx.rax = 0;
+        ctx.eflags &= !litebox_common_linux::EFLAGS_DF;
         Ok(())
     }
 }
--- a/litebox_common_linux/src/lib.rs
+++ b/litebox_common_linux/src/lib.rs
@@ -3035,6 +3035,9 @@ pub struct PtRegs {
     pub xss: usize,
 }
 
+#[cfg(any(target_arch = "x86_64", target_arch = "x86"))]
+pub const EFLAGS_DF: usize = 0x400;
+
 impl PtRegs {
     /// Get the `idx`th syscall argument.
     ///
--- a/litebox_platform_lvbs/src/host/linux.rs
+++ b/litebox_platform_lvbs/src/host/linux.rs
@@ -54,9 +54,6 @@ pub struct Timespec {
     pub tv_nsec: i64,
 }
 
-#[allow(non_camel_case_types)]
-pub type sigset_t = ::core::ffi::c_ulong;
-
 const BITS_PER_LONG: usize = 64;
 
 #[repr(C)]


jstarks:
@CvvT, how important is the `alarm`/`setitimer` functionality that this removes? Do we need to add that back before this can be merged?

CvvT:
> @CvvT, how important is the `alarm`/`setitimer` functionality that this removes? Do we need to add that back before this can be merged?

Not necessary. They are only needed for running UnixBench, which I haven’t upstreamed yet. It’s fine to remove them.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -366,9 +354,9 @@ pub trait HostInterface {
 
     /// For Punchthrough
```
wdcui: Should we remove this comment since it's not a punchthrough anymore?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -626,7 +626,7 @@ exception_callback:
     mov esp, gs:host_sp@ntpoff
     mov ebp, gs:host_bp@ntpoff
 
-    mov ecx, [esp] // pass thread_ctx
+    mov edi, [esp] // pass thread_ctx
```
wdcui, line 629: This bug didn't cause any problems because we don't have a test that triggers this path?

litebox_shim_linux/src/syscalls/signal/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+                sp: 0,
+                size: 0,
+            }
+            .into(),
+            // Preserve last exception
+            last_exception: self.last_exception.clone(),
+        }
+    }
+
+    pub(crate) fn reset_for_exec(&self, litebox: &LiteBox<Platform>) {
```
wdcui, line 86: What if this function is called simultaneously by multiple callers?

litebox_shim_linux/src/syscalls/signal/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+            // No entry. Synthesize a default siginfo.
+            Siginfo {
+                signo: signal.as_i32(),
+                errno: 0,
+                code: SI_USER,
+                data: SiginfoData { pad: [0; 29] },
+            }
+        };
+
+        if self.find(signal).is_none() {
```
wdcui: Do we expect that the `queue` could have multiple `siginfo` for the same signal?

litebox_shim_linux/src/syscalls/signal/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+            handler.immutable = true;
+        }
+    }
+
+    pub(crate) fn handle_exception_request(&self, info: &litebox::shim::ExceptionInfo) {
+        let signal = match info.exception {
+            Exception::DIVIDE_ERROR => Signal::SIGFPE,
+            Exception::BREAKPOINT => Signal::SIGTRAP,
+            Exception::INVALID_OPCODE => Signal::SIGILL,
+            _ => Signal::SIGSEGV,
```
wdcui, line 645: Is it okay to treat all other cases as `SIGSEGV`?

litebox_shim_linux/src/syscalls/signal/x86.rs
```diff
@@ -0,0 +1,223 @@
...
+
+impl SignalState {
+    pub(super) fn write_signal_frame(
+        &self,
+        frame_addr: usize,
+        siginfo: &Siginfo,
+        action: &SigAction,
+        ctx: &mut PtRegs,
+    ) -> Result<(), DeliverFault> {
+        if !action.flags.contains(SaFlags::RESTORER) {
```
wdcui, line 93: To double check, should we error when the `RESTORER` flag is set or not?

litebox_platform_linux_userland/src/lib.rs
jstarks: Right. Now that signals are hooked up, we do have a test for this, though.

litebox_shim_linux/src/syscalls/signal/mod.rs
jstarks: That's fine, if there are multiple processes sharing a signal handler set (e.g., because of multiple concurrent vforks that passed CLONE_SIGHAND for some reason), then each of them will get a unique set of reset handlers, and the other processes sharing the signal handlers will not see any modifications at all. 

(In practice this will never happen, lots of this stuff is legacy of the pre-POSIX Linux threading model... We could choose not to support it by requiring that the signal handlers be per-process (so no CLONE_SIGHAND without CLONE_THREAD). Not sure if that would simplify code/locking much or not.)

litebox_shim_linux/src/syscalls/signal/mod.rs
jstarks: Yes, this is explicitly possible for RT signals (but not "legacy" signals).

litebox_shim_linux/src/syscalls/signal/mod.rs
jstarks: That seems to be the behavior of Linux--most things get folded into `SIGSEGV`. I'm not sure I got all the weird FP exceptions right, though. Hard to test that right now since we don't support FPU state.

litebox_shim_linux/src/syscalls/signal/x86.rs
jstarks: I'll add a clarifying comment.

litebox_shim_linux/src/syscalls/signal/mod.rs
wdcui: Got it. Thanks.

litebox_shim_linux/src/syscalls/signal/mod.rs
wdcui: Sounds good for now!

litebox_shim_linux/src/syscalls/signal/x86.rs
wdcui: Thanks!

litebox_shim_linux/src/syscalls/signal/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+        let siginfo = if let Some(pos) = self.find(signal) {
+            self.queue.remove(pos).unwrap()
+        } else {
+            // No entry. Synthesize a default siginfo.
+            Siginfo {
+                signo: signal.as_i32(),
+                errno: 0,
+                code: SI_USER,
+                data: SiginfoData { pad: [0; 29] },
+            }
```
CvvT: We can directly remove the signal from `pending` and return `Siginfo` here without calling `find` again?

Why `remove` needs to return a default siginfo? Looks like it only happens to `SIGKILL` and `SIGSTOP`.

litebox_platform_linux_kernel/src/lib.rs
jstarks: Probably we can remove this host method altogether.

litebox_shim_linux/src/syscalls/signal/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+    /// Updates the blocked signal mask.
+    fn set_signal_mask(&self, mask: SigSet) {
+        self.blocked.set(mask);
+    }
+
+    /// Sets the alternate signal stack.
+    fn set_sigaltstack(&self, ss: SigAltStack) -> Result<(), Errno> {
+        if !ss
+            .flags
+            .difference(SsFlags::DISABLE | SsFlags::AUTODISARM)
```
CvvT: It's okay to have `ONSTACK` flag: https://elixir.bootlin.com/linux/v5.19.17/source/kernel/signal.c#L4193

litebox_shim_linux/src/syscalls/signal/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+            && !altstack.flags.contains(SsFlags::DISABLE);
+        let sp = if switch_stacks {
+            altstack.sp + altstack.size
+        } else {
+            sp
+        };
+
+        let frame_addr = arch::get_signal_frame(sp, action);
+
+        self.set_signal_mask(action.mask);
```
CvvT:
```
The signal being delivered is also added to the signal mask, 
unless SA_NODEFER was specified when registering the handler. 
These signals are thus blocked while the handler executes.
```
https://man7.org/linux/man-pages/man7/signal.7.html

litebox_shim_linux/src/syscalls/signal/x86_64.rs
```diff
@@ -0,0 +1,169 @@
...
+                    err: last_exception.error_code.into(),
+                    trapno: last_exception.exception.0.into(),
+                    oldmask: self.blocked.get().as_u64(),
+                    cr2: last_exception.cr2 as u64,
+                    fpstate: 0, // TODO
+                    reserved1: [0; 8],
+                },
+                sigmask: self.blocked.get(),
+            },
+            siginfo: siginfo.clone(),
```
CvvT, line 94: https://man7.org/linux/man-pages/man7/signal.7.html: `If the signal handler was installed using the
                   [sigaction(2)](https://man7.org/linux/man-pages/man2/sigaction.2.html) SA_SIGINFO flag, then the above
                   information is accessible via the ucontext_t object
                   that is pointed to by the third argument of the signal
                   handler.`

https://elixir.bootlin.com/linux/v5.19.17/source/arch/x86/kernel/signal.c#L490

We don't need to pass `siginfo` if `SIGINFO` is not specified. We don't have to follow it though.

jstarks: Yeah, I don't think it's worth the trouble to skip it.

litebox_shim_linux/src/syscalls/signal/mod.rs
jstarks: Good catch.

litebox_shim_linux/src/syscalls/signal/mod.rs
jstarks: Fixed this.

PR #527, opened 2025-11-26
Close pending sockets

CvvT:
#497 adds support for graceful close so that `close` becomes blocking when `SO_LINGER` is enabled and the socket has pending data to be sent. However, there is no notification after all the data are transferred, and thus `close` might be blocked indefinitely.

To fix it, we need to check and close those sockets when performing `check_and_update_events`. Closed sockets would then send out `Events::HUP` to unblock the caller. To determine whether a socket is closed or not, we should use `is_open` instead of just checking `state() == Closed` because `TimeWait` is also considered as closed (which is handled by `is_open`).

### DIFF
--- a/litebox_runner_linux_userland/src/lib.rs
+++ b/litebox_runner_linux_userland/src/lib.rs
@@ -204,10 +204,11 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
         in_mem.with_root_privileges(|fs| {
             let mode = Mode::RWXU | Mode::RWXG | Mode::RWXO;
             if let Err(err) = fs.mkdir("/tmp", mode) {
-                if let litebox::fs::errors::MkdirError::AlreadyExists = err {
-                    fs.chmod("/tmp", mode).expect("Failed to call chmod");
-                } else {
-                    panic!();
+                match err {
+                    litebox::fs::errors::MkdirError::AlreadyExists => {
+                        fs.chmod("/tmp", mode).expect("Failed to call chmod");
+                    }
+                    _ => panic!(),
                 }
             }
         });


litebox/src/net/mod.rs
```diff
@@ -636,14 +694,47 @@ where
...
+                    |udp_socket| {
+                        let has_pending_data = udp_socket.is_open() && udp_socket.send_queue() > 0;
+                        if !has_pending_data {
+                            udp_socket.close();
+                        }
+                        !has_pending_data
+                    },
+                );
+                if closed {
+                    socket_handle.consider_closed = false;
```
wdcui, line 733: When will the socket handle be removed from the descriptor table?

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -204,11 +204,10 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
         in_mem.with_root_privileges(|fs| {
             let mode = Mode::RWXU | Mode::RWXG | Mode::RWXO;
             if let Err(err) = fs.mkdir("/tmp", mode) {
-                match err {
-                    litebox::fs::errors::MkdirError::AlreadyExists => {
-                        fs.chmod("/tmp", mode).expect("Failed to call chmod");
-                    }
-                    _ => panic!(),
+                if let litebox::fs::errors::MkdirError::AlreadyExists = err {
```
wdcui: If this bug fix is unrelated, should we move it to a separate PR?

CvvT: This is not bug fix; it addresses some clippy warning.

litebox/src/net/mod.rs
CvvT: After this line, this socket would emit `Events::HUP`. if `close` is called again on the socket, it would succeed and then remove the socket handle.

litebox_runner_linux_userland/src/lib.rs
wdcui: Why does it trigger warnings in this PR when you don't touch this file?

CvvT: Interestingly, I reverted the change but CI does not complain about it. Will fix it in a separate PR.

CvvT:
> LGTM. Thank you for the quick fix, Weiteng. Should we reenable the flaky tcp test in CI since this PR fixes it?

We didn't disable it.

wdcui:
I resolved my comments. Please feel free to merge it.

PR #528, opened 2025-11-26
Support for OP-TEE ldelf

sangho2:
This PR adds support for OP-TEE `ldelf` to load and run TAs without relying on a (kernel-mode) elf loader. This includes adding several OP-TEE-specific syscalls for memory map and binary loading with padding, changing syscall table, and allocating guest TLS to overcome the compatibility issue introduced by the compiler (i.e., stack guard with TLS). For now, it uses the `elf_loader` crate to load `ldelf`.

### DIFF
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -103,11 +103,7 @@ pub fn get_ta_base_addr() -> Option<usize> {
 ///
 /// # Safety
 /// Ensure that `dst` is valid for `count` bytes.
-unsafe fn read_ta_bin(
-    dst: litebox::platform::common_providers::userspace_pointers::UserMutPtr<u8>,
-    offset: usize,
-    count: usize,
-) -> Option<()> {
+unsafe fn read_ta_bin(dst: UserMutPtr<u8>, offset: usize, count: usize) -> Option<()> {
     with_current_task(|task| {
         if let Some(ta_bin) = task.ta_bin.as_ref() {
             let end_offset = offset.checked_add(count)?;
@@ -135,9 +131,14 @@ pub(crate) fn litebox_page_manager<'a>() -> &'a PageManager<Platform, PAGE_SIZE>
     VMEM.get_or_init(|| alloc::boxed::Box::new(PageManager::new(litebox())))
 }
 
-// Convenience type aliases
-type ConstPtr<T> = <Platform as litebox::platform::RawPointerProvider>::RawConstPointer<T>;
-type MutPtr<T> = <Platform as litebox::platform::RawPointerProvider>::RawMutPointer<T>;
+type UserMutPtr<T> = litebox::platform::common_providers::userspace_pointers::UserMutPtr<
+    litebox::platform::common_providers::userspace_pointers::NoValidation,
+    T,
+>;
+type UserConstPtr<T> = litebox::platform::common_providers::userspace_pointers::UserConstPtr<
+    litebox::platform::common_providers::userspace_pointers::NoValidation,
+    T,
+>;
 
 pub struct OpteeShim;
 
@@ -455,10 +456,10 @@ fn handle_ldelf_syscall_request(ctx: &mut litebox_common_linux::PtRegs) -> Conti
 #[inline]
 fn handle_cipher_update_or_final<F>(
     state: TeeCrypStateHandle,
-    src: ConstPtr<u8>,
+    src: UserConstPtr<u8>,
     src_len: usize,
-    dst: MutPtr<u8>,
-    dst_len: MutPtr<u64>,
+    dst: UserMutPtr<u8>,
+    dst_len: UserMutPtr<u64>,
     syscall_fn: F,
 ) -> Result<(), TeeResult>
 where
--- a/litebox_shim_optee/src/loader/elf.rs
+++ b/litebox_shim_optee/src/loader/elf.rs
@@ -18,7 +18,7 @@ use once_cell::race::OnceBox;
 use thiserror::Error;
 
 use super::ElfLoadInfo;
-use crate::MutPtr;
+use crate::UserMutPtr;
 
 #[cfg(feature = "platform_linux_userland")]
 use crate::litebox_page_manager;
@@ -204,7 +204,7 @@ impl elf_loader::mmap::Mmap for ElfLoaderMmap {
                 .expect("fd_elf_map.read failed");
 
             crate::syscalls::mm::sys_mprotect(
-                MutPtr::from_usize(mapped_addr),
+                UserMutPtr::from_usize(mapped_addr),
                 len,
                 litebox_common_linux::ProtFlags::from_bits(prot.bits()).ok_or(
                     elf_loader::Error::MmapError {
@@ -357,7 +357,8 @@ fn load_trampoline(trampoline: TrampolineHdr, relo_off: usize, fd: i32) -> usize
     // TODO: For now, we unmap `ldelf`'s memory area to load the trampoline.
     // TAs might interact with `ldelf` to use its critical syscalls, so we might need to
     // figure out how to deal with this potential memory area overlap.
-    let _ = crate::syscalls::mm::sys_munmap(MutPtr::from_usize(start_addr), end_addr - start_addr);
+    let _ =
+        crate::syscalls::mm::sys_munmap(UserMutPtr::from_usize(start_addr), end_addr - start_addr);
     let ret = unsafe {
         ElfLoaderMmap::mmap(
             Some(start_addr),
@@ -386,7 +387,7 @@ fn load_trampoline(trampoline: TrampolineHdr, relo_off: usize, fd: i32) -> usize
     unsafe {
         placeholder.write(litebox_platform_multiplex::platform().get_syscall_entry_point());
     }
-    let ptr = crate::MutPtr::from_usize(start_addr);
+    let ptr = UserMutPtr::from_usize(start_addr);
     let pm = litebox_page_manager();
     unsafe { pm.make_pages_executable(ptr, end_addr - start_addr) }
         .expect("failed to make pages executable");
--- a/litebox_shim_optee/src/syscalls/ldelf.rs
+++ b/litebox_shim_optee/src/syscalls/ldelf.rs
@@ -1,5 +1,5 @@
+use crate::UserMutPtr;
 use litebox::mm::linux::PAGE_SIZE;
-use litebox::platform::common_providers::userspace_pointers::UserMutPtr;
 use litebox::platform::{RawConstPointer, RawMutPointer};
 use litebox_common_linux::{MapFlags, ProtFlags};
 use litebox_common_optee::{LdelfMapFlags, TeeResult, TeeUuid};
@@ -20,7 +20,7 @@ const DUMMY_HANDLE: u32 = 1;
 /// Memory regions between `start - pad_begin` and `start` and between
 /// `start + num_bytes` and `start + num_bytes + pad_end` are reserved and must not be used.
 pub fn sys_map_zi(
-    va: crate::MutPtr<usize>,
+    va: UserMutPtr<usize>,
     num_bytes: usize,
     pad_begin: usize,
     pad_end: usize,
@@ -83,7 +83,7 @@ pub fn sys_map_zi(
 
 /// OP-TEE's syscall to open a TA binary.
 #[expect(clippy::unnecessary_wraps)]
-pub fn sys_open_bin(ta_uuid: TeeUuid, handle: crate::MutPtr<u32>) -> Result<(), TeeResult> {
+pub fn sys_open_bin(ta_uuid: TeeUuid, handle: UserMutPtr<u32>) -> Result<(), TeeResult> {
     // TODO: This function requires an RPC from the secure world to the normal world to
     // open the TA binary identified by `ta_uuid` and return a handle to it in `handle`.
     // Since we don't have RPC implementation yet, we just return a dummy handle value.
@@ -125,7 +125,7 @@ pub fn sys_close_bin(handle: u32) -> Result<(), TeeResult> {
 
 /// OP-TEE's syscall to map a portion of a TA binary into memory.
 pub fn sys_map_bin(
-    va: crate::MutPtr<usize>,
+    va: UserMutPtr<usize>,
     num_bytes: usize,
     handle: u32,
     offs: usize,
--- a/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
+++ b/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
@@ -1,5 +1,4 @@
 use litebox::mm::linux::{PageFaultError, PageRange, VmFlags, VmemPageFaultHandler};
-use litebox::platform::common_providers::userspace_pointers::UserMutPtr;
 use litebox::platform::page_mgmt;
 use x86_64::{
     PhysAddr, VirtAddr,
@@ -17,6 +16,7 @@ use x86_64::{
     },
 };
 
+use crate::UserMutPtr;
 use crate::mm::{
     MemoryProvider,
     pgtable::{PageTableAllocator, PageTableImpl},
@@ -96,7 +96,7 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
                 .expect("Failed to handle page fault");
             }
         }
-        UserMutPtr::from_usize(range.start)
+        UserMutPtr::from_ptr(range.start as *mut u8)
     }
 
     /// Unmap 4KiB pages from the page table
@@ -207,7 +207,7 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
             new_start += 1;
         }
 
-        Ok(UserMutPtr::from_usize(new_range.start))
+        Ok(UserMutPtr::from_ptr(new_range.start as *mut u8))
     }
 
     pub(crate) unsafe fn mprotect_pages(
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -13,7 +13,6 @@ use core::{
     arch::asm,
     sync::atomic::{AtomicU32, AtomicU64},
 };
-use litebox::platform::common_providers::userspace_pointers::{UserConstPtr, UserMutPtr};
 use litebox::platform::page_mgmt::DeallocationError;
 use litebox::platform::{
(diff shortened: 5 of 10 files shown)


litebox_shim_optee/src/loader/elf.rs
```diff
@@ -343,6 +344,10 @@ fn load_trampoline(trampoline: TrampolineHdr, relo_off: usize, fd: i32) -> usize
     let start_addr = relo_off + trampoline.vaddr;
     let end_addr = (start_addr + trampoline.size).next_multiple_of(PAGE_SIZE);
     let mut need_copy = false;
+    // TODO: For now, we unmap `ldelf`'s memory area to load the trampoline.
```
jstarks, line 357: I don't understand why we need to do this. Is this because `ldelf` thinks the VA range used by the trampoline is unused and maps the TA (or heap or stack or something) on top of it?

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -0,0 +1,279 @@
+use litebox::mm::linux::PAGE_SIZE;
+use litebox::platform::{RawConstPointer, RawMutPointer};
+use litebox_common_linux::{MapFlags, ProtFlags};
+use litebox_common_optee::{LdelfMapFlags, TeeResult, TeeUuid};
+
+#[inline]
+fn align_up(addr: usize, align: usize) -> usize {
```
jstarks: This can just be `addr.next_multiple_of(align)`

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -0,0 +1,279 @@
...
+            .checked_add(pad_begin)
+            .and_then(|t| t.checked_add(pad_end))
+            .ok_or(TeeResult::BadParameters)?,
+        PAGE_SIZE,
+    );
+    if (*addr).checked_add(total_size).is_none() {
+        return Err(TeeResult::BadParameters);
+    }
+    let prot = ProtFlags::PROT_READ_WRITE;
+    let flags = if flags.contains(LdelfMapFlags::LDELF_MAP_FLAG_SHAREABLE) {
```
jstarks: What does this `LDELF_MAP_FLAG_SHAREABLE` really mean? It doesn't quite seem to correspond to `MAP_SHARED` since there's no underlying object that we're mapping from here...

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -0,0 +1,279 @@
...
+    litebox::log_println!(
+        litebox_platform_multiplex::platform(),
+        "sys_map_zi: va {:#x} (addr {:#x}), num_bytes {}, flags {:#x}",
+        va.as_usize(),
+        *addr,
+        num_bytes,
+        flags
+    );
+
+    if flags.contains(LdelfMapFlags::LDELF_MAP_FLAG_WRITEABLE)
```
jstarks: I think you want to fail if there is any flag other than `LDELF_MAP_FLAG_SHAREABLE`.

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -0,0 +1,279 @@
...
+                },
+                align_up(
+                    num_bytes + padded_start - align_down(padded_start, PAGE_SIZE),
+                    PAGE_SIZE,
+                ),
+                prot,
+            )
+            .expect("sys_map_zi: failed to set memory protection");
+            unsafe {
+                core::ptr::write_bytes(padded_start as *mut u8, 0, num_bytes);
```
jstarks: This needs to go through a raw pointer or it will fail when we enable SMAP.

jstarks: But also--this seems unnecessary, because freshly mapped memory will always be zero.

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -0,0 +1,279 @@
...
+                let _ = va.write_at_offset(0, padded_start);
+            }
+            Ok(())
+        }
+        Err(_) => Err(TeeResult::OutOfMemory),
```
jstarks: nit: use `.map_err()?`

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -0,0 +1,279 @@
...
+        offs,
+        pad_begin,
+        pad_end,
+        flags
+    );
+
+    assert!(handle == DUMMY_HANDLE, "invalid handle");
+    // TODO: check whether `handle` is valid
+
+    if flags.contains(LdelfMapFlags::LDELF_MAP_FLAG_SHAREABLE)
```
jstarks, line 168: Check for unknown/unsupported flags

litebox_shim_optee/src/lib.rs
```diff
@@ -66,6 +71,55 @@ pub fn get_session_id() -> u32 {
...
+///
+/// # Safety
+/// Ensure that `dst` is valid for `count` bytes.
+unsafe fn read_ta_bin(dst: *mut u8, offset: usize, count: usize) -> Option<()> {
+    with_current_task(|task| {
+        if let Some(ta_bin) = task.ta_bin.as_ref() {
+            let end_offset = offset.checked_add(count)?;
+            if end_offset <= ta_bin.len() {
+                unsafe {
+                    core::ptr::copy_nonoverlapping(ta_bin.as_ptr().add(offset), dst, count);
```
jstarks: Need to use user ptr.

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -0,0 +1,279 @@
...
+
+            let mut prot = ProtFlags::PROT_READ;
+            if flags.contains(LdelfMapFlags::LDELF_MAP_FLAG_WRITEABLE) {
+                prot |= ProtFlags::PROT_WRITE;
+            } else if flags.contains(LdelfMapFlags::LDELF_MAP_FLAG_EXECUTABLE) {
+                prot |= ProtFlags::PROT_EXEC;
+            }
+            crate::syscalls::mm::sys_mprotect(
+                litebox::platform::common_providers::userspace_pointers::UserMutPtr {
+                    inner: align_down(padded_start, PAGE_SIZE) as *mut u8,
```
jstarks: Use `from_usize` on the platform's user pointer.

litebox_shim_optee/src/loader/elf.rs
sangho2: Something like that. Without this, `load_trampoline()` fails because this area is already mapped for `ldelf`.

litebox_shim_optee/src/syscalls/ldelf.rs
sangho2: I think there is no significant difference between `LDELF_MAP_FLAG_SHAREABLE` and `MAP_SHARED`.

litebox_shim_optee/src/syscalls/ldelf.rs
sangho2: yes. we can remove this.

litebox_shim_optee/src/syscalls/ldelf.rs
sangho2: Dispatcher uses `from_bit_truncate` and thus unknown bits are discarded. We might need to have a checker there if needed.

litebox_shim_optee/src/syscalls/ldelf.rs
jstarks: But what does shared mean for a "map zi" operation, since OP-TEE doesn't support fork? How would those pages ever been shared with anything, so that they would be meaningfully different from anonymous/non-shared?

sangho2: Checked the doc again. You are right. Basically, it is about controlling cache behaviors, not memory sharing. dropped `MAP_SHARED`.

sangho2: This might be more about whether LiteBox should allow the guest to have more control over page tables. Currently, it only exposes Linux's `MapFlags` and  `ProtFlags` to the guest. It doesn't provide a way to control something like `PCD` and `PWT`. Linux applications don't need them, but OP-TEE TAs do.

litebox_shim_optee/src/syscalls/ldelf.rs
jstarks: I recommend against using truncating version of from_bits. The reason being that if you add another flag later it becomes a breaking change--you have to remember to find and update _all_ locations that call `from_bits_truncate`. That's hard to remember to do.

It's much safer for each call site to perform the validation of exactly which flags it supports. It's also clearer to the reader of the code.

sangho2: I don't have a strong preference here, but, anyhow, adding another flag would be a breaking change and could affect all other code regardless of whether we use `from_bits_retain` or `from_bits_trunate`.

litebox_shim_optee/src/syscalls/ldelf.rs
jstarks: Looking a little closer at the OP-TEE source, it looks like this is used to support one TA calling another--in that case, it looks like the kernel should implicitly map the parameter buffers from the source TA into the destination TA, and that this is only allowed for buffers mapped with `*_SHAREABLE`. So, I think that does imply that it's similar to `MAP_SHARED`, but with a kind of implicit creation of a shared memory object.

For now, since we don't support one TA calling another TA, I think we can ignore this flag.

litebox_shim_optee/src/syscalls/ldelf.rs
jstarks: I don't agree--if we consistently use `from_bits_retain` (or maybe better yet, just use `const _ = !0` like we do in the Linux code, so that there's no difference between retain and truncate), then defining a new flag is not a breaking change.

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -176,80 +166,67 @@ pub fn sys_map_bin(
...
+    } else if flags.contains(LdelfMapFlags::LDELF_MAP_FLAG_EXECUTABLE) {
+        prot |= ProtFlags::PROT_EXEC;
+    }
+    crate::syscalls::mm::sys_mprotect(
+        UserMutPtr::from_usize(align_down(padded_start, PAGE_SIZE)),
+        (num_bytes + padded_start - align_down(padded_start, PAGE_SIZE))
+            .next_multiple_of(PAGE_SIZE),
+        prot,
+    )
+    .expect("sys_map_bin: failed to set memory protection");
```
jstarks: Why not propagate the error?

litebox_shim_optee/src/syscalls/ldelf.rs
sangho2: I meant, if there are unknown flags, there are two options: ignore them or return an error. If we decide to return an error, running a TA assuming an unknown flag is not possible anyhow. We did use `const _ = !0` for OP-TEE code as well.

litebox_shim_optee/src/syscalls/ldelf.rs
sangho2: yes. time to minimize panics.

litebox_shim_optee/src/syscalls/ldelf.rs
jstarks: The point is to be able to tell at the use site which bits are supported for a given operation. If you have to consult two places, the type definition and the code using the type, then it's very easy to introduce regressions to existing code paths when a new field is added.

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -150,6 +161,17 @@ pub fn sys_map_bin(
         flags
     );
 
+    // OP-TEE ignores unknown flags. Log and truncate them.
```
jstarks: That doesn't appear to be true--at least for this syscall, OP-TEE returns `TEE_ERROR_BAD_PARAMETERS` for flags other than
```
LDELF_MAP_FLAG_SHAREABLE |
				      LDELF_MAP_FLAG_WRITEABLE |
				      LDELF_MAP_FLAG_BTI |
				      LDELF_MAP_FLAG_EXECUTABLE
```

sangho2: Got the point.

litebox_shim_optee/src/syscalls/ldelf.rs
sangho2: Looks like `LDELF_MAP_FLAG_SHAREABLE` has two different meanings. `map_zi` checks it to create either shareable or cacheable mapping. `map_bin` uses it to see whether it can reuse already mapped file slices. Anyhow, both are no-ops for now.

litebox_common_optee/src/lib.rs
```diff
@@ -707,6 +707,11 @@ impl TeeMemoryAccessRights {
             .map_err(|_| Errno::EINVAL)
             .and_then(|v| Self::from_bits(v).ok_or(Errno::EINVAL))
     }
+    pub fn try_from_usize_retain(value: usize) -> Result<Self, Errno> {
```
jstarks: Do we still need this?

litebox_common_optee/src/lib.rs
```diff
@@ -916,3 +921,170 @@ pub enum UserTaPropType {
...
+        const TEE_MATTR_PRWX = Self::TEE_MATTR_PRW.bits() | Self::TEE_MATTR_PX.bits();
+        const TEE_MATTR_UR = 0x80;
+        const TEE_MATTR_UW = 0x100;
+        const TEE_MATTR_UX = 0x200;
+        const TEE_MATTR_URW = Self::TEE_MATTR_UR.bits() | Self::TEE_MATTR_UW.bits();
+        const TEE_MATTR_URWX = Self::TEE_MATTR_URW.bits() | Self::TEE_MATTR_UX.bits();
+        const TEE_MATTR_PROT_MASK = Self::TEE_MATTR_PRWX.bits() | Self::TEE_MATTR_URWX.bits();
+        const TEE_MATTR_GLOBAL = 0x400;
+        const TEE_MATTR_SECURE = 0x800;
+    }
```
jstarks, line 1065: Shoudl we add `_ = !0`?

litebox_runner_optee_on_linux_userland/src/lib.rs
```diff
@@ -57,6 +62,22 @@ pub enum InterceptionBackend {
 /// panic. If it does actually panic, then ping the authors of LiteBox, and likely a better error
 /// message could be thrown instead.
 pub fn run(cli_args: CliArgs) -> Result<()> {
+    let ldelf_data: Vec<u8> = {
+        let ldelf = PathBuf::from(&cli_args.ldelf);
+        let data = std::fs::read(ldelf).unwrap();
+        #[allow(clippy::let_and_return)]
+        let data = if cli_args.rewrite_syscalls {
+            // capstone declares a global allocator in conflict with our own.
```
jstarks: This isn't true anymore

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -378,6 +393,44 @@ fn load_trampoline(trampoline: TrampolineHdr, relo_off: usize, fd: i32) -> usize
...
 
+/// Allocate the guest TLS for an OP-TEE TA to run it in an x86 environment.
+/// This function is required to overcome the compatibility issue coming from
+/// architectural and build toolchain differences (x86 vs. Arm, Linux vs OP-TEE OS).
+/// OP-TEE `ldelf` and TAs currently only support Arm and OP-TEE OS environment.
+/// To this end, they do not use syscalls like `arch_prctl` and the `FS` register to
+/// set up the TLS area. In contrast, we do use an x86 toolchain to compile OP-TEE TAs.
+/// This toolchain assumes a loader will set up the TLS area with the `FS` register for
+/// programs such that it generates binaries using the `FS` register for TLS access.
+/// To avoid this issue, we might need to modify `ldelf` to make it set up the `FS` register or
```
jstarks: It looks to me like OP-TEE TAs are always single threaded and so they don't use/need fs to find their TLS... they just use a global variable: <https://github.com/op-tee-ia/optee_os/blob/a3f9f466ae27ad10c91f374ce3b5ce496efb3dca/lib/libutee/arch/x86_64/tcb.c#L75>.

jstarks: But I assume you did this for some reason. Do TAs not run correctly without this?

sangho2: I took a look at our TA binaries and GCC flags. It turns out that our TA binaries are accessing `fs` due to GCC's stack protector since GCC-x86 assumes TLS. We might have to use `-fno-stack-protector` to avoid those punch throughs (but need another way to protect stack?).

jstarks: Ah, interesting.

It looks like the build system in optee_os handles this in some way, since they do support stack protection, search for `CFG_TA_STACK_PROTECTOR` and `__stack_chk_guard`. Since the system is single threaded, gcc just reads the guard value from the global `__stack_chk_guard`.

How is our build system for TAs different? Maybe gcc thinks it's building for a target triple that includes `-linux-`--IIRC, LLVM, at least, only uses the TLS-based stack protection for Linux targets. Maybe gcc is similar.

sangho2: It looks like gcc-arm uses `-mstack-protector-guard=global` by default (which uses the global `__stack_chk_guard`) whereas gcc-x86 uses `-mstack-protector-guard=tls` by default (which uses `fs`). OP-TEE's build system doesn't specify this flag which results in different behaviors. I tested `mstack-protector-guard=global` for our x86 build system and it seems that TAs no longer access `fs`. If we do not plan to use other compiler options (e.g., sanitizers), no TLS execution might be okay.

There are some issues with KMPP which relies on SymCrypt. SymCrypt doesn't define `__stack_chk_guard`, resulting in multiple undefined reference to `__stack_chk_guard`. We might have to manually modify SymCrypt source code.

jstarks: OK. I think this makes sense for now. It shouldn't be hard to add a syscall for ldelf to set fsbase in the future, if we find it to be necessary.

jstarks:
Needs a rebase.

litebox_common_optee/src/lib.rs
sangho2: forgot about this. will do.

litebox_runner_optee_on_linux_userland/src/lib.rs
sangho2: will remove this comment.

PR #529, opened 2025-12-01
Store multiple trusted keys in VTL1

athvu:
Parse the system certificates passed from VTL0. Store each certificate and use them for module and kexec signature verifications.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/mem_integrity.rs
+++ b/litebox_platform_lvbs/src/mshv/mem_integrity.rs
@@ -536,6 +536,7 @@ pub fn verify_kernel_pe_signature(
     if digest_algorithm_oid != ID_SHA_256 && digest_algorithm_oid != ID_SHA_512 {
         todo!("Unsupported digest algorithm: {:?}", digest_algorithm_oid);
     }
+    let mut signature_verified = false;
 
     // verify the authenticity of the signed attributes using the system certificate
     for cert in certs {
@@ -548,12 +549,14 @@ pub fn verify_kernel_pe_signature(
             continue;
         };
         if rsa_verifier.verify(&signed_attrs_der, &signature).is_ok() {
-            // check whether the computed digest matches the one in the Authenticode signature
-            let computed_digest = compute_authenticode_digest(kernel_blob, digest_algorithm_oid)?;
-            if authenticode_signature.digest() == computed_digest {
-                return Ok(());
-            }
+            signature_verified = true;
+            break;
         }
+    }
+    // check whether the computed digest matches the one in the Authenticode signature
+    let computed_digest = compute_authenticode_digest(kernel_blob, digest_algorithm_oid)?;
+    if signature_verified && authenticode_signature.digest() == computed_digest {
+        return Ok(());
     }
     Err(VerificationError::AuthenticationFailed)
 }


litebox_platform_lvbs/src/mshv/mem_integrity.rs
```diff
@@ -389,14 +389,20 @@ pub fn verify_kernel_module_signature(
...
-    let rsa_verifier = RsaPkcs1v15Verifier::new(rsa_pubkey, digest_alg)?;
-    rsa_verifier
-        .verify(module_data, &signature)
-        .map_err(|_| VerificationError::AuthenticationFailed)
+    for cert in certs {
+        let key_info = &cert.tbs_certificate.subject_public_key_info;
+        let Ok(rsa_pubkey) = RsaPublicKey::from_pkcs1_der(key_info.subject_public_key.raw_bytes())
+        else {
+            continue;
+        };
```
sangho2, line 397: What is the maximum number of keys supported? If we expect a large number of keys, we might want to maintain/cache decoded keys rather than extract them from certificates on demand.

litebox_platform_lvbs/src/mshv/mem_integrity.rs
```diff
@@ -532,22 +538,24 @@ pub fn verify_kernel_pe_signature(
...
+        let Ok(rsa_verifier) = RsaPkcs1v15Verifier::new(rsa_pubkey, digest_algorithm_oid) else {
+            continue;
+        };
+        if rsa_verifier.verify(&signed_attrs_der, &signature).is_ok() {
+            // check whether the computed digest matches the one in the Authenticode signature
+            let computed_digest = compute_authenticode_digest(kernel_blob, digest_algorithm_oid)?;
+            if authenticode_signature.digest() == computed_digest {
+                return Ok(());
+            }
```
sangho2: These statements are loop invariant. We can move them outside the loop.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -316,6 +316,37 @@ pub fn mshv_vsm_protect_memory(pa: u64, nranges: u64) -> Result<i64, Errno> {
     }
 }
 
+fn parse_certs(mut buf: &[u8]) -> Vec<Certificate> {
+    let mut certs = Vec::new();
+
+    while buf.len() >= 4 && buf[0] == 0x30 && buf[1] == 0x82 {
+        let der_len = ((buf[2] as usize) << 8) | (buf[3] as usize);
+        let total_len = der_len + 4;
```
sangho2, line 324: It might be better to add some comments to explain these DER fields.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -406,15 +437,18 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
...
         // The system certificate is loaded into VTL1 and locked down before `end_of_boot` is signaled.
         // Its integrity depends on UEFI Secure Boot which ensures only trusted software is loaded during
         // the boot process.
-        if let Ok(cert) = Certificate::from_der(&cert_buf) {
-            vtl0_info.set_system_certificate(cert);
-        } else {
-            serial_println!("VSM: Failed to parse system certificate");
-            return Err(Errno::EINVAL);
-        }
+        vtl0_info.set_system_certificates(certs.clone());
```
sangho2, line 450: `clone()` creates a deep copy of the vector which might be slow.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1143,12 +1177,13 @@ impl Vtl0KernelInfo {
         self.boot_done.load(core::sync::atomic::Ordering::SeqCst)
     }
 
-    pub(crate) fn set_system_certificate(&self, cert: Certificate) {
-        let _ = self.system_cert.set(alloc::boxed::Box::new(cert));
+    pub fn set_system_certificates(&self, certs: Vec<Certificate>) {
```
sangho2, line 1180: `certs: &[Certificate]` and `Box::new(certs)` might be better.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -316,6 +316,37 @@ pub fn mshv_vsm_protect_memory(pa: u64, nranges: u64) -> Result<i64, Errno> {
     }
 }
 
+fn parse_certs(mut buf: &[u8]) -> Vec<Certificate> {
```
sangho2, line 319: returning `Option<Vec<Certificate>>` and using `is_none()` might be better.

litebox_platform_lvbs/src/mshv/mem_integrity.rs
tgopinath-microsoft: So as of now we have been dealing with two or three keys. The module signing key generated during kernel build and some machine keys or uefi keys that may get passed.

athvu: I will note this down as a future improvement work item, but since we are only dealing with a couple keys, I think it is okay for now.

PR #530, opened 2025-12-01
Sequentially set PER_CPU_VARIABLE_ADDRESSES[] for each core

athvu:
Create a variable, PER_CPU_VARIABLE_ADDRESSES_IDX, and use it to sequentially set the index of PER_CPU_VARIABLE_ADDRESSES[] for each core instead of indexing by APIC ID.

### DIFF
--- a/dev_tests/src/ratchet.rs
+++ b/dev_tests/src/ratchet.rs
@@ -28,7 +28,7 @@ fn ratchet_globals() -> Result<()> {
             ("litebox/", 7),
             ("litebox_platform_linux_kernel/", 5),
             ("litebox_platform_linux_userland/", 5),
-            ("litebox_platform_lvbs/", 20),
+            ("litebox_platform_lvbs/", 21),
             ("litebox_platform_multiplex/", 1),
             ("litebox_platform_windows_userland/", 7),
             ("litebox_runner_linux_userland/", 1),


tgopinath-microsoft:
At some point we will need a apic id to cpu id api, but for now we do not need it. And when we need it , we should maintain the order in cpu online mask rather than just +1 from the last cpu.

sangho2:
Is this a workaround? A bit unclear why we need to sequentially initialize per cpu variables.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

athvu:
@sangho2 PER_CPU_VARIABLE_ADDRESSES[] was originally indexed using core_id, which is the APIC ID. Since APIC ID is not sequential and can be outside the range of [0, MAX_CORES), we were seeing issues when using a larger number of cores.
For example, with 96 CPUs, logical CPU ID 80 maps to APIC ID 96, so the access PER_CPU_VARIABLE_ADDRESSES[96] would cause a panic

sangho2:
> @sangho2 PER_CPU_VARIABLE_ADDRESSES[] was originally indexed using core_id, which is the APIC ID. Since APIC ID is not sequential and can be outside the range of [0, MAX_CORES), we were seeing issues when using a larger number of cores. For example, with 96 CPUs, logical CPU ID 80 maps to APIC ID 96, so the access PER_CPU_VARIABLE_ADDRESSES[96] would cause a panic

I see. If that's the issue, I guess this PR only deals with the initialization and further access with `core_id` still results in a panic.

tgopinath-microsoft:
> > @sangho2 PER_CPU_VARIABLE_ADDRESSES[] was originally indexed using core_id, which is the APIC ID. Since APIC ID is not sequential and can be outside the range of [0, MAX_CORES), we were seeing issues when using a larger number of cores. For example, with 96 CPUs, logical CPU ID 80 maps to APIC ID 96, so the access PER_CPU_VARIABLE_ADDRESSES[96] would cause a panic
> 
> I see. If that's the issue, I guess this PR only deals with the initialization and further access with `core_id` still results in a panic.

yes. But we are not using core id anywhere outside of boot for now . The per cpu address is stored in gsbase at init and we use it from there thereafter. I do see use of core id for debug purpose in LVBS  . For e.g in prints it will be useful to know which cpu is invoking the apis. But for this we should have a apic id to cpu id mapping based on online cpu mask. Because online cpu mask can have holes.

sangho2:
> > > @sangho2 PER_CPU_VARIABLE_ADDRESSES[] was originally indexed using core_id, which is the APIC ID. Since APIC ID is not sequential and can be outside the range of [0, MAX_CORES), we were seeing issues when using a larger number of cores. For example, with 96 CPUs, logical CPU ID 80 maps to APIC ID 96, so the access PER_CPU_VARIABLE_ADDRESSES[96] would cause a panic
> > 
> > 
> > I see. If that's the issue, I guess this PR only deals with the initialization and further access with `core_id` still results in a panic.
> 
> yes. But we are not using core id anywhere outside of boot for now . The per cpu address is stored in gsbase at init and we use it from there thereafter. I do see use of core id for debug purpose in LVBS . For e.g in prints it will be useful to know which cpu is invoking the apis. But for this we should have a apic id to cpu id mapping based on online cpu mask. Because online cpu mask can have holes.

Make sense. Using `gsbase` can work around this lack of mappings (we might need both `apic_id` to `cpu_id` and `apic_id` to `vp_id` mappings? not so sure). For now, this work because VTL1 is stateless, but it might change once we have stateful VTL1 (e.g., to support RPC from VTL1 to VTL0).

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -214,6 +214,7 @@ static mut BSP_VARIABLES: PerCpuVariables = PerCpuVariables {
 /// Instead of maintaining this map, we might be able to use a hypercall to directly program each core's GS register.
 static mut PER_CPU_VARIABLE_ADDRESSES: [RefCell<*mut PerCpuVariables>; MAX_CORES] =
     [const { RefCell::new(core::ptr::null_mut()) }; MAX_CORES];
+static mut PER_CPU_VARIABLE_ADDRESSES_IDX: usize = 0;
```
sangho2, line 217: Is this safe from data race? If not, using something like `AtomicUsize` might be better.

jstarks:
> For now, this work because VTL1 is stateless, but it might change once we have stateful VTL1 (e.g., to support RPC from VTL1 to VTL0).

I think for VTL0 to be able to take arbitrarily long to respond to these RPCs (e.g., by handing them out to a kernel thread or even a user-mode task), these pending RPCs will need to be stored in some kind of global table, indexed by some request ID. Presumably in this case, VTL1 would store the pending RPC in the table and then return to VTL0, rather than "call out" to VTL1--that way, the response can come in on a different CPU.

I think we'll need a similar mechanism to remember that a TA instance is running when VTL0 gets an interrupt--we'll want to return to VTL0 and have VTL0 call back when it's ready to continue running the TA.

I guess my point is that all of this needs to be global state so that the VTL0 scheduler can move VTL1 workloads around between CPUs. So, I hope that this won't affect per-CPU state too much.

sangho2:
> > For now, this work because VTL1 is stateless, but it might change once we have stateful VTL1 (e.g., to support RPC from VTL1 to VTL0).
> 
> I think for VTL0 to be able to take arbitrarily long to respond to these RPCs (e.g., by handing them out to a kernel thread or even a user-mode task), these pending RPCs will need to be stored in some kind of global table, indexed by some request ID. Presumably in this case, VTL1 would store the pending RPC in the table and then return to VTL0, rather than "call out" to VTL1--that way, the response can come in on a different CPU.
> 
> I think we'll need a similar mechanism to remember that a TA instance is running when VTL0 gets an interrupt--we'll want to return to VTL0 and have VTL0 call back when it's ready to continue running the TA.
> 
> I guess my point is that all of this needs to be global state so that the VTL0 scheduler can move VTL1 workloads around between CPUs. So, I hope that this won't affect per-CPU state too much.

Yes. I tested submission/completion queue like design before (for OP-TEE command messages, but technically there is no significant difference). I think we can implement this without assuming stateful kernel threads.

tgopinath-microsoft:
> I think for VTL0 to be able to take arbitrarily long to respond to these RPCs (e.g., by handing them out to a kernel thread or even a user-mode task), these pending RPCs will need to be stored in some kind of global table, indexed by some request ID. Presumably in this case, VTL1 would store the pending RPC in the table and then return to VTL0, rather than "call out" to VTL1--that way, the response can come in on a different CPU.

John.. We do not allow VTL0 to interrupt VTL1 for now. Atleast so far that is the design. Hyper-V does not voluntarily relinquish control to VTL0 till VTL1 explicitly issues a return . And we do not track VTL0 interrupts in VTL1 either. But yes we have not had TAs running as well. So this is new and if we need to I agree a global  state makes lot of sense

PR #532, opened 2025-12-03
Support UNIX stream socket

CvvT:
This PR adds support for [UNIX stream socket](https://man7.org/linux/man-pages/man7/unix.7.html). Most of the code were borrowed from VSBox with some adaptation. Some missing features I will port over next include:
 
  - Datagram socket
  - Ancillary message
  - Polling for UNIX socket
  - Abstract address
  - setsockopt/getsockopt
  - socketpair

### DIFF
--- a/litebox_shim_linux/src/syscalls/unix.rs
+++ b/litebox_shim_linux/src/syscalls/unix.rs
@@ -85,6 +85,7 @@ impl UnixSocketAddr {
                 } else {
                     OFlags::RDWR
                 };
+                // TODO: extend fs to support creating sock file (i.e., with type `InodeType::Socket`)
                 let file = task.global.fs.open(
                     path.as_str(),
                     flags,
@@ -156,21 +157,15 @@ impl UnixInitStream {
         Self { addr: None }
     }
 
-    /// Attempts to bind to an unnamed address (only succeeds if already bound).
-    fn bind_unnamed(&self, addr: UnixSocketAddr) -> Result<(), Errno> {
-        if !addr.is_unnamed() {
-            return Err(Errno::EINVAL);
-        }
-        Ok(())
-    }
-
     /// Binds this socket to the given address.
     fn bind(&mut self, task: &Task, addr: UnixSocketAddr) -> Result<(), Errno> {
-        if self.addr.is_some() {
-            return self.bind_unnamed(addr);
+        if self.addr.is_some() && !addr.is_unnamed() {
+            return Err(Errno::EINVAL);
         }
-        let bound_addr = addr.bind(task, true)?;
-        self.addr = Some(bound_addr);
+        if self.addr.is_none() {
+            let bound_addr = addr.bind(task, true)?;
+            self.addr = Some(bound_addr);
+        }
         Ok(())
     }
 


litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -0,0 +1,1047 @@
...
+    fn check_io_events(&self) -> litebox::event::Events {
+        todo!()
+    }
+}
+
+enum UnixEntry {
+    Stream(Arc<Backlog>),
+}
+
+fn unix_addr_table<'a>() -> &'a RwLock<crate::Platform, BTreeMap<UnixSocketAddrKey, UnixEntry>> {
```
jstarks: Please move into `GlobalState`.

litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -0,0 +1,1047 @@
...
+    Abstract(Vec<u8>),
+}
+
+/// Key type for indexing Unix socket addresses in the global address table.
+///
+/// This is used internally to track which addresses are currently bound
+/// by listening sockets.
+#[derive(PartialEq, Eq, Hash, Debug, Ord, PartialOrd)]
+enum UnixSocketAddrKey {
+    Path(String),
```
jstarks, line 58: Maybe add a TODO to use an inode reference or something in the future, once the file system supports that.

litebox/src/pipes.rs
jaybosamiya-ms, line 1: Just a quick comment to say that if you are making things `pub` (especially within `litebox` crate) they should come with doc comments, thanks!

litebox_shim_linux/src/syscalls/unix.rs
CvvT: Done

litebox_shim_linux/src/syscalls/unix.rs
CvvT: Good point! Done.

CvvT:
> Can you also add `socketpair` support?

Sure, I put it into a separate PR #544.

litebox/src/pipes.rs
CvvT: Got it. I realized that this change only help me avoid ~10 lines of duplicate code. Copying it to `litebox_shim_linux` to have a self-contained component might be better.

litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -0,0 +1,782 @@
...
+    /// Establishes a connection, blocking if necessary.
+    fn connect(
+        &self,
+        mut init: UnixInitStream,
+        is_nonblocking: bool,
+    ) -> Result<UnixConnectedStream, (UnixInitStream, Errno)> {
+        if is_nonblocking {
+            self.try_connect(init)
+        } else {
+            // TODO: use polling instead of busy loop
```
jstarks, line 295: When will this come?

CvvT: I'm working on this part. Hopefully it can be done by EOD.

PR #534, opened 2025-12-03
Remove `Synchronization` type

jstarks:
Use `Mutex::new`, etc. to construct synchronization objects across the repo. This removes the need to pass `LiteBox` references around in many places.

### DIFF
--- a/litebox/src/litebox.rs
+++ b/litebox/src/litebox.rs
@@ -56,7 +56,7 @@ impl<Platform: RawSyncPrimitivesProvider> LiteBox<Platform> {
             );
         }
 
-        // Enable lock tracing using this platform for time keeping and debug
+        // Enable lock tracing, using this platform for time keeping and debug
         // prints, if the feature is enabled.
         #[cfg(feature = "lock_tracing")]
         crate::sync::lock_tracing::LockTracker::init(platform);
--- a/litebox/src/sync/lock_tracing.rs
+++ b/litebox/src/sync/lock_tracing.rs
@@ -1,4 +1,31 @@
-//! Lock-tracing functionality
+//! Lock-tracing functionality for debugging lock ordering and contention
+//! issues.
+//!
+//! This module provides runtime tracking of lock acquisitions and releases to
+//! help detect potential deadlocks, lock ordering violations, and performance
+//! issues related to lock contention. When enabled, it records which locks are
+//! held by the current thread and can detect and report various problematic
+//! patterns.
+//!
+//! # Feature flag
+//!
+//! Lock tracing is only compiled in when the `lock_tracing` cargo feature is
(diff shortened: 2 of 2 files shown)


litebox/src/fd/mod.rs
```diff
@@ -31,12 +29,8 @@ impl<Platform: RawSyncPrimitivesProvider> Descriptors<Platform> {
     ///
     /// This is expected to be invoked only by [`crate::LiteBox`]'s creation method, and should not
     /// be invoked anywhere else in the codebase.
-    pub(crate) fn new_from_litebox_creation(litebox: &LiteBox<Platform>) -> Self {
-        let litebox = litebox.clone();
-        Self {
-            litebox,
-            entries: vec![],
-        }
+    pub(crate) fn new_from_litebox_creation() -> Self {
```
wdcui, line 32: Should we change the function name since it doesn't require `litebox` anymore?

litebox/src/litebox.rs
```diff
@@ -56,23 +56,17 @@ impl<Platform: RawSyncPrimitivesProvider> LiteBox<Platform> {
...
         }
 
-        let sync = Synchronization::new_from_platform(platform);
-        // We set `descriptors` to `None` and replace it out with a `Some` after creation due to a
-        // circular dependency between the two types for their initialization. The public interfaces
-        // here do not need to deal with any of this though; see the post-creation invariant
-        // guarantee written on `LiteBoxX::descriptors`.
-        let descriptors = sync.new_rwlock(None);
-        let ret = Self {
+        // Enable lock tracing using this platform for time keeping and debug
```
wdcui: Minor: does this comment miss some punctuation?

litebox/src/fd/mod.rs
jstarks: I interpreted this name to mean "new, called from the litebox creation path." I.e., don't call this outside of there.

jaybosamiya-ms: Yep, it is intentionally intended to be used in only one place (from `LiteBox`'s creation method) so the name is just indicating that other places in the codebase should not invoke it, that's all.

litebox/src/sync/lock_tracing.rs
jaybosamiya-ms, line 1: Minor: we should probably include docs in this module to state that lock tracking may not be particularly helpful/useful until the `LiteBox` object has been initialized.

litebox/src/fd/mod.rs
wdcui: Got it. Thanks!

litebox/src/sync/lock_tracing.rs
jstarks: I put copilot to work.

PR #535, opened 2025-12-03
Support socketcall for x86

CvvT:
Realized there is another syscall [`socketcall`](https://elixir.bootlin.com/linux/v5.19.17/source/net/socket.c#L2980) we didn't support when I was trying to add a C test for UNIX socket. This syscall is only available for x86.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -435,7 +435,7 @@ impl Task {
     }
 
     /// Handle syscall `close`
-    pub(crate) fn sys_close(&self, fd: i32) -> Result<usize, Errno> {
+    pub(crate) fn sys_close(&self, fd: i32) -> Result<(), Errno> {
         let Ok(fd) = u32::try_from(fd) else {
             return Err(Errno::EBADF);
         };
@@ -444,7 +444,7 @@ impl Task {
         match file_table.remove(fd) {
             Some(desc) => {
                 drop(file_table); // drop before potentially blocking `close`
-                self.do_close(desc).map(|()| 0)
+                self.do_close(desc)
             }
             None => Err(Errno::EBADF),
         }
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -871,7 +871,7 @@ impl Task {
         addr: Option<MutPtr<u8>>,
         addrlen: Option<MutPtr<u32>>,
         flags: SockFlags,
-    ) -> Result<usize, Errno> {
+    ) -> Result<u32, Errno> {
         let Ok(sockfd) = u32::try_from(sockfd) else {
             return Err(Errno::EBADF);
         };
@@ -886,7 +886,7 @@ impl Task {
                 return Err(err);
             }
         }
-        Ok(fd as usize)
+        Ok(fd)
     }
     fn do_accept(
         &self,
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -528,7 +528,7 @@ pub unsafe extern "C" fn open(pathname: ConstPtr<i8>, flags: u32, mode: u32) -> 
 
 /// Closes the file
 pub extern "C" fn close(fd: i32) -> i32 {
-    with_current_task(|task| task.sys_close(fd).map_or_else(Errno::as_neg, |_| 0))
+    with_current_task(|task| task.sys_close(fd).map_or_else(Errno::as_neg, |()| 0))
 }
 
 // This places size limits on maximum read/write sizes that might occur; it exists primarily to
@@ -609,6 +609,13 @@ impl Task {
     }
 
     fn do_syscall(&self, ctx: &mut litebox_common_linux::PtRegs) -> Result<usize, Errno> {
+        // Helper macro to unify the return value from `sys_*`.
+        macro_rules! syscall {
+            ($func:ident($($args:expr),*)) => {
+                self.$func($($args),*).to_syscall_result()
+            };
+        }
+
         #[cfg(target_arch = "x86")]
         let syscall_number = ctx.orig_eax;
         #[cfg(target_arch = "x86_64")]
@@ -677,18 +684,16 @@ impl Task {
                 Some(buf) => self.sys_write(fd, &buf, None),
                 None => Err(Errno::EFAULT),
             },
-            SyscallRequest::Close { fd } => self.sys_close(fd),
+            SyscallRequest::Close { fd } => syscall!(sys_close(fd)),
             SyscallRequest::Lseek { fd, offset, whence } => {
                 use litebox::utils::TruncateExt as _;
                 syscalls::file::try_into_whence(whence.truncate())
                     .map_err(|_| Errno::EINVAL)
                     .and_then(|seekwhence| self.sys_lseek(fd, offset, seekwhence))
             }
-            SyscallRequest::Mkdir { pathname, mode } => {
-                pathname.to_cstring().map_or(Err(Errno::EINVAL), |path| {
-                    self.sys_mkdir(path, mode).map(|()| 0)
-                })
-            }
+            SyscallRequest::Mkdir { pathname, mode } => pathname
+                .to_cstring()
+                .map_or(Err(Errno::EINVAL), |path| syscall!(sys_mkdir(path, mode))),
             SyscallRequest::RtSigprocmask {
                 how,
                 set,
@@ -704,7 +709,7 @@ impl Task {
             SyscallRequest::RtSigreturn => self.sys_rt_sigreturn(ctx),
             #[cfg(target_arch = "x86")]
             SyscallRequest::Sigreturn => self.sys_sigreturn(ctx),
-            SyscallRequest::Ioctl { fd, arg } => self.sys_ioctl(fd, arg).map(|v| v as usize),
+            SyscallRequest::Ioctl { fd, arg } => syscall!(sys_ioctl(fd, arg)),
             SyscallRequest::Pread64 {
                 fd,
                 buf,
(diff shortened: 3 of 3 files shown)


litebox_runner_linux_userland/src/lib.rs
wdcui, line 1: Should the changes in this file be in a separate PR?

CvvT: This is necessary for the new test. Without the change, some data might be lost once the program exits.

litebox_common_linux/src/lib.rs
```diff
@@ -2383,23 +2465,164 @@ impl<Platform: litebox::platform::RawPointerProvider> SyscallRequest<Platform> {
...
-                        let protocol: u8 = ctx.sys_req_arg(2);
-                        Some(Protocol::try_from(protocol).map_err(|_| {
-                            unsupported_einval(format_args!("socket(protocol = {protocol})"))
-                        })?)
-                    },
+                let protocol: u8 = ctx.sys_req_arg(2);
+                parse_socket(domain, type_and_flags, protocol, unsupported_einval)?
+            }
+            #[cfg(target_arch = "x86")]
+            Sysno::socketcall => {
```
jstarks: I have been meaning to bring this up...

I think we should consider changing the way we parse syscalls a bit... right now, we have three layers of syscall handling: the parsing code here which constructs the enum, the dispatch code which matches on the enum, and then the individual `sys_*` methods.

In some cases, we spread the syscall parsing and handling across all three places--we check some parameters here in the parsing code and return various errnos, we check some more parameters in the dispatch code (or even copy user parameters to kernel memory), and then we do yet more validation in the syscall itself.

It seems to me that we should flatten all of this and do all the validation and arg parsing in the `sys_*` methods. This is slightly less powerful, in that you can't reuse the parsing code with multiple syscall implementations.  But I don't think that's a realistic goal, anyway, because the parsing logic is very specific to the individual implementation.

In that world, the separate enum goes away, and the dispatch code is just a big table of syscall number to `sys_*` fn--some Rust trickery can handle converting the parameter types. But getting there is a big effort, and we need to agree to it first.

If this seems vaguely right to everyone, then I suggest that in the meantime, we stop adding complicated code to the parsing and dispatch layers. In this case, just have `socketcall` parse into `Socketcall { call: i32, args: ConstPtr<usize> }`, and have the dispatch layer call `sys_socketcall(call: i32, args: ConstPtr<usize>) -> _`. Then do this complicated parsing there.

jstarks: Otherwise this is yet more work to redo later with the flattened model.

litebox_runner_linux_userland/src/lib.rs
wdcui: I see. Thanks.

litebox_common_linux/src/lib.rs
CvvT: I had the same concern. The validation is separated into different places. The conclusion we had before was that we'd like to have a strong Rust type for syscalls and possibly reuse the parsing code. I don't have strong preference. @jaybosamiya-ms any thoughts?

jaybosamiya-ms: Previously, things were even more messy with respect to parsing, which is why I'd introduced `sys_req!` macro in #288.  Seemingly that macro _and_ this particular function have gotten more complex since then (I had not noticed that complexity had increased since then, thanks for bringing it to attention). But yeah, the intention is for this function to be _super_ simple ("trivially correct") and anything not covered by `sys_req!` should actually just get punted to the `sys_*` functions.  Also `sys_req!` itself should not be updated too often (its job is just to get number/pointer arguments right).  I guess I should have added docs to this function (and to `sys_req!`) to explain that it should not increase in complexity, but since I had not documented that during #288, I guess complexity has grown.

In short: things that fit into `sys_req!`, we should just use that.  Things that are more complex, we should just use `sys_req!` for the simper number/pointer mapping and then pass on to `sys_*` functions to then handle more.

In this specific case, `sys_socketcall(call: i32, args: ConstPtr<usize>) -> _` _is_ the right move, which would be equivalent to a `sys_req!(SocketCall { call, args:* })` or such and then the rest of the handling can be done there at `sys_socketcall` instead.

I'm happy to do a clean-up pass to push things towards a cleaner state and add the documentation that tells what the intentions are.  I don't think we should block this PR on it; feel free to convert this comment thread to an issue and assign me to it :)

---

Fwiw, the use of more strongly-typed values was originally intended more for punchthroughs (so as to not expose just raw Linux-y integers / bitfields / etc. to non-Linux platforms), but for stuff that is just happening _effectively_ just within LiteBox shim+litebox itself (i.e., not needing to deal with the platform being aware of the syscall), we should go with whatever is more convenient pragmatically, rather than be dogmatic.

wdcui: Thank you, @jstarks , for raising the issue. I agree with @jaybosamiya-ms 's proposal.

CvvT: Thanks for the discussion and comments. I have consolidated all argument parsing into `sys_*` functions for socket related syscalls in this PR per our discussion. For the rest and further improvement (i.e., the flattened model), I will create an issue to track it.

CvvT: @wdcui I made significant change to this PR as discussed. Do you want to review this PR again?

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -435,10 +435,13 @@ impl Task {
...
 
     /// Handle syscall `close`
-    pub fn sys_close(&self, fd: i32) -> Result<(), Errno> {
+    pub(crate) fn sys_close(&self, fd: i32) -> Result<usize, Errno> {
         let Ok(fd) = u32::try_from(fd) else {
             return Err(Errno::EBADF);
         };
+        self.sys_close_inner(fd).map(|()| 0)
+    }
+    pub(crate) fn sys_close_inner(&self, fd: u32) -> Result<(), Errno> {
```
wdcui: Does this function need to be `pub(crate)`?

litebox_shim_linux/src/syscalls/tests.rs
```diff
@@ -110,7 +110,7 @@ fn test_dup() {
 
     // test socket dup
     let socket_fd = task
-        .sys_socket(
+        .sys_socket_inner(
```
wdcui: Why do we want to expose an inner function?

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -927,11 +992,13 @@ impl Task {
...
     /// Handle syscall `listen`
-    pub(crate) fn sys_listen(&self, sockfd: i32, backlog: u16) -> Result<(), Errno> {
+    pub(crate) fn sys_listen(&self, sockfd: i32, backlog: u16) -> Result<usize, Errno> {
         let Ok(sockfd) = u32::try_from(sockfd) else {
             return Err(Errno::EBADF);
         };
-
+        self.sys_listen_inner(sockfd, backlog).map(|()| 0)
+    }
+    pub(crate) fn sys_listen_inner(&self, sockfd: u32, backlog: u16) -> Result<(), Errno> {
```
wdcui: Why does this one need to be `pub(crate)`?

litebox_shim_linux/src/syscalls/tests.rs
CvvT: Lots of `sys_*` functions accepts raw pointers instead of those nice Rust strong types. Using `sys_*_inner` functions is much more convenient for testing. As we discussed, I will keep all those inner functions private. For this particular test that requires `sys_socket_inner` from another file (and thus it cannot be private), I will move the test to the same file `sys_socket_inner` belongs to.

litebox_shim_linux/src/syscalls/net.rs
CvvT: Oh, this is a mistake. Changed it to private.

litebox_shim_linux/src/syscalls/file.rs
CvvT: Changed it to private.

litebox_shim_linux/src/syscalls/tests.rs
wdcui: Sounds good. Thanks!

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -435,10 +435,13 @@ impl Task {
...
 
     /// Handle syscall `close`
-    pub fn sys_close(&self, fd: i32) -> Result<(), Errno> {
+    pub(crate) fn sys_close(&self, fd: i32) -> Result<usize, Errno> {
         let Ok(fd) = u32::try_from(fd) else {
             return Err(Errno::EBADF);
         };
+        self.sys_close_inner(fd).map(|()| 0)
+    }
+    fn sys_close_inner(&self, fd: u32) -> Result<(), Errno> {
```
jstarks: I suggest a convention that we name functions like this `do_{syscall}`, e.g.. `do_close`. I'd like to keep `sys_` reserved for true syscall entrypoints.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -839,13 +869,32 @@ impl Task {
     pub(crate) fn sys_accept(
         &self,
         sockfd: i32,
-        peer: Option<&mut SocketAddress>,
+        addr: Option<MutPtr<u8>>,
+        addrlen: Option<MutPtr<u32>>,
         flags: SockFlags,
-    ) -> Result<u32, Errno> {
+    ) -> Result<usize, Errno> {
```
jstarks: I am OK with having non-usize in the return type, as long as it can be cast back to usize without truncation.

jstarks: Similarly, () is OK, we can have the syscall handling code turn this back into 0 via a trait.

jstarks: (Either way is OK with me)

CvvT: Thanks for the suggestion. I like the idea of having the syscall handling code to unify the result via a trait.

litebox_shim_linux/src/syscalls/file.rs
CvvT: Makes sense to me. Updated.

PR #541, opened 2025-12-06
LVBS: Revise VTL switch loop and add run_thread support

sangho2:
This PR revises VTL switch loop including VTL0 state saving/restoring, and
adds `run_thread` support.

This PR uses assembly code to carefully switch between VTLs while
saving/restoring VTL0 general-purpose registers and extended states. 

Also, it adds `run_thread` support for the LVBS platform and revises the syscall
handling routine based on that of the Linux userland platform.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vtl_switch.rs
+++ b/litebox_platform_lvbs/src/mshv/vtl_switch.rs
@@ -228,6 +228,16 @@ macro_rules! XRSTOR_ASM {
     };
 }
 
+/// VTL switch loop implemented in assembly.
+///
+/// # Register Assumptions
+///
+/// At each iteration start, this code only relies on `rip`, `rsp`, and `gs` which Hyper-V
+/// saves/restores across VTL switches. All other registers may contain VTL0 state and must
+/// be saved before use and restored before returning to VTL0.
+///
+/// VTL1 registers can be freely clobbered since this loop is stateless -- `rsp` is reset to
+/// the kernel stack each iteration and all state lives in per-CPU variables via `gs`.
 #[unsafe(naked)]
 unsafe extern "C" fn vtl_switch_loop_asm() -> ! {
     naked_asm!(


litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -209,11 +210,73 @@ static mut BSP_VARIABLES: PerCpuVariables = PerCpuVariables {
...
+/// TODO: Let's consider whether we should maintain these two types of TLS areas (for Rust and
+/// assembly, respectively). This design secures Rust-side access to `PerCpuVariables` with `RefCell`,
+/// but it might be unnecessarily complex. Instead, we could use assembly code in all cases, but
+/// this might be unsafe.
+#[repr(C)]
+#[derive(Clone)]
+pub struct RefCellWrapper<T> {
+    /// Make some TLS area be accessible via the GS register. This is for assembly code and
+    /// never meant to be used in Rust code (unsafe).
+    ktls: KernelTls,
```
jstarks: As written, Rust will assume that this type is immutable, so any Rust-side accesses to `ktls` will probably generate bad code. Right now I don't see any Rust accesses to `ktls`, so this is probably not going to cause a real problem. But I suggest fixing this properly to avoid weird behavior in the future.

There are two ways you could fix this. The easy way is to wrap the whole `KernelTls` with `UnsafeCell`. This will tell Rust that the type has interior mutability, so it will be legal for its values to change even just through an immutable reference. However, this approach makes it hard for Rust code to access any of these fields, again if you end up needing to access these through both Rust and asm.

The approach I took in the Windows userland platform, which has a similar TLS mechanism with sharing between asm and Rust, is to wrap the individual fields of `KernelTls` with `Cell<>`. That way, the individual fields are marked with interior mutability, you can still access them in asm just fine, but you can also use `Cell` methods to safely access them from the Rust side.

I suggest the `Cell`-based approach.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
jstarks, line 494: Should we just store the base of the `RefcellWrapper` in `gs` so that you don't have to use the negative offsets?

jstarks: One reason I ask is because your computations today seem to assume that the `RefCell` is _right_ after `KernelTls`. At the very least these should be updated to account for possible padding in the future.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: Thanks for the suggestion. Now it uses `Cell`.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: Revised. I used negative offsets because negative ones are more common in TLS, but we don't have to do that.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -273,19 +232,161 @@ static mut BSP_VARIABLES: PerCpuVariables = PerCpuVariables {
...
     vtl1_xsave_area_addr: VirtAddr::zero(),
-    vtl0_xsave_mask: 0,
-    vtl0_xsaved: false,
     vtl1_xsaved: false,
     tls: VirtAddr::zero(),
 };
 
+/// Specify the layout of PerCpuVariables for Assembly area.
+///
+/// Unlike userland platforms, this kernel platform does't rely on the `tbss` section
```
wdcui: I don't understand this comment. We can always set up kernel TLS if needed, right?

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -273,19 +232,161 @@ static mut BSP_VARIABLES: PerCpuVariables = PerCpuVariables {
...
+        offset_of!(PerCpuVariablesAsm, vtl0_xsaved)
+    }
+}
+
+/// Wrapper struct to maintain `RefCell` along with `PerCpuVariablesAsm`.
+/// This struct allows assembly code to read/write some PerCpuVariables area via the GS register (e.g., to
+/// save/restore RIP/RSP). Currently, `PerCpuVariables` is protected by `RefCell` such that
+/// assembly code cannot easily access it.
+///
+/// TODO: Let's consider whether we should maintain these two types of TLS areas (for Rust and
```
wdcui: Why do we call it "TLS areas"?

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -411,15 +538,44 @@ pub fn allocate_per_cpu_variables() {
         let per_cpu_variables = unsafe {
             let ptr = per_cpu_variables.as_mut_ptr();
             ptr.write_bytes(0, 1);
-            (*ptr).allocate_xsave_area();
             per_cpu_variables.assume_init()
         };
         unsafe {
-            PER_CPU_VARIABLE_ADDRESSES[i] = RefCell::new(Box::into_raw(per_cpu_variables));
+            PER_CPU_VARIABLE_ADDRESSES[i] = RefCellWrapper::new(Box::into_raw(per_cpu_variables));
+            // Allocate xsave area for this core, writing directly to its PerCpuVariablesAsm
```
wdcui, line 546: Why can we write them directly but have to use `with_per_cpu_variables_mut` above?

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -273,19 +232,161 @@ static mut BSP_VARIABLES: PerCpuVariables = PerCpuVariables {
...
+/// assembly code cannot easily access it.
+///
+/// TODO: Let's consider whether we should maintain these two types of TLS areas (for Rust and
+/// assembly, respectively). This design secures Rust-side access to `PerCpuVariables` with `RefCell`,
+/// but it might be unnecessarily complex. Instead, we could use assembly code in all cases, but
+/// this might be unsafe.
+#[repr(C)]
+pub struct RefCellWrapper<T> {
+    /// Make some PerCpuVariablesAsm area be accessible via the GS register. This is mainly for assembly code
+    pcv_asm: PerCpuVariablesAsm,
```
wdcui, line 357: If `pcv_asm` is only accessed through `gs`, why do we even put it in this `RefCellWrapper`?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -78,236 +85,216 @@ impl VtlState {
...
-        // VTL calls and intercepts (i.e., returns from synthetic interrupt handlers) land here.
+#[unsafe(naked)]
+unsafe extern "C" fn vtl_switch_loop_asm() -> ! {
+    naked_asm!(
+        "1:",
+        "mov rsp, gs:[{kernel_sp_off}]", // reset kernel stack pointer. Hyper-V saves/restores rsp and rip.
+        "cli", // disable VTL1 interrupts before returning to VTL0
+        VTL_RETURN_ASM!({vtl_ret_addr_off}),
+        // *** VTL1 resumes here regardless of the entry reason (VTL switch or intercept) ***
+        SAVE_VTL_STATE_ASM!({scratch_off}, {vtl0_state_top_addr_off}),
```
wdcui, line 249: Do we use this macro for both VTL0 and VTL1? If not, it might be helpful to specify it in the name.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -78,236 +85,216 @@ impl VtlState {
...
+        XSAVE_ASM!({vtl0_xsave_area_off}, {vtl0_xsave_mask_lo_off}, {vtl0_xsave_mask_hi_off}, {vtl0_xsaved_off}),
+        "mov rbp, rsp", // rbp contains VTL0's stack frame, so update it.
+        "sti", // enable VTL1 interrupts after saving VTL0 state
```
wdcui, line 252: @tgopinath-microsoft This is how you see we should disable/enable interrupts?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -910,6 +915,255 @@ impl<Host: HostInterface, const ALIGN: usize> VmapManager<ALIGN> for LinuxKernel
...
+        "mov rsp, gs:[{cur_kernel_sp_off}]",
+        // Handle the syscall. This will jump back to the user but
+        // will return if the thread is exiting.
+        "mov rdi, [rsp]", // pass `thread_ctx`
+        "call {syscall_handler}",
+        "jmp done",
+        // Exception and interrupt callback placeholders
+        // IDT handler functions will jump to these labels to
+        // handle user-mode exceptions/interrupts.
+        ".globl exception_callback",
```
wdcui, line 1087: The current code doesn't handle exceptions and interrupts?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -910,6 +915,255 @@ impl<Host: HostInterface, const ALIGN: usize> VmapManager<ALIGN> for LinuxKernel
...
+        "mov rax, cr3",
+        "mov cr3, rax",
+        "xor eax, eax",
+        // Restore user context from ctx.
+        "mov rsp, rdi",
+        RESTORE_USER_CONTEXT_ASM!(),
+        // clear the GS base register (as the `KernelGsBase` MSR contains 0)
+        // while writing the current GS base value to `KernelGsBase`.
+        "swapgs",
+        "iretq",
```
wdcui, line 1156: Do we go through this path when we enter the user mode for the first time?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -910,6 +915,255 @@ impl<Host: HostInterface, const ALIGN: usize> VmapManager<ALIGN> for LinuxKernel
...
+        "switch_to_user_end:",
+    );
+}
+
+// Note on user page table management:
+// The legacy platform code creates a new page table to load a program in a separate
+// address space and destroys it when the program terminates. This is why the old syscall
+// handler invokes `change_address_space()`. Previously, the platform does all these because
+// it is the one running the event loop. Once we have an `upcall` mechanism, the runner
```
wdcui, line 1165: We should make sure to update this comment when the `upcall` code is done.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: This comment is trying to explain why we can't use `.tbss` section unlike userland platforms.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: Forget to revise it. Will fix it.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: We can directly access `PER_CPU_VARIABLE_ADDRESSES` with `unsafe`. `with_per_cpu_variables_mut` to access this without `unsafe`.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: Both `PerCpuVariablesAsm` and `PerCpuVariables` are accessed through `gs`.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: maybe. I didn't put VTL0 there because it already has `vtl0_state_top_addr_off` as an argument.

sangho2: and we can use this macro for saving either of them (although we don't store VTL1 state yet).

litebox_platform_lvbs/src/lib.rs
sangho2: We capture exceptions/interrupts through IDT. For now, we only handle SINT.

litebox_platform_lvbs/src/lib.rs
sangho2: We use `iretq` always (first time, once a syscall is handled, ...). We did use `sysret` before, but @jstarks suggested us to use `iretq`.

litebox_platform_lvbs/src/lib.rs
sangho2: Will do.

litebox_platform_lvbs/src/lib.rs
wdcui: It might help to include this in the comment. Basically, you meant interrupts/exceptions won't reach these locations, right?

litebox_platform_lvbs/src/host/per_cpu_variables.rs
wdcui: Please resolve this one after you update the comment.

litebox_platform_lvbs/src/lib.rs
sangho2: True. Basically, these are placeholders to make this asm block consistent with that of `litebox_platform_linux_userland`. I'll revise the comment to clarify this.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
tgopinath-microsoft: yes. it is ! Thank you for including this

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -78,236 +85,216 @@ impl VtlState {
...
+        // A pending SINT can be fired here. Our SINT handler only executes `iretq` so returns to here immediately.
+        "call {loop_body}",
+        ".globl panic_vtl_switch",
```
tgopinath-microsoft, line 255: @sangho2 where is  panic_vtl_switch being invoked from ?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -78,236 +85,216 @@ impl VtlState {
...
+        "panic_vtl_switch:", // jump to here on panic to switch back to VTL0
+        XRSTOR_ASM!({vtl0_xsave_area_off}, {vtl0_xsave_mask_lo_off}, {vtl0_xsave_mask_hi_off}, {vtl0_xsaved_off}),
+        LOAD_VTL_STATE_ASM!({vtl0_state_top_addr_off}, {VTL_STATE_SIZE}),
+        // *** VTL0 state is recovered. Do not put any code tampering with them here ***
```
tgopinath-microsoft, line 259: so IUC, earlier you were saving VTL1 context when exiting VTL1 and restoring it back on entry. You are not doing that anymore. We do not do that in Linux as well. But any reason why you removed it here ?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: it's a placeholder. I added this because #586 might need it.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: Because storing VTL1 context in per-CPU variables is meaningless. Without RPC, VTL1 is stateless. With RPC, we shouldn't use Per-CPU variables to store VTL1 context because no one ensures that the same CPU core will be used again for this RPC request.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
wdcui: Thank you, Thara!

litebox_platform_lvbs/src/mshv/vtl_switch.rs
tgopinath-microsoft: ok thanks.. I agree 586 may need it. If not , we can remove it

litebox_platform_lvbs/src/mshv/vtl_switch.rs
tgopinath-microsoft: I am mostly ok with this. I have not disassembled the code and seen if we are using any register without loading it first. The issue with not saving and restoring VTL1 context is that ,the code or the compiler does not know about switching VTL contexts. So if there is any optimization that assumes a previously loaded value in a register after return from VTL0, we will run into issues.

sangho2: I see your concern. This `vtl_switch_loop_asm` is a very tight asm block which doesn't read arbitrary registers. I'll go over a compiled binary again to make sure it.

sangho2: confirm that there is no arbitrary register access (this is just an asm block and Rust compiler doesn't touch it).

PR #543, opened 2025-12-08
Use contiguous buffer for memory container

femiatl:
After data is copied from heki ranges into a memory container, it has to be copied again into a byte vector for parsing. For larger containers, this temporary double copy can exhaust LiteBox heap.

Instead, copy all data into one contiguous buffer, and keep track of heki range offsets into that buffer. Since most memory containers, (except patch info) store data for one VTL0 kernel variable, they can be readily processed as bytes slices.

The containers keep virtual and physical address metadata to help with future module symbol resolution and TOCTOU page protection.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -375,24 +375,22 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
     for heki_page in &heki_pages {
         for heki_range in heki_page {
             debug_serial_println!("VSM: Load kernel data {heki_range:?}");
-            let Some(mem) = (match heki_range.heki_kdata_type() {
-                HekiKdataType::SystemCerts => Some(&mut system_certs_mem),
+            match heki_range.heki_kdata_type() {
+                HekiKdataType::SystemCerts => system_certs_mem.extend_range(heki_range)?,
                 HekiKdataType::KexecTrampoline => {
                     kexec_trampoline_metadata.insert_heki_range(heki_range);
-                    None
                 }
-                HekiKdataType::PatchInfo => Some(&mut patch_info_mem),
-                HekiKdataType::KernelInfo => Some(&mut kinfo_mem),
-                HekiKdataType::KernelData => Some(&mut kdata_mem),
+                HekiKdataType::PatchInfo => patch_info_mem.extend_range(heki_range)?,
+                HekiKdataType::KernelInfo => kinfo_mem.extend_range(heki_range)?,
+                HekiKdataType::KernelData => kdata_mem.extend_range(heki_range)?,
                 HekiKdataType::Unknown => {
                     serial_println!("VSM: Invalid kernel data type",);
                     return Err(Errno::EINVAL);
                 }
-                _ => None,
-            }) else {
-                continue;
-            };
-            mem.extend_range(heki_range);
+                _ => {
+                    debug_serial_println!("VSM: Unsupported kernel data not loaded {heki_range:?}");
+                }
+            }
         }
     }
 
@@ -510,13 +508,13 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
 
     for heki_page in &heki_pages {
         for heki_range in heki_page {
-            let Some(mem) = (match heki_range.mod_mem_type() {
+            match heki_range.mod_mem_type() {
                 ModMemType::Unknown => {
                     serial_println!("VSM: Invalid module memory type");
                     return Err(Errno::EINVAL);
                 }
-                ModMemType::ElfBuffer => Some(&mut module_as_elf),
-                ModMemType::Patch => Some(&mut patch_info_for_module),
+                ModMemType::ElfBuffer => module_as_elf.extend_range(heki_range)?,
+                ModMemType::Patch => patch_info_for_module.extend_range(heki_range)?,
                 _ => {
                     // if input memory range's type is neither `Unknown` nor `ElfBuffer`, its addresses must be page-aligned
                     if !heki_range.is_aligned(Size4KiB::SIZE) {
@@ -524,12 +522,9 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
                         return Err(Errno::EINVAL);
                     }
                     module_memory_metadata.insert_heki_range(heki_range);
-                    module_in_memory.get_module_memory(heki_range.mod_mem_type())
+                    module_in_memory.extend_range(heki_range.mod_mem_type(), heki_range)?;
                 }
-            }) else {
-                continue;
-            };
-            mem.extend_range(heki_range);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


femiatl:
Tested by reducing secure kernel memory to minimum of 48MB using command-line option

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -375,69 +368,54 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
...
+                HekiKdataType::KernelInfo => Some(&mut kinfo_mem),
+                HekiKdataType::KernelData => Some(&mut kdata_mem),
+                HekiKdataType::Unknown => {
+                    serial_println!("VSM: Invalid kernel data type",);
+                    return Err(Errno::EINVAL);
+                }
+                _ => None,
+            }) else {
+                continue;
+            };
```
sangho2: Rather than using `&mut mem`, directly calling `extend_range` might be more readable (e.g., `HekiKdataType::SystemCerts => system_certs_mem.extend_range(heki_range)`.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -375,69 +368,54 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, Errno> {
...
                 }
-            }
+                HekiKdataType::PatchInfo => Some(&mut patch_info_mem),
+                HekiKdataType::KernelInfo => Some(&mut kinfo_mem),
+                HekiKdataType::KernelData => Some(&mut kdata_mem),
+                HekiKdataType::Unknown => {
+                    serial_println!("VSM: Invalid kernel data type",);
+                    return Err(Errno::EINVAL);
+                }
+                _ => None,
```
sangho2: Perhaps it's time to add a debug message here to note that we don't support some `HekiKdataType` yet.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -538,35 +504,58 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
...
+                        serial_println!("VSM: input address must be page-aligned");
+                        return Err(Errno::EINVAL);
+                    }
+                    module_memory_metadata.insert_heki_range(heki_range);
+                    module_in_memory.get_module_memory(heki_range.mod_mem_type())
+                }
+            }) else {
+                continue;
+            };
+            mem.extend_range(heki_range);
```
sangho2: similar

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -793,16 +734,37 @@ pub fn mshv_vsm_kexec_validate(pa: u64, nranges: u64, crash: u64) -> Result<i64,
...
+                    None
+                }
+                HekiKexecType::Unknown => {
+                    serial_println!("VSM: Invalid kexec type");
+                    return Err(Errno::EINVAL);
+                }
+            }) else {
+                continue;
+            };
+            mem.extend_range(heki_range);
```
sangho2: similar

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -793,16 +734,37 @@ pub fn mshv_vsm_kexec_validate(pa: u64, nranges: u64, crash: u64) -> Result<i64,
...
+        }
     }
+    kexec_image.write_bytes_from_heki_range().unwrap();
+    kexec_kernel_blob.write_bytes_from_heki_range().unwrap();
```
sangho2: let's use `map_err` or something similar to return an error rather than panic

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1496,216 +1389,116 @@ impl ModuleMemory {
...
     }
 
-    fn get_or_alloc_page(&mut self, addr: VirtAddr) -> &mut Box<AlignedPage> {
-        let page_base = addr.align_down(Size4KiB::SIZE);
-        self.pages
-            .entry(page_base)
-            .or_insert_with(|| Box::new(AlignedPage::new()))
+    pub(crate) fn extend_range(&mut self, heki_range: &HekiRange) {
+        let addr = VirtAddr::try_new(heki_range.va).unwrap();
+        let phys_addr = PhysAddr::try_new(heki_range.pa).unwrap();
```
sangho2: this function will panic if `heki_range` is invalid (e.g., a non-canonical address). It might be better to return an error if `try_new` fails or leave TODO to fix this issue later.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1496,216 +1389,116 @@ impl ModuleMemory {
...
     }
 
     /// Return the byte length of the memory container including all gaps (never-written virtual pages) it contains
     pub fn len(&self) -> usize {
-        if self.range.is_empty() {
-            0
-        } else {
-            usize::try_from(self.range.end - self.range.start).unwrap()
-        }
+        self.buf.len()
```
sangho2, line 1412: how does `self.buf.len()` have gaps?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1496,216 +1389,116 @@ impl ModuleMemory {
...
+                usize::try_from(Size4KiB::SIZE).unwrap() - src_offset,
+            );
+            let src = &page.0[src_offset..src_offset + src_len];
 
-        if num_bytes == data.len() {
-            self.extend_range(start, end);
-            Ok(())
-        } else {
-            Err(MemoryContainerError::WriteFailed)
+            self.buf.extend_from_slice(src);
```
sangho2, line 1489: `extend_from_slice` simply appends a slice to the current `Vec`. In that sense, my understanding is `buf` will contain data without gaps. Does it imply that we assume `range` will contain virtual addresses without any gaps? If there might be gaps, perhaps we need to maintain `Vec<Vec<u8>>` or `HashMap<Vec<u8>>`, or move `buf` to `MemoryRange`.

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Agreed. Will fix

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Done

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Fixed

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Fixed. Perhaps we should have some generic VsmError that uses thiserror package to coalesce errors from other libraries like elf, x86_64, etc.

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Updated function comment

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: There is one exception for PATCH_INFO which is not used today. Each patch type - jump label, ftrace and static call - has its own kernel data at a different virtual address. VTL0 kernel collects all the patch type data into PATCH_INFO data blob, with enough info to help VTL1 parse it. We only have jump label patch type today, so PATCH_INFO is virtually contiguous.

Ideally all patch data should be housed in one struct in VTL0 kernel before being copied over to reduce parsing and alignment issues. I am adding a TODO and println! to extend_range() to make discontiguous virtual address an error in the future

CvvT:
Looks like there is another flaky test [test_tun_tcp_connection_refused](https://github.com/microsoft/litebox/actions/runs/20080657897/job/57606635057). I will take a look and fix it ASAP. Meanwhile, please ignore it and try to merge this PR again.

PR #544, opened 2025-12-08
Add socketpair

CvvT:
Add [`socketpair`](https://www.man7.org/linux/man-pages/man2/socketpair.2.html) for UNIX sockets


wdcui:
I feel this PR is straightforward. We probably don't need to wait for John's review.

PR #545, opened 2025-12-09
Fix and improve VTL switching loop and state management

sangho2:
This PR fixes some bugs in the VTL switching loop and improve its reliability and performance.
Specifically, this PR uses assembly code to carefully switch between VTLs while saving/restoring
VTL0 registers. This PR removes some unreliable stack manipulation code we had before.
In addition, this PR makes it clear that we no longer store the VTL1 state in per-core storage.

Note: this PR and #541 should be merged together since #541 suffers from a crash due to
some bugs that this PR is fixing.

### DIFF
--- a/litebox_platform_lvbs/src/host/per_cpu_variables.rs
+++ b/litebox_platform_lvbs/src/host/per_cpu_variables.rs
@@ -192,17 +192,36 @@ static mut BSP_VARIABLES: PerCpuVariables = PerCpuVariables {
 
 /// Specify the layout of PerCpuVariables for Assembly area.
 ///
+/// Unlike userland platforms, this kernel platform does't rely on the `tbss` section
+/// to specify Kernel TLS because there is no ELF loader that will set up TLS for it.
+///
+/// Note that kernel & host and user & guest are interchangeable in this context.
+/// We use "kernel" and "user" here to emphasize that there must be hardware-enforced
+/// mode transitions (i.e., ring transitions through iretq/syscall) unlike userland
+/// platforms.
+///
 /// TODO: Consider unifying with `PerCpuVariables` if possible.
 #[non_exhaustive]
 #[cfg(target_arch = "x86_64")]
 #[repr(C)]
 #[derive(Clone)]
 pub struct PerCpuVariablesAsm {
-    kernel_stack_ptr: Cell<usize>,    // gs:[0x0]
+    /// Initial kernel stack pointer to reset the kernel stack on VTL switch
+    kernel_stack_ptr: Cell<usize>, // gs:[0x0]
+    /// Initial interrupt stack pointer for x86 IST
     interrupt_stack_ptr: Cell<usize>, // gs:[0x8]
-    vtl_return_addr: Cell<usize>,     // gs:[0x10]
-    scratch: Cell<usize>,             // gs:[0x18]
+    /// Return address for call-based VTL switching
+    vtl_return_addr: Cell<usize>, // gs:[0x10]
+    /// Scratch pad
+    scratch: Cell<usize>, // gs:[0x18]
+    /// Top address of VTL0 VTLState
     vtl0_state_top_addr: Cell<usize>, // gs:[0x20]
+    /// Current kernel stack pointer
+    cur_kernel_stack_ptr: Cell<usize>, // gs:[0x28]
+    /// Current kernel base pointer
+    cur_kernel_base_ptr: Cell<usize>, // gs:[0x30]
+    /// Top address of the user context area
+    user_context_top_addr: Cell<usize>, // gs:[0x38]
 }
 
 impl PerCpuVariablesAsm {
@@ -221,18 +240,30 @@ impl PerCpuVariablesAsm {
     pub fn set_vtl0_state_top_addr(&self, addr: usize) {
         self.vtl0_state_top_addr.set(addr);
     }
-}
-
-/// PerCpuVariablesAsm offsets. Difference between the GS value and an offset is used to access
-/// the corresponding variable.
-#[non_exhaustive]
-#[repr(usize)]
-pub enum PerCpuVariablesAsmOffset {
-    KernelStackPtr = offset_of!(PerCpuVariablesAsm, kernel_stack_ptr),
-    InterruptStackPtr = offset_of!(PerCpuVariablesAsm, interrupt_stack_ptr),
-    VtlReturnAddr = offset_of!(PerCpuVariablesAsm, vtl_return_addr),
-    Scratch = offset_of!(PerCpuVariablesAsm, scratch),
-    Vtl0StateTopAddr = offset_of!(PerCpuVariablesAsm, vtl0_state_top_addr),
+    pub const fn kernel_stack_ptr_offset() -> usize {
+        offset_of!(PerCpuVariablesAsm, kernel_stack_ptr)
+    }
+    pub const fn interrupt_stack_ptr_offset() -> usize {
+        offset_of!(PerCpuVariablesAsm, interrupt_stack_ptr)
+    }
+    pub const fn vtl_return_addr_offset() -> usize {
+        offset_of!(PerCpuVariablesAsm, vtl_return_addr)
+    }
+    pub const fn scratch_offset() -> usize {
+        offset_of!(PerCpuVariablesAsm, scratch)
+    }
+    pub const fn vtl0_state_top_addr_offset() -> usize {
+        offset_of!(PerCpuVariablesAsm, vtl0_state_top_addr)
+    }
+    pub const fn cur_kernel_stack_ptr_offset() -> usize {
+        offset_of!(PerCpuVariablesAsm, cur_kernel_stack_ptr)
+    }
+    pub const fn cur_kernel_base_ptr_offset() -> usize {
+        offset_of!(PerCpuVariablesAsm, cur_kernel_base_ptr)
+    }
+    pub const fn user_context_top_addr_offset() -> usize {
+        offset_of!(PerCpuVariablesAsm, user_context_top_addr)
+    }
 }
 
 /// Wrapper struct to maintain `RefCell` along with `PerCpuVariablesAsm`.
@@ -259,6 +290,9 @@ impl<T> RefCellWrapper<T> {
                 vtl_return_addr: Cell::new(0),
                 scratch: Cell::new(0),
                 vtl0_state_top_addr: Cell::new(0),
+                cur_kernel_stack_ptr: Cell::new(0),
+                cur_kernel_base_ptr: Cell::new(0),
+                user_context_top_addr: Cell::new(0),
             },
             inner: RefCell::new(value),
         }
--- a/litebox_platform_lvbs/src/mshv/vtl_switch.rs
+++ b/litebox_platform_lvbs/src/mshv/vtl_switch.rs
@@ -4,7 +4,7 @@ use crate::{
     host::{
         hv_hypercall_page_address,
         per_cpu_variables::{
-            PerCpuVariablesAsmOffset, with_per_cpu_variables, with_per_cpu_variables_asm_mut,
+            PerCpuVariablesAsm, with_per_cpu_variables, with_per_cpu_variables_asm_mut,
             with_per_cpu_variables_mut,
         },
     },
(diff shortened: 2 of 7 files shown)


litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -215,9 +196,25 @@ static mut BSP_VARIABLES: PerCpuVariables = PerCpuVariables {
 #[repr(C)]
 #[derive(Clone)]
 pub struct KernelTls {
-    pub kernel_stack_ptr: Cell<usize>,    // gs:[0x0]
-    pub interrupt_stack_ptr: Cell<usize>, // gs:[0x8]
-    pub vtl_return_addr: Cell<usize>,     // gs:[0x10]
+    kernel_stack_ptr: Cell<usize>,    // gs:[0x0]
```
wdcui: Is it okay to just use `u64`? The comment assumes each one is 8 bytes.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -80,237 +85,208 @@ impl VtlState {
...
+/// Loads VTL0 state from per-CPU storage to stack.
+///
+/// # Safety
+/// - `vtl_state` must be a valid pointer to a writable `VtlState` structure
+/// - Must be called from VTL1 context with valid per-CPU variables initialized
+unsafe extern "C" fn load_vtl0_state(vtl_state: *mut VtlState) {
+    with_per_cpu_variables(|pcv| {
+        let src = &raw const pcv.vtl0_state;
+        // SAFETY: src points to per-CPU storage with valid VtlState.
+        // vtl_state points to caller's stack frame with sufficient space.
```
wdcui: Does it have to be on the stack? If not, this comment is not accurate.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: let me specify the target arch (x86_64). `u64` should be fine, but we do use `usize` for storing addresses in many other places.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: you're right. This function doesn't need to assume that `vtl_state` is on the stack.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -80,237 +85,208 @@ impl VtlState {
...
-pub(crate) unsafe extern "C" fn jump_to_vtl_switch_loop() -> ! {
-    naked_asm!(
-        "call {loop}",
-        loop = sym vtl_switch_loop,
-    );
+/// Assembly macro to save VTL state using the stack as temporary storage.
+/// `fn_save_state` is the function to save VTL state stored in the stack.
+/// `scratch_off` is the offset of the kernel TLS which holds the scratch space to
+/// save/restore the current stack pointer.
+macro_rules! SAVE_VTL_STATE_ASM {
```
wdcui, line 104: Code like this should be moved out of `mshv` at some point?

wdcui: Do we need to handle xsave here?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -80,237 +85,208 @@ impl VtlState {
...
+            "\n",
+            "lea rdi, [rsp]\n",
+            "call ",
+            stringify!($fn_load_state),
+            "\n",
+            "mov rsp, gs:[",
+            stringify!($scratch_off),
+            "]\n", // anchor
+            "sub rsp, ",
+            stringify!($state_size),
```
wdcui: Is the `rsp` value after line 159 the same as after the call?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -80,237 +85,208 @@ impl VtlState {
...
-            per_cpu_variables.save_extended_states(HV_VTL_NORMAL);
-        });
+#[unsafe(naked)]
+unsafe extern "C" fn vtl_switch_loop_asm() -> ! {
+    naked_asm!(
+        "1:",
+        "mov rsp, gs:[{kernel_sp_off}]", // reset kernel stack pointer. Hyper-V saves/restores rsp and rip.
+        VTL_RETURN_ASM!({vtl_ret_addr_off}),
+        // *** VTL1 resumes here regardless of the entry reason (VTL switch or intercept) ***
+        SAVE_VTL_STATE_ASM!({save_vtl0_state}, {scratch_off}),
```
wdcui: Can we save the vtl0 state to the per cpu storage directly?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -80,237 +85,208 @@ impl VtlState {
...
+    // loading OP-TEE TAs). This should use global data structures since the core which makes the RPC
+    // can be different from the core where the VTL1 is running.
+    // TODO: Even if we don't have RPC from VTL1 to VTL0, we may still need to save VTL1's state for
+    // debugging purposes.
+
+    // Since we do not know whether the VTL0 kernel saves its extended states (e.g., if a VTL switch
+    // is due to memory or register access violation, the VTL0 kernel might not have saved
+    // its states), we conservatively save and restore its extended states on every VTL switch.
+    with_per_cpu_variables_mut(|per_cpu_variables| {
+        per_cpu_variables.save_extended_states(HV_VTL_NORMAL);
```
wdcui, line 195: How do we make sure all the code executed before we get to this point doesn't change the extended state?

litebox_platform_lvbs/src/syscall_entry.rs
```diff
@@ -175,20 +171,8 @@ fn syscall_entry(sysnr: u64, ctx_raw: *const SyscallContextRaw) -> usize {
...
-    });
-    unsafe {
-        asm!(
-            "mov rsp, rax",
-            "and rsp, -16",
-            in("rax") stack_top
-        );
-    }
-
+    // TODO: catch up up-to-date syscall handler's return path.
```
wdcui: Does any of your existing PRs cover this TODO?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: xsave and xstore are handled inside vtl_switch_loop_body because we don't particularly need assembly code for it.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: We save and restore rsp in Lines 156 and 177, respectively, using scrachpad.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: As far as I know, using stack as temporary storage to save the current registers is one of the most reliable and efficient ways.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: It only executes some assembly code and memory copy operations. I agree that some compiler optimization might be able to use SSE though.

litebox_platform_lvbs/src/syscall_entry.rs
sangho2: PR #566 covers it.

sangho2:
> It's not clear to me (1) why we have both per_cpu and ktls and (2) why we load/store vtl0 state from per_cpu via stack. I left some other minor comments.

The main reason I introduced `KernelTls` is that, assembly code cannot easily access `PerCpuVariables` because it is protected by `RefCell` (i.e., this is for Rust code not assembly code). `KernelTls` is named after `ShimTls` but at this point this is confusing because we no longer have `ShimTls`. Let me rename it (`PerCpuVariablesAsm` or something else).
I think using stack is most reliable and efficient way to save the current CPU registers. Also, as noted above, `PerCpuVariables` is protected by `RefCell` so assembly code cannot simply access it.

praveen-pk:
>this PR makes it clear that we no longer store the VTL1 state in per-core storage.

Is this being dropped because all VTL1 invocations can start with a fresh stack for now. Should this change be revisited once dynamic TAs are enabled?

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -444,6 +444,24 @@ pub fn allocate_per_cpu_variables() {
...
 
+/// Initialize the kernel and interrupt stack pointers in the Kernel TLS area with 16-byte alignment.
+///
+/// # Panics
+/// Panics if the per-CPU variables are not properly initialized.
+pub fn init_ktls_stack_pointers() {
+    const STACK_ALIGNMENT: usize = 16;
+    with_per_cpu_variables_mut(|per_cpu_variables| {
+        let kernel_sp =
+            usize::try_from(per_cpu_variables.kernel_stack_top()).unwrap() & !(STACK_ALIGNMENT - 1);
```
praveen-pk, line 498: Shouldn't kernel_stack_top always capture an aligned address?

Wouldn't it be better to `assert!` here rather than get the aligned address?

sangho2: This is to set up or initialize the kernel stack pointer. it doesn't need assertion (we can simply waste some bytes if the buffer is not aligned).

sangho2:
> > this PR makes it clear that we no longer store the VTL1 state in per-core storage.
> 
> Is this being dropped because all VTL1 invocations can start with a fresh stack for now. Should this change be revisited once dynamic TAs are enabled?

We shouldn't use per-core storage for saving the VTL1 state because, as discussed, we can't ensure that VTL1 is re-entered with the same core. We need a separate global data structure for this.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -80,237 +85,208 @@ impl VtlState {
...
+            "push rbx\n",
+            "push rax\n",
+            "push rbp\n",
+            "lea rdi, [rsp]\n",
+            "call ",
+            stringify!($fn_save_state),
+            "\n",
+            "mov rsp, gs:[",
+            stringify!($scratch_off),
+            "]\n",
```
CvvT: why not just do `add rsp, size_of::<VtlState>` instead of relying on scratch space?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -80,237 +85,208 @@ impl VtlState {
...
+            "lea rdi, [rsp]\n",
+            "call ",
+            stringify!($fn_load_state),
+            "\n",
+            "mov rsp, gs:[",
+            stringify!($scratch_off),
+            "]\n", // anchor
+            "sub rsp, ",
+            stringify!($state_size),
+            "\n",
```
CvvT: Is this necessary? `rsp` should be the same before and after the `call fn_load_state`.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -80,237 +85,208 @@ impl VtlState {
...
+            "pop r9\n",
+            "pop r10\n",
+            "pop r11\n",
+            "pop r12\n",
+            "pop r13\n",
+            "pop r14\n",
+            "pop r15\n",
+            "mov rsp, gs:[",
+            stringify!($scratch_off),
+            "]\n",
```
CvvT: Similarly, just do `add rsp`?

sangho2:
Let me add XSAVE/XRSTORE when I merge #541. doing this in this PR is a bit tricky due to complicated rebase.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
sangho2: Yes. This PR used to have unnecessary register saving/restoring and memory copy. It is revised to get rid of them.

PR #547, opened 2025-12-10
Add data structures for OP-TEE message protocol and SMC calling convention

sangho2:
This PR adds OP-TEE message protocol and SMC calling convention data structures to `litebox_common_optee`, which will be used by `litebox_shim_optee` and `litebox_platform_lvbs` for exchanging OP-TEE related commands and data with the normal world (VTL0) kernel.

### DIFF
--- a/litebox_common_optee/src/lib.rs
+++ b/litebox_common_optee/src/lib.rs
@@ -1161,7 +1161,7 @@ pub struct OpteeMsgParamValue {
     c: u64,
 }
 
-/// Parameter used totether with `OpteeMsgArg`
+/// Parameter used together with `OpteeMsgArg`
 #[derive(Clone, Copy)]
 #[repr(C)]
 pub union OpteeMsgParamUnion {
@@ -1332,7 +1332,7 @@ impl OpteeSmcFunction {
 
 /// OP-TEE SMC call result.
 /// OP-TEE SMC call uses CPU registers to pass input and output values.
-/// Thus, this structure is technically equivalent to `OpteeSmcArgs`, but we separete them for clarity.
+/// Thus, this structure is technically equivalent to `OpteeSmcArgs`, but we separate them for clarity.
 #[repr(align(4096))]
 #[derive(Clone, Copy)]
 #[repr(C)]


litebox_common_optee/src/lib.rs
```diff
@@ -1085,3 +1085,348 @@ impl LdelfArg {
...
+    /// Return value from the secure world
+    ret: u32,
+    /// Origin of the return value (e.g., OP-TEE OS, TA, ...)
+    ret_origin: u32,
+    /// Number of parameters contained in `params`
+    num_params: u32,
+    /// Parameters to be passed to the secure world. If `cmd == OpenSession`, the first two params contain
+    /// a TA UUID and they are not delivered to the TA.
+    /// Note that, originally, the length of this array is variable. We fix it to `TEE_NUM_PARAMS + 2` to
+    /// simply the implementation (our OP-TEE Shim supports up to four parameters as well).
```
wdcui: Minor: simply => simplify

litebox_common_optee/src/lib.rs
```diff
@@ -1085,3 +1085,348 @@ impl LdelfArg {
...
+    pub fn func_id(&self) -> Result<OpteeSmcFunction, Errno> {
+        OpteeSmcFunction::try_from(self.args[0] & OpteeSmcFunction::MASK).map_err(|_| Errno::EINVAL)
+    }
+
+    /// Get the physical address of `OpteeMsgArg`. The secure world is expected to map and copy
+    /// this structure.
+    /// # Panics
+    /// Panics if the computed address cannot be converted to `u64`.
+    pub fn optee_msg_arg_phys_addr(&self) -> Result<u64, Errno> {
+        // To avoid potential overflow issues, OP-TEE often only uses the low half of 64-bit variables.
```
wdcui: The comment says "often" but the code expects it otherwise it returns an error?

litebox_common_optee/src/lib.rs
sangho2: Fixed

litebox_common_optee/src/lib.rs
sangho2: Revised the comment. I did use "often" because the behavior is case by case. Better to clarify this in each function.

PR #548, opened 2025-12-10
Add unix datagram sockets

CvvT:
Add support for unix datagram socket and some unit tests similar to unix stream sockets.

Also fix an issue on double binding on the same address. Previously, I had a C test that showed it was okay to double bind on the same address. Turns out it was because of `unlink` that I overlooked. Now use `OFlags::EXCL` to ensure exclusive creation of sock file to address the issue.

### DIFF
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -2383,15 +2383,14 @@ mod unix_tests {
         task.do_socket(AddressFamily::UNIX, ty, flags, 0).unwrap()
     }
 
-    fn create_unix_server_socket(task: &Task, addr: &str, flags: SockFlags) -> u32 {
+    fn create_unix_server_socket(task: &Task, addr: &str, flags: SockFlags) -> Result<u32, Errno> {
         let server_fd = create_unix_socket(task, SockType::Stream, flags);
         task.do_bind(
             server_fd,
             SocketAddress::Unix(UnixSocketAddr::Path(addr.to_string())),
-        )
-        .unwrap();
-        task.do_listen(server_fd, 128).unwrap();
-        server_fd
+        )?;
+        task.do_listen(server_fd, 128)?;
+        Ok(server_fd)
     }
 
     fn close_socket(task: &crate::Task, fd: u32) {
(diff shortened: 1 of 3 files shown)


litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -2379,36 +2379,114 @@ mod unix_tests {
...
-        task.do_socket(AddressFamily::UNIX, SockType::Stream, flags, 0)
-            .unwrap()
+    fn create_unix_socket(task: &Task, ty: SockType, flags: SockFlags) -> u32 {
+        task.do_socket(AddressFamily::UNIX, ty, flags, 0).unwrap()
     }
 
-    fn create_unix_server_socket(task: &Task, addr: &str, flags: SockFlags) -> u32 {
-        let server_fd = create_unix_socket(task, flags);
+    fn create_unix_server_socket(task: &Task, addr: &str, flags: SockFlags) -> Result<u32, Errno> {
+        let server_fd = create_unix_socket(task, SockType::Stream, flags);
```
wdcui, line 2387: Why must it be a `Stream` socket?

litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -548,11 +560,272 @@ impl UnixConnectedStream {
...
+}
+
+/// Represents a Unix datagram socket.
+struct UnixDatagram {
+    /// The local address this socket is bound to, if any.
+    addr: Option<(UnixBoundSocketAddr, Arc<GlobalState>)>,
+    /// The read end of the local socket's channel.
+    reader: Option<ReadEnd<DatagramMessage>>,
+    /// The write end of the remote socket it is connected to, if any.
+    peer_writer: Option<WriteEnd<DatagramMessage>>,
```
wdcui, line 649: I'm not sure why the reader is called `reader` and the writer is called `peer_writer`?

litebox_shim_linux/src/syscalls/net.rs
CvvT: Only stream socket can call `listen` and `accept`.

litebox_shim_linux/src/syscalls/unix.rs
CvvT: these two are for different sockets. They are not paired. `peer_writer` refers to the write end of the other socket it connects to. `reader` cannot read data sent by `peer_writer`.

wdcui:
I resolved my comments. Thanks, Weiteng.

PR #550, opened 2025-12-10
Support nonblocking UNIX socket

CvvT:
This PR adds nonblocking mode support for UNIX socket operations (for both stream and datagram sockets), including `connect`, `accept`, `send/recv`, and etc. Current implementation leverages `WaitContext` and `Pollee` to replace previously temporary busy-looping solution.

Also updated some unit tests to test all nonblocking UNIX socket operations.

### DIFF
--- a/litebox_shim_linux/src/syscalls/unix.rs
+++ b/litebox_shim_linux/src/syscalls/unix.rs
@@ -292,7 +292,6 @@ impl Backlog {
     }
 
     fn check_io_events(&self) -> Events {
-        let limit = self.limit.load(Ordering::Relaxed);
         let sockets = self.sockets.lock();
         let Some(sockets) = &*sockets else {
             return Events::HUP;
@@ -301,7 +300,7 @@ impl Backlog {
         if !sockets.is_empty() {
             events |= Events::IN;
         }
-        if sockets.len() < limit as usize {
+        if sockets.len() < self.limit.load(Ordering::Relaxed) as usize {
             events |= Events::OUT;
         }
         events
@@ -401,8 +400,10 @@ struct Message {
 /// Represents a connected Unix stream socket.
 struct UnixConnectedStream {
     addr: AddrView,
-    reader: crate::channel::ReadEnd<Message>,
-    writer: crate::channel::WriteEnd<Message>,
+    /// The read end of the local socket's channel for receiving messages.
+    recv_channel: crate::channel::ReadEnd<Message>,
+    /// The write end of the connected peer socket for sending messages.
+    connected_send_channel: crate::channel::WriteEnd<Message>,
     pollee: Arc<Pollee<crate::Platform>>,
 }
 
@@ -417,21 +418,22 @@ impl UnixConnectedStream {
         let (addr1, addr2) = AddrView::new_pair(addr, peer);
         let pollee1 = pollee.unwrap_or(Arc::new(Pollee::new()));
         let pollee2 = Arc::new(Pollee::new());
-        let (writer_peer, reader) =
+        let (send_channel, recv_channel) =
             crate::channel::Channel::new(UNIX_BUF_SIZE, pollee2.clone(), pollee1.clone()).split();
-        let (writer, reader_peer) =
+        let (send_channel_peer, recv_channel_peer) =
             crate::channel::Channel::new(UNIX_BUF_SIZE, pollee1.clone(), pollee2.clone()).split();
         (
+            // Cross-wire: each socket keeps the other side's send channel.
             UnixConnectedStream {
                 addr: addr1,
-                reader,
-                writer,
+                recv_channel,
+                connected_send_channel: send_channel_peer,
                 pollee: pollee1,
             },
             UnixConnectedStream {
                 addr: addr2,
-                reader: reader_peer,
-                writer: writer_peer,
+                recv_channel: recv_channel_peer,
+                connected_send_channel: send_channel,
                 pollee: pollee2,
             },
         )
@@ -453,13 +455,13 @@ impl UnixConnectedStream {
 
     fn try_sendto(&self, msg: Message) -> Result<(), (Message, Errno)> {
         // TODO: write partial data?
-        self.writer.try_write_one(msg)
+        self.connected_send_channel.try_write_one(msg)
     }
 
     fn try_recvfrom(&self, mut buf: &mut [u8]) -> Result<usize, TryOpError<Errno>> {
         let mut total_read = 0;
         while !buf.is_empty() {
-            let n = match self.reader.peek_and_consume_one(|msg| {
+            let n = match self.recv_channel.peek_and_consume_one(|msg| {
                 if buf.len() >= msg.data.len() {
                     buf[..msg.data.len()].copy_from_slice(&msg.data);
                     Ok((true, msg.data.len()))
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -292,59 +303,75 @@ impl Backlog {
...
+        let limit = self.limit.load(Ordering::Relaxed);
+        let sockets = self.sockets.lock();
+        let Some(sockets) = &*sockets else {
+            return Events::HUP;
+        };
+        let mut events = Events::empty();
+        if !sockets.is_empty() {
+            events |= Events::IN;
+        }
+        if sockets.len() < limit as usize {
```
wdcui: Why is the number of sockets related to the `Events::OUT`?

litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -292,59 +303,75 @@ impl Backlog {
...
-        }
+        let mut state = Some(init);
+        cx.wait_on_events(
+            is_nonblocking,
+            Events::OUT,
+            |observer, mask| {
+                self.pollee.register_observer(observer, mask);
+                Ok(())
+            },
+            || match self.try_connect(state.take().unwrap()) {
```
wdcui: Minor: would it make sense to move the following code to `try_connect` to make it more like `try_accept`?

litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -671,22 +742,29 @@ impl UnixDatagram {
...
+            pollee: Arc::new(Pollee::new()),
         }
     }
 
     fn new_pair() -> (UnixDatagram, UnixDatagram) {
-        let (writer, reader) = crate::channel::Channel::new(UNIX_BUF_SIZE).split();
-        let (writer_peer, reader_peer) = crate::channel::Channel::new(UNIX_BUF_SIZE).split();
+        let pollee1 = Arc::new(Pollee::new());
+        let pollee2 = Arc::new(Pollee::new());
+        let (writer_peer, reader) =
```
wdcui: This code uses `pollee1` and `pollee2` in the opposite way as the `new_pair` function in `UnixConnectedStream`.

litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -1015,13 +1116,29 @@ impl UnixSocket {
         let is_nonblocking =
             flags.contains(ReceiveFlags::DONTWAIT) || self.get_status().contains(OFlags::NONBLOCK);
 
-        self.with_state_ref(|state| match state {
+        let ret = self.with_state_ref(|state| match state {
             UnixSocketState::InitStream(_) | UnixSocketState::ListenStream(_) => Err(Errno::EINVAL),
             UnixSocketState::ConnectedStream(connect) => {
-                connect.recvfrom(buf, is_nonblocking, source_addr)
+                if let Some(source_addr) = source_addr {
```
wdcui: Should we call a function like `ConnectedStream::recvfrom`?

litebox_shim_linux/src/syscalls/unix.rs
CvvT: For a listening socket with backlog limit, `Events:OUT` means it can accept more connections.

litebox_shim_linux/src/syscalls/unix.rs
CvvT: For `new_pair`, `pollee1` comes from the arguments so it has to be passed to the right channel. In this case, both `pollee1` and `pollee2` are newly created, so the order does not matter. For consistency, I updated it.

litebox_shim_linux/src/syscalls/unix.rs
CvvT: Sure, updated.

litebox_shim_linux/src/syscalls/unix.rs
CvvT: This one is tricky because it takes `init: UnixInitStream` and returns `Result<UnixConnectedStream, (UnixInitStream, Errno)>` and `try_op/try_connect` can be called multiple times. I will try to see how to improve it.

litebox_shim_linux/src/syscalls/unix.rs
wdcui: Got it. Thanks.

litebox_shim_linux/src/syscalls/unix.rs
wdcui: Thanks!

litebox_shim_linux/src/syscalls/unix.rs
wdcui: Thanks.

litebox_shim_linux/src/syscalls/unix.rs
wdcui: It's okay to keep it if it's not an easy fix.

CvvT:
Hi @wdcui I noticed one potential issue and made the following change. Please take a look again.
  1. Reordered operation sequence: Moved `wait_on_events` calls before state access (i.e., `with_state*` methods) in blocking sockets operations like `connect`. This allows multiple concurrent syscalls to access the same socket without holding locks during waits.
  2. Datagram sockets (`UnixDatagram` now have their own dedicated structure separate from stream sockets (`UnixStream`) to support the first point and provide clearner state management.

litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -292,59 +269,42 @@ impl Backlog {
...
+        let limit = self.limit.load(Ordering::Relaxed);
+        let sockets = self.sockets.lock();
+        let Some(sockets) = &*sockets else {
+            return Events::HUP;
+        };
+        let mut events = Events::empty();
+        if !sockets.is_empty() {
+            events |= Events::IN;
+        }
+        if sockets.len() < limit as usize {
```
wdcui: Should we read `limit` here? If we load `limit` before we have the lock of `sockets`, could it be changed before we check it here?

litebox_shim_linux/src/syscalls/unix.rs
```diff
@@ -715,6 +931,51 @@ impl UnixDatagram {
...
+        let pollee2 = Arc::new(Pollee::new());
+        let (writer_peer, reader) =
+            crate::channel::Channel::new(UNIX_BUF_SIZE, pollee2.clone(), pollee1.clone()).split();
+        let (writer, reader_peer) =
+            crate::channel::Channel::new(UNIX_BUF_SIZE, pollee1.clone(), pollee2.clone()).split();
+        (
+            UnixDatagram {
+                inner: RwLock::new(UnixDatagramInner {
+                    addr: None,
+                    reader: Some(reader),
```
wdcui: It might help to add some comments about the relations between `reader`, `writer`, `peer_reader` and `peer_writer`. It's hard for me to check this function without a good understanding of their relations.

litebox_shim_linux/src/syscalls/unix.rs
CvvT: Good point! Updated.

litebox_shim_linux/src/syscalls/unix.rs
CvvT: I replaced all `reader`, `peer_writer` with `recv_channel` and `connected_send_channel`, respectively. The key is that for a connected socket, it stores its own read end and the write end of its connected peer. Added some comment here as well.

CvvT:
Hi @jstarks Do you plan to take a look at this PR? If not, I will merge it later today.

wdcui:
@CvvT , should we merge this PR?

CvvT:
> @CvvT , should we merge this PR?

Sure, I will do that.

PR #552, opened 2025-12-11
Support tar files with `./` prefixes

jaybosamiya-ms:
Strip the `./` prefix from tar filenames if present.  This is helpful for tar files that have been created via `tar cvf foo.tar .`.


jaybosamiya-ms:
The actual use case for this comes out of some tooling I'm building for the benchmarking, where having the ability to just do `tar -C /abs/path -cvf /abs/path.tar .` is more convenient than listing things out.  But yeah, generically useful for us to handle this, so just a separate PR to fix this up.

I don't like the amount of duplication that this introduces, but a future PR after I'm done with the benchmarking stuff will be cleaning up the linear pass here, so at that point, it'll automatically clean things up.  This is mostly temporary duplication introduced until that fix lands (after the benchmarking lands).

PR #554, opened 2025-12-12
Introduce a benchmarking framework

jaybosamiya-ms:
This PR introduces `dev_bench`, to be run with `cargo run --bin dev_bench`.  Use `cargo run --bin dev_bench -- --help` to see the currently implemented features.

In short, the intention is that this tool makes it super easy to perform comparisons across various versions of LiteBox.  It is currently somewhat barebones in terms of _currently_ included benchmarks (see below), but these can be easily extended as we introduce new things we wish to compare.

I initially considered using existing benchmarking tools available for Rust, but each of them had some drawbacks that made them suboptimal for our use case, so instead developed this basic framework, which we can then evolve as we find useful.

For anyone working on perf improvements, they are expected to run with `--record main` on the `main` branch (the tool will remind you) and then on their own branch, running the benchmarking tool will produce a table similar to the following (in markdown syntax, to allow easy pasting into GitHub comments):

| Benchmark                  | main (ms) | joint-ribbonfish (ms) | Diff (ms) |
|:---------------------------|----------:|----------------------:|----------:|
| rewriter_hello_static      |        22 |                    22 |         0 |
| rewriter_node              |      1581 |                  1652 |        71 |
| run_rewritten_hello_static |        72 |                    71 |        -1 |
| run_rewritten_node         |       422 |                   305 |      -117 |

The `joint-ribbonfish` above is just an auto-generated name; you can pick names manually using `--record`.  The intention is to allow comparisons between various runs via `--compare` if needed (it always auto-compares to `main` if `main` has been recorded).

The framework also supports summarizing across multiple runs.  I may (in a future PR) set up an automated multi-run stabilization (otherwise system noise can overwhelm things esp in short-running benchmarks), but for now, you need to opt-in (again, check out `--help`).

### DIFF
--- a/dev_tests/src/ratchet.rs
+++ b/dev_tests/src/ratchet.rs
@@ -25,6 +25,7 @@ fn ratchet_transmutes() -> Result<()> {
 fn ratchet_globals() -> Result<()> {
     ratchet(
         &[
+            ("dev_bench/", 1),
             ("litebox/", 8),
             ("litebox_platform_linux_kernel/", 5),
             ("litebox_platform_linux_userland/", 5),


jaybosamiya-ms:
Just making a few notes in terms of TODO items that I think might be good to support at some point (not part of this PR; we should do these only as we hit the use cases for this):
- Add more benchmarks
- Add support for cross-technique benchmarks (e.g., compare "native" vs "litebox w/ rewriter" vs "litebox w/ seccomp").  We can temporarily just emulate this by having more rows (i.e., more benchmarks).
- Add support for benchmark groups (we may not want to run all benchmarks all the time).
- Add some support for `--profiler linux-perf` and similar, so that it will make it easier to get profiling traces.
- De-duplicate some of the commonality between the testing framework within `litebox_runner_linux_userland` and `dev_bench`. It is fairly little duplication, but I think we can make it nicer on both sides.  Not particularly high priority, since this is meant more for dev tooling, but I am aware of ways to clean it up.
- Add a `--debug-build` or similar that benchmarks the debug mode rather than the release mode.  _Can_ be helpful, but not a priority.
- Add `--run-for-time 60` or something that runs as many iterations it can in _roughly_ 60 seconds as a different way to denoise (might even be good to make this the default).
- Consider adding some version of this to the CI, so that every PR automatically gets a report added with the perf on it.  One worry is the noisiness of the CI, so I don't want to do this right now, and instead keep it limited to running it manually locally, but in the future, we probably want some version of this once we are able to make things more stable.  This _may_ be easier if we measure more stable metrics rather than time somehow (some valgrind based tools like cachegrind/callgrind support something like this), but that comes at other costs, so something to be thought about before doing it.
- Consider adding a "the main recording might be too old" or similar warning.  Not sure if we're going to need it, but yeah, it is feasible to believe that otherwise unrelated changes might be considered on a PR.
- Add support for checking for performance regressions (maybe this is just a "exit with correct exit code if `slower/faster` than `foo` by more than `X`%", so that it can work with `git bisect`).

PR #555, opened 2025-12-12
Add handlers for OP-TEE message protocol and SMC

sangho2:
This PR adds OP-TEE message protocol and SMC handlers (a handler for TA request is
covered in another PR).

Also, this PR adds traits for physical memory access to (copy from/to) the memory of
a remote environment (e.g., VTL0, normal world, ...), `PhysMutPtr` and `PhysConstPtr`.
These traits expect platforms will implement Linux kernel's `vmap`-like functions to
enable virtually contiguous mapping of non-contiguous physical pages on demand.

### DIFF
--- a/litebox_common_optee/src/lib.rs
+++ b/litebox_common_optee/src/lib.rs
@@ -1566,7 +1566,7 @@ impl From<OpteeSmcResult<'_>> for OpteeSmcArgs {
             }
             OpteeSmcResult::CallWithArg { .. } => {
                 panic!(
-                    "OpteeSmcResult::CallWithArg cannot be converted to OpteeSmcArgs directly. Handle the incorporate OpteeMsgArgs."
+                    "OpteeSmcResult::CallWithArg cannot be converted to OpteeSmcArgs directly. Handle the incorporated OpteeMsgArgs."
                 );
             }
         }


litebox_common_optee/src/lib.rs
```diff
@@ -1233,76 +1293,129 @@ impl OpteeMsgParam {
...
     /// Note that, originally, the length of this array is variable. We fix it to `TEE_NUM_PARAMS + 2` to
     /// simplify the implementation (our OP-TEE Shim supports up to four parameters as well).
-    params: [OpteeMsgParam; TEE_NUM_PARAMS + 2],
+    pub params: [OpteeMsgParam; TEE_NUM_PARAMS + 2],
 }
 
-/// OP-TEE SMC call arguments.
+impl OpteeMsgArg {
+    pub fn get_param_tmem(&self, index: usize) -> Result<OpteeMsgParamTmem, OpteeSmcReturn> {
+        if index >= self.params.len() || index >= self.num_params as usize {
```
wdcui: What's the relation between `params.len()` and `num_params`?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -0,0 +1,261 @@
...
+/// A scatter-gather list of OP-TEE shared physical pages in VTL0.
+#[derive(Clone, Copy)]
+#[repr(C)]
+struct ShmRefPagesData {
+    pub pages_list: [u64; Self::PAGELIST_ENTRIES_PER_PAGE],
+    pub next_page_data: u64,
+}
+impl ShmRefPagesData {
+    const PAGELIST_ENTRIES_PER_PAGE: usize =
+        PAGE_SIZE / core::mem::size_of::<u64>() - core::mem::size_of::<u64>();
```
wdcui, line 200: Why does it subtract 8?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -0,0 +1,261 @@
...
+        guard.remove(&shm_ref)
+    }
+
+    #[expect(unused)]
+    pub fn get(&self, shm_ref: u64) -> Option<ShmRefInfo> {
+        let guard = self.inner.lock();
+        guard.get(&shm_ref).cloned()
+    }
+
+    pub fn register_shm(
```
wdcui, line 256: Maybe add some comment for this function? It's not easy to follow this function.

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,204 @@
...
+
+use litebox::platform::{RawConstPointer, RawMutPointer};
+
+// TODO: use the one from the litebox crate
+pub trait ValidateAccess {}
+
+/// Trait to access a pointer to remote memory
+/// For now, we only consider copying the entire value before acccessing it.
+/// We do not consider byte-level access or unaligned access.
+pub trait RemoteMemoryAccess {
```
jaybosamiya-ms: Do you expect alternate `impl`s for this?  For now, I am only seeing in impl for `Vtl0PhysMemoryAccess` here, which itself is a placeholder impl, so I don't quite know how to evaluate whether this trait is being sufficiently helpful or not.  I can _guess_ why you're attempting something like this, but maybe it is not clear at the moment how you intend to use it for non-`Vtl0PhysMemoryAccess` and maybe it gets clearer then?

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn read_at_offset<T>(ptr: *mut T, count: isize) -> Option<T>;
+
+    fn write_at_offset<T>(ptr: *mut T, count: isize, value: T) -> Option<()>;
+
+    fn slice_from<T>(ptr: *mut T, len: usize) -> Option<alloc::boxed::Box<[T]>>;
```
jaybosamiya-ms: I'm not sure why this is a `Box<[T]>` and not a `Vec<T>` if you are already copying things over.

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,204 @@
...
+use litebox::platform::{RawConstPointer, RawMutPointer};
+
+// TODO: use the one from the litebox crate
+pub trait ValidateAccess {}
+
+/// Trait to access a pointer to remote memory
+/// For now, we only consider copying the entire value before acccessing it.
+/// We do not consider byte-level access or unaligned access.
+pub trait RemoteMemoryAccess {
+    fn read_at_offset<T>(ptr: *mut T, count: isize) -> Option<T>;
```
jaybosamiya-ms: Not sure why this is a `*mut T` and not a `*const T`.

Also, I believe each of these functions should come with more documentation.  Like how do we "check" that it is remote memory or not.  What are the situations where it can return a `None` or not.  Can it ever panic?  Can it ever lead to UB?

Also, this function should definitely be `unsafe`, since it is equivalent in some sense to a raw pointer deref.

jaybosamiya-ms: Unsurprisingly, these thoughts apply to the other relevant methods in this trait too.

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    fn write_at_offset<T>(ptr: *mut T, count: isize, value: T) -> Option<()>;
+
+    fn slice_from<T>(ptr: *mut T, len: usize) -> Option<alloc::boxed::Box<[T]>>;
+
+    fn copy_from_slice<T>(start_offset: usize, buf: &[T]) -> Option<()>;
```
jaybosamiya-ms: Copy from slice to _where_?  (there is no pointer argument here?).  Also what is expected to happen if only part of it can be copied over, does it get left in a partly copied state, or is exited early, or what?

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,204 @@
+//! Placeholders for specifying remote pointer access (e.g., reading data from
+//! VTL0 physical memory)
+//! TODO: Improve these and move these to the litebox crate later
+
+use litebox::platform::{RawConstPointer, RawMutPointer};
+
+// TODO: use the one from the litebox crate
+pub trait ValidateAccess {}
```
jaybosamiya-ms: A bit more of a general comment, do these (all across this file) need to be `pub` or can they be `pub(crate)`?  Anything made `pub` is something we are guaranteeing that users _outside this crate_ can depend upon, which has a much higher bar of designing the interfaces right (since any changes are _necessarily_ a breaking change), but anything that is internal to the crate is much more ok wrt changing as we understand it better.

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<V, A, T> Copy for RemoteConstPtr<V, A, T> {}
+
+impl<V, A, T: Clone> core::fmt::Debug for RemoteConstPtr<V, A, T> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_tuple("RemoteConstPtr").field(&self.inner).finish()
```
jaybosamiya-ms: Technically I think this should be `.finish_non_exhaustive()` right?

litebox_common_optee/src/lib.rs
sangho2: It was about validating the input. Added a separate function to do this.

litebox_shim_optee/src/msg_handler.rs
sangho2: I added detailed comments.

litebox_shim_optee/src/msg_handler.rs
sangho2: Done.

litebox_shim_optee/src/ptr.rs
sangho2: I expect alternative impls (at least in the future). A platform might need to implement certain versions of this (VTLs, VMPLs, TD partitions, different processes, or even DMA/RDMA). It's about letting two different address spaces exchange data without (persistent) memory mapping.

litebox_shim_optee/src/ptr.rs
sangho2: `Vec<T>` would work. I just wanted to point out that we don't need a dynamic array.

litebox_shim_optee/src/ptr.rs
sangho2: I'll add more documentation. `ValidateAccess` will do extra checks. How to check it heavily depends on the underlying platform and a bit unclear whether we can specify it here (e.g., it could be determined by a physical address range, page table flag, or something else).

litebox_shim_optee/src/ptr.rs
sangho2: Is it about a lack of documentation? I'll add it (technically, these function signatures are from `litebox/platform/mod.rs`).

litebox_shim_optee/src/ptr.rs
sangho2: This is from `platform/common_providers/userspace_pointers.rs`. Eventually, platforms need to implement these, and I plan to follow existing interfaces.

litebox_shim_optee/src/ptr.rs
sangho2: Let me see whether `finish_non_exhaustive` makes more sense here.

litebox_shim_optee/src/ptr.rs
jaybosamiya-ms: For this particular one it was not about lack of documentation, but more about there not being a `ptr` argument at all (neither `self` not `ptr: *mut T`), sorry my comment was not particularly clear.

litebox_shim_optee/src/ptr.rs
jaybosamiya-ms: Ah, it appears I misunderstood this particular module and why it was showing up here.  You are using it as staging ground for stuff to _eventually_ show up inside `litebox` itself.  Since ptr.rs has changed a significant amount, I will just re-review the entire file under the updated version.  Thanks for the clarification!

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,532 @@
...
+        }
+        Ok(Self {
+            pages,
+            offset,
+            count: size / core::mem::size_of::<T>(),
+            map_info: None,
+            _type: core::marker::PhantomData,
+        })
+    }
+    /// Create a new `PhysMutPtr` from the given contiguous physical address and length.
```
jaybosamiya-ms, line 133: Nit: blank lines between functions help with navigation

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,532 @@
...
+    ///
+    /// This is a shortcut for `try_from_page_array([align_down(pa), ..., align_up(align_down(pa) + bytes)], pa % ALIGN)`.
+    /// The caller must ensure that `pa`, ..., `pa+bytes` are both physically and virtually contiguous.
```
jaybosamiya-ms: Mostly to ask for clarification for myself to understand better:

Usually a "caller must ensure" implies that there is a downside to not ensuring it, which is mildly surprising for any function that is a `try_...`.  In this case, is it that this will cause an `Err` or will it cause a UB?  My understanding here is that it is the former, is this correct?  If so, the phrasing better suited here might be something closer to "Returns `Err(..)` if `pa`, ..., `pa+bytes` is not virtually contiguous." or similar to that, right?

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,532 @@
...
+    pub fn try_from_usize(pa: usize) -> Result<Self, PhysPointerError> {
+        Self::try_from_contiguous_pages(pa, core::mem::size_of::<T>())
+    }
+    /// Read the value at the given offset from the physical pointer.
+    ///
+    /// # Safety
+    ///
+    /// The caller should be aware that the given physical address might be concurrently accessed by
+    /// other entities (e.g., the normal world kernel) if there is no extra security mechanism
+    /// in place (e.g., by the hypervisor or hardware). That it, it might read corrupt data.
```
jaybosamiya-ms: Nit: That it -> That is?

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,532 @@
...
+    ///
+    /// Note: This module doesn't provide `as_usize` because LiteBox should not dereference physical addresses directly.
+    pub fn try_from_usize(pa: usize) -> Result<Self, PhysPointerError> {
+        Self::try_from_contiguous_pages(pa, core::mem::size_of::<T>())
+    }
+    /// Read the value at the given offset from the physical pointer.
+    ///
+    /// # Safety
+    ///
+    /// The caller should be aware that the given physical address might be concurrently accessed by
```
jaybosamiya-ms: Clarification: concurrent reads are not an issue, but writes are an issue, yes?

litebox_shim_optee/src/ptr.rs
sangho2: I'll do that.

litebox_shim_optee/src/ptr.rs
sangho2: Agree with you. I'll revise the comment.

litebox_shim_optee/src/ptr.rs
sangho2: Yes. Concurrent read would be fine. I'll clarify it.

sangho2:
If we merge #572 first, we can get rid of `PhysConstPtr` and `PhysMutPtr`'s usage of `litebox_platform_multiplex::platform()`.

litebox_common_linux/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+use thiserror::Error;
+
+/// A provider to map and unmap physical pages with virtually contiguous addresses.
+///
+/// `ALIGN`: The page frame size.
+///
+/// This provider is written to implement `litebox_shim_optee::ptr::PhysMutPtr` and
+/// `litebox_shim_optee::ptr::PhysConstPtr`. It can benefit other modules which need
+/// Linux kernel's `vmap()` and `vunmap()` functionalities (e.g., HVCI/HEKI, drivers).
+pub trait VmapProvider<const ALIGN: usize> {
```
wdcui: @CvvT @jaybosamiya-ms Can you two take a look at this new provider?

wdcui: Why do we need both this `vmap` provider and the other page management provider?

litebox_common_optee/src/lib.rs
```diff
@@ -1233,76 +1292,141 @@ impl OpteeMsgParam {
...
+
+/// A memory page to exchange OP-TEE SMC call arguments.
 /// OP-TEE assumes that the underlying architecture is Arm with TrustZone and
 /// thus it uses Secure Monitor Call (SMC) calling convention (SMCCC).
 /// Since we currently rely on the existing OP-TEE driver which assumes SMCCC, we translate it into
 /// our VTL switch convention.
 /// Specifically, OP-TEE SMC call uses up to nine CPU registers to pass arguments.
 /// However, since VTL call only supports up to four parameters, we allocate a VTL0 memory page and
 /// exchange all arguments through that memory page.
+/// TODO: Since this is LVBS-specific structure to facilitate the translation between VTL call convention,
```
wdcui, line 1393: Create an issue to track this?

litebox_common_optee/src/lib.rs
```diff
@@ -1332,64 +1457,94 @@ impl OpteeSmcFunction {
...
+                smc.args[0] = status as usize;
+                smc.args[1] = capabilities.bits();
+                smc.args[2] = max_notif_value;
+                smc.args[3] = data;
+                smc
+            }
+            OpteeSmcResult::Uuid { data } => {
+                let mut smc = OpteeSmcArgs::default();
+                for (i, arg) in smc.args.iter_mut().enumerate().take(4) {
+                    *arg = data[i] as usize;
```
wdcui, line 1535: Why do we convert it from `u32` to `usize`?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -755,6 +758,57 @@ impl<Host: HostInterface> StdioProvider for LinuxKernel<Host> {
...
+
+    unsafe fn vmap(
+        &self,
+        pages: Self::PhysPageAddrArray,
+        _perms: PhysPageMapPermissions,
+    ) -> Result<Self::PhysPageMapInfo, PhysPointerError> {
+        // TODO: Remove this check once this platform supports virtually contiguous
+        // non-contiguous physical page mapping.
+        check_contiguity(&pages)?;
+        todo!("use map_vtl0_phys_range()")
```
wdcui: If we don't need to implement these functions in this PR, should we move vmap related code to a separate PR?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -0,0 +1,335 @@
...
+//!
+//! Since the normal world is out of LiteBox's scope, the OP-TEE shim starts with handling
+//! an OP-TEE SMC call from the normal-world OP-TEE driver which consists of
+//! up to nine register values. By checking the SMC function ID, the shim determines whether
+//! it is for passing an OP-TEE message or a pure SMC function call (e.g., get OP-TEE OS
+//! version). If it is for passing an OP-TEE message/command, the shim accesses a normal world
+//! physical address containing `OpteeMsgArg` structure (the address is contained in
+//! the SMC call arguments). This `OpteeMsgArg` structure may contain references to normal
+//! world physical addresses to exchange a large amount of data. Also, like the OP-TEE
+//! SMC call, a certain OP-TEE message/command does not involve with any TA (e.g., register
```
wdcui: Nit: I don't follow this comment. What is "a certain OP-TEE message/command"?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -0,0 +1,335 @@
...
+    address & !(PAGE_SIZE as u64 - 1)
+}
+
+#[inline]
+fn page_align_up(len: u64) -> u64 {
+    len.next_multiple_of(PAGE_SIZE as u64)
+}
+
+/// The result of handling an OP-TEE SMC call along with an extracted OP-TEE message argument to handle.
+pub struct OpteeSmcHandled<'a> {
```
wdcui: I found this struct name a little confusing. It stores both the result and the args. What is it for?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -0,0 +1,335 @@
...
+
+/// This function handles `OpteeSmcArgs` passed from the normal world (VTL0) via an OP-TEE SMC call.
+/// It returns an `OpteeSmcResult` representing the result of the SMC call and
+/// an optional `OpteeMsgArg` if the SMC call involves with an OP-TEE messagewhich should be handled by
+/// `handle_optee_msg_arg` or `handle_ta_request`.
+///
+/// # Panics
+///
+/// Panics if the normal world physical address in `smc` cannot be converted to `usize`.
+pub fn handle_optee_smc_args(
```
wdcui, line 67: What is `OpteeSmcReturn`?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -0,0 +1,335 @@
...
+        &self,
+        shm_ref_pages_data_phys_addr: u64,
+        page_offset: u64,
+        aligned_size: u64,
+        shm_ref: u64,
+    ) -> Result<(), OpteeSmcReturn> {
+        if page_offset >= ALIGN as u64 || aligned_size == 0 {
+            return Err(OpteeSmcReturn::EBadAddr);
+        }
+        let num_pages = usize::try_from(aligned_size).unwrap() / ALIGN;
```
wdcui, line 266: Is `aligned_size` already aligned up?

litebox_shim_optee/src/ptr.rs
wdcui, line 1: My understanding is that @jaybosamiya-ms reviewed this file, so I will skip it.

litebox_common_linux/src/vmap.rs
sangho2: We can move these functions to `PageManagementProvider` if this makes sense but I think these two have quite different purposes (virtual vs physical, mandatory vs optional, ...)

litebox_common_optee/src/lib.rs
sangho2: smc.args mimic ARM SMC calling convention's registers. This assignment works like we write a value to a general-purpose register.

litebox_platform_lvbs/src/lib.rs
sangho2: I already implemented this function and can replace this with that code if that's better.

litebox_shim_optee/src/msg_handler.rs
sangho2: let me clarify this. In short, there are some OP-TEE messages that will be handled by OP-TEE OS (or OP-TEE Shim) not a TA.

litebox_shim_optee/src/msg_handler.rs
sangho2: Let me rename it. OP-TEE SMC can contain an OP-TEE message. We should decode/handle OP-TEE SMC first to see whether it contains an OP-TEE message and, if it does, we should call a handler against the contained OP-TEE message.

litebox_shim_optee/src/msg_handler.rs
sangho2: It's a return code.

litebox_shim_optee/src/msg_handler.rs
sangho2: yes

litebox_common_optee/src/lib.rs
wdcui: Got it. Thanks.

litebox_shim_optee/src/msg_handler.rs
wdcui: "Some OP-TEE messages/commands ..."

litebox_common_linux/src/vmap.rs
jaybosamiya-ms, line 1: Does this need to be its own module, or can the stuff in this file sit inside `page_mgmt.rs`?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -2146,6 +2146,41 @@ impl litebox::platform::CrngProvider for LinuxUserland {
     }
 }
 
+/// Dummy `VmapProvider`.
+///
+/// In general, userland platforms do not support `vmap` and `vunmap` (which are kernel functions).
+/// We might need to emulate these functions' behaviors using virtual addresses for development or
+/// testing, or use a kernel module to provide this functionality (if needed).
```
jaybosamiya-ms: If the expectation is handling of dummies via `UnsupportedOperation`, then you can just make it default functionality on the trait, such that traits don't actually need to implement this _unless_ they want to override it.

litebox_common_linux/src/vmap.rs
jaybosamiya-ms: Keeping them as separate traits I think is the correct move here

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+///
+/// This provider is written to implement `litebox_shim_optee::ptr::PhysMutPtr` and
+/// `litebox_shim_optee::ptr::PhysConstPtr`. It can benefit other modules which need
+/// Linux kernel's `vmap()` and `vunmap()` functionalities (e.g., HVCI/HEKI, drivers).
+pub trait VmapProvider<const ALIGN: usize> {
+    /// Data structure for an array of physical page addresses which are virtually contiguous.
+    type PhysPageAddrArray;
+
+    /// Data structure to maintain the mapping information returned by `vmap()`.
+    type PhysPageMapInfo;
```
jaybosamiya-ms: These types seem to always be assigned to 
```
    type PhysPageAddrArray = alloc::boxed::Box<[PhysPageAddr<ALIGN>]>;
    type PhysPageMapInfo = PhysPageMapInfo<ALIGN>;
```

Is there a reason to make these generic here, rather than just using the specific types directly?

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+use crate::platform::page_mgmt::MemoryRegionPermissions;
+use thiserror::Error;
+
+/// A provider to map and unmap physical pages with virtually contiguous addresses.
+///
+/// `ALIGN`: The page frame size.
+///
+/// This provider is written to implement `litebox_shim_optee::ptr::PhysMutPtr` and
+/// `litebox_shim_optee::ptr::PhysConstPtr`. It can benefit other modules which need
+/// Linux kernel's `vmap()` and `vunmap()` functionalities (e.g., HVCI/HEKI, drivers).
```
jaybosamiya-ms: Minor: is written to implement -> exists to service?

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+pub trait VmapProvider<const ALIGN: usize> {
+    /// Data structure for an array of physical page addresses which are virtually contiguous.
+    type PhysPageAddrArray;
+
+    /// Data structure to maintain the mapping information returned by `vmap()`.
+    type PhysPageMapInfo;
+
+    /// Map the given `PhysPageAddrArray` into virtually contiguous addresses with the given
+    /// [`PhysPageMapPermissions`] while returning [`PhysPageMapInfo`]. This function
+    /// expects that it can access and update the page table using `&self`.
```
jaybosamiya-ms: I don't understand what the "This function expects that it can access and update the page table using `&self`." is helping document: isn't that already documented in the type system via the function signature taking `&self`?

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// Multiple LiteBox threads might concurrently call this function with overlapping
+    /// physical pages, so the implementation should safely handle such cases.
+    unsafe fn vmap(
+        &self,
+        pages: Self::PhysPageAddrArray,
+        perms: PhysPageMapPermissions,
+    ) -> Result<Self::PhysPageMapInfo, PhysPointerError>;
+
+    /// Unmap the previously mapped virtually contiguous addresses ([`PhysPageMapInfo`]).
+    /// Use `&self` to access and update the page table.
```
jaybosamiya-ms: similar

litebox/src/platform/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+    unsafe fn vunmap(&self, vmap_info: Self::PhysPageMapInfo) -> Result<(), PhysPointerError>;
+
+    /// Validate that the given physical pages do not belong to LiteBox-owned memory.
+    /// Use `&self` to get the memory layout of the platform (i.e., the physical memory
+    /// range assigned to LiteBox).
+    ///
+    /// This function is a no-op if there is no other world or VM sharing the physical memory.
+    ///
+    /// Returns `Ok(())` if valid. If the pages are not valid, returns `Err(PhysPointerError)`.
+    fn validate(&self, pages: Self::PhysPageAddrArray) -> Result<(), PhysPointerError>;
```
jaybosamiya-ms: I don't think `validate` is a good name for this, perhaps `validate_unowned` or something like that?  Otherwise the direction of validation is unclear.

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+    ///
+    /// The caller should ensure that the virtual addresses in `vmap_info` are not in active
+    /// use by other entities.
+    unsafe fn vunmap(&self, vmap_info: Self::PhysPageMapInfo) -> Result<(), PhysPointerError>;
+
+    /// Validate that the given physical pages do not belong to LiteBox-owned memory.
+    /// Use `&self` to get the memory layout of the platform (i.e., the physical memory
+    /// range assigned to LiteBox).
+    ///
+    /// This function is a no-op if there is no other world or VM sharing the physical memory.
```
jaybosamiya-ms: How does the platform know there is nothing else around?  I am surprised by this no-op guarantee.  My understanding would be that it would keep track of "this is what I own, and everything else is unowned", which means that it wouldn't actually be a no-op, right?  Have I misunderstood what this no-op means?

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// Returns `Ok(())` if valid. If the pages are not valid, returns `Err(PhysPointerError)`.
+    fn validate(&self, pages: Self::PhysPageAddrArray) -> Result<(), PhysPointerError>;
+
+    /// Protect the given physical pages to ensure concurrent read or exclusive write access:
+    /// - Read protection: prevent others from writing to the pages.
+    /// - Read/write protection: prevent others from reading or writing to the pages.
+    /// - No protection: allow others to read and write the pages.
+    ///
+    /// This function can be implemented using EPT/NPT, TZASC, PMP, or some other hardware mechanisms.
+    /// It is a no-op if there is no other world or VM sharing the physical memory.
```
jaybosamiya-ms: This no-op is surprising.  Again, it is making an assertion about the outside world, which is quite surprising.

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+    ///
+    /// Returns `Ok(())` if it successfully protects the pages. If it fails, returns
+    /// `Err(PhysPointerError)`.
+    ///
+    /// # Safety
+    ///
+    /// Since this function is expected to use hypercalls or other privileged hardware features,
+    /// the caller must ensure that it is safe to perform such operations at the time of the call.
+    /// Also, the caller should unprotect the pages when they are no longer needed to be protected.
```
jaybosamiya-ms: This is a difficult-to-maintain safety contract for any caller.  How can the caller ensure that it is safe to perform privileged operations through a generic/parametric platform?

The more accurate safety contract here would be "if set to read, caller must not write to it; if set to none, caller must no longer read/write" etc, right?

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+/// Currently, this is an alias to `crate::mm::linux::NonZeroAddress`. This might change if
+/// we selectively conduct sanity checks based on whether an address is virtual or physical
+/// (e.g., whether a virtual address is canonical, whether a physical address is tagged with
+/// a valid key ID, etc.).
+pub type PhysPageAddr<const ALIGN: usize> = crate::mm::linux::NonZeroAddress<ALIGN>;
+
+/// Data structure to maintain the mapping information returned by `vmap()`.
+///
+/// `base` is the virtual address of the mapped region which is page aligned.
+/// `size` is the size of the mapped region in bytes.
```
jaybosamiya-ms: You can just place these as doc-comments at the fields themselves:
```rs
/// Data structure to maintain ...
pub struct Foo {
  /// The virtual address ....
  pub base: ...,
  /// The size of ...
  pub size: ...,
}
```

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// This module only supports READ and WRITE permissions. Both EXECUTE and SHARED
+    /// permissions are explicitly prohibited.
+    #[non_exhaustive]
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    pub struct PhysPageMapPermissions: u8 {
+        /// Readable
+        const READ = 1 << 0;
+        /// Writable
+        const WRITE = 1 << 1;
+        const _ = !0;
```
jaybosamiya-ms: Why is this one around?

litebox_common_optee/src/vmap.rs
```diff
@@ -0,0 +1,177 @@
...
+    pub size: usize,
+}
+
+bitflags::bitflags! {
+    /// Physical page map permissions which is a restricted version of
+    /// [`litebox::platform::page_mgmt::MemoryRegionPermissions`].
+    ///
+    /// This module only supports READ and WRITE permissions. Both EXECUTE and SHARED
+    /// permissions are explicitly prohibited.
+    #[non_exhaustive]
```
jaybosamiya-ms: Why is it non-exhaustive in that case?  Do we expect other permissions other than read/write to ever show up?

litebox_common_linux/src/vmap.rs
jaybosamiya-ms: Btw, if it helps unblock this PR, please feel free to move it into `litebox_common_optee`.  Since the current usage of it is only to provide `litebox_shim_optee::ptr::Phys*Ptr`, there is no reason to block this otherwise fairly-large PR on making sure the design is general enough to sit inside the main `litebox` crate.

sangho2: The reason we do separate this from PhysicalPtr is because this vmap feature should be implemented by the platform. My understanding is that we no longer want to have `cfg_if` for platforms in shims.

litebox_common_optee/src/vmap.rs
sangho2: No specific reason. we can use concrete ones.

litebox_common_optee/src/vmap.rs
sangho2: Let me remove `&self`. This is more about we hope a corresponding implementation can deal with page tables, but we don't need to mention it.

litebox/src/platform/vmap.rs
sangho2: Let me rename this function.

litebox_common_optee/src/vmap.rs
sangho2: You are right. this comment assumes that there is no other OS/kernel sharing RAM. It is not generic enough.

litebox_common_optee/src/vmap.rs
sangho2: Again, this is more about bare-metal, one OS assumption. Let me clarify this.

litebox_common_optee/src/vmap.rs
sangho2: Technically, we cannot make this fully safe and it is best effort. This is not about the caller, but about others sharing the same physical machine. In some sense, this is out of scope.

litebox_common_optee/src/vmap.rs
sangho2: future proof? I don't expect we'll support execute anyhow so it can be removed.

litebox_common_linux/src/vmap.rs
jaybosamiya-ms: My suggestion was just to move the _trait_ to `litebox_common_optee`, then the platform can implement the trait just fine, no?

sangho2: Ah. I misread it. `litebox_common_optee` would work.

litebox_common_optee/src/lib.rs
```diff
@@ -1111,57 +1108,68 @@ pub enum OpteeMessageCommand {
...
+    pub offs: u64,
     /// Size of the buffer
-    size: u64,
+    pub size: u64,
     /// Shared memory reference or identifier
-    shm_ref: u64,
+    pub shm_ref: u64,
 }
 
 /// FF-A memory reference parameter
```
praveen-pk, line 1159: As I understand it, FF-A is not used today. Until and unless Litebox is used on a hardware platform we don't need this.

Is this here to support FF-A in the future?

litebox_common_optee/src/lib.rs
```diff
@@ -1111,57 +1108,68 @@ pub enum OpteeMessageCommand {
...
+        match cmd {
+            OpteeMessageCommand::OpenSession => Ok(UteeEntryFunc::OpenSession),
+            OpteeMessageCommand::CloseSession => Ok(UteeEntryFunc::CloseSession),
+            OpteeMessageCommand::InvokeCommand => Ok(UteeEntryFunc::InvokeCommand),
+            _ => Err(OpteeSmcReturn::EBadCmd),
+        }
+    }
 }
 
 /// Temporary reference memory parameter
```
praveen-pk: Are `Temporary reference memory parameter` and `Registered memory reference parameter` OPTEE terminology? If not, could you please expand on what these are?

litebox_common_optee/src/lib.rs
```diff
@@ -1236,76 +1292,141 @@ impl OpteeMsgParam {
...
+/// TODO: Since this is LVBS-specific structure to facilitate the translation between VTL call convention,
+/// we might want to move it to the `litebox_platform_lvbs` crate later.
 #[repr(align(4096))]
 #[derive(Clone, Copy)]
 #[repr(C)]
+pub struct OpteeSmcArgsPage {
+    pub args: [usize; Self::NUM_OPTEE_SMC_ARGS],
+}
+impl OpteeSmcArgsPage {
+    const NUM_OPTEE_SMC_ARGS: usize = 9;
```
praveen-pk, line 1404: Seems like SMCCC uses r0-r7 for the first eight args and r8 as an extra parameter and to return.


Would be nice to document a similar convention here as well.

litebox_common_optee/src/lib.rs
```diff
@@ -1236,76 +1292,141 @@ impl OpteeMsgParam {
...
+        OpteeSmcFunction::try_from(self.args[0] & OpteeSmcFunction::MASK)
+            .map_err(|_| OpteeSmcReturn::EBadCmd)
     }
 
     /// Get the physical address of `OpteeMsgArg`. The secure world is expected to map and copy
     /// this structure.
-    pub fn optee_msg_arg_phys_addr(&self) -> Result<usize, Errno> {
+    pub fn optee_msg_arg_phys_addr(&self) -> Result<u64, OpteeSmcReturn> {
         // To avoid potential sign extension and overflow issues, OP-TEE stores the low and
         // high 32 bits of a 64-bit address in `args[2]` and `args[1]`, respectively.
```
praveen-pk, line 1434: If this is the convention, why have the check below?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -0,0 +1,335 @@
...
+
+/// The result of handling an OP-TEE SMC call along with an extracted OP-TEE message argument to handle.
+pub struct OpteeSmcHandled<'a> {
+    pub result: OpteeSmcResult<'a>,
+    pub msg_to_handle: Option<OpteeMsgArg>,
+}
+
+/// This function handles `OpteeSmcArgs` passed from the normal world (VTL0) via an OP-TEE SMC call.
+/// It returns an `OpteeSmcResult` representing the result of the SMC call and
+/// an optional `OpteeMsgArg` if the SMC call involves with an OP-TEE messagewhich should be handled by
```
praveen-pk: nit: fix `messagewhich`

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -0,0 +1,335 @@
...
+/// `handle_optee_msg_arg` or `handle_ta_request`.
+///
+/// # Panics
+///
+/// Panics if the normal world physical address in `smc` cannot be converted to `usize`.
```
praveen-pk: Is this even possible with a properly functioning OPTEE client? 

I am wondering if this case should be handled.

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,570 @@
...
+// Licensed under the MIT license.
+
+//! Physical Pointer Abstraction with On-demand Mapping
+//!
+//! This module adds supports for accessing physical addresses (e.g., VTL0 or
+//! normal-world physical memory) from LiteBox with on-demand mapping.
+//! In the context of LVBS and OP-TEE, accessing physical memory is necessary
+//! because VTL0 and VTL1 as well as normal world and secure world do not share
+//! the same virtual address space, but they still have to share data through memory.
+//! VTL1 and secure world receive physical addresses from VTL0 and normal world,
```
praveen-pk, line 11: nit: Isn't VTL1 the same as secure world? This sentence seems to imply they are 2 different things.

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,570 @@
...
+//! BIOS/UEFI such that it does not have complete knowledge of the physical memory
+//! layout. In the case of LVBS, LiteBox obtains the physical memory information
+//! from VTL0 including the total physical memory size and the memory range assigned
+//! to VTL1/LiteBox. Thus, this module can at least confirm a given physical address
+//! does not belong to VTL1's physical memory.
+//!
+//! This module should allow byte-level access while transparently handling page
+//! mapping and data access across page boundaries. This could become complicated
+//! when we consider multiple page sizes (e.g., 4 KiB, 2 MiB, 1 GiB). Also,
+//! unaligned access is matter to be considered.
```
praveen-pk: nit:  unaligned access is `a` matter to be considered.

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,570 @@
...
+//!
+//! In addition, often times, this physical pointer abstraction is involved with
+//! a list of physical addresses (i.e., scatter-gather list). For example, in
+//! the worse case, a two-byte value can span across two non-contiguous physical
+//! pages (the last byte of the first page and the first byte of the second page).
+//! Thus, to enhance the performance, we may need to consider mapping multiple pages
+//! at once, copy data from/to them, and unmap them later.
+//!
+//! When this module needs to access data across physical page boundaries, it assumes
+//! that those physical pages are virtually contiguous in VTL0 or normal-world address
```
praveen-pk, line 55: >those physical pages are virtually contiguous in VTL0

Why does this matter? VTL1 should be able to create a vmap using the physical pages passed by OPTEE. What is the dependency on VTL0's virtual addresses in VTL1?

litebox_shim_optee/src/ptr.rs
```diff
@@ -0,0 +1,570 @@
...
+}
+
+#[inline]
+fn align_up(len: usize, align: usize) -> usize {
+    len.next_multiple_of(align)
+}
+
+/// Represent a physical pointer to an object with on-demand mapping.
+/// - `pages`: An array of page-aligned physical addresses. Physical addresses in this array should be
+///   virtually contiguous.
```
praveen-pk, line 81: Are the pages virtually contiguous in VTL1?

litebox_common_optee/src/lib.rs
sangho2: Just for future proofing. We don't particularly need this for now.

litebox_common_optee/src/lib.rs
sangho2: They are from OP-TEE: TMEM and RMEM. I'll add a comment to refer `optee_os/core/include/optee_msg.h`.

litebox_common_optee/src/lib.rs
sangho2: My understanding is that using r8 to return a value is about Arm's original SMCCC to interact with the EL3 security monitor (normal world interacts with secure world through this security monitor). According to https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/include/sm/optee_smc.h, OP-TEE SMC use `a0` to return `OPTEE_SMC_RETURN_x`. Anyhow, we can further clarify this later.

litebox_common_optee/src/lib.rs
sangho2: because we don't trust the normal world. Malformed input implies that the normal world is misbehaving.

litebox_shim_optee/src/msg_handler.rs
sangho2: This `# Panics` is mainly for making `cargo clippy` happy. It doesn't happen in practice (unless we target 32-bit machines?)

litebox_shim_optee/src/ptr.rs
sangho2: Technically the same. My goal is making this module is general enough to support both LVBS and OP-TEE use cases somewhat independently. LVBS doesn't need to aware of OP-TEE and vice versa.

litebox_shim_optee/src/ptr.rs
sangho2: This is just an assumption. We can still vmap them (without correct order, of course).

litebox_shim_optee/src/ptr.rs
sangho2: There is a list of physical pages and we contiguously map them to virtual address space while sequentially reading the list.

litebox_common_optee/src/lib.rs
wdcui: Please create an issue to track this before we resolve this comment.

sangho2: Done

litebox_common_optee/src/lib.rs
```diff
@@ -1338,64 +1475,102 @@ impl OpteeSmcFunction {
...
+            } => {
+                let mut smc = OpteeSmcArgs::default();
+                smc.args[0] = status as usize;
+                smc.args[1] = shm_upper32;
+                smc.args[2] = shm_lower32;
+                smc
+            }
+            OpteeSmcResult::CallWithArg { .. } => {
+                panic!(
+                    "OpteeSmcResult::CallWithArg cannot be converted to OpteeSmcArgs directly. Handle the incorporate OpteeMsgArg."
```
sangho2: typo: incorporate -> incorporated

litebox_common_optee/src/lib.rs
```diff
@@ -1239,76 +1308,143 @@ impl OpteeMsgParam {
 #[repr(C)]
 pub struct OpteeMsgArg {
```
wdcui: I found this struct's name `OpteeMsgArg` confusing. The name makes me think it's for a single argument. However, it looks like that it's all the arguments in a single message. What makes it worse is that we use `OpteeSmcArgs` to represent multiple arguments. Would it make sense to change it to `OpteeMsgArgs`?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -2187,6 +2188,13 @@ impl litebox::platform::CrngProvider for LinuxUserland {
...
+impl<const ALIGN: usize> VmapProvider<ALIGN> for LinuxUserland {}
```
wdcui: What's our decision on the `VmapProvdier`? Did we agree to add it as a new platform provider?

wdcui: I thought `vmap` is only used by `ptr.rs` in `litebox_shim_optee`, so we can move this provider there? I don't remember the details. @jaybosamiya-ms @CvvT Thoughts?

litebox_common_optee/src/lib.rs
sangho2: I agree that many struct names including these are confusing. I wanted to maintain the original structure names if possibe. There are `struct optee_msg_arg` and `struct thread_smc_args` in Linux kernel/OP-TEE OS. Anyway, I think adding `s` wouldn't confuse ones familiar with OP-TEE anyhow.

litebox_platform_linux_userland/src/lib.rs
sangho2: We cannot have `vmap` in `litebox_shim_*` because platforms implement this function and they should not depend on shims.

Conclusion is having this in `litebox_common_*`. I tried `litebox_common_optee` first, but it has some issues with clippy target all, so I moved it to `litebox_common_linux`. `vmap` is a Linux kernel function, so I think using `litebox_common_linux` makes a sense.

wdcui: I see. After we move it to `litebox_common_linux`, should we still call it a `Provider`?

sangho2: `VmapManager`? I think using `Provider` only for the traits in the `litebox` crate might be better.

wdcui: Sounds good.

litebox_platform_linux_userland/src/lib.rs
sangho2: `VmapManager` now has default implementation.

litebox_common_optee/src/vmap.rs
sangho2: Now we use concrete types.

litebox_common_optee/src/vmap.rs
sangho2: I've revised comments to avoid misunderstanding.

litebox_common_optee/src/vmap.rs
sangho2: I've revised comments to avoid misunderstanding.

litebox_common_optee/src/vmap.rs
sangho2: I've re-placed the comments.

litebox_common_optee/src/vmap.rs
sangho2: I removed `const _ = !0;`

litebox_common_optee/src/vmap.rs
sangho2: I removed `#[non_exhaustive]`

litebox_common_optee/src/lib.rs
sangho2: Let's maintain FF-A stuffs for now.

litebox_common_optee/src/lib.rs
sangho2: I've revised the comments.

litebox_common_optee/src/lib.rs
sangho2: I've created an issue for this.

litebox_shim_optee/src/msg_handler.rs
sangho2: Fixed.

litebox_shim_optee/src/msg_handler.rs
sangho2: Now we use `truncate` to remove this panic doc.

litebox_shim_optee/src/ptr.rs
sangho2: Fixed.

litebox_common_optee/src/lib.rs
sangho2: Fixed

litebox_common_optee/src/vmap.rs
sangho2: Removed `&self`.

litebox_common_optee/src/vmap.rs
sangho2: Removed `&self`

litebox/src/platform/vmap.rs
sangho2: Now we are using `validate_unowned`

litebox_common_optee/src/vmap.rs
sangho2: Revised the comment.

PR #556, opened 2025-12-13
Add `upcall` for delegation

sangho2:
This PR adds the `upcall` trait to let platforms delegate handling of unknown
messages/requests (e.g., OP-TEE messages from the normal-world/VTL0
kernel) to other layers of LiteBox (i.e., runner or shim).

### DIFF
--- a/litebox/src/upcall.rs
+++ b/litebox/src/upcall.rs
@@ -1,28 +1,53 @@
-//! Types and traits to provide upcalls for platforms.
-//! A platform may get some messages or requests from the host or
-//! remote parties that it does not know how to handle (e.g.,
-//! an OP-TEE SMC call from the normal-world kernel). Rather than
-//! making the platform handle all these messages or requests itself,
-//! we implement upcalls to let the platform delegate handling of
-//! unknown messages or requests to other layers of LiteBox (i.e.,
-//! runner or shim).
+//! Upcall for platforms
+//!
+//! This module defines types and trait to provide upcalls for platforms.
+//! A platform may receive some messages or requests from the host, devices,
+//! or remote parties that it does not know how to handle. Rather than
+//! making the platform be aware of all these messages or requests itself,
+//! we implement upcalls to let platforms delegate handling of unknown
+//! messages or requests to other layers of LiteBox (i.e., runner or shim).
+//! Examples of such messages or requests include HVCI/Heki requests from
+//! VTL0 and OP-TEE SMC calls from the normal world.
+//!
+//! # Security considerations
+//!
+//! Unlike other upcalls like hyperupcalls from the hypervisor to a guest VM,
+//! this upcall is handled within LiteBox's TCB (i.e., by either runner or shim).
+//! Therefore, the security considerations for this upcall are similar to function
+//! calls within LiteBox. However, care must be taken to ensure that the upcall's
+//! parameters and return values are properly validated and sanitized to prevent
+//! potential security vulnerabilities. This is because the parameters might be
+//! provided by untrusted sources and return values might contain sensitive
+//! information. We assume that the upcall providers must implement necessary
+//! security checks and validations. The platform would simply pass the parameters
+//! and return values between untrusted sources and upcall providers (because it
+//! does not have semantics to validate them). We can specify a function for early
+//! validation at the platform side if needed but its advantages are not clear at
+//! this moment (since there is no costly context switch within LiteBox).
+
 use thiserror::Error;
 
+/// An interface for upcalls from the platform to other layers of LiteBox.
 pub trait Upcall {
-    /// The upcall parameter type
+    /// The upcall parameter type to be passed from the platform.
     type Parameter;
 
-    /// The upcall return type
+    /// The upcall return type to be returned to the platform.
     type Return;
 
-    /// Initialize the upcall handler
+    /// Initialize the upcall handler. Must be called by the platform exactly once.
+    /// Per-thread initialization is possible but all threads must share the same
+    /// upcall handler.
     fn init(
         &self,
     ) -> alloc::boxed::Box<
         dyn crate::upcall::Upcall<Parameter = Self::Parameter, Return = Self::Return>,
     >;
 
-    /// Execute the upcall with the given parameter
+    /// Execute the upcall with the given parameter. Since we do not expect that the
+    /// platform validates the parameters, the implementation of `execute` must validate
+    /// parameters to avoid potential security vulnerabilities. Also, it must sanitize
+    /// the return values before returning them to the platform.
     fn execute(&self, ctx: &mut Self::Parameter) -> Result<Self::Return, UpcallError>;
 }
 
@@ -34,4 +59,6 @@ pub enum UpcallError {
     Failure,
     #[error("Upcall needs to be retried")]
     Retry,
+    #[error("Upcall parameter is invalid")]
+    InvalidParameter,
 }


litebox/src/upcall.rs
```diff
@@ -0,0 +1,29 @@
...
+    type Context;
+
+    /// Initialize the upcall handler
+    fn init(&self) -> alloc::boxed::Box<dyn crate::upcall::Upcall<Context = Self::Context>>;
+
+    /// Execute the upcall with the given context
+    fn execute(&self, ctx: &mut Self::Context) -> Result<Self::Context, UpcallError>;
+}
+
+/// The operation to perform after returning from a shim handler
```
wdcui: The comment does not match the code?

sangho2: Oops. A copy and paste error. I'll fix it.

litebox_platform_lvbs/src/mshv/vsm_optee_smc.rs
```diff
@@ -18,7 +18,12 @@ pub(crate) fn optee_smc_dispatch(optee_smc_args_pfn: u64) -> i64 {
         with_per_cpu_variables_mut(|per_cpu_variables| {
             per_cpu_variables.save_extended_states(HV_VTL_SECURE);
         });
-        // TODO: Implement OP-TEE SMC for TA command invocation here.
+
+        // Placeholder for now
+        let upcall = crate::UPCALL.get().expect("OP-TEE upcall not registered");
+        let mut ctx = litebox_common_linux::PtRegs::default();
+        let _ = upcall.execute(&mut ctx);
```
praveen-pk, line 25: Did you intend to ignore the return the value here?

Seems like this should be properly handled/propagated.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -99,6 +102,30 @@ pub fn run(platform: Option<&'static Platform>) -> ! {
     vtl_switch_loop_entry(platform)
 }
 
+pub struct OpteeMsgHandler;
+impl litebox::upcall::Upcall for OpteeMsgHandler {
+    type Parameter = litebox_common_linux::PtRegs;
+    type Return = litebox_common_linux::PtRegs;
```
praveen-pk, line 108: Curious to know why you need the registers here?

Do you intend to run a thread during upcall.execute?

litebox_platform_lvbs/src/mshv/vsm_optee_smc.rs
sangho2: As wrote in the comment, this is just a placeholder. we should handle the return value.

litebox_runner_lvbs/src/lib.rs
sangho2: This is just for defining parameter and return types. We don't specify any particular data types here yet. The reason I use `PtRegs` here is because it is the most common data type used in many places in the LiteBox code base.

wdcui:
@jaybosamiya-ms It would be good to have you take a look at this PR since it changes litebox.

litebox/src/upcall.rs
```diff
@@ -0,0 +1,64 @@
+//! Upcall for platforms
+//!
+//! This module defines types and trait to provide upcalls for platforms.
+//! A platform may receive some messages or requests from the host, devices,
+//! or remote parties that it does not know how to handle. Rather than
+//! making the platform be aware of all these messages or requests itself,
+//! we implement upcalls to let platforms delegate handling of unknown
+//! messages or requests to other layers of LiteBox (i.e., runner or shim).
+//! Examples of such messages or requests include HVCI/Heki requests from
+//! VTL0 and OP-TEE SMC calls from the normal world.
```
jaybosamiya-ms, line 10: Would signals from the outside world be a similar situation (i.e., should they be handled through this mechanism)?

litebox/src/upcall.rs
```diff
@@ -0,0 +1,64 @@
...
+//!
+//! # Security considerations
+//!
+//! Unlike other upcalls like hyperupcalls from the hypervisor to a guest VM,
+//! this upcall is handled within LiteBox's TCB (i.e., by either runner or shim).
+//! Therefore, the security considerations for this upcall are similar to function
+//! calls within LiteBox. However, care must be taken to ensure that the upcall's
+//! parameters and return values are properly validated and sanitized to prevent
+//! potential security vulnerabilities. This is because the parameters might be
```
jaybosamiya-ms, line 19: "care must be taken" -> by whom?

litebox/src/upcall.rs
```diff
@@ -0,0 +1,64 @@
...
+//! provided by untrusted sources and return values might contain sensitive
+//! information. We assume that the upcall providers must implement necessary
+//! security checks and validations. The platform would simply pass the parameters
+//! and return values between untrusted sources and upcall providers (because it
+//! does not have semantics to validate them). We can specify a function for early
+//! validation at the platform side if needed but its advantages are not clear at
+//! this moment (since there is no costly context switch within LiteBox).
```
jaybosamiya-ms, line 26: Nit: this sentence is surprising to see as part of the doc string.  Might be better just as a comment, rather than being in the doc string.

litebox/src/upcall.rs
```diff
@@ -0,0 +1,64 @@
...
+pub trait Upcall {
+    /// The upcall parameter type to be passed from the platform.
+    type Parameter;
+
+    /// The upcall return type to be returned to the platform.
+    type Return;
+
+    /// Initialize the upcall handler. Must be called by the platform exactly once.
+    /// Per-thread initialization is possible but all threads must share the same
+    /// upcall handler.
```
jaybosamiya-ms, line 40: This comment is confusing.  What happens if the platform does not invoke this?  (I might have missed something but I didn't see the platform invoke this here).  I also don't understand the "per-thread initialization is possible but ..." comment here.  Should each thread invoke this?  Or should different threads within the platform coordinate to make sure they have not invoked this more than once?

A slightly easier-to-use interface is likely something like "get singleton", so that the platform doesn't need to store/manage this, and the runner side can choose to make it more/less performant as it needs.  Alternatively, one can just eliminate initialization and set up the `execute` to initialize on first use?

litebox/src/upcall.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn init(
+        &self,
+    ) -> alloc::boxed::Box<
+        dyn crate::upcall::Upcall<Parameter = Self::Parameter, Return = Self::Return>,
+    >;
```
jaybosamiya-ms, line 45: You need `self` to get _essentially_ a `Box<dyn itself>`?  This is a confusing signature.  Where do you get teh original `Upcall` to even invoke this initialization?  Why is it returning a boxed version of itself?

litebox/src/upcall.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// Execute the upcall with the given parameter. Since we do not expect that the
+    /// platform validates the parameters, the implementation of `execute` must validate
+    /// parameters to avoid potential security vulnerabilities. Also, it must sanitize
+    /// the return values before returning them to the platform.
```
jaybosamiya-ms, line 50: Minor: might be good to point at `See [security considerations](crate::upcall)` or such

litebox/src/upcall.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn execute(&self, ctx: &mut Self::Parameter) -> Result<Self::Return, UpcallError>;
```
jaybosamiya-ms, line 51: I'm mildly surprised by seeing the `&mut` here for the parameter.  Due to lifetime elision and such, `Self::Return`'s lifetime cannot depend on that mutable reference anyways.  A slightly cleaner design might be to just pass the parameter as a (consumed/owned) value?  Alternatively, the `Return` should have a generic lifetime that gets bound to the lifetime of the parameter (or even more generally, it would get bound to a new lifetime that is itself bounded by the lifetimes of `&self` and `ctx`).

litebox/src/upcall.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/// The error type for upcalls
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum UpcallError {
+    #[error("Upcall failed")]
+    Failure,
```
jaybosamiya-ms, line 59: Failed due to?  It becomes impossible in the current design for the upcall to give more useful information to the place below.  Would be good to make this generic in some way (can do it generically, or can do a `Failure(Box<dyn Error>)` might be enough).

litebox/src/upcall.rs
```diff
@@ -0,0 +1,64 @@
...
+    #[error("Upcall needs to be retried")]
+    Retry,
```
jaybosamiya-ms, line 61: If the upcall needs to be retried, it needs to mention _when_ it needs to be retried, right?  Otherwise, if "retry immediately" is allowed, might as well do the retrying internally.  The failure case is that "it cannot be handled right now, please try later", but yeah, something else feels necessary here in the output so that a platform can know when to retry.

litebox_platform_lvbs/src/mshv/vsm_optee_smc.rs
jaybosamiya-ms: Minor nit: keeping a `TODO:`/`XXX:`/... or such signifier makes it easier to grep in the future quickly

litebox_platform_lvbs/src/lib.rs
```diff
@@ -389,6 +389,19 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
+
+    /// Register the upcall.
+    pub fn register_upcall(
+        upcall: &'static (
+                     dyn litebox::upcall::Upcall<
+            Parameter = litebox_common_linux::PtRegs,
+            Return = litebox_common_linux::PtRegs,
+        > + Send
+                         + Sync
+                 ),
```
jaybosamiya-ms, line 401: oof, this looks like rustfmt did a _horrifyingly_ bad job here.  Consider putting the type into a type alias (`type Upcall = ...`) to make this be nicer maintained?

sangho2:
> Minor comments throughout, but I have a more general comment about the design/idea itself: this is roughly equivalent to the functionality that already exists in `litebox::shim::...` right? This feels like adding a parallel way of handling _roughly_ similar things, which seems suboptimal. Until we can clearly articulate when that should be used and when this should be used, I think this PR should not be merged, which is why I am marking this as "Request changes".
> 
> I am happy to discuss this IRL more if it would help tweak the design of the existing `litebox::shim::...` to support the use case (or maybe it already can support the use case?). Currently, this PR is only introducing the new functionality without the use case, which makes it a little hard to understand why the required use case cannot be handled by existing `litebox::shim::EnterShim` and related bits.
> 
> CC: @jstarks I expect you'll have some useful insights on this PR

We can definitely discuss this later. This design is based on some discussion between John, Weidong, and me. Long story short, we need a runner or something below the shim to support this stuff. For example, two of this upcall's main usages are loading ELF and invoking `run_thread` that shim cannot do, like https://github.com/microsoft/litebox/pull/564/changes#diff-3eeb171b29f4643a455abd0bf670819afc33087f009eb4f6300fff0364f19d6bR127

Anyhow, what we need is there is a function in the runner which should be invoked by the platform through a simple interface.

litebox/src/upcall.rs
sangho2: Unlike the Linux shim case that all signals will be handled by the shim, certain exceptions/interrupts will be handled by the LVBS platform whereas some other ones are handled by the OP-TEE shim.

litebox/src/upcall.rs
sangho2: the upcall handler. will clarify this.

litebox/src/upcall.rs
sangho2: Basically, this is a callback registration. I agree that the comment is confusing. I'll fix this.

litebox/src/upcall.rs
sangho2: Yes. Technically, we don't need to struggle with lifetime. let me use values.

litebox/src/upcall.rs
sangho2: Let me do that.

litebox/src/upcall.rs
sangho2: Yes. Technically, there is no easy way to know when the platform can try it again. Let me think.

litebox_platform_lvbs/src/mshv/vsm_optee_smc.rs
sangho2: Didn't put TODO here because the actual implementation already exists in another PR.
https://github.com/microsoft/litebox/pull/564/changes#diff-3eeb171b29f4643a455abd0bf670819afc33087f009eb4f6300fff0364f19d6bR127

litebox/src/upcall.rs
jaybosamiya-ms: That's fine right?  The platform can just decide not to forward along things to the shim.  There is nothing in the design that says that the platform _must_ forward everything over, right?  Maybe I am misunderstanding why the signals world and the upcall world are drastically different.  Might be easier to chat in-person?

sangho2: Yes, I think the current explanation is misleading. It is more about there are something the platform doesn't know how to handle and delegates them to the others if applicable.

litebox/src/upcall.rs
sangho2: the runner has a function and registers it to the platform through this (`Platform::register_upcall`).

sangho2:
Discussion result: Instead of defining a new interface, we will add one function to `EnterShim` to support this feature.

sangho2:
Turns that using `EnterShim` is not straightforward or may be not possible. The reason is that our current implementation passes `shim` to the platform via `run_thread` (we no longer explicitly register a shim to the platform). Perhaps we need to use `litebox_common_optee` for now (sounds like `litebox_common_*` are becoming a collection of second-class providers).

sangho2:
We decided to move `vtl_switch_loop` and its sub functions to `litebox_runner_lvbs`. To this end, we no longer need `upcall`.

PR #558, opened 2025-12-16
Add tracing support

### LINKED ISSUE #171: Add tracing support

jaybosamiya-ms:
Resolves #171.

This PR enables `tracing` for LiteBox, and set up a subscriber for the Linux userland.  Specifically, this adds `LITEBOX_LOG` environment variable to the `litebox_runner_linux_userland`, which can be used like the following:
- `LITEBOX_LOG=debug`: Show debug and higher-level logs for all crates and modules
- `LITEBOX_LOG=litebox::fs=debug`: Specifically `litebox::fs` code at debug and higher-level
- `LITEBOX_LOG=litebox=debug,litebox::fs=trace`: Multiple filters at different levels

This was previously not quite possible because we could not have real Rust thread locals (which `tracing` uses internally), but now that that works, we can enable `tracing`.

Currently, I have not added much in the form of tracing logs/messages much in the code and just added it for `LiteBox` object initialization.  Future PRs will enable it in more places so that we can more easily see a whole bunch more of what is happening easily.

Here's a test output:
```console
$ LITEBOX_LOG=litebox=trace cargo run --bin litebox_runner_linux_userland -- --unstable --interception-backend rewriter target/dev_bench/hello_static_rewritten
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
warning: the following packages contain code that will be rejected by a future version of Rust: num-bigint-dig v0.8.4
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running `target/debug/litebox_runner_linux_userland --unstable --interception-backend rewriter target/dev_bench/hello_static_rewritten`
   0.001065242s DEBUG litebox::litebox: LiteBox instance initialized
argv[0] = target/dev_bench/hello_static_rewritten
envp[0] = LD_AUDIT=/lib/litebox_rtld_audit.so
Elapsed time: 0.088816 seconds
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -796,6 +796,7 @@ dependencies = [
  "spin 0.9.8",
  "tar-no-std",
  "thiserror",
+ "tracing",
  "windows-sys 0.60.2",
 ]
 
@@ -946,6 +947,8 @@ dependencies = [
  "litebox_syscall_rewriter",
  "memmap2",
  "sha2",
+ "tracing",
+ "tracing-subscriber",
  "walkdir",
 ]
 
--- a/litebox/Cargo.toml
+++ b/litebox/Cargo.toml
@@ -18,6 +18,7 @@ ringbuf = { version = "0.4.8", default-features = false, features = ["alloc"] }
 buddy_system_allocator = { version = "0.11.0", default-features = false, features = ["use_spin"] }
 # Depend on (currently unreleased) slabmalloc `main`, which contains some fixes on top of `0.11.0`
 slabmalloc = { git = "https://github.com/gz/rust-slabmalloc.git", rev = "19480b2e82704210abafe575fb9699184c1be110" }
+tracing = { version = "0.1.43", default-features = false }
 
 [target.'cfg(windows)'.dependencies]
 windows-sys = { version = "0.60.2", features = [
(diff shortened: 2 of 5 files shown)


jstarks:
I have some concerns with `tracing` for a project at this level of the stack--in particular, the generated code for a trace point is quite large, and performance of the subscriber model is fundamentally bad. I have made some improvements in the latest release (see tokio-rs/tracing#3398), but I am not optimistic about further improvements without making breaking changes. And the `tracing` maintainers are severely overworked, so I don't think we can count on that even if we do the work.

So, I think for embedded-ish kernel projects, we'll likely need to turn `tracing` off completely at compile time to avoid code bloat, defeating the purpose of the traces. I wonder if it might be better to look at `log` or `defmt` or even our own set of litebox logging macros to avoid this limitation.

An advantage of litebox-specific logging macros is that we could back them with different implementations depending on compile-time configuration. So, we could back them by `tracing` for the userland platforms for ease of use where binary size doesn't matter so much, but use `defmt` or something similar for kernel platforms.

What do you think?

jstarks:
(Note that litebox-specific code becomes more complicated if you want to support spans and not just events. But in the context of LiteBox, I am skeptical that spans have value commensurate with their runtime cost.)

jaybosamiya-ms:
Briefly summarizing a longer discussion on Teams b/w @jstarks and myself: we are now considering a facade over `log` rather than depend on `tracing` directly in our library crates.

jaybosamiya-ms:
@jstarks I've now made a facade crate PR at https://github.com/microsoft/litebox/pull/665

jaybosamiya-ms:
Superseded by #773, closing

PR #562, opened 2025-12-16
Improve network throughput

CvvT:
1. Instead of using a fixed timeout value (50ms), call `poll_at` from smoltcp which would determine when the next network operation should occur.
2. Instead of only acquiring lock once at the beginning of `perform_platform_interaction`, now each network operation would acquire and release the lock so that concurrent syscalls (e.g., read) won't get blocked for too long.
3. `perform_platform_interaction` now processes multiple ingress packets before processing egress packets to avoid unnecessary egress processing (e.g. fewer ACK packets).

With the above change, the throughput can be improved from 22.2 Mbits/sec to 141 Mbits/sec. It's still 50x slower than native. ~~One big factor is the receive window (i.e., socket buffer size). smoltcp does not support dynamically increasing the socket buffer size and the default value we set is 64KB. When testing on Linux, the window size can be as large as 6MB.~~

 4. In addition to the network background worker that processes network in a loop, before and after each network related syscalls like `send` and `receive`, we perform some network operations in automatic mode (it's disabled in manual mode when we have a separate network worker). I found that if we do both, it helps increase the throughput to 272 Mbits/sec. My hypothesis is that those are the points where we have packets to process immediately.
 5. Double the socket buffer size could further increase the throughput to 360 Mbits/sec, but I'm not sure if it is the right move.

### DIFF
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -531,14 +531,6 @@ pub enum PollDirection {
     /// Both directions
     Both,
 }
-impl PollDirection {
-    fn ingress(self) -> bool {
-        matches!(self, PollDirection::Ingress | PollDirection::Both)
-    }
-    fn egress(self) -> bool {
-        matches!(self, PollDirection::Egress | PollDirection::Both)
-    }
-}
 
 /// Advice on when to invoke [`Network::perform_platform_interaction`] again.
 ///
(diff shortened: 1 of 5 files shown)


litebox/src/net/mod.rs
```diff
@@ -589,12 +593,32 @@ where
...
             "Requires manual-mode interactions"
         );
-        self.internal_perform_platform_interaction(PollDirection::Both)
+        self.internal_perform_platform_interaction(direction)
+    }
+
+    /// Return a _soft timeout_ (duration to wait) before calling [`Self::perform_platform_interaction`] again.
+    ///
+    /// Returns `None` if there is no pending timeout (i.e., no scheduled work requiring network operations).
+    pub fn poll_at(&mut self) -> Option<core::time::Duration> {
```
jaybosamiya-ms: Quick comment: rather than introducing `poll_at` maybe we should consider if updating `PlatformInteractionReinvocationAdvice` makes sense.  I have not thought through the implications of this yet though, but yeah, rather than introducing it at the shim as `NetworkInteractionAdvice`, maybe `PlatformInteractionReinvocationAdvice::WaitOnDeviceOrSocketInteraction` needs to become `WaitOnDeviceOrSocketInteractionOrTimeout(Option<Duration>)`? (the exact naming is not great, but I hope what I'm suggesting is clear enough?)

CvvT: I'm still experimenting with different setups. I have some hardcoded heuristics in linux shim: processing 128 incoming packets first then 128 times on egress packets and lastly calling `poll_at` to get the wait time. If I find this strategy is universal, I can move it to LiteBox without exposing `poll_at`.

https://github.com/microsoft/litebox/blob/1b7aee9d008bc67363131c7aa0ece97f4298052a/litebox_shim_linux/src/lib.rs#L279-L300

CvvT:
Replaced by #585

PR #563, opened 2025-12-16
Optimize file system initialization

jaybosamiya-ms:
This PR optimizes our initialization sequence for `in_mem` file system, which ends up improving our `node` runtime by cutting its initialization time by more than half.

Essentially, a large amount of our initialization time was spent in the initial `write` calls that were feeding in the data.  So now there is a new `initialize_primarily_read_heavy_file` function aimed purely for initializing files that are meant to be read-heavy (such as program data).  It essentially postpones the cost of the memcpy to the time that an actual `write` occurs on the file (which we expect to be never for program data).

The other change is that for the runner, we load in the program data via an `mmap` call rather than via a file read (similar to what we were doing for tar files).  This helps reduce that particular write too.

Along the way, I've also fixed a minor bug in `litebox::fs::in_mem::FileSystem::truncate` when a longer size was passed (previously it would just ignore it).

The benchmark summary from current `main` (summarized across 100 invocations):
| Benchmark                  | main (ms) | profitable-mola (ms) | Diff (ms) |
|:---------------------------|----------:|---------------------:|----------:|
| rewriter_hello_static      |        14 |                   14 |         0 |
| rewriter_node              |      1055 |                 1056 |         1 |
| run_rewritten_hello_static |        45 |                   44 |        -1 |
| run_rewritten_node         |       242 |                  109 |      -133 |

(i.e., speeds up node invocation by ~55%).

### DIFF
--- a/litebox_runner_linux_userland/src/lib.rs
+++ b/litebox_runner_linux_userland/src/lib.rs
@@ -189,7 +189,7 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
                         mode,
                     )
                     .unwrap();
-                fs.initialize_primarily_read_heavy_file(&fd, prog_data.into());
+                fs.initialize_primarily_read_heavy_file(&fd, prog_data);
                 fs.close(&fd).unwrap();
             };
         let last = ancestor_modes_and_users.last().unwrap();


jaybosamiya-ms:
Good question, the actual loading of the executable is later in the execution pipeline, and actually is a _different_ bottleneck (not handled in this PR---this PR is handling the main bottleneck _before_ it, after this PR is merged, that one becomes the main bottleneck).  Since the way we implement `mmap` on files (_inside_ LiteBox) is currently essentially through a memcpy, the actual executable file loading is fully independent of whether we make this particular area writable or not, but yes, we will have to keep that in mind when optimizing that part (to make sure we are not accidentally leading to a situation where we write back to disk---the approach will have to be via a copy-on-write aliasing or similar).

For this particular PR, the reason I use a `Cow` is to make sure that we are "postponing" the cost of writability to the indefinite future (which will never happen in reality, since nothing is expected to rewrite, for example, the `node` binary on the disk).

PR #564, opened 2025-12-17
Add handlers for OP-TEE TA request and execution result

sangho2:
This PR adds handlers for OP-TEE TA request and its execution result. It also has a runner-side
controller that leverages all OP-TEE SMC/messages handlers to deal with OP-TEE calls from
the normal world. Currently, this controller is no-op because it should be invoked via
an upcall from the platform and it relies on run_thread, both are in different PRs.

PR #566, opened 2025-12-18
Add run_thread to LVBS platform and revise its syscall handling

sangho2:
This PR adds `run_thread` support for LVBS platform and revise its syscall handling routine based on
Linux userland platform's syscall handling. It has suppressed new page table creation for now
because it needs to be done by the runner via an upcall (userspace support still exists, but doesn't
create a separate one for each TA instance for now).

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -825,15 +825,18 @@ macro_rules! RESTORE_CALLEE_SAVED_REGISTERS_ASM {
     };
 }
 
-/// Save user context to the memory area pointed by the current stack pointer (`rsp`).
+/// Save user context right after `syscall`-driven mode transition to the memory area
+/// pointed by the current stack pointer (`rsp`).
 ///
-/// `rsp` can point to a real stack or the *top address* of a memory area which
+/// `rsp` can point to the current CPU stack or the *top address* of a memory area which
 /// has enough space for storing the `PtRegs` structure using the `push` instructions
 /// (i.e., from high addresses down to low ones).
 ///
-/// Prerequisite: `r11` contains user `rsp`.
+/// Prerequisite:
+/// - Store user `rsp` in `r11` before calling this macro.
+/// - Store the userspace return address in `rcx` (`syscall` does this automatically).
 #[cfg(target_arch = "x86_64")]
-macro_rules! SAVE_USER_CONTEXT_ASM {
+macro_rules! SAVE_SYSCALL_USER_CONTEXT_ASM {
     () => {
         "
         push 0x2b       // pt_regs->ss = __USER_DS
@@ -866,7 +869,7 @@ macro_rules! SAVE_USER_CONTEXT_ASM {
 /// This macro uses the `pop` instructions (i.e., from low addresses up to high ones) such that
 /// it requires the start address of the memory area (not the top one).
 ///
-/// Prerequisite: The memory area has `PtRegs` structure saved by `SAVE_USER_CONTEXT_ASM`.
+/// Prerequisite: The memory area has `PtRegs` structure containing user context.
 #[cfg(target_arch = "x86_64")]
 macro_rules! RESTORE_USER_CONTEXT_ASM {
     () => {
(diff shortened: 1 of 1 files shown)


praveen-pk:
>runner via an upcall

What do you mean by an upcall here?

sangho2:
> > runner via an upcall
> 
> What do you mean by an upcall here?

This is related to #556 and #564. Basically, we want to make the platform be unaware of OP-TEE stuffs. The platform delegates any OP-TEE commands to the runner and the shim via an upcall (technically, just a callback), and they will handle those commands.

praveen-pk:
LGTM!. I am not familiar with the register handling, so not approving the PR.

litebox_platform_lvbs/src/lib.rs
```diff
@@ -755,6 +759,237 @@ impl<Host: HostInterface> StdioProvider for LinuxKernel<Host> {
...
+    ctx: &'a mut litebox_common_linux::PtRegs,
+}
+
+fn run_thread_inner(
+    shim: &dyn litebox::shim::EnterShim<ExecutionContext = litebox_common_linux::PtRegs>,
+    ctx: &mut litebox_common_linux::PtRegs,
+) {
+    let ctx_ptr = core::ptr::from_mut(ctx);
+    let mut thread_ctx = ThreadContext { shim, ctx };
+    unsafe { run_thread_arch(&mut thread_ctx, ctx_ptr) };
```
wdcui, line 793: Why do we have `ctx` in `thread_ctx` AND have `ctx_ptr`?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -755,6 +759,237 @@ impl<Host: HostInterface> StdioProvider for LinuxKernel<Host> {
...
+        push rax // orig_rax
+        push rdi
+        push rsi
+        push rdx
+        push rcx
+        push -38
+        push r8
+        push r9
+        push r10
+        push [rsp + 88]
```
wdcui: This is `r11`? Why not push it directly?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -755,6 +759,237 @@ impl<Host: HostInterface> StdioProvider for LinuxKernel<Host> {
...
+/// Calls `f` in order to call into a shim entrypoint.
+impl ThreadContext<'_> {
+    fn call_shim(
+        &mut self,
+        f: impl FnOnce(
+            &dyn litebox::shim::EnterShim<ExecutionContext = litebox_common_linux::PtRegs>,
+            &mut litebox_common_linux::PtRegs,
+        ) -> ContinueOperation,
+    ) {
+        // TODO: clear the interrupt flag before calling the shim
```
wdcui: Do we need to handle this TODO in this PR? Would it break anything if we don't?

litebox_platform_lvbs/src/lib.rs
sangho2: `run_thread_arch` is written in assembly. `ctx_ptr` is to ease its access to CPU context.

litebox_platform_lvbs/src/lib.rs
sangho2: This is for storing `rflags` to `r11` as `syscall` does.

litebox_platform_lvbs/src/lib.rs
sangho2: Currently, we don't handle interrupts/exception for OP-TEE Shim/TA. So, it might be better to handle them all together later.

litebox_platform_lvbs/src/lib.rs
wdcui: Let's add a comment.

litebox_platform_lvbs/src/lib.rs
wdcui: Add a comment.

litebox_platform_lvbs/src/lib.rs
sangho2: removed TODO

litebox_platform_lvbs/src/lib.rs
sangho2: Done

litebox_platform_lvbs/src/lib.rs
sangho2: Done

sangho2:
Note. This PR still uses `KernelTls` to avoid conflicts. I'll rename this structure through #545 (or #541)

Issue #559, opened 2025-12-16
Open sourcing tasks


jstarks:
Tasks that we should consider completing before open sourcing:

- [x] #561
- [ ] #560

jaybosamiya-ms:
Closing as completed despite #560 not being closed yet, since we've successfully reduced the biggest contributor of `unsafe` in #592 to the point that I don't think it is a blocker to open-sourcing anymore.  #560 will remain open to track further reduction of `unsafe`.

PR #567, opened 2025-12-23
Add component detection manifest + NOTICE for `sync::Mutex` and `sync::RwLock`

jaybosamiya-ms:
Even though our implementations have diverged a significant amount, the original code was from the Rust source.  In preparation for open-sourcing LiteBox, the `cgmanifest.json` helps us document the source of these in a machine-readable form.  Additionally, `NOTICE.txt` covers the notice for this at the root of the dir.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,0 +1,33 @@
+NOTICES
+
+This repository incorporates material as listed below or described in the code.
+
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
Open-sourcing tracking issue #559

Issue #574, opened 2026-01-06
No suitable text section to insert trampoline metadata


CvvT:
I found another .so file that rewriter failed to rewrite: https://microsoft-my.sharepoint.com/:u:/p/weitengchen/IQDT8Num0nB-QbYGgdDxi-RyAc2Cr4wHfdwWFP1r9MwZmPM?e=gWul0h

It returned an error at the line:

https://github.com/microsoft/litebox/blob/949f37479d48e2afc42d375e87ac96a2e956efcf/litebox_syscall_rewriter/src/lib.rs#L163

```
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x0000000000000040 0x0000000000000040
                 0x0000000000000268 0x0000000000000268  R      0x8
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000136414 0x0000000000136414  R      0x1000
  LOAD           0x0000000000136414 0x0000000000137414 0x0000000000137414
                 0x000000000033438c 0x000000000033438c  R E    0x1000
  LOAD           0x000000000046a7a0 0x000000000046c7a0 0x000000000046c7a0
                 0x0000000000032918 0x0000000000033860  RW     0x1000
  LOAD           0x000000000049d0b8 0x00000000004a00b8 0x00000000004a00b8
                 0x00000000000052f0 0x00000000000055d8  RW     0x1000
  TLS            0x000000000046a7a0 0x000000000046c7a0 0x000000000046c7a0
                 0x0000000000000020 0x0000000000000070  R      0x8
  DYNAMIC        0x000000000049c7a8 0x000000000049e7a8 0x000000000049e7a8
                 0x00000000000001f0 0x00000000000001f0  RW     0x8
  GNU_RELRO      0x000000000046a7a0 0x000000000046c7a0 0x000000000046c7a0
                 0x0000000000032918 0x0000000000033860  R      0x1
  GNU_EH_FRAME   0x00000000000ead68 0x00000000000ead68 0x00000000000ead68
                 0x0000000000009184 0x0000000000009184  R      0x4
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x0
  NOTE           0x00000000000002a8 0x00000000000002a8 0x00000000000002a8
[...]

PR #577, opened 2026-01-08
Fix syscall rewriter

CvvT:
I found several binaries that our rewriter failed to rewrite due to no space between segments: see #574 for one example.

The reason why we need some space between segments is because we need to store some trampoline information, including a magic number, file offset of the trampoline data, and the size of the trampoline data (only 24 bytes are required). This is fine in most cases but there are some popular python libs our rewriter failed to rewriter, and thus we need to fix it.

Turns out we don't need to insert that trampoline info into any segments. Instead, we can just repurpose some fields (i.e., sh_addr, sh_offset, and sh_entsize) in the section header to store it.


litebox_common_linux/src/loader.rs
```diff
@@ -238,31 +230,28 @@ impl ElfParsedFile {
...
-            return Ok(());
-        }
-        file.read_at(shdr.sh_offset, data.as_mut_bytes())
-            .map_err(ElfParseError::Io)?;
+        // Note these fields are repurposed to store our trampoline info.
+        // See litebox_syscall_rewriter for details.
+        let magic_number = shdr.sh_addr;
+        let tramp_offset = shdr.sh_offset;
+        let tramp_size = shdr.sh_entsize;
         // TODO: check section name instead of magic number
```
jstarks: Should we use the opportunity to implement this TODO?

CvvT: Sure, seems easy to do. I asked copilot agent to finish it.

PR #579, opened 2026-01-12
Implement initial VTL1 ring buffer

karissanchez:
Creates a ring buffer for VTL1 using physical contiguous memory allocated in VTL0. Every print statement gets added to the ring buffer automatically, and the physical memory where the ring buffer resides is protected as read only for VTL0.


karissanchez:

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -67,6 +73,65 @@ fn line_status(port: u16) -> u8 {
     inb(port + IN_LINE_STATUS_OFFSET)
 }
 
+pub struct RingBuffer {
```
sangho2: I think it is better to move `RingBuffer` to some other files under `src/mshv` because this `RingBuffer` is independent of x86 and COM port.

Also, it seems that we don't need `pub` here.

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -67,6 +73,65 @@ fn line_status(port: u16) -> u8 {
...
+    data: Box<[u8]>,
```
sangho2: I think using `inner: Mutex<Box[u8]>` and letting each member function call `lock()` is more common pattern to hide internal details (e.g., we could use existing ringbuffer implementation for better performance later).
(I have to admit that the current `com()` is not that in good shape...)

Using `Box[u8]` here is not semantically correct because, in Rust, `Box` is specified to deal with heap allocation.

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -67,6 +73,65 @@ fn line_status(port: u16) -> u8 {
...
+    /// Create a new RingBuffer mapped to the given physical address and size
+    ///
+    /// # Panics
+    ///
+    /// Panics if the mapped size is less than the requested size
+    pub fn new(phys_addr: x86_64::PhysAddr, size: usize) -> Self {
+        let platform = crate::platform_low();
+
+        // Map the physical memory to virtual space
+        let core::prelude::v1::Ok((virt_ptr, mapped_size)) = platform.map_vtl0_phys_range(
```
sangho2: I think `let Ok(...)` is enough.

This relies on persistent mapping which we want to avoid. For now, it might be fine because VTL0 kernel cannot modify this area and we don't deallocate it. It might better to leave some comment here to revisit this later.

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -67,6 +73,65 @@ fn line_status(port: u16) -> u8 {
...
+        let slice = unsafe { slice::from_raw_parts_mut(virt_ptr, size) };
+        slice.fill(0);
+
+        // Convert to boxed slice
+        let data = unsafe { Box::from_raw(slice) };
+
+        Self { data, head_lpos: 0 }
+    }
+
+    // #[inline]
```
sangho2: do we need this comment?

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -67,6 +73,65 @@ fn line_status(port: u16) -> u8 {
...
+    pub fn size(&self) -> usize {
+        self.data.len()
+    }
+
+    #[inline]
+    fn mask(&self) -> usize {
+        self.size() - 1
+    }
+
+    pub fn write(&mut self, buf: &[u8]) -> usize {
```
sangho2: this might be a bit slow. It could be replaced by up to two memory copy functions (`copy_from_slice` or `copy_nonoverlapping`).

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -141,17 +206,38 @@ fn com() -> &'static Mutex<ComPort> {
     })
 }
 
+fn ringbuffer() -> &'static Mutex<RingBuffer> {
+    static RINGBUFFER_ONCE: Once<Mutex<RingBuffer>> = Once::new();
+    RINGBUFFER_ONCE.call_once(|| {
+        let ring_buffer = RingBuffer::new(
+            get_ringbuffer_phys_addr().unwrap(),
+            get_ringbuffer_size().unwrap(),
```
sangho2: IMO we don't need to rely on these two global variables to initialize a ring buffer. This kind of indirection is error prone.
For instance, we can move `static RINGBUFFER` to the outside and define two functions like `set_ringbuffer(addr, size)` and `ringbuffer()` for initialization and usage, as `set_platform()` and `platform()` do.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -945,6 +968,21 @@ fn apply_vtl0_text_patch(heki_patch: HekiPatch) -> Result<(), Errno> {
     Ok(())
 }
 
+fn mshv_vsm_allocate_ringbuffer_memory(phys_addr: u64, size: usize) -> Result<i64, Errno> {
+    // Store the physical address and size globally
+    RINGBUFFER_PHYS_ADDR.store(phys_addr, Ordering::SeqCst);
+    RINGBUFFER_SIZE.store(size, Ordering::SeqCst);
```
sangho2: We could call a function like `set_ringbuffer` here to allocate buffer here.

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -67,6 +73,65 @@ fn line_status(port: u16) -> u8 {
...
+        Self { data, head_lpos: 0 }
+    }
+
+    // #[inline]
+    pub fn size(&self) -> usize {
+        self.data.len()
+    }
+
+    #[inline]
+    fn mask(&self) -> usize {
```
sangho2: bitmask works only if the ring buffer size is a power of two. We might need to check or adjust the size.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -964,6 +1002,9 @@ pub fn vsm_dispatch(func_id: VsmFunction, params: &[u64]) -> i64 {
         VsmFunction::CopySecondaryKey => mshv_vsm_copy_secondary_key(params[0], params[1]),
         VsmFunction::KexecValidate => mshv_vsm_kexec_validate(params[0], params[1], params[2]),
         VsmFunction::PatchText => mshv_vsm_patch_text(params[0], params[1]),
+        VsmFunction::AllocateRingbufferMemory => {
```
praveen-pk, line 981: Do we need to do this in reponse to a vtl call? Ring buffer is really valuable no matter what platform is used within Litebox. I am wondering if it makes sense to make this a default litebox feature and LVBS can just use it.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -945,6 +968,21 @@ fn apply_vtl0_text_patch(heki_patch: HekiPatch) -> Result<(), Errno> {
...
+fn mshv_vsm_allocate_ringbuffer_memory(phys_addr: u64, size: usize) -> Result<i64, Errno> {
+    // Store the physical address and size globally
+    RINGBUFFER_PHYS_ADDR.store(phys_addr, Ordering::SeqCst);
+    RINGBUFFER_SIZE.store(size, Ordering::SeqCst);
+    protect_physical_memory_range(
+        PhysFrame::range(
+            PhysFrame::containing_address(PhysAddr::new(phys_addr)),
+            PhysFrame::containing_address(PhysAddr::new(phys_addr + (size as u64) - 1)),
+        ),
+        MemAttr::MEM_ATTR_READ, // Read-only
```
praveen-pk: Does this make this memory region RO to VTL0? Is VTL0 expected to read from this memory directly?

litebox_platform_lvbs/src/arch/x86/ioport.rs
praveen-pk: what do you mean persistent mapping here? Does this mapping persist across reboots Or for the life of VTL1?

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -67,6 +73,65 @@ fn line_status(port: u16) -> u8 {
...
+        self.size() - 1
+    }
+
+    pub fn write(&mut self, buf: &[u8]) -> usize {
+        let mut written = 0;
+        while written < buf.len() {
+            let index = self.head_lpos & self.mask();
+            self.data[index] = buf[written];
+            self.head_lpos += 1;
```
praveen-pk: Byte by byte copying would be slow. Please filling the buffer till the end in a single copy and wrap around if `buf` doesn't completely fit.

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -141,17 +206,38 @@ fn com() -> &'static Mutex<ComPort> {
...
 impl fmt::Write for ComPort {
     fn write_str(&mut self, s: &str) -> fmt::Result {
         self.write_string(s);
         Ok(())
     }
 }
 
+impl fmt::Write for RingBuffer {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        self.write(s.as_bytes());
```
praveen-pk: Any reason for not writing string to the buffer? Any binary data written means, the reader should also know the format and that adds new depedencies.

litebox_platform_lvbs/src/arch/x86/ioport.rs
sangho2: for the life of VTL1. In many places, we do on-demand mapping to copy from/to data VTL0. On-demand mapping itself isn't particularly safer than persistent mapping (especially because VTL0 cannot modify this ring buffer due to the SLAT protection), but on-demand mapping would remind us that we should copy data rather than directly accessing it.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Writing a good ring buffer is not easy (especially because many ones want to have a lock-free one nowadays). Some other parts of LiteBox are using the `ringbuf` crate and we might want to use something similar for this purpose as well.

litebox_platform_lvbs/src/mshv/vsm.rs
karissanchez: Correct, it makes the region RO for VTL0. For now, VTL0 is expected to read directly from memory, but I imagine this might change as we develop more uses for the ring buffer

litebox_platform_lvbs/src/arch/x86/ioport.rs
sangho2: This implements the `https://doc.rust-lang.org/core/fmt/trait.Write.html#tymethod.write_str` trait. We use `core::fmt::Write::write_fmt()` without overriding it (Lines 237 and 239). This function internally does necessary formatting and call this `write_str` in the end to print out the final string.

One performance concern is, we call `write_fmt()` with the same format string/arguments redundantly. This might be less important for now.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: If VTL0 is expected to read out the ring buffer, VTL1 should share `head_lpos` with it (with atomic memory operations). My guess is we are using this ring buffer only for crash dumping? (i.e., VTL0 is not an active consumer of this ring buffer which advances the tail index?)

litebox_platform_lvbs/src/arch/x86/ioport.rs
praveen-pk: As this memory range will be marked RO for VTL0, using persistent mapping makes more sense here. If on-demand mapping requires from/to data copying, it would unnecessarily the buffer acces.

litebox_platform_lvbs/src/arch/x86/ioport.rs
praveen-pk: Sounds good!

litebox_platform_lvbs/src/mshv/mod.rs
```diff
@@ -8,6 +8,7 @@ pub mod hvcall;
 mod hvcall_mm;
 mod hvcall_vp;
 mod mem_integrity;
+pub mod ringbuffer;
```
sangho2: `pub(crate)` would be enough.

litebox_platform_lvbs/src/mshv/ringbuffer.rs
```diff
@@ -0,0 +1,76 @@
...
+        if buf.len() >= self.size {
+            let single_slice = &buf[(buf.len() - self.size)..];
+            unsafe {
+                crate::platform_low().copy_slice_to_vtl0_phys(self.rb_pa, single_slice);
+            }
+            self.write_offset = 0;
+            return;
+        }
+
+        // Otherwise, calculate if wrapround needed
```
sangho2: typo:  wrapround

litebox_platform_lvbs/src/mshv/ringbuffer.rs
```diff
@@ -0,0 +1,76 @@
...
+    RINGBUFFER_ONCE.call_once(|| {
+        let ring_buffer = RingBuffer::new(pa, size);
+        Mutex::new(ring_buffer)
+    })
+}
+
+// Ring buffer will have already been set, so inputs are not used
+pub fn ringbuffer() -> &'static Mutex<RingBuffer> {
+    set_ringbuffer(PhysAddr::zero(), 0)
+}
```
sangho2: We can slightly revise these two functions and further `ringbuffer()` usage and allocation check like below:

```rust
static RINGBUFFER_ONCE: Once<Mutex<RingBuffer>> = Once::new();
pub(crate) fn set_ringbuffer(pa: PhysAddr, size: usize) {
    RINGBUFFER_ONCE.call_once(|| {
        let ring_buffer = RingBuffer::new(pa, size);
        Mutex::new(ring_buffer)
    });
}
pub(crate) fn ringbuffer() -> Option<&'static Mutex<RingBuffer>> {
    RINGBUFFER_ONCE.get()
}
```
Once we write `ringbuffer()` like above, we can write `print` like below while removing `RINGBUFFER_ALLOCATED` and `is_ringbuffer_allocated` (avoid having one redundant global variable).
```rust
pub fn print(args: ::core::fmt::Arguments) {
    use core::fmt::Write;
    let _ = com().lock().write_fmt(args);
    if let Some(rb) = ringbuffer() {
        let _ = rb.lock().write_fmt(args);
    }
}
```

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -945,6 +951,21 @@ fn apply_vtl0_text_patch(heki_patch: HekiPatch) -> Result<(), Errno> {
     Ok(())
 }
 
+fn mshv_vsm_allocate_ringbuffer_memory(phys_addr: u64, size: usize) -> Result<i64, Errno> {
```
sangho2, line 949: We might want to have some sanity checks here to confirm whether `[phys_addr, phys_addr+size)` belongs to VTL0 (TODO?).

Issue #580, opened 2026-01-12
Interrupt handlers for the LVBS kernel platform


sangho2:
The LVBS platform currently relies on the `x86_64` crate to implement interrupt handlers and program the interrupt descriptor table (IDT). This has several issues including blind CPU state saving/restoring, a lack of detailed control flow management, and requiring a nightly feature. We need our own interrupt handling routines which require some assembly code for low-level control.

PR #581, opened 2026-01-16
IDT without unstable abi_x86_interrupt

sangho2:
This PR revises LVBS's IDT to avoid using the unstable `abi_x86_interrupt` feature (#580).
We still need a nightly toolchain to build `litebox_runner_lvbs` since it requires unstable `build-std`
for floating-point operations and SSE.

### DIFF
--- a/litebox_platform_lvbs/src/arch/x86/interrupts.S
+++ b/litebox_platform_lvbs/src/arch/x86/interrupts.S
@@ -21,7 +21,7 @@
  *   r15, r14, r13, r12, rbp, rbx,     (callee-preserved)
  *   r11, r10, r9, r8, rax, rcx, rdx, rsi, rdi,  (callee-clobbered)
  *   orig_rax (error code),
- *   rip, cs, eflags, rsp, ss          (iret frame)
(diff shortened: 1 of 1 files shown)


litebox_platform_lvbs/src/arch/x86/interrupts.S
```diff
@@ -0,0 +1,166 @@
...
+ *   [rsp+24] RFLAGS
+ *   [rsp+16] CS
+ *   [rsp+8]  RIP
+ *   [rsp+0]  Error code (if applicable, otherwise pushed as 0)
+ *
+ * After push_regs, the stack matches PtRegs layout from litebox_common_linux:
+ *   r15, r14, r13, r12, rbp, rbx,     (callee-preserved)
+ *   r11, r10, r9, r8, rax, rcx, rdx, rsi, rdi,  (callee-clobbered)
+ *   orig_rax (error code),
+ *   rip, cs, eflags, rsp, ss          (iret frame)
```
wdcui: NIT: `rflags`?

PR #582, opened 2026-01-16
Add a lock visualizer to lock_tracing

jaybosamiya-ms:
This PR introduces a lock visualizer, along with some functionality to the `lock_tracing` feature to be able to actually record locking attempts away to a file.  Due to compile-time flags, it does not impact actual run time, but when `--features lock_tracing` is enabled, the runner will dump to `/tmp/locks.jsonl`, which can then be analyzed via the locks viewer dev tool (run with `uv run python3 server.py` inside `dev_tools/lock_viewer`).

This viewer lets you select individual (or collections of) locks, and see all the lock/unlock attempts in a nice way:

image

### DIFF
--- a/litebox/src/sync/lock_tracing.rs
+++ b/litebox/src/sync/lock_tracing.rs
@@ -283,8 +283,8 @@ impl RecordedEvent {
 pub struct RecordingSummary {
     /// Number of events that were recorded.
     pub recorded_events: u64,
-    /// Number of events that were dropped due to buffer overflow.
-    pub dropped_events: u64,
+    /// Number of oldest events that were evicted due to buffer overflow.
+    pub evicted_events: u64,
 }
 
 impl RecordingSummary {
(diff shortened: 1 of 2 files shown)


jaybosamiya-ms:
For review purposes, the only bits that impact actual functionality are inside `litebox/`.  Basically everything in the visualizer is not critical in any way to LiteBox itself, and should be considered mostly a debugging tool more than anything else.  It simply consumes this `/tmp/locks.jsonl` file, and then does the fancy visualization.  Practically the entirety of the visualization piece (i.e., `dev_tools/lock_viewer`) is vibe-coded by GitHub Copilot w/ Opus 4.5, with quite a bunch of steering to get it to a usable enough state that I am happy with.

litebox/src/sync/lock_tracing.rs
```diff
@@ -153,6 +217,242 @@ impl Locked {
...
+        if CONFIG_ENABLE_RECORDING {
+            self.record_unconditionally(event);
+        }
+    }
+
+    fn record_unconditionally(&mut self, event: RecordedEvent) {
+        self.total_recorded += 1;
+        if self.events.len() == CONFIG_MAX_RECORDED_EVENTS {
+            self.events.pop_front();
+            self.dropped_events += 1;
```
wdcui: NIT: `dropped_events` makes me think of events that were not recorded, but they were actually recorded then removed to make space for new events. Is there a better name for it?

jaybosamiya-ms: Thanks, indeed these mean "recorded and then dropped to make space for new events".  Probably "evicted" might be better term.  I'll update it.

PR #583, opened 2026-01-16
Fix and improve XSAVE usage in LVBS

sangho2:
This PR fixes LVBS platform's XSAVE bugs and improve its XSAVE usage. Specifically, 
this PR fixes incorrect usage of XSAVE mask for VTL0's states which results in GP. 
Also, it optimizes XSAVE area for VTL1 while clarifying how to deal with the XCR0
register, XSAVE/XSAVEOPT instructions, as well as why we don't use compaction in VTL1.

### DIFF
--- a/litebox_platform_lvbs/src/host/per_cpu_variables.rs
+++ b/litebox_platform_lvbs/src/host/per_cpu_variables.rs
@@ -182,6 +182,8 @@ impl PerCpuVariables {
     }
 
     /// Restore the extended states of each core (VTL0 or VTL1).
+    ///
+    /// This function is a no-op if the state was never saved.
     pub(crate) fn restore_extended_states(&self, vtl: u8) {
         if self.vtl0_xsave_area_addr.is_null()
             || self.vtl1_xsave_area_addr.is_null()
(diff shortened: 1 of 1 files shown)


litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -113,24 +127,50 @@ impl PerCpuVariables {
...
         if self.vtl0_xsave_area_addr.is_null() || self.vtl1_xsave_area_addr.is_null() {
             panic!("XSAVE areas are not allocated");
         } else {
-            let xsave_area_addr = match vtl {
-                HV_VTL_NORMAL => self.vtl0_xsave_area_addr.as_u64(),
-                HV_VTL_SECURE => self.vtl1_xsave_area_addr.as_u64(),
+            let (xsave_area_addr, mask, initialized) = match vtl {
+                HV_VTL_NORMAL => (
+                    self.vtl0_xsave_area_addr.as_u64(),
+                    xgetbv0(),
```
wdcui: What if vtl0 changes its xcr0 after vtl0 xsave area is allocated?

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -113,24 +127,50 @@ impl PerCpuVariables {
...
+                    &mut self.vtl0_xsave_area_initialized,
+                ),
+                HV_VTL_SECURE => (
+                    self.vtl1_xsave_area_addr.as_u64(),
+                    Self::VTL1_XSAVE_MASK,
+                    &mut self.vtl1_xsave_area_initialized,
+                ),
                 _ => panic!("Invalid VTL value: {}", vtl),
             };
+            // The first instruction to save the extended states must be XSAVE,
```
wdcui, line 158: The comment and the code are not consistent?

litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -363,3 +412,22 @@ fn get_xsave_area_size() -> usize {
         .expect("Failed to get cpuid extended state info");
     usize::try_from(sinfo.xsave_area_size_enabled_features()).unwrap()
 }
+
+#[allow(clippy::inline_always)]
+#[inline(always)]
+fn xgetbv0() -> u64 {
```
wdcui, line 443: If XCR0 cannot be modified by VTL0, shall we just cache it directly instead of calling `xgetbv`?

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: it is consistent. the default value of `initialized` is `false`. let me clarify it.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: In LVBS, VTL1 locks down XCR0. VTL0's any attempt to modify XCR0 results in a synthetic interrupt.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: That's a good idea. Let me do that.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
wdcui: Got it. Thanks.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
wdcui: Ah, mistook `*` as `!`. :-)

PR #585, opened 2026-01-21
Improve network throughput

CvvT:
This PR makes some significant changes to the network. One major bottleneck I found was that syscalls like `send` and `receive` need to compete with the background network worker for lock. To address this, this PR introduces another layer that uses ringbuf as proxy between shim and network stack. With this new design, `send` and `receive` now become lock-free (i.e., no lock contention with background network worker). This also makes the event notification of socket easier.

Another primary bottleneck is the packet dispatching algorithm in smoltcp, which performs a linear search to find the corresponding socket. We create a large number of listening sockets for backlog when user calls `listen` syscall (backlog is usually 4k), which introduces significant overhead. This PR set the upper limit to 8 (probably fine in most cases?).

Some other optimizations I found helpful include:
1. Instead of using a fixed timeout value (50ms), call `poll_at` from `smoltcp` which would determine when the next network operation should occur.
2. `perform_platform_interaction` now calls `poll` from `smoltcp` that processes multiple ingress packets before processing egress packets to avoid unnecessary egress processing (e.g. fewer ACK packets).
3. Enlarge the `SOCKET_BUFFER_SIZE` from 65536 to 65536 * 4.
4. pin the network thread to a CPU

With above changes, the throughput on my machine is 5.78 Gbits/sec.

### DIFF
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -52,8 +52,8 @@ const MAX_PACKET_COUNT: usize = 32;
 /// An important decision that must be made by a user of a `Network` is decided by
 /// [`set_platform_interaction`](Self::set_platform_interaction), whose docs explain this further.
 ///
-/// A user of `Network` who does not care about [events](crate::event) can choose to have a trivial
-/// provider for [`platform::RawMutexProvider`] that panics on all calls except `underlying_atomic`.
+/// A user of `Network` who care about [events](crate::event) should call [set_socket_proxy](Self::set_socket_proxy)
+/// to set up a proxy for each socket created, so that events can be notified properly.
 pub struct Network<Platform>
 where
     Platform:
@@ -396,9 +396,11 @@ pub enum PlatformInteractionReinvocationAdvice {
     /// control back to you to prevent unbounded length waits (crucial to prevent in
     /// non-pre-emptible environments), but otherwise has more work it anticipates it can do.
     CallAgainImmediately,
-    /// You don't need to call again until more packets arrive on the device's receive side or
-    /// the given timeout expires.
-    WaitOnDeviceOrSocketInteraction(Option<core::time::Duration>),
+    /// You don't need to call again until more packets arrive on the device's receive side (i.e., `timeout` is `None`),
+    /// or the given `timeout` expires.
+    WaitOnDeviceOrSocketInteraction {
+        timeout: Option<core::time::Duration>,
+    },
 }
 impl PlatformInteractionReinvocationAdvice {
     /// Convenience function to match against [`Self::CallAgainImmediately`]
@@ -448,7 +450,9 @@ where
             }
             smoltcp::iface::PollResult::None => {
                 let poll_at = self.poll_at();
-                PlatformInteractionReinvocationAdvice::WaitOnDeviceOrSocketInteraction(poll_at)
+                PlatformInteractionReinvocationAdvice::WaitOnDeviceOrSocketInteraction {
+                    timeout: poll_at,
+                }
             }
         }
     }
@@ -690,6 +694,9 @@ where
     }
 
     /// Creates a socket.
+    ///
+    /// By default, the created socket has no associated proxy; to set a proxy, use
+    /// [`set_socket_proxy`](Self::set_socket_proxy).
     pub fn socket(&mut self, protocol: Protocol) -> Result<SocketFd<Platform>, SocketError> {
         let handle = match protocol {
             Protocol::Tcp => self.socket_set.add(tcp::Socket::new(
@@ -766,18 +773,27 @@ where
     }
 
     /// Set the network proxy for the socket at `fd`
+    ///
+    /// Associating a proxy enables event notification and sending/receiving data without accessing
+    /// [`Network`] (which may help avoid lock contention but still requires a periodic call to
+    /// [`perform_platform_interaction`](Self::perform_platform_interaction) to move data between smoltcp
+    /// and the socket channels though).
+    ///
+    /// If no proxy is set, then the socket can still be used for sending/receiving data via [`Network`]
+    /// interfaces like [`send`](Self::send)/[`receive`](Self::receive), but no events will be notified.
+    #[must_use]
     pub fn set_socket_proxy(
         &mut self,
         fd: &SocketFd<Platform>,
         proxy: alloc::sync::Arc<NetworkProxy<Platform>>,
-    ) -> Result<(), alloc::sync::Arc<NetworkProxy<Platform>>> {
+    ) -> bool {
         let descriptor_table = self.litebox.descriptor_table();
         let Some(mut table_entry) = descriptor_table.get_entry_mut(fd) else {
-            return Err(proxy);
+            return false;
         };
         let socket_handle = &mut table_entry.entry;
         socket_handle.proxy = Some(proxy);
-        Ok(())
+        true
     }
 
     /// Close the socket at `fd`
@@ -1235,6 +1251,9 @@ where
     /// Accept a new incoming connection on a listening socket.
     ///
     /// If `peer` is provided, it is filled with the remote address of the accepted connection.
+    ///
+    /// Note that the returned new socket has no associated proxy; to set a proxy, use
+    /// [`set_socket_proxy`](Self::set_socket_proxy).
     pub fn accept(
         &mut self,
         fd: &SocketFd<Platform>,
--- a/litebox/src/net/socket_channel.rs
+++ b/litebox/src/net/socket_channel.rs
@@ -203,28 +203,6 @@ impl<Platform: RawSyncPrimitivesProvider + TimeProvider> NetworkProxy<Platform> 
             NetworkProxy::Raw => false,
         }
     }
-
-    /// Get the local port for datagram sockets (lock-free).
-    ///
-    /// Returns `Some(port)` for datagram sockets (0 if unbound), `None` for other socket types.
-    pub fn datagram_local_port(&self) -> Option<u16> {
-        match self {
-            NetworkProxy::Datagram(channel) => Some(channel.local_port()),
-            _ => None,
-        }
-    }
-
-    /// Set the local port for datagram sockets (lock-free, one-time operation).
-    ///
-    /// Uses compare-and-swap to ensure only one thread sets the port.
-    /// Returns `Ok(())` if set successfully, `Err(current_port)` if already set.
-    /// Returns `Ok(())` for non-datagram sockets (no-op).
-    pub fn set_datagram_local_port(&self, port: u16) -> Result<(), u16> {
-        match self {
-            NetworkProxy::Datagram(channel) => channel.set_local_port(port),
-            _ => Ok(()),
-        }
-    }
 }
 
 /// A channel for stream (TCP) socket communication.
@@ -232,11 +210,6 @@ impl<Platform: RawSyncPrimitivesProvider + TimeProvider> NetworkProxy<Platform> 
 /// This channel provides lock-free data transfer between user threads and the network
 /// worker. User threads write to the TX buffer and read from the RX buffer, while
 /// the network worker drains TX to smoltcp and fills RX from smoltcp.
-///
-/// # Thread Safety
-///
-/// The channel uses lock-free ring buffers for data transfer, with atomic counters
-/// for tracking buffer availability. This allows concurrent access without blocking.
 pub struct StreamSocketChannel<Platform: RawSyncPrimitivesProvider + TimeProvider> {
     inner: StreamChannelInner<Platform>,
 }
@@ -798,6 +771,32 @@ impl<Platform: RawSyncPrimitivesProvider + TimeProvider> DatagramSocketChannel<P
     pub fn is_writable(&self) -> bool {
         self.inner.tx_space.load(Ordering::Acquire) > 0
     }
+
+    /// Get the local port the socket is bound to.
+    ///
+    /// Returns 0 if the socket is not yet bound.
+    pub fn local_port(&self) -> u16 {
+        self.inner.local_port.load(Ordering::Acquire)
+    }
+
+    /// Set the local port the socket is bound to.
+    ///
+    /// This should be called when the socket is bound (either explicitly or via auto-binding).
+    /// Uses compare-and-swap to ensure only one thread can set the port.
+    ///
+    /// Returns `Ok(())` if the port was set successfully, or `Err(current_port)` if
+    /// a port was already set.
+    pub fn set_local_port(&self, port: u16) -> Result<(), u16> {
+        debug_assert!(port != 0, "Port 0 is not a valid bound port");
+        match self
+            .inner
+            .local_port
+            .compare_exchange(0, port, Ordering::AcqRel, Ordering::Acquire)
+        {
+            Ok(_) => Ok(()),
+            Err(current) => Err(current),
+        }
+    }
 }
 
 impl<Platform: RawSyncPrimitivesProvider + TimeProvider> IOPollable
@@ -909,32 +908,6 @@ impl<Platform: RawSyncPrimitivesProvider + TimeProvider> DatagramSocketChannel<P
             self.inner.rx_count.store(1, Ordering::Release);
         } else {
             self.inner.rx_count.store(0, Ordering::Release);
-        }
-    }
-
-    /// Get the local port the socket is bound to.
-    ///
-    /// Returns 0 if the socket is not yet bound.
-    /// This is a lock-free operation.
-    fn local_port(&self) -> u16 {
-        self.inner.local_port.load(Ordering::Acquire)
-    }
-
-    /// Set the local port the socket is bound to.
-    ///
-    /// This should be called when the socket is bound (either explicitly or via auto-binding).
-    /// Uses compare-and-swap to ensure only one thread can set the port.
-    ///
-    /// Returns `Ok(())` if the port was set successfully, or `Err(current_port)` if
-    /// a port was already set.
-    fn set_local_port(&self, port: u16) -> Result<(), u16> {
-        match self
-            .inner
-            .local_port
-            .compare_exchange(0, port, Ordering::AcqRel, Ordering::Acquire)
-        {
-            Ok(_) => Ok(()),
-            Err(current) => Err(current),
         }
     }
 
--- a/litebox_runner_linux_userland/src/lib.rs
+++ b/litebox_runner_linux_userland/src/lib.rs
@@ -279,7 +279,7 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
                 let timeout = loop {
                     match shim.perform_network_interaction() {
                         litebox::net::PlatformInteractionReinvocationAdvice::CallAgainImmediately => {}
-                        litebox::net::PlatformInteractionReinvocationAdvice::WaitOnDeviceOrSocketInteraction(timeout) => {
+                        litebox::net::PlatformInteractionReinvocationAdvice::WaitOnDeviceOrSocketInteraction{ timeout } => {
                             break timeout;
                         }
                     }
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -233,8 +233,8 @@ impl GlobalState {
         assert!(old.is_none());
         drop(dt);
 
-        if self.net.lock().set_socket_proxy(fd, proxy.clone()).is_err() {
-            unreachable!("failed to set socket proxy for initialized socket");
+        if !self.net.lock().set_socket_proxy(fd, proxy.clone()) {
+            unreachable!("failed to set socket proxy for a newly-created socket");
         }
         proxy
     }
@@ -695,7 +695,9 @@ impl GlobalState {
         // Auto-bind UDP sockets if not already bound (Linux behavior: sendto() on an unbound
         // UDP socket implicitly binds it to an ephemeral port before sending).
         // This is mostly lock-free: we only take the network lock if we need to allocate a port.
-        if let Some(0) = proxy.datagram_local_port() {
+        if let NetworkProxy::Datagram(proxy) = proxy.as_ref()
+            && proxy.local_port() == 0
+        {
             // UDP socket is unbound - bind to an ephemeral port
             let mut net = self.net.lock();
             // Bind with port 0 to get an ephemeral port
@@ -719,7 +721,7 @@ impl GlobalState {
             // Get the assigned port
             let local_addr = net.get_local_addr(fd).map_err(Errno::from)?;
             // If another thread already set a port, that's fine - we'll use theirs
-            let _ = proxy.set_datagram_local_port(local_addr.port());
+            let _ = proxy.set_local_port(local_addr.port());
         }
 
         // Convert `SendFlags` to `litebox::net::SendFlags`


CvvT:
I used `iperf3` to measure the network throughput.

```
cargo test --package litebox_runner_linux_userland --test run --release -- test_tun_and_runner_with_iperf3 --exact --nocapture

throughput: 2.03 Gbps
```

Running `iperf3` natively:

```
throughput: 10.8 Gbits/sec
```

To understand how much slowdown was caused by smoltcp. I tried to benchmark smoltcp. However, it does not work directly with iperf3. The closest thing is the [benchmark](https://github.com/smoltcp-rs/smoltcp/blob/main/examples/benchmark.rs) from smoltcp:

```
cargo run --release --example benchmark -- --tun tun99 writer

throughput: 6.074 Gbps
```

However, this test only uses single thread, which is different from our design. I made some [change](https://gist.github.com/CvvT/79d2cda47e8a0968fa905093ac72c110) to it so that packet processing is done in a separate thread and then ran the above command again.

```
throughput: 2.542 Gbps
```

CvvT:
> I used `iperf3` to measure the network throughput.
> 
> ```
> cargo test --package litebox_runner_linux_userland --test run --release -- test_tun_and_runner_with_iperf3 --exact --nocapture
> 
> thoughput: 2.03 Gbps
> ```
> 
> To understand how much slowdown was caused by smoltcp. I tried to benchmark smoltcp. However, it does not work directly with iperf3. The closest thing is the [benchmark](https://github.com/smoltcp-rs/smoltcp/blob/main/examples/benchmark.rs) from smoltcp:
> 
> ```
> cargo run --release --example benchmark -- --tun tun99 writer
> 
> throughput: 6.074 Gbps
> ```
> 
> However, this test only uses single thread, which is different from our design. I made some [change](https://gist.github.com/CvvT/79d2cda47e8a0968fa905093ac72c110) to it so that packet processing is done in a separate thread and then ran the above command again.
> 
> ```
> throughput: 2.542 Gbps
> ```

The performance seems unstable. I made some changes to reduce memory copying but somehow the performance drops to 1.78 Gbits/sec. The current PR is already big enough; I won't add any further optimization to it.

Also, the comparison may not be fair because the socket buffer size used in LiteBox is 4x of that in smoltcp benchmark. If I reset it back 65536, the performance drops to ~1Gbits/s. I believe the bottleneck is that processing packets from kernel --> smoltcp --> intermediate layer (ringbuf) is too slow.

litebox/src/net/mod.rs
```diff
@@ -597,96 +442,92 @@ where
...
+        self.remove_dead_sockets();
+        self.close_pending_sockets();
+
+        self.drain_all_socket_channel_buffers();
+        // Process ingress.
+        let mut res = smoltcp::iface::PollResult::None;
         let timestamp = self.now();
-        let mut socket_state_changed = false;
-        let ingress_advice = if direction.ingress() {
+        let mut n = 1000;
```
wdcui: Add a comment to explain the choice of this constant?

litebox/src/net/mod.rs
```diff
@@ -597,96 +442,92 @@ where
...
-            PlatformInteraction::Automatic => {
-                while self
-                    .internal_perform_platform_interaction(direction)
-                    .call_again_immediately()
-                {
-                    // We just loop until all platform interaction is completed, _roughly_ analogous
-                    // to smoltcp's `poll`.
-                }
+        // Process egress.
+        match self
```
wdcui: Why do we need to repeatedly poll ingress but only need to poll egress once?

litebox/src/net/mod.rs
```diff
@@ -597,96 +442,92 @@ where
...
-            let events = PollableSocketHandle {
-                socket_handle,
-                socket_set: &self.socket_set,
-            }
-            .check_socket_events(true);
-            if !events.is_empty() {
-                socket_handle.pollee.notify_observers(events);
+    /// (Internal-only API) Perform the queued interactions.
+    fn automated_platform_interaction(&mut self, _direction: PollDirection) {
+        match self.platform_interaction {
```
wdcui, line 491: Where do we set `platform_interaction`?

litebox/src/net/mod.rs
```diff
@@ -737,6 +585,126 @@ where
...
+                // Drain RX buffer: from smoltcp directly to ring buffer
+                while tcp_socket.can_recv() {
+                    let received = proxy
+                        .push_rx_data_with(|buf| tcp_socket.recv_slice(buf).unwrap_or_default());
+                    if received == 0 {
+                        break;
+                    }
+                }
+
+                if let tcp::State::Established = tcp_socket.state() {
```
wdcui, line 598: Do we need to handle other states here as well?

litebox/src/net/mod.rs
```diff
@@ -737,6 +585,126 @@ where
...
+                }
+            }
+            (Protocol::Udp, NetworkProxy::Datagram(udp_proxy)) => {
+                let udp_socket = socket_set.get_mut::<udp::Socket>(socket_handle.handle);
+                let remote_endpoint = socket_handle.udp().remote_endpoint;
+
+                // Drain TX queue: try to send datagrams, consume only on success
+                while udp_socket.can_send() {
+                    // Try to send - consumes datagram only if closure returns true
+                    let result = udp_proxy.try_send_datagram_with(|data, addr| {
```
wdcui, line 631: Why does proxy need to have different APIs for TCP and UDP?

litebox/src/net/mod.rs
```diff
@@ -737,6 +585,126 @@ where
...
+                        let destination = addr
+                            .map(|s| match s {
+                                SocketAddr::V4(addr) => smoltcp::wire::IpEndpoint::from(addr),
+                                SocketAddr::V6(_) => unimplemented!(),
+                            })
+                            .or(remote_endpoint);
+                        if let Some(endpoint) = destination {
+                            udp_socket.send_slice(data, endpoint).is_ok()
+                        } else {
+                            // No destination - discard
```
wdcui, line 641: In what case whould this happen?

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -352,6 +362,19 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
...
 
+/// Pin the current thread to a specific CPU core
+fn pin_thread_to_cpu(cpu: usize) {
+    unsafe {
+        let mut set = std::mem::zeroed();
+        libc::CPU_ZERO(&mut set);
+        libc::CPU_SET(cpu, &mut set);
+
+        if libc::sched_setaffinity(0, std::mem::size_of::<libc::cpu_set_t>(), &raw const set) != 0 {
+            eprintln!("Warning: Failed to pin thread to CPU core {cpu}");
```
wdcui, line 373: Should we exit early if we can't pin the thread to a core?

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -571,28 +571,32 @@ impl Task {
...
-                                None,
-                            )
-                        })
-                    },
-                    |_fd| todo!("pipes"),
-                )
-                .flatten(),
+            Descriptor::LiteBoxRawFd(raw_fd) => {
+                let raw_fd = *raw_fd;
+                drop(locked_file_descriptors);
```
wdcui: Add a comment for this `drop` call?

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -169,6 +211,28 @@ impl GlobalState {
...
+                NetworkProxy::Stream(proxy)
+            }
+            SockType::Datagram => {
+                let proxy = litebox::net::socket_channel::new_datagram_channel();
+                NetworkProxy::Datagram(proxy)
+            }
+            SockType::Raw => NetworkProxy::Raw,
+            _ => unimplemented!(),
+        };
+        let proxy = Arc::new(proxy);
```
wdcui, line 231: Why do we need to set proxy in both entry metadata and socket?

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1380 @@
...
+//!          ▼                                            ▼
+//! ┌─────────────────────────────────────────────────────────────┐
+//! │                    SocketChannel                            │
+//! │  ┌─────────────────┐          ┌─────────────────────────┐   │
+//! │  │   RX Channel    │◄─────────│  Network Worker writes  │   │
+//! │  │  (lock-free)    │          │  data from smoltcp      │   │
+//! │  │  User reads ────┼──────────►                         │   │
+//! │  └─────────────────┘          └─────────────────────────┘   │
+//! │                                                             │
+//! │  ┌─────────────────┐          ┌─────────────────────────┐   │
```
wdcui: Should the arrows in TX channel be bi-direction as the RX channel?

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1380 @@
...
+//! ```
+//!
+//! # Benefits
+//!
+//! - **No lock contention**: User read/write operations and network processing can proceed
+//!   concurrently without blocking each other.
+//! - **Better latency**: User operations complete faster as they don't need to wait for
+//!   the network worker to release locks.
+//! - **Improved throughput**: Network worker can process packets continuously without
+//!   being blocked by user operations.
```
wdcui: Does this change the semantics of blocking IO?

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1380 @@
...
+    ///
+    /// Returns the number of bytes written, or an error if the buffer is full
+    /// or the socket is in an invalid state.
+    pub fn try_write(
+        &self,
+        buf: &[u8],
+        flags: super::SendFlags,
+        destination: Option<SocketAddr>,
+    ) -> Result<usize, SendError> {
+        if !flags.is_empty() {
```
wdcui, line 153: What does this check tell us? Is send buffer is empty?

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1380 @@
...
+    for StreamSocketChannel<Platform>
+{
+    fn register_observer(&self, observer: alloc::sync::Weak<dyn Observer<Events>>, mask: Events) {
+        self.inner.pollee.register_observer(observer, mask);
+    }
+
+    fn check_io_events(&self) -> Events {
+        let mut events = Events::empty();
+
+        if self.inner.rx_available.load(Ordering::Acquire) > 0 {
```
wdcui: Should we use `is_readable` and `is_writable`?

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1380 @@
...
+    where
+        F: FnMut(&mut [u8]) -> usize,
+    {
+        let mut rx_prod = self.inner.rx_prod.lock();
+        let (first, second) = (*rx_prod).vacant_slices_mut();
+
+        // SAFETY: We're treating maybe_uninit<u8> slices as &mut [u8].
+        // This is safe because:
+        // 1. u8 has no drop implementation or invalid bit patterns
+        // 2. The closure will write to these bytes before we advance the write index
```
wdcui: Nit: "write to these bytes" => "write to the buffer"?

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1380 @@
...
+    pub(super) fn notify_io_event(&self, events: Events) {
+        self.inner.pollee.notify_observers(events);
+    }
+}
+
+/// Create a new stream socket channel with default buffer sizes.
+///
+/// The channel is created with default buffer sizes [`super::SOCKET_BUFFER_SIZE`] for both
+/// RX and TX buffers.
+pub fn new_stream_channel<Platform: RawSyncPrimitivesProvider + TimeProvider>()
```
wdcui: Why not use a new() function for StreamSocketChannel?

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1380 @@
...
+//! └─────────────────────────────────────────────────────────────┘
+//! ```
+//!
+//! # Benefits
+//!
+//! - **No lock contention**: User read/write operations and network processing can proceed
+//!   concurrently without blocking each other.
+//! - **Better latency**: User operations complete faster as they don't need to wait for
+//!   the network worker to release locks.
+//! - **Improved throughput**: Network worker can process packets continuously without
```
wdcui: What happens when we run out of the ring buffer?

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1380 @@
...
+            rx_cons.pop_slice(buf)
+        };
+
+        if let Some(source_addr) = source_addr {
+            // TCP is connection-oriented, so no need to provide a source address
+            *source_addr = None;
+        }
+
+        // Update available count
+        self.inner.rx_available.fetch_sub(n, Ordering::Release);
```
wdcui, line 342: The `try_read` function for UDP has a notification at the end. Do we need one here as well?

CvvT:
I figured out another bottleneck caused by the linear search of packet dispatching in smoltcp (see discussion in https://github.com/asterinas/asterinas/issues/1267 and https://github.com/smoltcp-rs/smoltcp/issues/973). When we create a server socket that has 4k listening sockets in backlog, smoltcp needs to do a linear search to find out the correct socket when it tries to dispatch a raw packet.

It seems that the maintainers claimed that smoltcp is only designed for embedded system where there are typically 5-10 sockets around, and thus they are not willing to change it.

litebox/src/net/mod.rs
CvvT: Looks like 1000 is big enough to process all incoming packets. I removed this and changed to use another API `poll` that process all incoming packets.

litebox/src/net/mod.rs
CvvT: I changed to use smoltcp's API `poll` that does the same thing. I think the rationale is that batching the process of incoming packets help reduce the number of outgoing packets, and thus improve the performance.

litebox/src/net/mod.rs
CvvT: There is an interface `set_platform_interaction` that should be called by runners.

litebox/src/net/mod.rs
CvvT: Only two states really matter: `Connected` and `Closed`. `Closed` is set in the following line.

litebox/src/net/mod.rs
CvvT: For connected TCP sockets, we don't need to save the destination/source address, while for UDP we need to save both data and address.

litebox/src/net/mod.rs
CvvT: If user calls `send` without specifying a destination and no `connect` was called beforehand, we should return an error. However, in the background worker, there is no way to return an error to users (send already finished), so I just ignore this error.

To fix it, I added some check at `send`: if the socket is not connected and user does not specify the destination, it returns an error.

litebox/src/net/socket_channel.rs
CvvT: Good point. The current one is confusing. Let me update it.

litebox/src/net/socket_channel.rs
CvvT: You mean allocation fails or the buffer becomes full. In the former case, the program crashes, and in the latter case, send/receive would return EAGAIN.

litebox/src/net/socket_channel.rs
CvvT: It is handled already (by implementing the `IOPollable` interface). All operations in this file are nonblocking.

litebox/src/net/socket_channel.rs
CvvT: Flags `DONTWAIT` and `NOSIGNAL` are handled at upper level, but there are some other flags I have not supported like `DONTROUTE`. Probably most can be ignored directly.

litebox/src/net/socket_channel.rs
CvvT: Good catch. I forgot the remove that for UDP. Notification should only be done by the network worker when moving data from/to smoltcp.

litebox/src/net/socket_channel.rs
CvvT: Sure, fixed.

litebox/src/net/socket_channel.rs
CvvT: updated

litebox/src/net/socket_channel.rs
CvvT: This is just a wrapper function for `StreamChannelInner::new`.

litebox_runner_linux_userland/src/lib.rs
CvvT: probably fine? The performance does not drop too much.

litebox_shim_linux/src/syscalls/file.rs
CvvT: Sure.

litebox_shim_linux/src/syscalls/net.rs
CvvT: network socket needs to access it at many places and read/write syscalls also need to access it without locking network. It is possible to get rid of the one saved in socket (but seems to be non-trivial to make it work). Performance-wise, it is also better at the cost of a single copy.

litebox/src/net/socket_channel.rs
wdcui: I don't follow here. Why can't we use a member function like `StreamSocketChannel::new()`?

litebox_shim_linux/src/syscalls/net.rs
wdcui: Makes sense. Please add a comment for this design choice.

litebox/src/net/socket_channel.rs
CvvT: You mean we should move this function to `StreamSocketChannel` as a member function?

CvvT: done

litebox_shim_linux/src/syscalls/net.rs
CvvT: Sure, updated

litebox/src/net/socket_channel.rs
wdcui: Thanks!

CvvT:
@jaybosamiya-ms Would you like to take a look at this PR as it touches litebox module?

jaybosamiya-ms:
> Would you like to take a look at this PR as it touches litebox module?

Taking a look at the `litebox` module level changes now, will put review comments shortly, thanks for the heads up!

litebox/src/net/mod.rs
```diff
@@ -840,6 +765,21 @@ where
         self.litebox.descriptor_table_mut().insert(socket_handle)
     }
 
+    /// Set the network proxy for the socket at `fd`
+    pub fn set_socket_proxy(
```
jaybosamiya-ms, line 785: I think this might benefit from a bit more of a detailed comment.  Something along the lines of what you've already got on the `proxy` field might already be more helpful:
```
enable lock-free data transfer
    /// and event notification
```

It might also help to call out what happens if you _don't_ invoke this (i.e., things will work normally, it is just all operations then require holding the network lock?).

litebox/src/net/mod.rs
```diff
@@ -840,6 +765,21 @@ where
...
+        &mut self,
+        fd: &SocketFd<Platform>,
+        proxy: alloc::sync::Arc<NetworkProxy<Platform>>,
+    ) -> Result<(), alloc::sync::Arc<NetworkProxy<Platform>>> {
```
jaybosamiya-ms: I'm mildly surprised by this type signature, I'm not quite sure why the `Err` side gives back the `proxy`.  This would make sense if it were not an `Arc` so that it is not dropped, but in this case, it is an `Arc` so presumably if the callee wanted to keep it, they could have just maintained a clone of the `Arc`.  The correct move here is probably a `#[must_use]` attribute on the function, along with a `-> bool` that says success/failure.  _Technically_ to keep things in line with the other functionality in this, the correct move is probably a `Result<(), ProxySetError>` or such, but that seems like it is not going to help much more than a `#[must_use]`+`bool`.

litebox/src/net/mod.rs
```diff
@@ -551,9 +396,9 @@ pub enum PlatformInteractionReinvocationAdvice {
     /// control back to you to prevent unbounded length waits (crucial to prevent in
     /// non-pre-emptible environments), but otherwise has more work it anticipates it can do.
     CallAgainImmediately,
-    /// You don't need to call again until more packets arrive on the device's receive side or if
-    /// any socket interaction has occurred.
-    WaitOnDeviceOrSocketInteraction,
+    /// You don't need to call again until more packets arrive on the device's receive side or
+    /// the given timeout expires.
+    WaitOnDeviceOrSocketInteraction(Option<core::time::Duration>),
```
jaybosamiya-ms: What does "None" mean should be documented here.  Because if `None` shows up, then the user won't know whether that means (effectively) 0 or infinity.  I _think_ it is equivalent to infinity, but you should put the correct side of it.

litebox/src/net/mod.rs
jaybosamiya-ms: I think this comment is no longer accurate, should we remove it?

litebox/src/net/mod.rs
```diff
@@ -1333,22 +1288,16 @@ where
...
                         local_port,
                         server_socket: None,
-                        previously_acceptable: AtomicBool::new(false),
-                        was_connection_initiated: AtomicBool::new(true),
                         immediate_close: AtomicBool::new(false),
                     }),
-                    pollee: crate::event::polling::Pollee::new(),
-                    previously_sendable: AtomicBool::new(true),
-                    recv_queue: AtomicUsize::new(0),
+                    proxy: None,
```
jaybosamiya-ms, line 1319: I think the docs for `accept` should call out the implications on `proxy` explicitly (i.e., that the new FD will have no proxy assigned to it)

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1406 @@
...
+    ///
+    /// Uses compare-and-swap to ensure only one thread sets the port.
+    /// Returns `Ok(())` if set successfully, `Err(current_port)` if already set.
+    /// Returns `Ok(())` for non-datagram sockets (no-op).
+    pub fn set_datagram_local_port(&self, port: u16) -> Result<(), u16> {
+        match self {
+            NetworkProxy::Datagram(channel) => channel.set_local_port(port),
+            _ => Ok(()),
+        }
+    }
```
jaybosamiya-ms: This feels like the wrong place for this functionality to be.  The enum is already `pub`, which means that the callees can already do the `match`.  The names of the functions _also_ are saying "only do this if datagram", which feels overly specific, so why not just make the `DatagramSocketChannel` have their `local_port` and `set_local_port` be `pub`?

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1406 @@
...
+/// A channel for stream (TCP) socket communication.
+///
+/// This channel provides lock-free data transfer between user threads and the network
+/// worker. User threads write to the TX buffer and read from the RX buffer, while
+/// the network worker drains TX to smoltcp and fills RX from smoltcp.
+///
+/// # Thread Safety
+///
+/// The channel uses lock-free ring buffers for data transfer, with atomic counters
+/// for tracking buffer availability. This allows concurrent access without blocking.
```
jaybosamiya-ms: Usually a safety note is a recommendation for the caller to pay attention to some details.  This one seems like it is just saying "we implemented it correctly, you don't need to worry" which seems like it is not a helpful comment, no?  Can be removed.

litebox/src/net/socket_channel.rs
```diff
@@ -0,0 +1,1406 @@
...
+            self.inner.rx_count.store(1, Ordering::Release);
+        } else {
+            self.inner.rx_count.store(0, Ordering::Release);
+        }
+    }
+
+    /// Get the local port the socket is bound to.
+    ///
+    /// Returns 0 if the socket is not yet bound.
+    /// This is a lock-free operation.
```
jaybosamiya-ms: This sentence here _implies_ that other operations might take locks.  Any reason this is explicitly mentioned?

litebox/src/net/mod.rs
CvvT: Right, I updated it.

litebox/src/net/mod.rs
CvvT: Good point, updated. The first half of the sentence refers to `None`. I made it clear.

litebox/src/net/mod.rs
CvvT: Good point, changed to `must_use + bool`

litebox/src/net/mod.rs
CvvT: Right, updated.

litebox/src/net/mod.rs
CvvT: Sure, I also updated `Network::socket` to make it clear.

litebox/src/net/socket_channel.rs
CvvT: Good point, fixed.

litebox/src/net/socket_channel.rs
CvvT: Sure, removed.

litebox/src/net/socket_channel.rs
CvvT: Good catch! I was comparing against an alternative solution that needs to acquire the lock on network. This is removed.

PR #586, opened 2026-01-21
Raise GP fault in VTL0 if VTL1 panics

athvu:
Raise a GP fault in VTL0 if VTL1 panics so that there is a reboot.

### DIFF
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -12,7 +12,7 @@ use litebox_platform_lvbs::{
     mshv::{
         hvcall,
         vsm_intercept::raise_vtl0_gp_fault,
-        vtl_switch::{jump_to_vtl_switch_loop_with_stack_cleanup, vtl_switch_loop_entry},
+        vtl_switch::{panic_vtl_switch, vtl_switch_loop_entry},
         vtl1_mem_layout::{
             PAGE_SIZE, VTL1_INIT_HEAP_SIZE, VTL1_INIT_HEAP_START_PAGE, VTL1_PML4E_PAGE,
             VTL1_PRE_POPULATED_MEMORY_SIZE, get_heap_start_address,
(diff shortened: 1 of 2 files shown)


litebox_runner_lvbs/src/lib.rs
```diff
@@ -105,5 +106,6 @@ pub fn run(platform: Option<&'static Platform>) -> ! {
 #[panic_handler]
 fn panic(info: &PanicInfo) -> ! {
     serial_println!("{}", info);
-    hlt_loop()
+    raise_vtl0_gp_fault().expect("Failed to raise VTL0 GP fault on panic");
```
athvu: How should I handle this causing a loop in the case that raise_vtl0_gp_fault() fails?

sangho2: `expect()` calls `panic()` if the function returns an error. You'd better to suppress the error like `let _ = raise_vtl0_gp_fault();`.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -105,5 +106,6 @@ pub fn run(platform: Option<&'static Platform>) -> ! {
...
+    jump_to_vtl_switch_loop_with_stack_cleanup()
```
sangho2: Perhaps it's better to explicitly do vtl_return here. At this point, there might be nothing we can save/restore.
FYI. #545 heavily revises the VTL switch loop and this implementation might be affected by this PR.

litebox_platform_lvbs/src/mshv/vsm_intercept.rs
```diff
@@ -48,6 +48,12 @@ pub enum InterceptedRegisterName {
     Unknown = 0xffff_ffff,
 }
 
+/// # Panics
+///
+/// Panics if:
+/// - Failed to get intercept message type
+/// - Failed to raise VTL0 GP fault
+/// - Intercepted write to unknown MSR/register
```
sangho2, line 56: Does this mean that we need to define types and MSRs/registers we don't cover yet? TODO might be needed in that case.

Do we know why we might fail to raise VTL0 GP fault?

athvu: We currently protect only a subset of the control registers. I am not sure if in the future that will change and we will need to define the rest of the registers. I will discuss that with Thara

raise_vtl0_gp_fault() calls hvcall_set_vp_vtl0_registers() and that hypercall can fail for various reasons such as having invalid access, state, or parameters

Issue #505, opened 2025-11-19
RFC: "Raw" pointer improvements


jstarks:
LiteBox has evolved such that "raw" pointers essentially mean "guest"/"user" pointers--they represent untrusted addresses under user control, outside of the Rust memory model. Although one can imagine uses of LiteBox where the pointers _can_ be trusted, no one writing generic shim code can _assume_ this.

As such, there are several challenges with the current design:

## Safety

Many of the pointer trait methods are `unsafe`, but the safety contract is unclear and/or unenforceable. These pointers are inherently untrusted--they come from code on the other side of the security barrier (the Linux process, the OPTEE TA).

And there is no mechanism that the shim could use to validate a address ahead of time, before constructing a pointer, since in most cases this would be a TOCTOU bug--the user process can change the page mappings between pointer validation and pointer use. The best we can do is validate that the address is outside of the kernel address range... but we might as well defer that to access time if we're going to perform the check at all (on some architectures, it can be combined with the access).

We should fix this so that all trait methods on raw pointers are safe, relying on the platform to ensure this. For userland platforms, where this is not practical and there is no security boundary, this will have to be best-effort.

### COW

The pointer trait methods provide ways to get borrowed `Cow` references to the user data but, as discussed above, this is always unsafe and undefined behavior--guest memory lives outside the Rust memory model. We should eliminate this altogether.

### Safe transmutability

Currently, the pointer trait access methods do not require that the underlying data can be safely transmuted to/from bytes--it's up to the caller to not read an `Arc<u32>` or `&'static Platform` from guest memory. We can leverage the `zerocopy` crate to provide useful bounds while still allowing safe access of user types.

### Atomic accesses

Some code (currently, just the futex manager) needs a guarantee that a given memory access is performed atomically. There is no mechanism for this today, and so code that has this requirement currently transmutes the address to an `&AtomicU32` and hopes for the best.

## Ergonomics

The current pointers are difficult to use correctly for some use cases. 

* There are multiple overlapping methods to read/write data, but no simple `read`/`write` methods for the common case where you just want to access the pointed-to data.
* It is difficult to cast a pointer between types or to construct a pointer that is offset from another one. You can cast through `usize`, but the traits warn you not to do this.
* Accessing a slice of guest memory requires extra care since you are responsible for all bounds checks. There's no mechanism to create a "raw slice".
[...]

PR #587, opened 2026-01-22
Refactor the pointer API to let the `Cow`s out to pasture

jaybosamiya-ms:
This PR changes the interfaces for raw pointer APIs to remove the `Cow`s that existed, making all the reading operations return the owned variants always.  We already were not using the "borrowed" variants much, and the existence of the borrowed variants makes maintaining the safety contracts quite difficult.

Technically, this _slightly_ increases the number of blocks that require wrapping with `unsafe` but the _vast_ majority of the situations _already_ have a wrapped `unsafe` already, so practically nothing much changes this.  A future PR is going to clean these `unsafe`s up to make things much safer through the use of `zerocopy` traits. Future work _after_ that will work on an even nicer design (see #505).

Along the way, this PR also fixes up a bug in `litebox/src/platform/trivial_providers.rs` where it (under certain sized reads) would ignore the offset counts.

### DIFF
--- a/litebox/src/platform/common_providers/userspace_pointers.rs
+++ b/litebox/src/platform/common_providers/userspace_pointers.rs
@@ -101,10 +101,7 @@ impl<V, T: Clone> core::fmt::Debug for UserConstPtr<V, T> {
 ///
 /// Note that this is fallible only if recovering from exceptions (e.g., page fault or SIGSEGV)
 /// is supported.
-unsafe fn read_at_offset<'a, V: ValidateAccess, T: Clone>(
-    ptr: *const T,
-    count: isize,
-) -> Option<alloc::borrow::Cow<'a, T>> {
+unsafe fn read_at_offset<V: ValidateAccess, T: Clone>(ptr: *const T, count: isize) -> Option<T> {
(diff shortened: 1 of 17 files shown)


jaybosamiya-ms:
Opus did an _ok_ job at this change, its first set of changes were not amazing (e.g., even when I told it to do an owned/boxed slice, it _really_ wanted to make everything a `Vec`).  After a bunch of guidance, it did the right stuff.  Might have been faster if I had just done it manually from the start.

Anyways, the code as it stands is ready for review.

jaybosamiya-ms:
I had marked this one as auto-merge before because I didn't think this one needed a deep review, this is not a drastic design change (yes it is an API change, but the change is a fairly conservative change that had already been discussed).  The auto-merge failed due to today's release of new Rust Stable (1.93.0), which means there are new clippy lints on `main` that are currently failing.  I'll make a separate PR to fix those up, and then we can merge this PR after that one is merged.

wdcui:
sounds good, @jaybosamiya-ms . I agree that we don't need multiple reviews of this PR. Please merge it when it's ready. Thanks!

jaybosamiya-ms:
#588 has the fixups for the new Rust version, once that is merged, we can merge this PR

PR #591, opened 2026-01-26
LVBS: Add support for OP-TEE SMC call for TA loading/running

sangho2:
This PR add supports for OP-TEE SMC call for TA loading and running.

It includes a few more OP-TEE handlers to decode TA request information, write output back to the normal world, and maintain shared memory. Also, it implements a tentative OP-TEE SMC handler inside `litebox_runner_lvbs` to load/run a (KMPP) TA via ldelf. This includes some changes in `litebox_shim_optee` to deal with a lack of demand paging. It only supports a single TA for now. Additionally, it now maintains three extended states areas for VTL0, VTL1 kernel, and VTL1 user and use XSAVE/XSAVEOPT properly.

### DIFF
--- a/litebox_platform_lvbs/src/host/per_cpu_variables.rs
+++ b/litebox_platform_lvbs/src/host/per_cpu_variables.rs
@@ -55,7 +55,7 @@ impl PerCpuVariables {
         &raw const self.interrupt_stack as u64 + (self.interrupt_stack.len() - 1) as u64
     }
 
-    pub(crate) fn hv_vp_assist_page_as_ptr(&self) -> *const HvVpAssistPage {
+    pub fn hv_vp_assist_page_as_ptr(&self) -> *const HvVpAssistPage {
         (&raw const self.hv_vp_assist_page).cast::<HvVpAssistPage>()
     }
 
@@ -79,7 +79,7 @@ impl PerCpuVariables {
         &raw mut self.hvcall_output
     }
 
-    pub(crate) fn set_vtl_return_value(&mut self, value: u64) {
+    pub fn set_vtl_return_value(&mut self, value: u64) {
         self.vtl0_state.r8 = value; // LVBS uses R8 to return a value from VTL1 to VTL0
     }
 
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -5,8 +5,12 @@
 
 extern crate alloc;
 
-use core::panic::PanicInfo;
-use litebox::{mm::linux::PAGE_SIZE, utils::TruncateExt};
+use core::{ops::Neg, panic::PanicInfo};
+use litebox::{
+    mm::linux::PAGE_SIZE,
+    utils::{ReinterpretSignedExt, TruncateExt},
+};
+use litebox_common_linux::errno::Errno;
 use litebox_common_optee::{
     OpteeMessageCommand, OpteeMsgArgs, OpteeSmcArgs, OpteeSmcReturnCode, UteeParams,
 };
@@ -16,9 +20,10 @@ use litebox_platform_lvbs::{
     host::{bootparam::get_vtl1_memory_info, per_cpu_variables::allocate_per_cpu_variables},
     mm::MemoryProvider,
     mshv::{
-        hvcall,
+        NUM_VTLCALL_PARAMS, VsmFunction, hvcall,
+        vsm::vsm_dispatch,
         vsm_intercept::raise_vtl0_gp_fault,
-        vtl_switch::{panic_vtl_switch, vtl_switch_loop_entry},
+        vtl_switch::{vtl_switch, vtl_switch_init},
         vtl1_mem_layout::{
             VTL1_INIT_HEAP_SIZE, VTL1_INIT_HEAP_START_PAGE, VTL1_PML4E_PAGE,
             VTL1_PRE_POPULATED_MEMORY_SIZE, get_heap_start_address,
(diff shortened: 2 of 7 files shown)


litebox_platform_lvbs/src/host/per_cpu_variables.rs
```diff
@@ -276,11 +213,28 @@ pub struct PerCpuVariablesAsm {
...
+    ///   2: restored at least once - XSAVE uses xsaveopt (safe), XRSTOR executes
+    /// Reset to 0 at each VTL1 entry (OP-TEE SMC call) since returning to VTL0 invalidates CPU tracking.
+    vtl1_kernel_xsaved: Cell<u8>,
+    /// XSAVE/XRSTOR state tracking for VTL1 user (see `vtl1_kernel_xsaved` for state values and reset).
+    vtl1_user_xsaved: Cell<u8>,
 }
 
 impl PerCpuVariablesAsm {
+    /// VTL1 XSAVE mask: let XSAVE and XRSTOR deal with x87 and SSE states
+    pub const VTL1_XSAVE_MASK: u64 = 0b11;
```
wdcui: Nit: should be put the mask next to the const `VTL1_XSAVE_AREA_SIZE`?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -931,6 +931,28 @@ pub unsafe fn run_thread(
...
+/// times on the same shim instance. `run_thread` does not support this because the given
+/// shim instance is *moved* to `run_thread`.
+///
+/// CAUTION: This function doesn't work for multi-threaded processes, so it is only for
+/// OP-TEE use cases.
+///
+/// # Safety
+/// The context must be valid user context.
+#[cfg(feature = "optee_syscall")]
+pub unsafe fn run_thread_with_shim_ref(
```
wdcui: Minor: maybe it's okay to just call it `run_thread_with_shim`?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -1093,11 +1206,18 @@ unsafe extern "C" fn run_thread_arch(
         "done:",
         "mov rbp, gs:[{cur_kernel_bp_off}]",
         "mov rsp, gs:[{cur_kernel_sp_off}]",
+        XRSTOR_VTL1_ASM!({vtl1_kernel_xsave_area_off}, {vtl1_xsave_mask_lo_off}, {vtl1_xsave_mask_hi_off}, {vtl1_kernel_xsaved_off}),
```
wdcui, line 1217: Where do we restore vtl1 user extended states?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -176,9 +184,213 @@ pub fn handle_optee_msg_arg(msg_arg: &OpteeMsgArgs) -> Result<(), OpteeSmcReturn
...
+    pub session: u32,
+    pub entry_func: UteeEntryFunc,
+    pub cmd_id: u32,
+    pub params: [UteeParamOwned; UteeParamOwned::TEE_NUM_PARAMS],
+    pub out_shm_info: [Option<ShmInfo<ALIGN>>; UteeParamOwned::TEE_NUM_PARAMS],
+}
+
+/// This function decodes a TA request contained in `OpteeMsgArgs`.
+///
+/// It copies the entire parameter data from the normal world shared memory into LiteBox's memory
```
wdcui: Nit: "LiteBox's memory" => "the secure world's memory"?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -176,9 +184,213 @@ pub fn handle_optee_msg_arg(msg_arg: &OpteeMsgArgs) -> Result<(), OpteeSmcReturn
...
+                }
+            }
+            TeeParamType::MemrefOutput | TeeParamType::MemrefInout => {
+                if let Ok(Some((addr, len))) = ta_params.get_values(index) {
+                    // SAFETY
+                    // `addr` is expected to be a valid address of a TA and `addr + len` does not
+                    // exceed the TA's memory region.
+                    use litebox::platform::RawConstPointer;
+                    let ptr = crate::UserConstPtr::<u8>::from_usize(addr.truncate());
+                    let slice = unsafe { ptr.to_owned_slice(len.truncate()) }
```
wdcui, line 380: @jaybosamiya-ms is this kind of unsafe block unavoidable? Can this case leverage any of the code you are writing?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -303,3 +547,75 @@ fn shm_ref_map() -> &'static ShmRefMap<PAGE_SIZE> {
     static SHM_REF_MAP: OnceBox<ShmRefMap<PAGE_SIZE>> = OnceBox::new();
     SHM_REF_MAP.get_or_init(|| Box::new(ShmRefMap::new()))
 }
+
+/// Get the normal world shared memory information (physical addresses and page offset) from `OpteeMsgParamTmem`.
+///
+/// Note that we use this function for handing TA requests and in this context
```
wdcui: Minor: "handing" => "handling"

litebox_runner_lvbs/src/lib.rs
```diff
@@ -102,6 +115,307 @@ pub fn run(platform: Option<&'static Platform>) -> ! {
...
+    }
+}
+
+fn optee_smc_handler_entry_inner(
+    smc_args_pfn: u64,
+) -> Result<i64, litebox_common_linux::errno::Errno> {
+    let smc_args_pfn: usize = smc_args_pfn.truncate();
+    let smc_args_addr = smc_args_pfn << litebox_platform_lvbs::mshv::vtl1_mem_layout::PAGE_SHIFT;
+    match optee_smc_handler(smc_args_addr) {
+        Ok(smc_arg) => {
```
wdcui: Why is it called "smc_arg" instead of "smc_return"? It looks like some information returned from the secure world to the normal world after an smc call is handled.

litebox_platform_lvbs/src/host/per_cpu_variables.rs
sangho2: Let me do that.

litebox_platform_lvbs/src/lib.rs
sangho2: the original `run_thread` also gets `shim`. The difference is that in `run_thread`, `shim` is moved to the platform while this function borrows `shim`. let me rename this function (`run_thread_borrow_shim` or something).

litebox_platform_lvbs/src/lib.rs
sangho2: we restore vtl1 user extended states in `switch_to_user`.

litebox_shim_optee/src/msg_handler.rs
jaybosamiya-ms: My new zerocopy-based PR will make this bit of code fully safe.  I'll be opening that PR soon (EDIT: opened #592).  Please don't consider this blocked one way or the other here the change will literally just be `unsafe { ptr.to_owned_slice(len.truncate()) }` -> `ptr.to_owned_slice(len.truncate())` at this line (because it is a slice of bytes here based on the type provided to `ptr`, this will trivially satisfy the requirements, so nothing special needed here), and the Rust compiler will just remind you for this part.

litebox_runner_lvbs/src/lib.rs
sangho2: this is an SMC call with (message) arguments, and we need to run an additional handler for these arguments.

wdcui: `smc_arg` is returned to the normal world. I don't see it's passed to another handler?

sangho2: Ah. my bad. `optee_smc_handler` calls that handler internally.

I'll improve the comment. Basically, we don't return code (`OpteeSmcReturnCode`). What we return or copy back is a page containing nine register values (`OpteeSmcArgs`).

litebox_platform_lvbs/src/lib.rs
sangho2: now we have `run_thread_with_reenter`

litebox_runner_lvbs/src/lib.rs
```diff
@@ -102,6 +115,330 @@ pub fn run(platform: Option<&'static Platform>) -> ! {
...
+    }
+}
+
+fn optee_smc_handler_entry_inner(
+    smc_args_pfn: u64,
+) -> Result<i64, litebox_common_linux::errno::Errno> {
+    let smc_args_pfn: usize = smc_args_pfn.truncate();
+    let smc_args_addr = smc_args_pfn << litebox_platform_lvbs::mshv::vtl1_mem_layout::PAGE_SHIFT;
+    match optee_smc_handler(smc_args_addr) {
+        Ok(smc_arg_ret) => {
```
wdcui: Nit: call it `smc_args`?

Issue #560, opened 2025-12-16
Reduce/eliminate unsafe code in the shims


jstarks:
In order to make useful safety claims about LiteBox as a whole, I think we need to better layer our use of unsafe across the LiteBox stack. Considering the four layers:

* Platform code requires lots of unsafe, since it is integrating with OS primitives that the Rust borrow checker and even Verus cannot easily reason about.
* Runner code's use of unsafe depends on how well the associated platform manages to provide safe methods on top of the unsafe code.
* `litebox` code requires unsafe to implement mutexes and other low-level data structures... but this type of code is probably formally verifiable via Verus.
* Shim code should require no unsafe code--it should be built on top of abstractions provided by `litebox`.

This issue is most concerned with shim code. We should drive all uses of unsafe out of the shims and into `litebox`.

The biggest offender is user-mode memory access. We will need to consider #505 as part of this.

jaybosamiya-ms:
The biggest offender has been covered by #592.  I'm keeping this issue open since we'd like to cut down our use of `unsafe` further.

PR #592, opened 2026-01-28
Safely eliminating a large amount of `unsafe` for raw pointers

jaybosamiya-ms:
This PR is a major refactor of our raw pointers API in order to eliminate ~84% of the usage of `unsafe` from our Linux shim and ~38% of the usage of `unsafe` from our OPTEE shim.

It does so by switching the interfaces to use [`zerocopy`](http://docs.rs/zerocopy)-based traits in order to guarantee the validity of the values being read or written (e.g., to make sure that one does not attempt to read from a `RawConstPointer<bool>` which would be unsound because not all byte values are safe for `bool`).  We previously used to do this by keeping the interfaces `unsafe` and by _manually_ looking at the types, but this is not long-term sustainable.  `zerocopy` already has the necessary machinery to confirm the relevant validity, thus restricting things this was is _almost_ a no-brainer.

There are a few challenges however that must be overcome to properly migrate things over.  Indeed, some of our prior usage was _technically_ unsound.  This PR fixes any such detected unsoundness by tweaking the types.  There are also some issues that one hits when using `zerocopy` and this PR needs to employ a few workarounds for some such cases.  Ones where I am not the happiest with the workaround, I have explicitly annotated with a `FUTURE`/`NOTE`/similar annotation.

While a large part of this PR would become essentially repetitive refactoring _after_ all the important decisions have been made, it is somewhat impossible to decompose this PR into different steps for easier reviewing, so instead I am just documenting important decisions here that I think deserve extra attention when reviewing, and noting that reviewing each commit in this PR separately is probably easier than trying to review the whole thing at once.

Important decisions:
1. APIs for raw pointer providers now _require_ that the pointers point to a `FromBytes` and `IntoBytes` type.  These traits (provided by zerocopy) _roughly_ mean that "all byte representations are valid for this type".  See the [`zerocopy` docs](https://docs.rs/zerocopy) for a more precise notion of this.  This is the key constraint that lets us remove the `unsafe` requirement from these functions.
2. Some types required their padding to become concretized into the type, so that it is obvious that there is padding.  I have attempted to check against all the C types that exist on Linux to make sure I am maintaining the correct types, but I cannot be 100% sure despite my best efforts for some types because there sometimes was already some amount of divergence.  Find all the padding-relevant structs by looking for `__pad` fields.  I am not particularly happy about the way the `__pad` needs to be handled at call sites, but I didn't want to diverge the code too far in this PR.  Future PR can fix this up if we'd like.
3. `zerocopy` does _not_ currently support `FromBytes`/`IntoBytes` on raw `*const T`/`*mut T` types, even though it would be sound to do so, due to some provenance footguns.  I've documented this in our `userspace_pointers.rs‎`, where I have tweaked the implementation to use explicitly exposed provenance.  This can be quite subtle, but I believe is the correct move for us.  Fwiw, our usage of provenance info throughout LiteBox has stuck to the exposed-provenance model wherever we have had to deal with provenance details, so this is just roughly sticking to the lines of decisions made before, but nonetheless, it is important to note that this has to be done via the `UserConstPtr`/`UserMutPtr` types being _effectively_ abstraction wrappers on `usize` that handle the provenance at their interface methods.  Similarly with transparent pointers (sidenote: maybe we should drop transparent pointers in a future PR).
4. Multiple types needed to be marked as `repr(packed)`.  This is usually necessary in situations with significant polymorphism where it is not possible for `zerocopy` to be aware of sizes of those polymorphic types.  In this scenario, we were _previously_ leaving it as undefined sizes, but now are being explicitly precise.  There _is_ a downside to marking structs as `packed` through, in that one cannot take references to any fields directly, and must copy those fields out (to guarantee alignment requirements on refs).  This is usually not an issue (the Rust compiler will correctly complain if you forget to do this), since the fields tend to just be `u32`s/`u64`s or similar tiny sized things, but it _does_ lead to the rare `let bar = foo.bar; bar.baz()` in cases where one might use `foo.bar.baz()` previously.
5. The `bitflags` macro does not nicely support the `derive` for the zerocopy traits so we need to separate things out.  This is not a major issue, just took a while to figure out how to convince `zerocopy` and `bitflags` to work together.  Similarly, `bitfield` macro and the `derive` seem to clash a little.
6. `litebox_common_optee` had some enums that were _technically_ unsound even if the unsoundness was not triggered right now.  Adding the trait constraints exposed this possible unsoundness.  To fix this up, I've introduced a new `open_enum` macro (currently only inside `litebox_common_optee`, but we may in the future expose it more generally if we find it useful elsewhere).  Generally speaking, an `enum` can be `FromBytes` and `IntoBytes` iff the enum is "complete" in its representation (this is similar to the `bool` mentioned above).  The `open_enum` macro guarantees completeness by switching the enum to a `struct` + `const`s.  While we could (in theory) use the `open_enum` crate for this, it does not interact nicely enough with the zerocopy derives as we need it to, so instead I've rolled my own version of it.
7. `rustc` gets confused by overly recursive types when we place the bounds on it.  This seems to be only really triggered on the Linux `RobustList` and `RobustListHead` types, which are indeed _heavily_ recursive.  The implementation that uses these types itself heavily keeps converting back and forth with `usize`s, so for now, I've switched these to just have `usize`s instead of deeply recursive pointers, and placed explanations around the types.  We may wish to clean this up, but I think that would need to happen alongside cleanup of the robust list head functions themselves.  I don't think that needs to be particularly high priority though.
8. We cannot pass `unions` cleanly through raw pointers. This primarily impacts only the siginfo data, which needed some tweaks to work correctly.  Apparently, there was also an implicit dependency on padding there that also needed to be fixed up.
9. The ratchet test for `transmute` was reading comments too, which would have unnecessarily bumped up the ratchet, so I fixed the ratchet test to only count `transmute`s that are in real code, rather than stuff in comments.

---

Related:
  - #560: this PR cleans up the biggest source of `unsafe` in the shims
  - #505: discussion towards better pointer APIs

### DIFF
--- a/litebox_shim_optee/src/msg_handler.rs
+++ b/litebox_shim_optee/src/msg_handler.rs
@@ -377,7 +377,8 @@ pub fn prepare_for_return_to_normal_world(
                     // exceed the TA's memory region.
                     use litebox::platform::RawConstPointer;
                     let ptr = crate::UserConstPtr::<u8>::from_usize(addr.truncate());
-                    let slice = unsafe { ptr.to_owned_slice(len.truncate()) }
+                    let slice = ptr
+                        .to_owned_slice(len.truncate())
                         .ok_or(OpteeSmcReturnCode::EBadAddr)?;
 
                     if slice.is_empty() {


litebox/src/platform/mod.rs
```diff
@@ -376,16 +377,16 @@ pub trait DebugLogProvider {
...
-    type RawMutPointer<T: Clone>: RawMutPointer<T>;
+    type RawConstPointer<T: Clone + FromBytes>: RawConstPointer<T>;
+    type RawMutPointer<T: Clone + FromBytes + IntoBytes>: RawMutPointer<T>;
 }
 
 /// A read-only raw pointer, morally equivalent to `*const T`.
 ///
 /// See [`RawPointerProvider`] for details.
-pub trait RawConstPointer<T>: Copy + core::fmt::Debug
+pub trait RawConstPointer<T>: Copy + core::fmt::Debug + FromBytes + IntoBytes
```
wdcui, line 387: I thought we need `IntoBytes` only when it's a mut pointer?

litebox_common_linux/src/signal/mod.rs
```diff
@@ -267,27 +286,37 @@ pub const SI_DETHREAD: i32 = -7;
...
     pub data: SiginfoData,
 }
 
-#[repr(C)]
-#[derive(Clone, Copy)]
-pub union SiginfoData {
-    pub pad: [u32; 29],
-    pub addr: usize,
+#[repr(C, packed)]
+#[derive(Clone, Copy, FromBytes, IntoBytes)]
```
wdcui, line 311: How do we decide what structs need to derive `FromBytes` and `IntoBytes`?

litebox/src/platform/mod.rs
jaybosamiya-ms: This is a difference of requirement and guarantee.

- `T` (the base value) _requires_ that it is `FromBytes` (see the `where` clause in the next lines)
- `RawConstPointer<T>` (the pointer) _guarantees_ that it is both `FromBytes` and `IntoBytes`.

litebox_common_linux/src/signal/mod.rs
jaybosamiya-ms: Currently I have only added the derives for places where it was absolutely necessary (i.e., the Rust compiler would not compile without them).  Roughly speaking, it is necessary only if it is (possibly recursively) showing up somewhere inside a `T` for `Raw*Pointer<T>` type.  The `FromBytes` and `IntoBytes` derivations check and guarantee that the type is actually valid for them before succeeding to add the trait to the type.  It would not be incorrect to add them _everywhere_ where the macros succeed, but for this PR at least I wanted to keep it minimal.

jaybosamiya-ms:
> Would the rust compiler or clippy catch any unnecessary unsafe closures in the code?

Yes, that's the `unused_unsafe` lint, and is automatically caught.

litebox/src/platform/mod.rs
```diff
@@ -376,16 +377,16 @@ pub trait DebugLogProvider {
...
 where
-    T: Clone,
+    T: Clone + FromBytes,
```
jstarks: Is `Clone` necessary anymore?

Also, I wonder if these `T` bounds should be on the individual methods instead of on the trait. I am guessing that might allow fewer generic bounds in some other places, where the ability to read/write the pointer is actually incidental.

litebox/src/platform/common_providers/userspace_pointers.rs
```diff
@@ -101,13 +128,20 @@ impl<V, T: Clone> core::fmt::Debug for UserConstPtr<V, T> {
 ///
 /// Note that this is fallible only if recovering from exceptions (e.g., page fault or SIGSEGV)
 /// is supported.
-unsafe fn read_at_offset<V: ValidateAccess, T: Clone>(ptr: *const T, count: isize) -> Option<T> {
+fn read_at_offset<V: ValidateAccess, T: Clone + FromBytes>(
+    ptr: *const T,
+    count: isize,
+) -> Option<T> {
     let src = unsafe { ptr.add(usize::try_from(count).ok()?) };
```
jstarks: Soundness issue here? `wrapping_add` instead?

litebox/src/platform/mod.rs
jaybosamiya-ms: Good point, we may not need `Clone` anymore.

I'll play around a bit with the bounds-on-individual methods and see how it works out.  You're right that it can help reduce the number of needed generic bounds.

litebox/src/platform/common_providers/userspace_pointers.rs
jaybosamiya-ms: Good catch, indeed `wrapping_add` is the correct move here, will update, thanks!

litebox/src/platform/mod.rs
wdcui: Got it. Thanks!

litebox_common_linux/src/signal/mod.rs
wdcui: Makes sense. Thanks.

litebox/src/platform/mod.rs
jaybosamiya-ms: I played around with it a little bit.  Removing the `Clone` was definitely the right move, but for the rest, I am not yet sure of the ergonomics of it for usage, and I don't want to keep this PR open for too long blocked on that.  I will open a tracking issue for it.

PR #594, opened 2026-01-29
Move VTL switch to litebox_runner_lvbs

sangho2:
This PR moves the VTL switch function (specifically, its control) to the `litebox_runner_lvbs` to let the runner deal with HVCI/HEKI functions and OP-TEE calls. Its basic design is similar to that of `run_thread`. The runner calls the `vtl_switch` function in the platform. If the function encounters an event that it cannot handle (e.g., an OP-TEE SMC call), it returns to the caller (i.e., the runner). The runner handles the event and calls the function again.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vtl_switch.rs
+++ b/litebox_platform_lvbs/src/mshv/vtl_switch.rs
@@ -213,52 +213,6 @@ macro_rules! LOAD_VTL_STATE_ASM {
     };
 }
 
-/// Restore VTL0 state, return to VTL0, and save VTL0 state when VTL1 resumes.
-///
-/// This function:
-/// 1. Restores VTL0 extended states (XRSTOR) and general-purpose registers
-/// 2. Disables interrupts and returns to VTL0
-/// 3. When VTL1 resumes, saves VTL0 general-purpose registers and extended states
-/// 4. Re-enables interrupts and returns to the caller
-///
-/// # Safety
-///
-/// Must be called in a loop to perform VTL switching. The caller must save/restore
-/// VTL1's callee-saved registers around this call.
-///
-/// TODO: We must save/restore VTL1's state when there is RPC from VTL1 to VTL0 (e.g., dynamically
-/// loading OP-TEE TAs). This should use global data structures since the core which makes the RPC
-/// can be different from the core where the VTL1 is running.
-///
-/// TODO: Even if we don't have RPC from VTL1 to VTL0, we may still need to save VTL1's state for
-/// debugging purposes.
-#[unsafe(naked)]
-#[rustfmt::skip]
-pub unsafe extern "C" fn vtl_switch_asm() {
-    #[cfg(target_arch = "x86_64")]
-    core::arch::naked_asm!(
-        XRSTOR_VTL0_ASM!({vtl0_xsave_area_off}, {vtl0_xsave_mask_lo_off}, {vtl0_xsave_mask_hi_off}),
-        LOAD_VTL_STATE_ASM!({scratch_off}, {vtl0_state_top_addr_off}, {VTL_STATE_SIZE}),
-        // *** VTL0 state is restored. Return to VTL0 immediately ***
-        "cli", // disable VTL1 interrupts before returning to VTL0
-        VTL_RETURN_ASM!({vtl_ret_addr_off}),
-        // *** VTL1 resumes here regardless of the entry reason (VTL switch or intercept) ***
-        // Hyper-V restored VTL1's rip and rsp, so we're back on the original stack.
-        SAVE_VTL_STATE_ASM!({scratch_off}, {vtl0_state_top_addr_off}),
-        XSAVE_VTL0_ASM!({vtl0_xsave_area_off}, {vtl0_xsave_mask_lo_off}, {vtl0_xsave_mask_hi_off}),
-        "sti", // enable VTL1 interrupts after saving VTL0 state
-        // A pending SINT can be fired here. Our SINT handler only executes `iretq` so returns to here immediately.
-        "ret",
-        vtl_ret_addr_off = const { PerCpuVariablesAsm::vtl_return_addr_offset() },
-        scratch_off = const { PerCpuVariablesAsm::scratch_offset() },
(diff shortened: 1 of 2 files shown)


litebox_runner_lvbs/src/lib.rs
```diff
@@ -113,7 +123,34 @@ pub fn init() -> Option<&'static Platform> {
...
+    match func_id {
+        VsmFunction::Unknown => Errno::EINVAL.as_neg().into(),
+        VsmFunction::OpteeMessage => {
+            // Reset VTL1 xsaved flags before calling the OP-TEE SMC handler. The CPU's XSAVEOPT
+            // tracking is global - it only tracks one buffer at a time. At this point, the CPU's
+            // tracking might rely on VTL0's buffer (if VTL0 called XRSTOR). Thus, we shouldn't
+            // use XSAVEOPT until XRSTOR re-establishes tracking for VTL1's buffer.
+            // This is only needed for OpteeMessage which does VTL1 kernel/user xsave/xrstor in
+            // run_thread_with_reenter.
+            with_per_cpu_variables_asm(PerCpuVariablesAsm::reset_vtl1_xsaved);
```
wdcui: Overall the code changes look good to me.

For this `reset_vtl1_xsaved`, should we do it inside the platform function `switch_to_guest` (i.e., right before we go back to the guest)?

sangho2: This function is for clearing extended states tracking for VTL0. We can't do this clearing explicitly, so just make VTL1 use `xsave` like it doesn't save anything before. Ideally, we need to call it just before the first invocation of run_thread (to avoid unnecessary clearing). I am going to move it to `vtl_switch` to call it before returning to the runner (such that hide internals).

sangho2:
As expected, some comments are misleading or wrong. I'll fix them, revise some code if necessary, and mark it ready for review.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -207,132 +206,114 @@ macro_rules! LOAD_VTL_STATE_ASM {
...
+///
+/// This function must only be called after `vtl_switch_asm` has saved VTL0 state.
+/// The caller must ensure that VTL0 general-purpose registers have been saved to
+/// per-CPU variables
+fn handle_vtl_entry() -> Option<[u64; NUM_VTLCALL_PARAMS]> {
+    let reason = get_vtl_entry_reason()?;
+    match reason {
+        VtlEntryReason::VtlCall => Some(get_vtlcall_params()),
         VtlEntryReason::Interrupt => {
             vsm_handle_intercept();
```
sangho2, line 239: For now, we handle SINT/SIMP here. Unlike HVCI/HEKI functions and OP-TEE SMC calls, SINT handling is a bit more platform-specific because it controls VTL0's architectural state like injecting GP or advancing RIP via hypercalls.

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -207,132 +206,114 @@ macro_rules! LOAD_VTL_STATE_ASM {
...
+            "mov rsp, gs:[",
+            stringify!($scratch_off),
+            "]\n",
         )
     };
 }
 
-/// VTL switch loop implemented in assembly.
+/// Restore VTL0 state, return to VTL0, and save VTL0 state when VTL1 resumes.
+///
```
sangho2: remove or replace this

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -207,132 +206,114 @@ macro_rules! LOAD_VTL_STATE_ASM {
...
-        VsmFunction::OpteeMessage => {
-            // Reset VTL1 xsaved flags before calling the OP-TEE SMC handler. The CPU's XSAVEOPT
-            // tracking is global - it only tracks one buffer at a time. At this point, the CPU's
-            // tracking might rely on VTL0's buffer (if VTL0 called XRSTOR). Thus, we shouldn't
-            // use XSAVEOPT until XRSTOR re-establishes tracking for VTL1's buffer.
-            // This is only needed for OpteeMessage which does VTL1 kernel/user xsave/xrstor in
-            // run_thread_with_reenter.
-            with_per_cpu_variables_asm(PerCpuVariablesAsm::reset_vtl1_xsaved);
+/// Get the VTL call parameters from the saved VTL0 state.
+fn get_vtlcall_params() -> [u64; NUM_VTLCALL_PARAMS] {
```
sangho2, line 259: inline?

litebox_platform_lvbs/src/mshv/vtl_switch.rs
```diff
@@ -349,11 +330,57 @@ pub(crate) fn mshv_vsm_get_code_page_offsets() -> Result<(), Errno> {
...
+        unsafe {
+            #[cfg(target_arch = "x86_64")]
+            core::arch::asm!(
+                // Save VTL1's callee-saved registers (only relies on rsp)
+                "push rbx",
+                "push rbp",
+                "push r12",
+                "push r13",
+                "push r14",
+                "push r15",
```
sangho2: we could do this in `vtl_switch_asm` ("callee-saved") or include the entire `vtl_switch_asm` here.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -434,6 +461,9 @@ const TA_BINARY: &[u8] = &[0u8; 0];
 #[panic_handler]
 fn panic(info: &PanicInfo) -> ! {
     serial_println!("{}", info);
-    let _ = raise_vtl0_gp_fault();
-    unsafe { panic_vtl_switch() }
+    match raise_vtl0_gp_fault() {
+        Ok(result) => vtl_switch(Some(result.reinterpret_as_signed())),
+        Err(err) => vtl_switch(Some((err as u32).reinterpret_as_signed().neg().into())),
```
sangho2, line 466: Assumption. Once a core panics, there is no "reliable" way to return to it. Switch back to VTL0 immediately, do crash dump, and reboot the machine.

PR #632, opened 2026-02-02
Add support for running multiple TA instances with page table isolation

sangho2:
This PR adds support for running **multiple TAs** concurrently with **per-instance page table isolation** on LVBS.

  **Page Table Management**
  - `PageTableManager` manages base page table and task-specific page tables
  - Each TA instance gets its own task page table
  - Base page table contains only VTL1 kernel mappings
  
  **Session/Instance Management**
  - New `session.rs` module manages TA session/instance
  - Each instance tracks its page table ID for memory isolation
  - Sessions can share the same instance with a try-lock
  - Proper cleanup of page tables when sessions/instances are closed or TA panics

  **OP-TEE Shim Updates**
  - Update message handler and syscalls to work with per-instance page tables
  - Refactor pointer validation (`ptr.rs`) for the new isolation model
  - Enhance ELF loader integration with session-aware memory management

  **Runner Updates**
  - Significant refactoring to support multi-TA dispatch

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -94,7 +94,7 @@ pub struct PageTableManager {
     base_page_table_frame: PhysFrame<Size4KiB>,
     /// Task page tables indexed by their ID (starting from 1).
     /// Each contains kernel mappings + task-specific user-space mappings.
-    task_page_tables: spin::Mutex<HashMap<usize, mm::PageTable<PAGE_SIZE>>>,
+    task_page_tables: spin::Mutex<HashMap<usize, alloc::boxed::Box<mm::PageTable<PAGE_SIZE>>>>,
     /// Reverse lookup: physical frame -> page table ID (for O(1) CR3 lookup).
     /// Only contains task page tables (base page table is checked separately).
     frame_to_id: spin::Mutex<HashMap<PhysFrame<Size4KiB>, usize>>,
@@ -148,11 +148,15 @@ impl PageTableManager {
         if let Some(id) = task_pt_id {
             let task_pts = self.task_page_tables.lock();
             if let Some(pt) = task_pts.get(&id) {
-                // SAFETY: This page table is the current CR3, so `delete_task_page_table`
-                // will refuse to remove it (returns EBUSY). The PageTableManager itself
-                // is 'static, so the HashMap won't be deallocated. Together, these
-                // guarantee the page table remains valid for the lifetime of the reference.
-                return unsafe { &*core::ptr::from_ref(pt) };
+                // SAFETY: Three invariants guarantee this reference remains valid:
+                // 1. The PageTable is Box-allocated, so HashMap rehashing does not
+                //    move the PageTable itself (only the Box pointer moves).
+                // 2. This page table is the current CR3, so `delete_task_page_table`
+                //    will refuse to remove it (returns EBUSY).
+                // 3. The PageTableManager is 'static, so neither it nor the HashMap
+                //    will be deallocated.
+                let pt_ref: &mm::PageTable<PAGE_SIZE> = pt;
+                return unsafe { &*core::ptr::from_ref(pt_ref) };
             }
         }
 
@@ -275,6 +279,7 @@ impl PageTableManager {
             return Err(Errno::ENOMEM);
         }
 
+        let pt = alloc::boxed::Box::new(pt);
         let phys_frame = pt.get_physical_frame();
 
         let mut task_pts = self.task_page_tables.lock();
@@ -489,7 +494,9 @@ impl<Host: HostInterface> LinuxKernel<Host> {
     ///
     /// Note: VTL0 physical memory is external memory not owned by LiteBox (similar to MMIO).
     /// LiteBox accesses it by creating a temporary non-shared mapping, copying data to/from a
-    /// LiteBox-owned buffer, and unmapping immediately.
+    /// LiteBox-owned buffer, and unmapping immediately. No Rust references are created to the
+    /// mapped VTL0 memory; all accesses use raw pointer operations (read_volatile /
+    /// copy_nonoverlapping) to avoid violating Rust's aliasing model.
     fn unmap_vtl0_pages(
         &self,
         page_addr: *const u8,
@@ -612,9 +619,13 @@ impl<Host: HostInterface> LinuxKernel<Host> {
             let dst_ptr = page_addr.wrapping_add(page_offset).cast::<T>();
             assert!(dst_ptr.is_aligned(), "dst_ptr is not properly aligned");
 
-            // Safety: dst_ptr points to valid VTL0 memory that was just mapped
-            let dst = unsafe { core::slice::from_raw_parts_mut(dst_ptr, value.len()) };
-            dst.copy_from_slice(value);
+            // Safety: dst_ptr points to mapped VTL0 memory with enough space for value.len()
+            // elements. We use copy_nonoverlapping instead of creating a slice reference
+            // because VTL0 memory is external (similar to MMIO/DMA) and may be concurrently
+            // modified, which would violate Rust's aliasing model for references.
+            unsafe {
+                core::ptr::copy_nonoverlapping(value.as_ptr(), dst_ptr, value.len());
+            }
 
             assert!(
                 self.unmap_vtl0_pages(page_addr, length).is_ok(),
@@ -650,9 +661,11 @@ impl<Host: HostInterface> LinuxKernel<Host> {
             let src_ptr = page_addr.wrapping_add(page_offset).cast::<T>();
             assert!(src_ptr.is_aligned(), "src_ptr is not properly aligned");
 
-            // Safety: src_ptr points to valid VTL0 memory that was just mapped
-            let src = unsafe { core::slice::from_raw_parts(src_ptr, buf.len()) };
-            buf.copy_from_slice(src);
+            // Safety: see copy_slice_to_vtl0_phys for why we use copy_nonoverlapping
+            // instead of creating a slice reference to VTL0 memory.
+            unsafe {
+                core::ptr::copy_nonoverlapping(src_ptr, buf.as_mut_ptr(), buf.len());
+            }
 
             assert!(
                 self.unmap_vtl0_pages(page_addr, length).is_ok(),
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -18,7 +18,7 @@ use litebox_common_optee::{
     TeeResult, UteeEntryFunc, UteeParams,
 };
 use litebox_platform_lvbs::{
-    arch::{gdt, get_core_id, interrupts},
+    arch::{gdt, get_core_id, instrs::hlt_loop, interrupts},
     debug_serial_println,
     host::{bootparam::get_vtl1_memory_info, per_cpu_variables::allocate_per_cpu_variables},
     mm::MemoryProvider,
@@ -1074,5 +1074,5 @@ fn panic(info: &PanicInfo) -> ! {
     // We assume that once this VTL1 kernel panics, we don't try to resume its execution.
     // This is because, after the panic, the kernel is in an undefined state.
     // Switch back to VTL0, do crash dump, and reboot the machine.
-    unreachable!()
+    hlt_loop()
 }
--- a/litebox_shim_optee/src/session.rs
+++ b/litebox_shim_optee/src/session.rs
@@ -261,7 +261,7 @@ impl Default for SingleInstanceCache {
 
 /// Allocate a new unique session ID.
 ///
-/// Delegates to [`SessionIdPool::allocate`] for unified session ID management.
+/// Delegates to `SessionIdPool::allocate` for unified session ID management.
 /// Returns `None` if all session IDs are exhausted.
 pub fn allocate_session_id() -> Option<u32> {
     SessionIdPool::allocate()
@@ -269,7 +269,7 @@ pub fn allocate_session_id() -> Option<u32> {
 
 /// Recycle a session ID for potential future reuse.
 ///
-/// Delegates to [`SessionIdPool::recycle`].
+/// Delegates to `SessionIdPool::recycle`.
 pub fn recycle_session_id(session_id: u32) {
     SessionIdPool::recycle(session_id);
 }


wdcui:
## 🔍 PR #632 Review Summary: Multi-TA Page Table Isolation

### Critical Findings Consensus

| Category | Severity | Issue | Location |
|----------|----------|-------|----------|
| **Correctness/Security** | 🔴 Critical | Arc reference counting bugs in cleanup logic cause resource leaks | lib.rs:442, 795, 915 |
| **Correctness** | 🔴 Critical | `instance_count()` double-counts single-instance TAs | lib.rs:198 |
| **Compatibility** | 🔴 Critical | TeeOrigin enum values changed (0→1 based) - ABI break | litebox_common_optee/lib.rs:818-821 |
| **Compatibility** | 🔴 Critical | Platform thread function signatures changed (return type removed) | litebox_platform_lvbs/lib.rs:1251 |
| **Security** | 🟠 High | Race condition between Arc::strong_count check and cleanup | lib.rs:915 |
| **Security** | 🟠 High | Page table deletion while references may exist (use-after-free risk) | litebox_platform_lvbs/lib.rs:230-240 |
| **Compatibility** | 🟠 High | TeeResult::Unknown variant removed | lib.rs:931-1022 |
| **Compatibility** | 🟠 High | OpteeMsgParamFmem.offs_high changed u32→u16 | lib.rs:1310 |
| **Performance** | 🟠 High | O(N) page table lookup on every syscall | litebox_platform_lvbs/lib.rs:124-131 |
| **Performance** | 🟡 Medium | Two lock acquisitions per OpenSession for counting | lib.rs:197-199 |

---

### 📋 Consolidated Recommendations

**Must Fix (Blockers):**
1. **Fix Arc reference counting logic** - The strong_count checks are fundamentally broken. Remove items from cache *before* checking count, or use dedicated reference tracking
2. **Fix instance_count() calculation** - Currently double-counts single-instance TAs
3. **Revert TeeOrigin enum values** to 0-based to maintain OP-TEE ABI compatibility
4. **Restore TeeResult::Unknown** or make enum `#[non_exhaustive]`
5. **Restore original thread function signatures** or provide deprecation path

**Should Fix:**
- Add reverse lookup HashMap for CR3→page table ID (O(1) vs O(N))
- Cache physical frame in PageTable struct
- Add safety comments to all unsafe blocks in paging.rs
- Document TOCTOU assumptions in pointer validation

**Consider:**
- Atomic counter for instance count instead of dual-lock approach
- RwLock for session map to allow concurrent reads

---

### ✅ Positive Aspects Noted
- Proper memory isolation via separate page tables
- Clean session/instance lifecycle management design
- Proper cleanup of page tables on session close
- Try-lock pattern for preventing concurrent TA access

---

**Overall Verdict:** The PR implements valuable multi-TA isolation features, but has **critical bugs in cleanup logic** and **breaking ABI/API changes** that must be addressed before merge.

---
*Review conducted by 5 specialized agents analyzing: correctness, security, performance, compatibility, and code quality.*

litebox_shim_optee/src/session.rs
```diff
@@ -0,0 +1,350 @@
...
+//!
+//! This provides transparent retry behavior for client applications while keeping
+//! the waiting logic in normal world (where scheduling is appropriate), without
+//! requiring RPCs that would give untrusted code control over secure world execution.
+//!
+//! Reference: <https://optee.readthedocs.io/en/latest/architecture/trusted_applications.html#multi-session>
+//!
+//! ## OP-TEE OS Thread IDs and RPC
+//!
+//! In OP-TEE OS, **thread IDs** and **session IDs** serve different purposes:
```
sangho2: minor: switch thread IDs and session IDs

litebox_common_optee/src/lib.rs
```diff
@@ -1701,19 +1809,126 @@ impl From<OpteeSmcReturnCode> for litebox_common_linux::errno::Errno {
...
 
+/// Parse the `.ta_head` section from a raw ELF binary.
+///
+/// This function searches for the `.ta_head` section in the ELF and parses the `TaHead`
+/// structure from it. Returns `None` if the section is not found or cannot be parsed.
+///
+/// # Arguments
+/// * `elf_data` - Raw bytes of the ELF binary
+#[allow(clippy::cast_possible_truncation)] // ELF offsets fit in usize on 64-bit
+pub fn parse_ta_head(elf_data: &[u8]) -> Option<TaHead> {
```
wdcui, line 1801: Can we use the existing elf loader to do this kind of parsing?

litebox_common_optee/src/lib.rs
```diff
@@ -19,6 +19,28 @@ use zerocopy::{FromBytes, IntoBytes};
 
 pub mod syscall_nr;
 
+/// Maximum virtual address (exclusive) for user-space allocations.
+/// This is set to (1 << 47) - PAGE_SIZE (upper limit of 4-level paging).
+pub const USER_ADDR_MAX: usize = 0x7FFF_FFFF_F000;
```
wdcui: Why do we need these page table related definitions here?

wdcui:
@CvvT it would be very helpful to have you take a look at the low-level code changes in litebox_platform_lvbs. Thanks!

litebox_platform_lvbs/src/arch/x86/interrupts.S
```diff
@@ -75,33 +75,47 @@
...
  */
 .macro isr_no_err_code name:req handler:req
 .global \name
 \name:
     cld
-    push    0           /* Push dummy error code */
+    push    0                   /* Push dummy error code */
     push_regs
-    mov     rdi, rsp    /* Pass pointer to saved registers as first arg */
+    sub     rsp, 8              /* Align stack for call (RSP % 16 == 0) */
```
wdcui: Would it be easier to use `and` to do alignment in case `push_regs` is changed in the future?

wdcui: Why didn't it crash previously with misaligned stack pointer?

litebox_shim_optee/src/lib.rs
```diff
@@ -705,17 +746,30 @@ impl Task {
             -1,
             0,
         )?;
-        let punchthrough = litebox_common_linux::PunchthroughSyscall::SetFsBase {
-            addr: addr.as_usize(),
-        };
+        // Store TLS address for later restoration
+        self.tls_base_addr.set(addr.as_usize());
```
wdcui, line 750: Do we need to add a target_arch check here?

litebox_shim_optee/src/lib.rs
```diff
@@ -1199,50 +1286,48 @@ pub(crate) enum ThreadInitState {
...
+/// Global session ID pool.
+///
+/// This is a global singleton that manages session IDs across all shim instances.
+/// Session IDs are globally unique and can be recycled after a session is closed.
+pub struct SessionIdPool;
+
+/// Global session ID counter shared across all shim instances.
+static NEXT_SESSION_ID: AtomicU32 = AtomicU32::new(1);
+/// Global recycle pool for reusing session IDs.
+static RECYCLE_POOL: spin::mutex::SpinMutex<VecDeque<u32>> =
```
wdcui: Does OPTEE-OS do recycle?

litebox_shim_optee/src/ptr.rs
```diff
@@ -183,27 +183,13 @@ impl<T: Clone, const ALIGN: usize> PhysMutPtr<T, ALIGN> {
...
-        unsafe {
-            self.map_range(start, end, PhysPageMapPermissions::READ)?;
-        }
-        // Don't forget to call unmap() before returning to the caller
-        let Some(src) = self.base_ptr() else {
-            unsafe {
-                self.unmap()?;
-            }
-            return Err(PhysPointerError::NoMappingInfo);
+        let src = unsafe {
```
wdcui: Just to double check: is it possible to avoid this unsafe?

litebox_shim_optee/src/ptr.rs
```diff
@@ -392,10 +319,40 @@ impl<T: Clone, const ALIGN: usize> PhysMutPtr<T, ALIGN> {
                 );
             }
         }
+        Ok(())
+    }
+
+    /// Compute byte offset, map the required pages, and return the pointer.
+    ///
+    /// `count` is the element index, `size` is the total byte size to map.
```
wdcui: element of what?

litebox_shim_optee/src/ptr.rs
```diff
@@ -443,27 +398,26 @@ impl<T: Clone, const ALIGN: usize> PhysMutPtr<T, ALIGN> {
...
-        Some(map_info.base.wrapping_add(self.offset).cast::<T>())
-    }
 }
 
+/// Ensures physical pages are unmapped when `PhysMutPtr` goes out of scope.
+///
+/// This type is designed for single-use access: create, read/write once, then drop.
+/// The `Drop` implementation guarantees that mapped pages are always released,
+/// preventing resource leaks and adhering to the "minimize persistent mapping"
+/// security principle. Errors during unmapping are silently ignored since we
```
wdcui, line 451: Should we panic in debug mode to catch this error?

litebox_shim_optee/src/session.rs
```diff
@@ -0,0 +1,375 @@
...
+//! on the same TA instance concurrently, we use `try_lock()` and return
+//! `OPTEE_SMC_RETURN_ETHREAD_LIMIT` at the SMC level if the lock is held.
+//!
+//! ### Difference from OP-TEE OS
+//!
+//! OP-TEE OS uses RPC-based waiting: when a TA is busy, it returns to normal world
+//! via `mutex_lock()` issuing an RPC, allowing the Linux kernel to schedule other
+//! work while waiting. This is efficient but potentially insecure because:
+//! - Normal world is untrusted and could delay or never resume the waiting thread
+//! - It creates a denial-of-service vector where normal world controls secure world scheduling
```
wdcui: Normal world can always do denial of service, right?

litebox_shim_optee/src/session.rs
```diff
@@ -0,0 +1,375 @@
...
+//!
+//! In OP-TEE OS, multiple threads can have pending operations on a session, but only one
+//! **executes** at a time—others wait via RPC (suspended in normal world).
+//!
+//! ### LiteBox Status (TODO)
+//!
+//! - We return `ETHREAD_LIMIT` instead of RPC-based waiting
+//! - RPC needed for secure storage—will require:
+//!   - Saving CPU context (registers, stack, etc.) when suspending for RPC
+//!   - Indexing saved contexts by an identifier (passed via `a3-a7`)
```
wdcui, line 90: This part is similar to the OPTEE OS, right? It's potentially vulnerable to the manipulation of these values by the normal world?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -233,225 +259,796 @@ static PERSISTENT_TA: SpinMutex<Option<TaState>> = SpinMutex::new(None);
...
+    if ta_req_info.entry_func != UteeEntryFunc::OpenSession {
+        return Err(OpteeSmcReturnCode::EBadCmd);
+    }
+
+    let ta_uuid = ta_req_info.uuid.ok_or(OpteeSmcReturnCode::EBadCmd)?;
+    let client_identity = ta_req_info.client_identity;
+    let params = &ta_req_info.params;
+
+    // Try to reuse existing single-instance TA, or create a new instance
+    // If the TA is busy (lock held), return EThreadLimit - driver will wait and retry
```
wdcui: Move this comment down to `open_session_single_instance`?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -233,225 +259,796 @@ static PERSISTENT_TA: SpinMutex<Option<TaState>> = SpinMutex::new(None);
...
+
+    // Check if this was the last session using the TA instance by counting
+    // remaining sessions that reference this instance.
+    let remaining_sessions = session_manager()
+        .sessions()
+        .count_sessions_for_instance(&instance_arc);
+    let is_last_session = remaining_sessions == 0;
+
+    // If this was the last session using the TA instance, clean up (unless keep_alive is set)
+    if is_last_session {
```
wdcui: Can we check if `remaining_session == 0` directly here?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -50,19 +48,291 @@ pub mod mm;
...
+    /// - The target page table contains valid mappings for all memory that will be accessed
+    ///   after the switch (including the code being executed and stack)
+    /// - No references to the previous address space's memory are held across the switch
+    ///
+    /// # Returns
+    ///
+    /// `Ok(())` if the switch was successful, or `Err(Errno::ENOENT)` if the
+    /// specified page table ID does not exist.
+    pub unsafe fn load_task(&self, task_pt_id: usize) -> Result<(), Errno> {
+        if task_pt_id == BASE_PAGE_TABLE_ID {
```
wdcui, line 234: Should we change the code to require the callers to always call `load_base` (so it's more explicit)?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -50,19 +48,291 @@ pub mod mm;
...
+    ///
+    /// # Arguments
+    ///
+    /// * `vtl1_phys_frame_range` - The physical frame range of VTL1 kernel memory to map
+    ///
+    /// # Returns
+    ///
+    /// The ID of the newly created task page table, or `Err(Errno::ENOMEM)` if
+    /// allocation fails or the ID space is exhausted.
+    pub fn create_task_page_table(
```
wdcui, line 261: Does our page fault handler support dynamic loading?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -373,24 +654,90 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
-        let pt = unsafe { mm::PageTable::new_top_level() };
-        if pt
-            .map_phys_frame_range(
-                self.vtl1_phys_frame_range,
-                PageTableFlags::PRESENT | PageTableFlags::WRITABLE,
-            )
-            .is_err()
-        {
-            panic!("Failed to map VTL1 physical memory");
+    /// Create a new task page table for VTL1 user space and returns its ID.
```
wdcui, line 681: We have similar comments above. Do we need them in both places?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -373,24 +654,90 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
+    /// - `Ok(())` if successful
+    /// - `Err(Errno::EINVAL)` if the page table is the base page table
+    /// - `Err(Errno::ENOENT)` if the page table doesn't exist
+    /// - `Err(Errno::EBUSY)` if the page table is currently active
+    pub unsafe fn delete_task_page_table(&self, task_pt_id: usize) -> Result<(), Errno> {
+        // Safety: caller guarantees no dangling references
+        unsafe { self.page_table_manager.delete_task_page_table(task_pt_id) }
+    }
+
+    /// Switches the current page table to the specified one.
```
wdcui: Nit: "Switch to"?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -922,7 +1290,7 @@ impl<Host: HostInterface, const ALIGN: usize> VmapManager<ALIGN> for LinuxKernel
 ///
 /// # Safety
 /// The context must be valid user context.
-pub unsafe fn run_thread<T>(shim: T, ctx: &mut litebox_common_linux::PtRegs) -> T
+pub unsafe fn run_thread<T>(shim: T, ctx: &mut litebox_common_linux::PtRegs)
```
wdcui, line 1306: If run_thread is not being called anymore, should we remove it?

wdcui (commented):
Thank you for this big code change. I left some comments below.

litebox_common_optee/src/lib.rs
```diff
@@ -19,6 +19,28 @@ use zerocopy::{FromBytes, IntoBytes};
...
+/// This separation allows easy identification during cleanup and supports
+/// future designs where kernel VAs may be in higher addresses.
+pub const USER_ADDR_MIN: usize = USER_ADDR_MAX - USER_ADDR_RANGE_SIZE;
+
+/// Default low address for loading TA binaries.
+///
+/// This must be >= `USER_ADDR_MIN` because user memory is mapped in the
+/// range [`USER_ADDR_MIN`, `USER_ADDR_MAX`) for easy identification during cleanup.
+/// The binary grows upwards from this address.
+pub const TA_DEFAULT_LOW_ADDR: usize = USER_ADDR_MIN + 0x1000usize;
```
praveen-pk: I understand the need to keep this above `USER_ADDR_MIN`. I am curious why the 0x1000 hole here? what is expected to occupy this space?

litebox_common_optee/src/lib.rs
```diff
@@ -571,33 +593,99 @@ impl TeeUuid {
...
+
+    /// Returns true if multiple sessions can share the TA instance.
+    ///
+    /// Note: This flag is only meaningful when `SINGLE_INSTANCE` is also set.
+    /// For non-single-instance TAs, each session gets its own instance anyway.
+    pub fn is_multi_session(&self) -> bool {
+        self.contains(TaFlags::MULTI_SESSION)
+    }
+
+    /// Returns true if the TA instance should persist after all sessions close.
```
praveen-pk, line 640: Does `TA instance` here mean, the full memory footprint of the TA within VTL1?

Or is limited to TA state without any session/memory information?

litebox_common_optee/src/lib.rs
sangho2: These specify TA's address range. Specifically, we need to define `TA_DEFAULT_LOW_ADDR` for `ElfLoader` which depends on `USER_ADDR_MIN`. A sort of negotiation between platform and shim is needed to define the user address range.

litebox_common_optee/src/lib.rs
sangho2: Let me try it.

litebox_platform_lvbs/src/arch/x86/interrupts.S
sangho2: Let me revise it. What I observed is that, oftentimes, CPU just runs a function even if the stack is not aligned with some weird states and crashes in the end.

litebox_platform_lvbs/src/lib.rs
sangho2: Agree with you. Let me fix it (return an error).

litebox_platform_lvbs/src/lib.rs
sangho2: It does but for the LVBS platform, we don't use demand paging (yet). Currently, we pre-allocate pages.

litebox_platform_lvbs/src/lib.rs
sangho2: Let me remove this and reference the above one.

litebox_platform_lvbs/src/lib.rs
sangho2: run_thread is for Linux shim. We can remove it from the LVBS platform because we don't support Linux shim (yet). But we need to add it again if we support Linux shim (in the future).

litebox_runner_lvbs/src/lib.rs
sangho2: Let me revise it.

litebox_shim_optee/src/lib.rs
sangho2: Let me add it.

litebox_shim_optee/src/lib.rs
sangho2: It does because session IDs are 32-bit integers. I plan to revise this recycling a bit to improve its randomness. Currently, session IDs are too frequently reused.

litebox_shim_optee/src/ptr.rs
sangho2: We need to closely revisit the physical pointer design/implementation along with the Rust memory model. Better to assume it unsafe before doing the analysis.

litebox_shim_optee/src/ptr.rs
sangho2: slice: [T]. Let me add more description.

litebox_shim_optee/src/ptr.rs
sangho2: let me add debug_assert.

litebox_shim_optee/src/session.rs
sangho2: You're right. Let me revise this. It is mostly about let's not rely on the normal world if possible.

litebox_shim_optee/src/session.rs
sangho2: Yes. OP-TEE OS stores plaintext info in `a3-a7` while assuming that the normal world will not touch it.
Instead, what we can do is storing an encrypted identifier with AES-GCM or some other AEAD (we do have 40 bytes storage, so it's doable). This allows us to detect manipulation and replay attacks.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -314,30 +314,39 @@ impl litebox::platform::Provider for LinuxUserland {}
...
 /// # Safety
 /// The context must be valid guest context.
-pub unsafe fn run_thread<T>(shim: T, ctx: &mut litebox_common_linux::PtRegs) -> T
+pub unsafe fn run_thread<T>(shim: T, ctx: &mut litebox_common_linux::PtRegs)
 where
     T: litebox::shim::EnterShim<ExecutionContext = litebox_common_linux::PtRegs>,
 {
     run_thread_inner(&shim, ctx, false);
-    shim
 }
```
CvvT, line 322: Should we just remove this and always use `run_thread_ref`?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -314,30 +314,39 @@ impl litebox::platform::Provider for LinuxUserland {}
...
 
-/// Re-enters a guest thread using the provided shim and the given initial context.
+/// Run a guest thread using a reference to the shim.
 ///
-/// This will run until the thread terminates or returns.
+/// Unlike `run_thread`, this version takes a reference instead of ownership,
+/// avoiding struct moves that could invalidate internal state.
```
CvvT, line 327: What state becomes invalid?

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -145,6 +154,132 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
...
+    ///    (because each user page table has its own PT frame allocations,
+    ///    including for the kernel identity mapping)
+    ///
+    /// Kernel data frames (physical memory) are NOT deallocated - only their
+    /// page table entries are cleaned up.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that no references to the unmapped pages exist.
+    pub(crate) unsafe fn cleanup_user_mappings(&self, user_addr_min: usize, user_addr_max: usize) {
```
CvvT, line 173: We could probably simply this by calling `unmap_pages` to deallocate all data frames and [`clean_up`](https://github.com/rust-osdev/x86_64/blob/2c86471e77ec379874ece167ae2014759665b8a6/src/structures/paging/mapper/mapped_page_table.rs#L566) for table frames.

litebox_platform_lvbs/src/arch/x86/interrupts.S
CvvT: Maybe use rbp?
```
mov rbp, rsp
and rsp, -16
call ...
mov rsp, rbp
```

litebox_platform_lvbs/src/lib.rs
```diff
@@ -50,19 +48,291 @@ pub mod mm;
...
+            if let Some(pt) = task_pts.get(&id) {
+                // Safety: The page table won't be removed while it's the current CR3,
+                // and the PageTableManager lives for 'static.
+                // We extend the lifetime here because the lock prevents removal.
+                return unsafe { &*core::ptr::from_ref(pt) };
+            }
+        }
+
+        // CR3 doesn't match any known page table - this shouldn't happen
+        panic!(
```
CvvT: `unreachable!` instead?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -50,19 +48,291 @@ pub mod mm;
...
+            let frame_to_id = self.frame_to_id.lock();
+            frame_to_id.get(&cr3_frame).copied()
+        };
+
+        if let Some(id) = task_pt_id {
+            let task_pts = self.task_page_tables.lock();
+            if let Some(pt) = task_pts.get(&id) {
+                // Safety: The page table won't be removed while it's the current CR3,
+                // and the PageTableManager lives for 'static.
+                // We extend the lifetime here because the lock prevents removal.
```
CvvT: I didn't follow this. You mean the lock at the line 136? which is dropped after this function returns.

litebox_platform_lvbs/src/lib.rs
```diff
@@ -50,19 +48,291 @@ pub mod mm;
...
+            return BASE_PAGE_TABLE_ID;
+        }
+
+        let frame_to_id = self.frame_to_id.lock();
+        if let Some(&id) = frame_to_id.get(&cr3_frame) {
+            return id;
+        }
+
+        // CR3 doesn't match any known page table - this shouldn't happen
+        panic!(
```
CvvT: similar

CvvT (commented):
The page table manager seems to be useful for SNP as well. Current implementation of SNP runner can only run one program because it can only set one page table once. I'm considering having a common crate to share it.

litebox_shim_optee/src/ptr.rs
wdcui: @jaybosamiya-ms What's your thought about this issue? We may not need to solve it in this PR, but it would be good to put this on your radar.

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -341,26 +349,43 @@ fn build_memref_inout(
...
+        msg_args.session = session_id;
+    }
+
+    let Some(ta_params) = ta_params else {
+        return Ok(());
+    };
+    let Some(ta_req_info) = ta_req_info else {
+        return Ok(());
+    };
     for index in 0..UteeParams::TEE_NUM_PARAMS {
```
praveen-pk, line 381: Form what I understand the `params` within OpteeMsgArgs varies by request. The number of params in the incoming requested is captured in `num_params`.

This loop seems to run `TEE_NUM_PARAMS` number of times, irrespective of the incoming number of params. Couldn't this run over and cause a violation in VTL1?

litebox_common_optee/src/lib.rs
sangho2: AFAIU, it would be code + data (heap) except for stacks (no session). A keep alive TA should be always in memory (it is expected to be re-loaded if it crashes).

litebox_common_optee/src/lib.rs
sangho2: this is about whether it is inclusive or exclusive.

litebox_shim_optee/src/msg_handler.rs
sangho2: It doesn't. This is because we copy a fix amount of data from VTL0 to VTL1. We might read garbage data, but there is no fault. Regarding `num_params` we can't trust it anyhow (https://nvd.nist.gov/vuln/detail/CVE-2022-46152). However, we should compare these two and select smaller one. Let me fix it.

litebox_shim_optee/src/ptr.rs
jaybosamiya-ms: Cannot be resolved in this PR. I agree with Sangho, should be discussed separately, and a close revisit is needed.  This new `unsafe` is due to a function introduced by this PR itself, and that one is marked `unsafe` (correctly) because it _on its own_ cannot guarantee the requirements of its own callees (I have not explored deeper to check if the callees of _that_ can be eliminated).

litebox_common_optee/src/lib.rs
sangho2: now use the elf crate.

litebox_common_optee/src/lib.rs
sangho2: reverted this change based on our discussion.

litebox_platform_lvbs/src/arch/x86/interrupts.S
sangho2: now use the conventional approach as Weiteng suggested.

litebox_shim_optee/src/lib.rs
sangho2: ah. in fact, we don't need to. this function is gated.

litebox_shim_optee/src/lib.rs
sangho2: I rewrote session id recycling, based on LInux kernel's pidmap (old one. recent one is overkill).

litebox_shim_optee/src/ptr.rs
sangho2: revised the comment.

litebox_shim_optee/src/ptr.rs
sangho2: added

litebox_shim_optee/src/session.rs
sangho2: removed DoS.

litebox_shim_optee/src/session.rs
sangho2: briefly mentioned authenticated encryption to deal with this attack.

litebox_runner_lvbs/src/lib.rs
sangho2: done

litebox_runner_lvbs/src/lib.rs
sangho2: revised

litebox_platform_lvbs/src/lib.rs
sangho2: fixed

litebox_platform_lvbs/src/lib.rs
sangho2: Issue #637

litebox_platform_lvbs/src/lib.rs
sangho2: done

litebox_platform_lvbs/src/lib.rs
sangho2: done

litebox_platform_lvbs/src/lib.rs
sangho2: As discussed, we'll maintain it.

litebox_platform_linux_userland/src/lib.rs
sangho2: As mentioned above, this is a placeholder for future LinuxShim support.

litebox_platform_linux_userland/src/lib.rs
sangho2: Fixed. This comment was for an intermediate version of `run_thread`. I tried to copy in and copy out shim, which result in different internal addresses.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: Revised.

litebox_platform_lvbs/src/lib.rs
sangho2: done

litebox_platform_lvbs/src/lib.rs
sangho2: done

litebox_platform_lvbs/src/lib.rs
sangho2: you are right. Revised the comment.

litebox_shim_optee/src/msg_handler.rs
sangho2: turns out that, we already check `num_params` (Line 253). Basically, we do initialize unspecified ones with `None`.

litebox_platform_linux_userland/src/lib.rs
CvvT: Why LinuxShim needs this function? We can update LinuxShim to use `run_thread_ref` instead?

sangho2: I tested `run_thread_ref` with linux_shim on linux_userland weeks ago. Multi-threaded test cases hanged. Didn't take a look at in detail, so don't know whether this is an easy fix or not.

litebox_platform_lvbs/src/lib.rs
```diff
@@ -190,13 +475,18 @@ impl<Host: HostInterface> LinuxKernel<Host> {
...
+                .current_page_table()
+                .map_phys_frame_range(frame_range, flags)?,
             usize::try_from(frame_range.len()).unwrap() * PAGE_SIZE,
         ))
     }
 
     /// This unmaps VTL0 pages from the page table. Allocator does not allocate frames
     /// for VTL0 pages (i.e., it is always shared mapping), so it must not attempt to deallocate them.
+    ///
+    /// Note: VTL0 physical memory is external/remote memory that this Rust binary doesn't own,
```
sangho2: and we don't create a reference to it

litebox_shim_optee/src/msg_handler.rs
praveen-pk: >This is because we copy a fix amount of data from VTL0 to VTL1. 
In the case SW(Secure World) advertises support for RPC params, which litebox does, the NW places `rpc_arg` right next to `optee_msg_arg` in memory. This copying of fixed amount of data will corrupt the contents of `rpc_arg`.

It is better to write to min(num_args, TEE_NUM_PARAMS) params here.

Lastly, line 253, seems to use `num_params`, but it does not validate it to be less than `TEE_NUM_PARAMS`.

litebox_common_optee/src/lib.rs
praveen-pk: > this is about whether it is inclusive or exclusive.

Could you please explain this? I don't follow.

sangho2: get confused. + 0x1000 is not needed. anyhow, this is old one and we decided to use constant for now.

litebox_shim_optee/src/msg_handler.rs
sangho2: doable, but there is no huge perf gain I'd say.

`param.iter().take(num_params)` takes `min(num_params, UteeParamOwned::TEE_NUM_PARAMS)` elements.

current design is like, we don't care `num_params` and always assume there are a fixed number of parameters (4) and unused ones should be `None`. This aligns with TA side sanity check: https://github.com/OP-TEE/optee_test/blob/b27648ea8472cceceb8dda368a965c709066f7aa/ta/crypt/ta_entry.c#L300 . TAs don't get `num_params`.

litebox_platform_linux_userland/src/lib.rs
CvvT: I see. I will take a look. We can keep it for now.

litebox_shim_optee/src/msg_handler.rs
sangho2: Created issue #641

litebox_shim_optee/src/ptr.rs
wdcui: Thank you, Jay. Should we create an issue to track this?

litebox_shim_optee/src/lib.rs
wdcui: The latest fix looks good to me. Thanks!

judithseefeldt719-a11y:
``

PR #634, opened 2026-02-03
fs: Add support for O_APPEND flag in open()

wdcui:
Implement the O_APPEND flag for the filesystem, which causes all writes to append to the end of the file regardless of the current file position.

Changes:
- in_mem: Add append_mode field to Descriptor::File, update write() to seek to end before writing when in append mode. pwrite (explicit offset) ignores append mode per POSIX.
- layered: Accept O_APPEND flag and delegate to underlying filesystem
- tar_ro: Accept O_APPEND flag (read-only, so no behavioral change)
- devices: Strip O_APPEND flag (not applicable to device files)
- tests: Add comprehensive tests for append behavior

### DIFF
--- a/litebox/src/fs/devices.rs
+++ b/litebox/src/fs/devices.rs
@@ -160,7 +160,7 @@ impl<
         let flags = flags - OFlags::DIRECTORY;
         let nonblocking = flags.contains(OFlags::NONBLOCK);
         let flags = flags - OFlags::NONBLOCK;
-        // ignore NOCTTY, NOFOLLOW, and APPEND (APPEND doesn't apply to device files)
+        // ignore NOCTTY, NOFOLLOW, and APPEND
         let flags = flags - OFlags::NOCTTY - OFlags::NOFOLLOW - OFlags::APPEND;
         let truncate = flags.contains(OFlags::TRUNC);
         let flags = flags - OFlags::TRUNC;
--- a/litebox/src/fs/tests.rs
+++ b/litebox/src/fs/tests.rs
@@ -715,45 +715,6 @@ mod in_mem {
     }
 
     #[test]
-    fn o_append_flag_multiple_writes() {
-        let litebox = LiteBox::new(MockPlatform::new());
-        let mut fs = in_mem::FileSystem::new(&litebox);
-
-        fs.with_root_privileges(|fs| {
-            fs.chmod("/", Mode::RWXU | Mode::RWXG | Mode::RWXO)
-                .expect("Failed to chmod /");
-        });
-
-        // Create a file with O_APPEND from the start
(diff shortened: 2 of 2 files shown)


litebox/src/fs/devices.rs
```diff
@@ -160,8 +160,8 @@ impl<
         let flags = flags - OFlags::DIRECTORY;
         let nonblocking = flags.contains(OFlags::NONBLOCK);
         let flags = flags - OFlags::NONBLOCK;
-        // ignore NOCTTY and NOFOLLOW
-        let flags = flags - OFlags::NOCTTY - OFlags::NOFOLLOW;
+        // ignore NOCTTY, NOFOLLOW, and APPEND (APPEND doesn't apply to device files)
```
jaybosamiya-ms: If we have an explanation for APPEND, we should have an explanation for the others too, no?

litebox/src/fs/tests.rs
```diff
@@ -623,6 +623,278 @@ mod in_mem {
...
+        let bytes_read = fs
+            .read(&fd, &mut buffer, None)
+            .expect("Failed to read from file");
+        assert_eq!(bytes_read, 9);
+        assert_eq!(&buffer[..bytes_read], b"ABCDEF123");
+        fs.close(&fd).expect("Failed to close file");
+    }
+
+    #[test]
+    fn o_append_flag_multiple_writes() {
```
jaybosamiya-ms: This is not carrying any particularly extra defining power over `o_append_flag_basic` (if we just did two writes there), can be merged

litebox/src/fs/tests.rs
```diff
@@ -1649,6 +1921,91 @@ mod layered {
             Err(RmdirError::NotADirectory)
         ));
     }
+
+    #[test]
+    fn o_append_flag_with_migration() {
```
jaybosamiya-ms: Are either of the two layered FS tests teaching us anything new about the append code?  The layered FS changes are purely just forwarding things through, so we _already_ know that migration and upper-layer and such already work, so we don't need to add these tests that do not test a different edge case, but slow down our CI by running unnecessary stuff.

jaybosamiya-ms (approved):
Generally seems good, a few minor comments to clean things up before merging, thanks!

litebox/src/fs/devices.rs
wdcui: Good point. Removed.

litebox/src/fs/tests.rs
wdcui: Removed.

litebox/src/fs/tests.rs
wdcui: Removed.

PR #635, opened 2026-02-03
Fix SNP runner

CvvT:
Address some remaining TODOs in `litebox_runner_snp`:

1. Add new [host interface](https://github.com/MSRSSP/sandbox_driver/blob/9dadee243433faeac58dcb0794f7737374c0ae5c/src/snp-vmpl.c#L659) to support `ThreadProvider`
2. Add new [host interface](https://github.com/MSRSSP/sandbox_driver/blob/9dadee243433faeac58dcb0794f7737374c0ae5c/src/snp-vmpl.c#L656) to support loading a file from host to `in_mem` fs. Because kernel cannot allocate memory larger than 4MB, this only works for small files. Next, I will implement the 9p fs to address it.
3. Support fallible memory copy from user to kernel
4. Fix some bugs like stack alignment.
5. Fix `SystemInfoProvider`

This needs to work with a specific branch of [sandbox_driver](https://github.com/MSRSSP/sandbox_driver/tree/weiteng/snp), which still has a lot of legacy code I haven't cleaned up.

### DIFF
--- a/litebox_runner_snp/src/main.rs
+++ b/litebox_runner_snp/src/main.rs
@@ -113,7 +113,9 @@ pub extern "C" fn sandbox_kernel_init(
     litebox_platform_linux_kernel::host::snp::snp_impl::HostSnpInterface::return_to_host();
 }
 
-/// Pre-defined file paths to load from host
+/// Pre-defined file paths to load from host for testing purposes.
+/// Will remove this after we migrate 9p fs from VSBox to LiteBox, as we can directly access host files then.
+#[cfg(debug_assertions)]
 const HOST_FILE_PATHS: &[&str] = &[
     "/out/hello",
     "/out/efault",
@@ -122,6 +124,9 @@ const HOST_FILE_PATHS: &[&str] = &[
 ];
 
 /// Load pre-set files from host into the in-memory filesystem
+///
+/// This is a temporary solution before we migrate 9p fs from VSBox to LiteBox.
+#[cfg(debug_assertions)]
 fn load_host_files_into_fs<Platform: litebox::sync::RawSyncPrimitivesProvider>(
     in_mem_fs: &mut litebox::fs::in_mem::FileSystem<Platform>,
 ) {
@@ -190,9 +195,18 @@ pub extern "C" fn sandbox_process_init(
     litebox_platform_multiplex::set_platform(platform);
     let mut shim_builder = litebox_shim_linux::LinuxShimBuilder::new();
     let litebox = shim_builder.litebox();
-    let mut in_mem_fs = litebox::fs::in_mem::FileSystem::new(litebox);
-    load_host_files_into_fs(&mut in_mem_fs);
-
+    let in_mem_fs = {
+        #[cfg(debug_assertions)]
+        {
+            let mut fs = litebox::fs::in_mem::FileSystem::new(litebox);
+            load_host_files_into_fs(&mut fs);
+            fs
+        }
+        #[cfg(not(debug_assertions))]
+        {
+            litebox::fs::in_mem::FileSystem::new(litebox)
+        }
+    };
     let tar_ro =
         litebox::fs::tar_ro::FileSystem::new(litebox, litebox::fs::tar_ro::EMPTY_TAR_FILE.into());
     shim_builder.set_fs(shim_builder.default_fs(in_mem_fs, tar_ro));


litebox_common_linux/src/loader.rs
```diff
@@ -209,6 +209,11 @@ impl ElfParsedFile {
         file: &mut F,
         syscall_entry_point: usize,
     ) -> Result<(), ElfParseError<F::Error>> {
+        if syscall_entry_point == 0 {
+            // No syscall entry point; This may happen when running in kernel mode
+            // where we don't need a trampoline.
```
sangho2: This comment is a bit misleading. We invoke parse_trampoline like `parsed.parse_trampoline(&mut &file, task.global.platform.get_syscall_entry_point())?`. In that sense, syscall_entry_point is zero if we haven't set up a syscall entry point in the platform.

CvvT:
> **1. Missing Bounds Check on Untrusted Host Return Value (HIGH)**
> 
> * **File:** `litebox_platform_linux_kernel/src/host/snp/snp_impl.rs` - `load_file_from_host()`
> * The `bytes_read` value from the untrusted host is used without validation before slicing the buffer
> * A malicious host could return a value larger than `CHUNK_SIZE` (4096), causing an out-of-bounds access
> * This violates SNP's security model where the VM should be protected from a malicious host
> * **Suggested fix:**

Host is trusted. I added an assertion anyway.

CvvT:
> **2. File Size Check After Allocation (Medium)**
> 
> * The `MAX_FILE_SIZE` limit is enforced after appending data, allowing allocation up to 4MB + 4KB
> * **Suggested fix:** Check `(result.len() + bytes_read) > MAX_FILE_SIZE` BEFORE `extend_from_slice()`

This makes sense.

litebox_common_linux/src/loader.rs
CvvT: I updated the comment. For platform that does not need trampoline, we should probably return None instead of 0. I will have a separate PR for it.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
```diff
@@ -338,7 +338,71 @@ pub struct SyscallN<const N: usize, const ID: u32> {
     args: [u64; N],
 }
 
+/// Page-aligned buffer for reading file chunks from host
+#[repr(C, align(4096))]
+struct PageAlignedBuffer([u8; 4096]);
+
+/// Maximum file size that can be loaded (4MB)
```
wdcui, line 345: What's this file for?

litebox_runner_snp/src/main.rs
```diff
@@ -95,7 +113,66 @@ pub extern "C" fn sandbox_kernel_init(
     litebox_platform_linux_kernel::host::snp::snp_impl::HostSnpInterface::return_to_host();
 }
 
-const ROOTFS: &[u8] = include_bytes!("./test.tar");
+/// Pre-defined file paths to load from host
+const HOST_FILE_PATHS: &[&str] = &[
+    "/out/hello",
```
wdcui, line 120: What are these files for?

litebox_runner_snp/src/main.rs
```diff
@@ -95,7 +113,66 @@ pub extern "C" fn sandbox_kernel_init(
...
+    "/out/efault",
+    "/out/thread_exit",
+    // Add more paths as needed
+];
+
+/// Load pre-set files from host into the in-memory filesystem
+fn load_host_files_into_fs<Platform: litebox::sync::RawSyncPrimitivesProvider>(
```
wdcui, line 130: Why do we need to do this?

litebox_runner_snp/src/main.rs
```diff
@@ -154,7 +234,17 @@ pub extern "C" fn sandbox_process_init(
         );
     };
     let shim = shim_builder.build();
-    let program = match shim.load_program(platform.init_task(boot_params), &program, argv, envp) {
+    unsafe { SHIM = Some(shim) };
```
wdcui, line 251: Why do we need to do this?

litebox_runner_snp/src/main.rs
CvvT: Previously we just embedded all test programs with code using `include_bytes`, but the problem was that if kernel becomes larger than 4MB; we cannot allocate physically contiguous memory to load it. I had a workaround that export some internal Linux kernel to use some special memory region, but I don't want to modify Linux. This PR fixes it by loading file one by one from the host so that it only requires each file is smaller than 4MB. Still not perfect. My next step is to migrate 9p from VSBox to LiteBox.

litebox_runner_snp/src/main.rs
CvvT: For testing, will be removed one we have 9p fs.

litebox_platform_linux_kernel/src/host/snp/snp_impl.rs
CvvT: See comments below.

litebox_runner_snp/src/main.rs
CvvT: We need to set up SHIM once. Loading a program may trigger page faults which requires access to SHIM, so we need to set SHIM before loading the program.

litebox_runner_snp/src/main.rs
wdcui: It would be helpful to update the comment so it's clear these files are currently for tests and a future plan is to make them accessible through 9p.

Is it possible to wrap the code related to testing under `#[test]`?

litebox_runner_snp/src/main.rs
wdcui: We got rid of the global variable for shim in Linux userland, but we need it here?

CvvT: In Linux userland, shim is saved on per-thread stack. In SNP, we save it in TLS and this global variable. It is possible to get rid of this global variable. I will try to fix it in the next PR.

litebox_runner_snp/src/main.rs
CvvT: Updated. `#[test]` is for rust unit test, not end-to-end testing. I used `#[cfg(debug_assertion)]` here.

litebox_runner_snp/src/main.rs
wdcui: If this fix doesn't need to be done soon, you can create an issue to track it.

PR #636, opened 2026-02-04
Remove section header insertion from syscall rewriter

wdcui:
Instead of inserting a `.trampolineLB0` section header into the ELF file, store trampoline metadata directly at the end of the ELF file. This fixes issues when there is no space in sections or symbol tables. A future PR will directly patch the ELF file without invoking any other crates to reassemble the ELF file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1121,6 +1121,7 @@ dependencies = [
  "similar",
  "tempfile",
  "thiserror",
+ "zerocopy",
 ]
 
 [[package]]
--- a/litebox_syscall_rewriter/Cargo.toml
+++ b/litebox_syscall_rewriter/Cargo.toml
@@ -10,6 +10,7 @@ iced-x86 = "1.21"
 memmap2 = "0.9"
 object = { version = "0.36.7", default-features = false, features = ["elf", "build", "std"] }
 thiserror = { version = "2.0.6", default-features = false }
+zerocopy = { version = "0.8", features = ["derive"] }
 
 [lints]
 workspace = true
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -17,6 +17,7 @@
 use std::collections::HashSet;
 
 use thiserror::Error;
+use zerocopy::{FromBytes, Immutable, IntoBytes};
 
 /// Possible errors during hooking of `syscall` instructions
 #[derive(Error, Debug)]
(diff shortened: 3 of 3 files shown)


.vscode/settings.json
CvvT, line 1: Please remove this file.

docs/design-cpu-resource-management.md
CvvT, line 1: Please remove this. On a side note, maybe we would like to have a design doc at high level (i.e., users story mentioned in Cheng's talk today. We could try something similar: https://zfhuang99.github.io/rust/claude%20code/codex/contracts/spec-driven%20development/2025/12/01/rust-with-ai.html#lightweight-spec-driven-development).

litebox_common_linux/src/loader.rs
```diff
@@ -200,113 +201,98 @@ impl ElfParsedFile {
...
-        let shdr = elf::section::SectionHeader::parse_at(
-            self.header.endianness,
-            self.header.class,
-            &mut 0,
-            &buf,
-        )?;
+        // Search backwards from the end of the file at page boundaries for the magic.
+        // We limit the search to a reasonable number of pages to avoid scanning the
+        // entire file.
+        let min_offset = file_size.saturating_sub((MAX_SEARCH_PAGES * PAGE_SIZE) as u64);
```
CvvT: Why not put those entries at the end of the file (end of trampoline code instead of head of it) so that we only need to read the last 32/20 bytes for metadata?

.vscode/settings.json
wdcui: Done. Sorry about it!

docs/design-cpu-resource-management.md
wdcui: I agree that we can start documenting our design plans and pricinpals as well as lessons from previous code reviews.

litebox_common_linux/src/loader.rs
wdcui: Great suggestion! Working on it.

jaybosamiya-ms:
I haven't reviewed the code, but quick note: we use a constant like `LITEBOX0` so that we can bump the versioning when there is a non-confirming change.  We have not yet claimed that things are stable right now, so I'll let you make a decision on it, but technically speaking, this would require bumping the magic constant to `LITEBOX1`, since it is an incompatible change.

wdcui:
I thought about if we should bump the version from LITEBOX0 to LITEBOX1 and decided to just keep LITEBOX0. Let me know if anyone thinks we should increase it to LITEBOX1. I'm okay with either way.

litebox_common_linux/src/loader.rs
```diff
@@ -200,113 +203,100 @@ impl ElfParsedFile {
...
+                .map_err(|_| ElfParseError::BadTrampoline)?;
+            (file_offset, vaddr, trampoline_size)
+        } else {
+            // 32-bit: [0..8] magic, [8..12] file_offset, [12..16] vaddr, [16..20] code_size
+            let file_offset = u64::from(u32::from_le_bytes(header_buf[8..12].try_into().unwrap()));
+            let vaddr = u32::from_le_bytes(header_buf[12..16].try_into().unwrap()) as usize;
+            let trampoline_size =
+                u32::from_le_bytes(header_buf[16..20].try_into().unwrap()) as usize;
+            (file_offset, vaddr, trampoline_size)
+        };
```
CvvT: This is fine, but I believe a better way is to define a struct for it and then directly converting raw bytes to a struct.

litebox_common_linux/src/loader.rs
```diff
@@ -200,113 +203,100 @@ impl ElfParsedFile {
...
+
+        // Validate trampoline size bounds
+        if trampoline_size == 0 || trampoline_size > MAX_TRAMPOLINE_SIZE {
```
CvvT: Why we need `MAX_TRAMPOLINE_SIZE`? Looks like a small number to me.

litebox_rtld_audit/rtld_audit.c
```diff
@@ -27,10 +27,18 @@
 #define SYS_exit_group 231
 #define AT_FDCWD -100
 
+// Maximum number of pages to search for trampoline
+#define MAX_SEARCH_PAGES 16
+// Maximum allowed trampoline size (must match Rust loader)
+#define MAX_TRAMP_SIZE (MAX_SEARCH_PAGES * 0x1000)
```
CvvT: Similar

litebox_common_linux/src/loader.rs
wdcui: Removed.

litebox_rtld_audit/rtld_audit.c
wdcui: Removed.

litebox_common_linux/src/loader.rs
wdcui: Added a struct.

PR #639, opened 2026-02-04
mm: Add support for MAP_FIXED_NOREPLACE flag in mmap

wdcui:
## Summary

Add support for the `MAP_FIXED_NOREPLACE` mmap flag, which allows creating a mapping at a specific address while failing with `EEXIST` if the range overlaps with existing mappings (unlike `MAP_FIXED` which silently replaces them).

This matches the Linux kernel behavior for `MAP_FIXED_NOREPLACE` (added in Linux 4.17).

## Changes

- Add `NOREPLACE` flag to `CreatePagesFlags` (used in combination with `FIXED_ADDR`)
- `MAP_FIXED_NOREPLACE` now sets both `FIXED_ADDR` and `NOREPLACE` flags, matching Linux semantics where `MAP_FIXED_NOREPLACE` implies `MAP_FIXED` behavior for address selection
- Return `EINVAL` for `MAP_FIXED_NOREPLACE` with `addr=0` (Linux requires a non-zero address)
- Add comprehensive syscall-level test covering:
  - Full overlap detection
  - Partial overlap at start/end
  - Adjacent (non-overlapping) mappings
  - Zero address validation

### DIFF
--- a/litebox_shim_linux/src/syscalls/mm.rs
+++ b/litebox_shim_linux/src/syscalls/mm.rs
@@ -126,10 +126,6 @@ impl Task {
         if !offset.is_multiple_of(PAGE_SIZE) || !addr.is_multiple_of(PAGE_SIZE) || len == 0 {
             return Err(Errno::EINVAL);
         }
-        // MAP_FIXED_NOREPLACE requires a non-zero address
-        if flags.contains(MapFlags::MAP_FIXED_NOREPLACE) && addr == 0 {
-            return Err(Errno::EPERM);
-        }
         if flags.intersects(
             MapFlags::MAP_SHARED
                 | MapFlags::MAP_32BIT
--- a/.gitignore
+++ b/.gitignore
@@ -131,11 +131,6 @@ tags
 
 ### VisualStudioCode ###
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
-!.vscode/extensions.json
-!.vscode/*.code-snippets
 
 # Local History for Visual Studio Code
 .history/
--- a/litebox/src/mm/linux.rs
+++ b/litebox/src/mm/linux.rs
@@ -417,8 +417,11 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
         fixed_address_behavior: FixedAddressBehavior,
     ) -> Result<Platform::RawMutPointer<u8>, AllocationError> {
         let (start, end) = (suggested_range.start, suggested_range.end);
-        if start < Platform::TASK_ADDR_MIN || end > Platform::TASK_ADDR_MAX {
-            return Err(AllocationError::InvalidRange);
+        if start < Platform::TASK_ADDR_MIN {
+            return Err(AllocationError::BelowMinAddress);
+        }
+        if end > Platform::TASK_ADDR_MAX {
+            return Err(AllocationError::AboveMaxAddress);
         }
         let platform_fixed_address_behavior = match fixed_address_behavior {
             FixedAddressBehavior::Hint => FixedAddressBehavior::Hint,
@@ -626,7 +629,11 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
                     | AllocationError::AddressInUseByPlatform
                     | AllocationError::AddressPartiallyInUse,
                 ) => return Err(VmemResizeError::RangeOccupied(range.into())),
-                Err(AllocationError::Unaligned | AllocationError::InvalidRange) => unreachable!(),
+                Err(
+                    AllocationError::Unaligned
+                    | AllocationError::BelowMinAddress
+                    | AllocationError::AboveMaxAddress,
+                ) => unreachable!(),
             }
             return Ok(());
         }
@@ -859,6 +866,10 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
             {
                 return Some(suggested_address.0);
             }
+        } else if fixed_addr {
+            // MAP_FIXED with addr=0: return 0 so insert_mapping rejects it
+            // via the TASK_ADDR_MIN check (BelowMinAddress → EPERM).
+            return Some(0);
         }
 
         // top down
--- a/litebox/src/platform/page_mgmt.rs
+++ b/litebox/src/platform/page_mgmt.rs
@@ -116,7 +116,8 @@ pub trait PageManagementProvider<const ALIGN: usize>: RawPointerProvider {
                     RemapError::AlreadyAllocated
                 }
                 AllocationError::Unaligned
-                | AllocationError::InvalidRange
+                | AllocationError::BelowMinAddress
+                | AllocationError::AboveMaxAddress
                 | AllocationError::AddressPartiallyInUse => unreachable!(),
             })?;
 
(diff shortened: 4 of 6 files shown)


litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -127,6 +127,10 @@ impl Task {
         if !offset.is_multiple_of(PAGE_SIZE) || !addr.is_multiple_of(PAGE_SIZE) || len == 0 {
             return Err(Errno::EINVAL);
         }
+        // MAP_FIXED_NOREPLACE requires a non-zero address
+        if flags.contains(MapFlags::MAP_FIXED_NOREPLACE) && addr == 0 {
+            return Err(Errno::EINVAL);
+        }
```
CvvT: I asked copilot to generate C test for it:

```
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <errno.h>
#include <sys/mman.h>
#include <unistd.h>

#define PAGE_SIZE 4096

int main(void) {
    void *addr;

    /* Case 1: mmap(NULL, ...) without MAP_FIXED
     * NULL is treated as a hint; kernel picks a suitable address.
     */
    printf("=== Case 1: mmap(NULL, ...) — no MAP_FIXED ===\n");
    addr = mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE,
                MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
    if (addr == MAP_FAILED) {
        printf("  FAILED: %s (errno=%d)\n", strerror(errno), errno);
    } else {
        printf("  OK: mapped at %p (kernel chose address)\n", addr);
        munmap(addr, PAGE_SIZE);
    }

    /* Case 2: mmap(0, ..., MAP_FIXED, ...)
     * MAP_FIXED forces the kernel to map exactly at address 0.
     * On modern Linux, vm.mmap_min_addr prevents this → EPERM.
     * If it somehow succeeded, it would also unmap any existing
     * mapping at that address.
     */
    printf("\n=== Case 2: mmap(0, ..., MAP_FIXED) ===\n");
    addr = mmap(0, PAGE_SIZE, PROT_READ | PROT_WRITE,
                MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0);
    if (addr == MAP_FAILED) {
        printf("  FAILED: %s (errno=%d)\n", strerror(errno), errno);
        if (errno == EPERM)
            printf("  → Blocked by vm.mmap_min_addr (cannot map the zero page)\n");
    } else {
        printf("  OK: mapped at %p (zero page mapped!)\n", addr);
        munmap(addr, PAGE_SIZE);
    }

    /* Case 3: mmap(0, ..., MAP_FIXED_NOREPLACE, ...)
     * MAP_FIXED_NOREPLACE forces address 0, but won't replace
     * existing mappings. Same vm.mmap_min_addr restriction applies.
     */
    printf("\n=== Case 3: mmap(0, ..., MAP_FIXED_NOREPLACE) ===\n");
    addr = mmap(0, PAGE_SIZE, PROT_READ | PROT_WRITE,
                MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED_NOREPLACE, -1, 0);
    if (addr == MAP_FAILED) {
        printf("  FAILED: %s (errno=%d)\n", strerror(errno), errno);
        if (errno == EPERM)
            printf("  → Blocked by vm.mmap_min_addr (cannot map the zero page)\n");
        else if (errno == EEXIST)
            printf("  → Address 0 already mapped (NOREPLACE refuses to overwrite)\n");
    } else {
        printf("  OK: mapped at %p (zero page mapped!)\n", addr);
        munmap(addr, PAGE_SIZE);
    }

    /* Case 4: Compare MAP_FIXED vs MAP_FIXED_NOREPLACE at a valid address.
     * This shows the key behavioral difference between the two flags.
     */
    printf("\n=== Case 4: MAP_FIXED vs MAP_FIXED_NOREPLACE at a valid address ===\n");

    /* First, get a valid mapping */
    void *base = mmap(NULL, PAGE_SIZE * 2, PROT_READ | PROT_WRITE,
                      MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
    if (base == MAP_FAILED) {
        perror("  initial mmap");
        return 1;
    }
    printf("  Initial mapping at %p (2 pages)\n", base);
    memset(base, 'A', PAGE_SIZE * 2);

    /* MAP_FIXED at the same address: silently replaces the existing mapping */
    addr = mmap(base, PAGE_SIZE, PROT_READ | PROT_WRITE,
                MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0);
    if (addr == MAP_FAILED) {
        printf("  MAP_FIXED replace FAILED: %s\n", strerror(errno));
    } else {
        printf("  MAP_FIXED at %p: OK (replaced existing mapping, data zeroed: first byte = 0x%02x)\n",
               addr, ((unsigned char *)addr)[0]);
    }

    /* MAP_FIXED_NOREPLACE at the same address: should fail with EEXIST */
    addr = mmap(base, PAGE_SIZE, PROT_READ | PROT_WRITE,
                MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED_NOREPLACE, -1, 0);
    if (addr == MAP_FAILED) {
        printf("  MAP_FIXED_NOREPLACE at %p: FAILED: %s (errno=%d)\n",
               base, strerror(errno), errno);
        if (errno == EEXIST)
            printf("  → Refuses to overwrite existing mapping (safe behavior)\n");
    } else {
        printf("  MAP_FIXED_NOREPLACE at %p: OK (unexpected)\n", addr);
    }

    munmap(base, PAGE_SIZE * 2);

    printf("\n=== Summary ===\n");
    printf("- mmap(NULL) without MAP_FIXED: kernel chooses address (NULL is just a hint)\n");
    printf("- mmap(0, MAP_FIXED): tries to map at address 0 → blocked by mmap_min_addr\n");
    printf("- mmap(0, MAP_FIXED_NOREPLACE): same → blocked by mmap_min_addr\n");
    printf("- MAP_FIXED: silently replaces existing mappings at the target address\n");
    printf("- MAP_FIXED_NOREPLACE: fails with EEXIST if address is already mapped\n");

    return 0;
}
```

The correct error code should be `EPERM`. We already have check at:

https://github.com/microsoft/litebox/blob/90c7ed92101a8e320f7f990c0a73badbe91049a2/litebox/src/mm/linux.rs#L416-L418

But `get_unmmaped_area` incorrectly returns an available address when suggested addr is zero and MAP_FIEXED flag is set.

https://github.com/microsoft/litebox/blob/90c7ed92101a8e320f7f990c0a73badbe91049a2/litebox/src/mm/linux.rs#L515-L525

We should probably fix these functions and new error for `AllocationError`.

wdcui: Thank you for catching these issues. I resolved them in the latest commit. Please take a look when you get a chance. Thanks!

CvvT: I was suggesting we should fix `get_unmmaped_area`: https://github.com/microsoft/litebox/blob/90c7ed92101a8e320f7f990c0a73badbe91049a2/litebox/src/mm/linux.rs#L515-L525

Currently, when `get_unmmaped_area` is called with 0 and FIXED_ADDR, it would return a different available address. It should return zero directly, and pass it to `insert_mapping`, which would return an error. We just need to convert that error to EPERM.

CvvT: For any address that is not within the range, we should return EPERM (not just zero).

wdcui: Good points! I fixed the code. Thanks!

PR #640, opened 2026-02-04
Add support for getpgrp/getpgid/setpgid/TIOCGPGRP/TIOCSPGRP

wdcui:
This PR adds the support for getpgrp/getpgid/setpgid/TIOCGPGRP/TIOCSPGRP.


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1275,6 +1275,15 @@ impl Task {
                 Ok(0)
             }
             IoctlArg::TCSETS(_) => Ok(0), // TODO: implement
+            IoctlArg::TIOCGPGRP(pgrp) => {
+                // Get the process group ID of the foreground process group on the terminal.
+                // In LiteBox, processes are simplified: each process forms its own process group
+                // where the process group ID equals the process ID. This is a simplification
+                // compared to standard Unix where processes can explicitly join different groups.
+                let pid = self.sys_getpid();
```
CvvT: `TIOCGPGRP` is not equivalent to `getpgrp`, it returns the `pgid` of the foreground process on the terminal. If litebox is running in background, it should not return its own `pid`. I'm not sure here what value we should return as the foreground process on the terminal is outside LiteBox. Maybe just log and return an error code for now?

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -1148,10 +1148,22 @@ impl Task {
...
     }
 
+    /// Handle syscall `getpgrp`.
+    ///
+    /// Returns the process group ID. For simplicity, this implementation returns
+    /// the process ID, which is the default behavior for a process that hasn't
+    /// explicitly joined another process group via `setpgid`.
+    pub(crate) fn sys_getpgrp(&self) -> i32 {
+        // In a full implementation, we'd track pgid separately. For now, return pid
+        // which is the default pgid for a new process.
```
CvvT: For a forked process, the `pgid` is the same as its parent, but right now we don't support `fork` yet. Should we just implement `setpgid` if it is simple for an agent to do?

litebox_shim_linux/src/syscalls/file.rs
wdcui: Fixed. I added TIOCSPGRP.

litebox_shim_linux/src/syscalls/process.rs
wdcui: Yes, added the support for setpgid.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -1148,10 +1149,47 @@ impl Task {
...
+
+    /// Handle syscall `getpgid`.
+    pub(crate) fn sys_getpgid(&self, pid: i32) -> Result<i32, Errno> {
+        if pid < 0 {
+            return Err(Errno::EINVAL);
+        }
+        if pid == 0 || pid == self.pid {
+            Ok(self.pgrp.load(core::sync::atomic::Ordering::Relaxed))
+        } else {
+            Err(Errno::ESRCH)
```
CvvT, line 1170:
```
ERRORS         [top](https://man7.org/linux/man-pages/man3/getpgid.3p.html#top_of_page)
       The getpgid() function shall fail if:

       EPERM  The process whose process ID is equal to pid is not in the
              same session as the calling process, and the implementation
              does not allow access to the process group ID of that
              process from the calling process.

       ESRCH  There is no process with a process ID equal to pid.

       The getpgid() function may fail if:

       EINVAL The value of the pid argument is invalid.
```

Add some comment here? Currently there is only process so we always return ESRCH.

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -1148,10 +1149,47 @@ impl Task {
...
+        if pid < 0 {
+            return Err(Errno::EINVAL);
+        }
+        let target_pid = if pid == 0 { self.pid } else { pid };
+        let target_pgid = if pgid == 0 { target_pid } else { pgid };
+        if target_pgid < 0 {
+            return Err(Errno::EINVAL);
+        }
+        if target_pid != self.pid {
+            return Err(Errno::ESRCH);
```
CvvT, line 1186: Similar

litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -1148,10 +1149,47 @@ impl Task {
...
+        }
+        self.pgrp
+            .store(target_pgid, core::sync::atomic::Ordering::Relaxed);
```
CvvT, line 1189: We should add some check to ensure that `pgid` is valid. Since we only have one process, we can just return an error if `pgid` differs from the current one. Don't need to change `pgid` at all.

litebox_shim_linux/src/lib.rs
```diff
@@ -194,6 +194,7 @@ impl LinuxShimBuilder {
             boot_time: self.platform.now(),
             load_filter: self.load_filter,
             next_thread_id: 2.into(), // start from 2, as 1 is used by the main thread
+            fg_pgrp: core::sync::atomic::AtomicI32::new(1),
```
CvvT, line 197: In LiteBox,  there is no concept of forground and background process. Initializing it to 1 indicates that LiteBox always runs in foreground, but I can also run it in background (`cargo run -p litebox_runner_linux_userland ... &`). In the latter case, LiteBox does not control the foreground process. Depending on how applications use this syscall, we may have different designs. For now, maybe we could just return some error code and print out some warnings.

PR #652, opened 2026-02-09
Fixes to simplify VTL1 boot sequence

tgopinath-microsoft:
Simplify VTL1 boot sequence by
1. Removing the need for cmdline and bootparams pages
2. Removing the need for boot signal page to bring up secondary processors

### DIFF
--- a/litebox_platform_lvbs/src/host/bootparam.rs
+++ b/litebox_platform_lvbs/src/host/bootparam.rs
@@ -1,200 +1,49 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-//! VTL1 kernel boot parameters (compatible with Linux kernel's boot_params structure and command line)
-
-use crate::mshv::vtl1_mem_layout::{
-    VTL1_BOOT_PARAMS_PAGE, VTL1_CMDLINE_PAGE, VtlMemoryError, get_address_of_special_page,
-};
-use core::ffi::{CStr, c_char};
-use num_enum::TryFromPrimitive;
+use litebox_common_linux::errno::Errno;
 use spin::Once;
-
-#[cfg(debug_assertions)]
-use crate::debug_serial_println;
-
-// This module defines a simplified Linux boot params structure
-// (based on arch/x86/include/uapi/asm/bootparam.h and
-// arch/x86/include/uapi/asm/e820.h). We need this because VTL0 kernel
-// passes memory information to VTL1 kernel via boot params.
-
-const E820_MAX_ENTRIES: usize = 128;
-
-const E820_RAM: u32 = 1;
-const E820_RESERVED: u32 = 2;
-const E820_ACPI: u32 = 3;
-const E820_NVS: u32 = 4;
-const E820_UNUSABLE: u32 = 5;
-const E820_PMEM: u32 = 7;
-const E820_PRAM: u32 = 12;
-const E820_RESERVED_KERN: u32 = 128;
 
 static POSSIBLE_CPUS: Once<u32> = Once::new();
 static VTL1_MEMORY_INFO: Once<(u64, u64)> = Once::new();
 
-#[derive(Default, Clone, Copy)]
-#[repr(C, packed)]
-struct BootE820Entry {
-    pub addr: u64,
-    pub size: u64,
-    pub typ: u32,
-}
-
-#[derive(Clone, Copy)]
-#[repr(C, packed)]
-struct BootParams {
-    _unused0: [u8; 720], // fields in this area are not used.
-    e820_table: [BootE820Entry; E820_MAX_ENTRIES],
-    _unused1: [u8; 816], // fields in this area are not used.
-}
-
-impl BootParams {
-    pub fn new() -> Self {
-        Self {
-            _unused0: [0; 720],
-            e820_table: [BootE820Entry::default(); E820_MAX_ENTRIES],
-            _unused1: [0; 816],
-        }
-    }
-
-    #[cfg(debug_assertions)]
-    pub fn dump(&self) {
-        for entry in self.e820_table {
-            let typ_val = entry.typ;
-
-            if E820Type::try_from(typ_val).unwrap_or(E820Type::Unknown) == E820Type::Unknown {
-                break;
-            } else {
-                let addr_val = entry.addr;
-                let size_val = entry.size;
-                debug_serial_println!(
-                    "addr: {:#x}, size: {:#x}, type: {:?}",
-                    addr_val,
-                    size_val,
-                    typ_val
-                );
-            }
-        }
-    }
-
-    pub fn memory_info(&self) -> Result<(u64, u64), VtlMemoryError> {
-        for entry in self.e820_table {
-            let typ_val = entry.typ;
-
-            match E820Type::try_from(typ_val).unwrap_or(E820Type::Unknown) {
-                E820Type::Ram => {
-                    let addr_val = entry.addr;
-                    let size_val = entry.size;
-                    return Ok((addr_val, size_val));
-                }
-                E820Type::Unknown => {
-                    return Err(VtlMemoryError::InvalidBootParams);
-                }
-                _ => {}
-            }
-        }
-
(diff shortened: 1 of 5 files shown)


tgopinath-microsoft:
This need corresponding VTL0 Linux changes. Do not merge till that is ready. This was done as part of clean up for submitting the initial set of VTL0 patches upstream

litebox_platform_lvbs/src/host/bootparam.rs
```diff
@@ -3,198 +3,51 @@
...
-    Ram = E820_RAM,
-    Reserved = E820_RESERVED,
-    Acpi = E820_ACPI,
-    Nvs = E820_NVS,
-    Unusable = E820_UNUSABLE,
-    Pmem = E820_PMEM,
-    Pram = E820_PRAM,
-    ReservedKern = E820_RESERVED_KERN,
-    Unknown = 0xffff_ffff,
+pub fn parse_boot_info(possible_cpus: u32, mem_pa: u64, mem_size: u64) {
```
sangho2: Do we expect `mem_pa != _memory_base`?

tgopinath-microsoft: we don't since we build this from the same structure in vtl0

sangho2: asked this because whether we need to get `mem_pa` through a CPU register. a bit unnecessary at this point, but perhaps fine to possibly deal with future changes.

sangho2: btw, I think `parse_boot_info` is no longer a good name (it doesn't parse anything now).

litebox_platform_lvbs/src/host/bootparam.rs
```diff
@@ -1,200 +1,49 @@
...
         Ok(pair)
     } else {
-        Err(VtlMemoryError::InvalidBootParams)
+        Err(Errno::EINVAL)
     }
 }
 
 /// Funtion to get the number of possible cpus from the command line (Linux kernel's num_possible_cpus())
-pub fn get_num_possible_cpus() -> Result<u32, VtlMemoryError> {
+pub fn get_num_possible_cpus() -> Result<u32, Errno> {
```
sangho2, line 20: Seems that no one is using this function anymore.

litebox_runner_lvbs/src/main.rs
sangho2, line 164: This documentation needs to be updated (i.e., page numbers). I can fix this if you want.

litebox_runner_lvbs/src/main.rs
```diff
@@ -430,15 +430,16 @@ unsafe extern "C" fn common_start(is_bsp: bool) -> ! {
...
+pub unsafe extern "C" fn _start(possible_cpus: u64, mem_pa: u64, mem_size: u64) -> ! {
     unsafe {
         // Phase 1a: Fix GOT entries from link-time (base 0x0) to PA-based VAs.
         // The binary is linked at address 0x0 but loaded by VTL0 at an arbitrary
         // physical address (memory_base). This pass rewrites every GOT slot so
         // that globals and function pointers resolve correctly under the
         // identity map (VA == PA) that VTL0 left us with.
         apply_relocations();
-
+        parse_boot_info(possible_cpus as u32, mem_pa, mem_size);
```
sangho2: Not so sure whether this saved data survive across address remapping. I'll test it from my side as well.

sangho2 (commented):
Thanks! Overall, looks good to me. There is one uncertain change; let me test it as well.

litebox_platform_lvbs/src/host/bootparam.rs
tgopinath-microsoft: > asked this because whether we need to get `mem_pa` through a CPU register. a bit unnecessary at this point, but perhaps fine to possibly deal with future changes.

Yes. I think we should have it, atleast for now. Esp since code can be loaded with at an offset . Also like you mentioned for other changes. Plus when we upstream, we would like to say that this is an abi for any secure kernel

litebox_runner_lvbs/src/main.rs
tgopinath-microsoft: ys please

litebox_runner_lvbs/src/main.rs
tgopinath-microsoft: you mean does it survive across remap_to_high_canonical ? If so, it does. During my testing I invoked get_num_possible_cpus and get_vtl1_memory_info in kernel_main after "Hello from Litebox" print . And it is retrieved correctly

litebox_platform_lvbs/src/host/bootparam.rs
tgopinath-microsoft: yes you are right. Would mind though if we keep this a little while longer ,please ? I want to do the multi instance support for TA and dynamic TA loading and if we still do not need it, we can remove it

tgopinath-microsoft:
> Thanks! Overall, looks good to me. There is one uncertain change; let me test it as well.

Thanks for the review ! If you want to test it , use the thara/project/lvbs/6.6/main branch in https://dev.azure.com/msazure/LSG-linux/_git/LSG-linux-rolling?_a=history . And use azl3_lvbs_main_config to build the config

litebox_runner_lvbs/src/main.rs
tgopinath-microsoft: image

sangho2: seems fine. context: I tried to use another complicated data structure across remap, but it didn't survive. This one uses simple static Once, so would work.

litebox_platform_lvbs/src/host/bootparam.rs
sangho2: maintaining it should be fine if there is upcoming usage (documentation might be worthwhile)

litebox_runner_lvbs/src/main.rs
sangho2: done

litebox_platform_lvbs/src/arch/x86/ioport.rs
```diff
@@ -8,7 +8,7 @@ use core::{arch::asm, fmt};
 use spin::{Mutex, Once};
 
 // LVBS uses COM PORT 2 for printing out debug messages
-const COM_PORT_2: u16 = 0x2F8;
+const COM_PORT_2: u16 = 0x3F8;
```
sangho2: COM_PORT_1? It is related to #787. We might need to decide which one should be the default (COM PORT 1 or 2).

sangho2 (approved):
Looks good to me. Seems that we need to revert or change the default COM port number.

sangho2:
> > Thanks! Overall, looks good to me. There is one uncertain change; let me test it as well.
> 
> Thanks for the review ! If you want to test it , use the thara/project/lvbs/6.6/main branch in https://dev.azure.com/msazure/LSG-linux/_git/LSG-linux-rolling?_a=history . And use azl3_lvbs_main_config to build the config

Thanks! Let me test it. One question: does this branch have OP-TEE support you've made as well?

tgopinath-microsoft:
No. though Angelina has a PR against this branch. Once this gets checked in, I will check in the other branch as well and then push in Angelina's PR

litebox_platform_lvbs/src/arch/x86/ioport.rs
tgopinath-microsoft: no. I should remove this. So you have to approve again !

sangho2:
> No. though Angelina has a PR against this branch. Once this gets checked in, I will check in the other branch as well and then push in Angelina's PR

That sounds like a chicken-and-egg problem because I need the OP-TEE support for adding features and fixing some bugs. Anyway, seems that I need to maintain my own branch for a while.

litebox_runner_lvbs/src/main.rs
sangho2: Looks like my changes on documentation are reverted. Will make a separate PR later.

sangho2: #799

PR #653, opened 2026-02-09
 Parse attributes in extended VTL0 APIs

femiatl:
VTL0 sends data to the secure kernel during and after initial boot. Use the new attributes parameter to extend load_kdata, allowing it to process data after boot and use the improved format for sending data from VTL0. Small, simple data buffers like certificates can be sent efficiently with the same APIs used for larger aggregated data like module info.

The new data format is leveraged to reduce the size of the kernel string table passed from ~10MB (all of rodata) to ~250KB.

The validate_module and validate_kexec APIs are extended to use the new attributes.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/heki_data.rs
+++ b/litebox_platform_lvbs/src/mshv/heki_data.rs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 //! VTL0 data format and parsing
+#![allow(clippy::trivially_copy_pass_by_ref)]
 
 use modular_bitfield::{
     bitfield,


sangho2:
@femiatl Could you please resolve the conflicts such that CI can work on this PR.

sangho2:
It seems that this change is based on new ABI to simplify data transfer. Good idea! As @tgopinath-microsoft noted in #652, this cannot be merged yet.

One overall feedback or requested change: This PR needs some detailed comments to explain new ABI and data structures. Also, some functions are with old comments (e.g., these functions no longer get `nranges`), and some other functions are without any comments.

litebox_platform_lvbs/src/mshv/heki_data.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+#[repr(C)]
+#[derive(FromBytes, Immutable, KnownLayout)]
+pub struct HekiDataRange {
+    pub va: u64,
+    pub pa: u64,
+    pub epa: u64,
+    data_type: u16,
+    pub attr: u16,
```
sangho2: Is this `MemAttr`? a bit misleading because we do have `HekiDataAttr` above.

litebox_platform_lvbs/src/mshv/heki_data.rs
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// modular_bitfield generates warning. There is an updated crate
+// with fix, but we are locked to current version.
```
sangho2: We could try to use the updated one.

litebox_platform_lvbs/src/mshv/heki_data.rs
```diff
@@ -0,0 +1,180 @@
...
+            .field("epa", &format_args!("{epa:#x}"))
+            .field("attr", &format_args!("{attr:#x}"))
+            .field("type", &format_args!("{data_type}"))
+            .field("size", &format_args!("{:?}", self.epa - self.pa))
+            .finish_non_exhaustive()
+    }
+}
+
+/// `HekiDataRange` is a generic container for various types of memory ranges.
+/// It has a context-specific `attributes`
```
sangho2: Seems this "context-specific attributes" is no longer valid. `data_type` is now context-specific and `attr` is just `MemAttr`.

litebox_platform_lvbs/src/mshv/heki_data.rs
```diff
@@ -0,0 +1,180 @@
...
+            .finish_non_exhaustive()
+    }
+}
+
+impl HekiDataPage {
+    const SIZE: usize = PAGE_SIZE;
+    const MAX_RANGE_COUNT: usize =
+        (Self::SIZE - size_of::<HekiDataHdr>()) / size_of::<HekiDataRange>();
+
+    pub fn from_bytes(bytes: &[u8]) -> Result<&Self, VsmError> {
```
sangho2: `try_from_bytes` would be better.

litebox_platform_lvbs/src/mshv/heki_data.rs
```diff
@@ -0,0 +1,180 @@
...
+        let (data_page, _) =
+            HekiDataPage::ref_from_prefix(bytes).map_err(|_| VsmError::DataPageInvalid)?;
+        Ok(data_page)
+    }
+
+    pub fn kdata_type(&self) -> HekiKdataType {
```
sangho2, line 129: Both `HekiDataHdr` and `HekiDataRange` have `data_type`. Do they serve different purposes?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -254,12 +256,14 @@ pub fn mshv_vsm_end_of_boot() -> i64 {
 
 /// VSM function for protecting certain memory ranges (e.g., kernel text, data, heap).
 /// `pa` and `nranges` specify a memory area containing the information about the memory ranges to protect.
```
sangho2: `nranges` is no longer used.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -254,12 +256,14 @@ pub fn mshv_vsm_end_of_boot() -> i64 {
...
-pub fn mshv_vsm_protect_memory(pa: u64, nranges: u64) -> Result<i64, VsmError> {
+pub fn mshv_vsm_protect_memory(pa: u64, va: u64, attr: u64) -> Result<i64, VsmError> {
+    if attr == 0 {
```
sangho2, line 267: Fine for now, but breaking ABI change like this should be gracefully handled in the future.

litebox_platform_lvbs/src/mshv/heki_data.rs
```diff
@@ -0,0 +1,180 @@
...
+        HekiKdataType::try_from(self.hdr.data_type).unwrap_or(HekiKdataType::Unknown)
+    }
+
+    pub fn next(&self) -> Option<(u64, u64, usize)> {
```
sangho2: better to rename this function because `next()` looks like this is for iterator, but it isn't.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -270,39 +274,36 @@ pub fn mshv_vsm_protect_memory(pa: u64, nranges: u64) -> Result<i64, VsmError> {
...
-
-            protect_physical_memory_range(
-                PhysFrame::range(
-                    PhysFrame::containing_address(PhysAddr::new(pa)),
-                    PhysFrame::containing_address(PhysAddr::new(epa)),
-                ),
-                mem_attr,
-            )?;
         }
+        let Some(next_data_page_addr) = data_page.next() else {
```
sangho2: `next_data_page_addr` is misleading because this is a tuple.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -332,128 +333,29 @@ fn parse_certs(mut buf: &[u8]) -> Result<Vec<Certificate>, VsmError> {
...
-                    return Err(VsmError::KernelDataTypeInvalid);
-                }
-                _ => {
-                    debug_serial_println!("VSM: Unsupported kernel data not loaded {heki_range:?}");
-                }
-            }
-        }
-    }
+    let attr = HekiDataAttr::from_bytes(attr.to_le_bytes());
+    let data_type = attr.dtype_or_err().unwrap_or(HekiKdataType::Unknown);
```
sangho2: If this is unknown data, better to return an error here.

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Agreed. Let me update the PR with some doc for breaking API as well

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Fixed

femiatl: Do we want to remove the Unknown option from the kdata enum as well? Otherwise we still have to handle it in a match somewhere

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: reworded the variable. I also moved the loop over datapage into its function

litebox_platform_lvbs/src/mshv/heki_data.rs
femiatl: Renamed to next_page_params

litebox_platform_lvbs/src/mshv/heki_data.rs
femiatl: Renamed to mem_attr

litebox_platform_lvbs/src/mshv/heki_data.rs
femiatl: Updated platform_lvbs Cargo.toml's modular-bitfield v0.13.1

litebox_platform_lvbs/src/mshv/heki_data.rs
femiatl: attr field uses only 4 of the 16 bits for RWX permissions, so we could use the other bits for other info in the future. But I updated the comments and field name to reflect what we have now.

litebox_platform_lvbs/src/mshv/heki_data.rs
femiatl: Fixed

litebox_platform_lvbs/src/mshv/heki_data.rs
femiatl: They are different, usually. HekiDataHdr can have a type of ModuleInfo, while the ranges can have Subtypes of ModuleInfo eg Elf, Init, etc

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: I updated the function parameter name, as well as the documentation for the changed functions

litebox_platform_lvbs/src/mshv/vsm.rs
femiatl: Perhaps we can add one byte each for a major/minor version to each of the APIs, check those in the dispatch function.

PR #654, opened 2026-02-09
Demand paging and fallible memory copy for LVBS

### LINKED ISSUE #637: Page fault handler for the LVBS platform
Currently, the LVBS platform does not support a page fault handler. We need it to support demand paging and `memcpy_fallible` which suppresses memory copy errors between kernel and user, capturing some issues like use after free.

#632

sangho2:
Adds demand paging support for user-space addresses on the LVBS platform, handling user-space page faults from both user-mode and kernel-mode contexts (e.g., syscall handler copying not-yet mapped or invalid user pages using `memcpy_fallible`). This PR resolves #637.

Changes
- ISR stubs check CS RPL to route user- and kernel-mode exceptions to the shim with different CPU context savings; kernel-mode page faults at user-space addresses trigger demand paging or address fixup
- VTL0 memory copies replaced with `memcpy_fallible` + RAII `Vtl0MappedGuard`; `copy_from_vtl0_phys` uses `zerocopy::FromBytes` for memory safety
- Per-CPU separate stacks for double-fault (TSS.IST1) and use-mode exception (TSS.RSP0)
- Shim interface extended with `kernel_mode` field on `ExceptionInfo` and `ResumeKernelPlatform`/`ExceptionFixup` variants on `ContinueOperation`

### DIFF
--- a/litebox_platform_lvbs/src/arch/x86/interrupts.S
+++ b/litebox_platform_lvbs/src/arch/x86/interrupts.S
@@ -168,49 +168,19 @@ isr_with_err_code isr_general_protection_fault general_protection_fault_handler_
 
 /* Vector 14: Page Fault (#PF) - Error code
  *
- * Custom ISR stub to support kernel-mode page faults at user-space addresses.
- *
- * For kernel-mode page faults, page_fault_handler_impl returns:
- *   0 = handled (exception table fixup applied)
- *   1 = user-space address fault; route to shim for demand paging or fixup
- *   (panics for unrecoverable faults — never returns)
- *
- * When routing to the shim is needed, the stub restores GPRs and jumps to
- * kernel_exception_callback (in run_thread_arch), which calls the shim's
- * exception handler and then iretq back to the faulting kernel instruction.
+ * Both kernel-mode and user-mode page faults are routed through the shim's
+ * exception_handler, which handles demand paging, exception table fixup,
+ * and panic in order. The ISR stub just pushes the vector number and jumps
+ * to the appropriate callback.
  */
 .global isr_page_fault
 isr_page_fault:
     cld
     test    qword ptr [rsp + 16], 0x3  /* Check CS RPL bits */
+    push    14                  /* Pass vector number (push does not affect flags) */
     jnz     .Luser_isr_page_fault
-
-    /* --- Kernel-mode page fault --- */
-    push_regs
-    mov     rbp, rsp            /* Save stack pointer */
-    and     rsp, -16            /* Align stack to 16 bytes for call */
-    mov     rdi, rbp            /* Pass pointer to saved registers as first arg */
-    call    page_fault_handler_impl
-    test    eax, eax
-    jnz     .Lkernel_demand_page
-
-    /* Handled (exception table fixup) — return to fixup address */
-    mov     rsp, rbp
-    pop_regs
-    add     rsp, 8              /* Skip error code */
-    iretq
-
-.Lkernel_demand_page:
-    /* User-space address fault. Exception info already stored in per-CPU vars by
-     * page_fault_handler_impl. GPRs are already saved on the stack by push_regs
-     * (same layout as SAVE_CPU_CONTEXT_ASM). Jump directly to
-     * kernel_exception_regs_saved, skipping the redundant pop_regs/push cycle.
-     */
-    mov     rsp, rbp
-    jmp     kernel_exception_regs_saved
-
+    jmp     kernel_exception_callback
 .Luser_isr_page_fault:
-    push    14                  /* Pass vector number to exception_callback */
     jmp     exception_callback
 
 /* Vector 16: x87 Floating-Point Exception (#MF) - No error code */
--- a/litebox_platform_lvbs/src/arch/x86/interrupts.rs
+++ b/litebox_platform_lvbs/src/arch/x86/interrupts.rs
@@ -171,60 +171,6 @@ extern "C" fn general_protection_fault_handler_impl(regs: &PtRegs) {
     );
 }
 
-/// Result from the kernel-mode page fault handler, consumed by the ISR stub.
-#[repr(u8)]
-enum PageFaultResult {
-    /// Fault was handled via exception table fixup.
-    Handled = 0,
-    /// Fault at a user-space address; route to the shim's exception handler
-    /// for demand paging or exception-table fixup on invalid access.
-    RouteToShim = 1,
-}
-
-/// Kernel-mode page fault handler (vector 14).
-///
-/// Returns [`PageFaultResult`] to the ISR stub.
-/// For unrecoverable faults, this function panics and never returns.
-#[unsafe(no_mangle)]
-extern "C" fn page_fault_handler_impl(regs: &mut PtRegs) -> PageFaultResult {
-    use crate::host::per_cpu_variables::with_per_cpu_variables_asm;
-    use crate::{USER_ADDR_MAX, USER_ADDR_MIN};
-    use litebox::mm::exception_table::search_exception_tables;
-    use litebox::utils::TruncateExt as _;
-    use x86_64::registers::control::Cr2;
-
-    let fault_addr: usize = Cr2::read_raw().truncate();
-    let error_code = regs.orig_rax;
-
-    // Kernel-mode page fault at a user-space address (e.g., a syscall handler has accessed
-    // not-yet mapped or invalid user memory): route to the shim's exception handler for
-    // demand paging or exception table fixup.
-    if (USER_ADDR_MIN..USER_ADDR_MAX).contains(&fault_addr) {
-        with_per_cpu_variables_asm(|pcv| {
-            pcv.set_exception_info(
-                litebox::shim::Exception::PAGE_FAULT,
-                error_code.truncate(),
-                fault_addr,
-                regs.rip,
-            );
-        });
-        return PageFaultResult::RouteToShim;
-    }
-
-    // Safety net for fallible kernel memory operations (e.g., copying to/from
-    // VTL0 addresses) where the target address may be unmapped due to bugs.
-    if let Some(fixup_addr) = search_exception_tables(regs.rip) {
-        regs.rip = fixup_addr;
-        return PageFaultResult::Handled;
-    }
-
-    // Kernel-mode page fault at kernel-space addresses — unrecoverable
-    panic!(
-        "EXCEPTION: PAGE FAULT\nAccessed Address: {:#x}\nError Code: {:#x}\n{:#x?}",
-        fault_addr, error_code, regs
-    );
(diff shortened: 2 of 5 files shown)


litebox/src/shim.rs
```diff
@@ -96,6 +96,11 @@ pub enum ContinueOperation {
     ResumeGuest,
     /// Exit the current thread.
     ExitThread,
+    /// The exception was triggered in kernel mode; resume kernel execution.
+    ResumeKernel,
```
wdcui: "ReturnToPlatform"?

litebox/src/shim.rs
```diff
@@ -96,6 +96,11 @@ pub enum ContinueOperation {
...
+    /// The exception was not handled. The platform should apply an
+    /// exception-table fixup if one exists.
+    ExceptionFixup,
```
wdcui, line 113: What is exception table fixup?

litebox_platform_lvbs/src/arch/x86/interrupts.rs
```diff
@@ -97,174 +97,150 @@ pub fn init_idt() {
     idt().load();
 }
 
-// TODO: carefully handle exceptions/interrupts. If an exception or interrupt is due to userspace code,
-// we should destroy the corresponding user context rather than halt the entire kernel.
+// TODO: Let's consider whether we can recover some of the below exceptions instead of panicking.
```
wdcui, line 100: Let's log an issue to track this.

litebox_platform_lvbs/src/arch/x86/interrupts.rs
```diff
@@ -97,174 +97,150 @@ pub fn init_idt() {
...
-        regs
+    panic!(
+        "EXCEPTION: GENERAL PROTECTION FAULT (Error Code: {:#x})\n{:#x?}",
+        regs.orig_rax, regs
     );
 }
 
-/// Rust handler for page fault exception (vector 14).
-/// Called from assembly stub with pointer to saved register state.
+/// Fault was handled via exception table fixup. The ISR stub should
```
wdcui: "Fault was handled" is confusing. Below is just a constant. What is the context for this comment?

wdcui: You probably want to start saying these are possible return values from the kernel page fault handler.

litebox_platform_lvbs/src/arch/x86/interrupts.rs
```diff
@@ -97,174 +97,150 @@ pub fn init_idt() {
...
-    todo!(
-        "EXCEPTION [{}]: PAGE FAULT\nAccessed Address: {:?}\nError Code: {:#x}\n{:#x?}",
-        mode_str(regs),
-        Cr2::read(),
-        regs.orig_rax,
-        regs
+    let fault_addr: usize = Cr2::read_raw().truncate();
+    let error_code = regs.orig_rax;
+
+    // Kernel-mode page fault at a user-space address: route to the shim's
```
wdcui: Do we route user-mode page faults to the shim directly?

litebox_platform_lvbs/src/arch/x86/interrupts.S
```diff
@@ -109,51 +129,98 @@
...
+    /* --- Kernel-mode page fault --- */
+    push_regs
+    mov     rbp, rsp            /* Save stack pointer */
+    and     rsp, -16            /* Align stack to 16 bytes for call */
+    mov     rdi, rbp            /* Pass pointer to saved registers as first arg */
+    call    page_fault_handler_impl
+    test    eax, eax
+    jnz     .Lkernel_demand_page
+
+    /* Handled (exception table fixup) — return to fixed-up instruction */
```
wdcui: "fixed-up instruction" or "fix-up instruction"?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -1061,8 +1061,12 @@ fn write_msg_args_to_normal_world(
 }
 
 // use include_bytes! to include ldelf and (KMPP) TA binaries
-const LDELF_BINARY: &[u8] = &[0u8; 0];
-const TA_BINARY: &[u8] = &[0u8; 0];
+// const LDELF_BINARY: &[u8] = &[0u8; 0];
```
wdcui: Remove these lines?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -1621,16 +1776,59 @@ unsafe extern "C" fn run_thread_arch(
...
+    use crate::host::per_cpu_variables::{PerCpuVariablesAsm, with_per_cpu_variables_asm};
+    let info = with_per_cpu_variables_asm(|pcv| litebox::shim::ExceptionInfo {
+        exception: pcv.get_exception(),
+        error_code: pcv.get_exception_error_code(),
+        cr2: pcv.get_exception_cr2(),
+        kernel_mode,
+    });
+
+    if kernel_mode {
+        // Call the shim directly instead of using `call_shim` because:
```
wdcui: I don't follow this comment.

wdcui (commented):
I left some comments below.

litebox/src/shim.rs
sangho2: This enum specifies the operation that the platform should perform next (it always returns to the platform). `ResumePlatform`, `ResumePlatformCode`, or `ResumeKernelExceptionHandler`?

litebox/src/shim.rs
sangho2: if `memcpy_fallible` fails, we return an error and advance to the next instruction.

litebox_platform_lvbs/src/arch/x86/interrupts.rs
sangho2: Let me create an issue. These are real kernel panics, so recovery might not be possible, though.

litebox_platform_lvbs/src/arch/x86/interrupts.rs
sangho2: Yes. This specifies the meaning of the return values of `page_fault_handler_impl`. Let me clarify it.

litebox_platform_lvbs/src/arch/x86/interrupts.rs
sangho2: yes. Assembly stubs check CPL and jump to `exception_callback` directly if the exception is a user-mode exception.

litebox_platform_lvbs/src/arch/x86/interrupts.S
sangho2: fixed-up instruction. if this fault is not due to demand paging and the faulting address is at `memcpy_fallible`, `page_fault_handler_impl` fixes `rip` in `PtRegs` and `iretq` uses it to return to the fixed address. 

`fixup address` is more common.

litebox_runner_lvbs/src/lib.rs
sangho2: This is a mistake. Let me fix it.

litebox_platform_lvbs/src/lib.rs
sangho2: In most places, we call the shim functions like `thread_ctx.call_shim(|shim, ctx| shim.exception(ctx, &info))`. This `call_shim` is designed to return to the guest. However, this case is different because we resume the trapped kernel execution. I think this is fine at least for now.

litebox_platform_lvbs/src/arch/x86/interrupts.rs
sangho2: #655

litebox/src/shim.rs
wdcui: `ResumePlatform` sounds good.

litebox/src/shim.rs
wdcui: If memcpy fails, why can we continue the execution?

sangho2: This is for dealing with use-after-free or unmapped memory access, ensuring Rust memory safety. We use `memcpy_fallible` for accessing VTL0 and VTL1 userspace (`UserConst/MutPtr`) which is non-Rust memory. There is an exception table `litebox/src/mm/exception_table.rs`. If a faulting address is within `memcpy_fallible`, we adjust `rip` to a fixup instruction which returns an error to the caller of `memcpy_fallible` rather than panics the kernel.

litebox_platform_lvbs/src/arch/x86/interrupts.rs
```diff
@@ -97,174 +97,151 @@ pub fn init_idt() {
...
+                litebox::shim::Exception::PAGE_FAULT,
+                error_code.truncate(),
+                fault_addr,
+                regs.rip,
+            );
+        });
+        return PageFaultResult::DemandPage;
+    }
+
+    // Handle fallible memory operations like memcpy_fallible that access
```
sangho2: Once GPA range locking is implemented, this fixup becomes no longer meaningful. For now, this is mainly for dealing with potential race condition.

litebox/src/shim.rs
wdcui: ResumeKernelPlatform

wdcui (approved):
LGTM. Thank you!

litebox/src/shim.rs
```diff
@@ -90,12 +90,27 @@ pub trait EnterShim {
...
     ExitThread,
+    /// The shim successfully handled an exception which was triggered by
+    /// the kernel platform (e.g., a syscall handler's copy_from_user against
+    /// demand-pageable user memory); Resume the kernel platform's execution.
+    ResumeKernelPlatform,
+    /// The shim failed to handle the exception (e.g., invalid memory access).
+    /// The kernel platform will apply a fixup via
+    /// [`search_exception_tables`](crate::mm::exception_table::search_exception_tables)
+    /// if one exists.
+    ExceptionFixup,
```
CvvT, line 113: We should probably reconsider the `ContinueOperation` as it is priarily used by userland before. I'm suggesting that we should have something like the following:

```
pub enum ContinueOperation {
    /// Resume to the previous execution
    Resume,
    /// Terminate this thread or the operation had failed (depending on the context)
    ExitOrFail
}
```

On kenrel platform, `Resume` does not necessarily return back to guest code, it may also return to kernel code.

`ExitOrFail` means either exiting current thread for `init` and `syscall` or failing to handle the request for `exception` and `interrupt` (I don't have a better naming for it). This is because `ExitThread` does not make any sense for exception/interrupt handler in kernel mode.

With this change, we don't need `ResumeKernelPlatform` and `ExceptionFixup`, i.e., `ResumeKernelPlatform` is merged into `Resume` and `ExceptionFixup` is merged into `ExitOrFail`. Of course, lvbs platform needs to be updated to support a single `Resume`.

PR #656, opened 2026-02-10
litebox_syscall_rewriter: use direct ELF editing instead of builder round-trip

wdcui:
## Summary

Replace `object::build::elf::Builder` (read → modify → write round-trip) with `object::read` (read-only parse) + direct byte patching.

Since all `.text` modifications are in-place (same offsets, same sizes) and the trampoline is appended outside any existing segment, no ELF headers need modification. The ELF format has no checksum or integrity field, so in-place edits are safe.

## Changes

- **Cargo.toml**: Switch `object` features from `["elf", "build", "std"]` to `["elf", "read", "std"]`
- **lib.rs**: Use `object::File::parse()` for read-only ELF metadata extraction; patch section bytes directly in a mutable copy of the input
- Add `TextSectionInfo` struct and `section_slice`/`section_slice_mut` helpers with checked bounds (returns errors instead of panicking)
- Remove `GenerateObjFileError` variant (no write step)
- Remove `unsafe` block from old alignment workaround (replaced with safe zerocopy-based version)
- Drop 3 transitive dependencies: `crc32fast`, `indexmap`, `hashbrown 0.16`

## What stays the same

- All instruction-level patching logic (`hook_syscalls_in_section`, `hook_syscall_and_after`, `hook_syscall_before_and_after`)
- Trampoline code generation and header format
- CLI interface (`main.rs`)
- Output binary layout: `[original ELF with patched .text][padding][trampoline][header]`

## Testing

Both snapshot tests pass (32-bit and 64-bit). Clippy clean, fmt clean.

### DIFF
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -16,7 +16,8 @@
 
 use std::collections::HashSet;
 
-use object::read::{Object as _, ObjectSection as _, ObjectSegment as _, ObjectSymbol as _};
+use object::read::elf::{ElfFile, ProgramHeader as _};
+use object::read::{Object as _, ObjectSection as _, ObjectSymbol as _};
 use thiserror::Error;
 use zerocopy::{FromBytes, Immutable, IntoBytes};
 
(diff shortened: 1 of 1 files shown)


litebox_syscall_rewriter/src/lib.rs
```diff
@@ -454,44 +456,37 @@ fn hook_syscalls_in_section(
     Ok(())
 }
 
-fn find_addr_for_trampoline_code(builder: &object::build::elf::Builder<'_>) -> u64 {
-    // Find the highest virtual address among all sections in executable segments
-    let max_virtual_addr = builder
-        .segments
-        .iter()
-        .filter(|seg| seg.p_type == object::elf::PT_LOAD)
```
CvvT, line 462: why is this (filter) removed?

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -155,17 +151,17 @@ pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<u64>) -> Res
         trampoline_data.extend_from_slice(&trampoline.to_le_bytes());
     }
 
+    // Make a mutable copy of the input binary for in-place patching
+    let mut out = input_binary.to_vec();
```
CvvT: Minor optimization: If we are going to have a mutable copy anyway, we could remove the copy at the beginning.

CvvT (approved):
Overall LGTM, left some comments.

PR #662, opened 2026-02-11
Make Linux shim generic on file system

CvvT:
I was trying to add 9p fs and use it for SNP runner, then I realized our Linux shim decides the fs to use not runner. This PR makes Linux shim generic on fs so that runner can pick whatever fs.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -172,9 +172,7 @@ impl<FS: ShimFS> LinuxShimBuilder<FS> {
     ) -> DefaultFS {
         default_fs(&self.litebox, in_mem_fs, tar_ro_fs)
     }
-}
 
-impl<FS: ShimFS> LinuxShimBuilder<FS> {
     /// Set the load filter, which can augment envp or auxv when starting a new program.
     pub fn set_load_filter(&mut self, callback: LoadFilter) {
         self.load_filter = Some(callback);


litebox_shim_linux/src/lib.rs
```diff
@@ -59,7 +59,11 @@ pub(crate) type LinuxFS = litebox::fs::layered::FileSystem<
     >,
 >;
 
-pub(crate) type FileFd = litebox::fd::TypedFd<LinuxFS>;
+pub(crate) type FileFd<FS> = litebox::fd::TypedFd<FS>;
+
+/// A trait required for file systems to be used in the shim.
+pub trait ShimFS: litebox::fs::FileSystem + Send + Sync + 'static {}
+impl<T: litebox::fs::FileSystem + Send + Sync + 'static> ShimFS for T {}
```
jaybosamiya-ms, line 66: This portion is the key change, as I understand it, right?

This is similar to a (broader) refactoring I was planning do but haven't had a chance to finish it up. That refactoring can build on top of this one and can clean it up, so please feel free to merge this as is.

jaybosamiya-ms: I assume you've already tested this, but any particular reason it needs to be `Send + Sync + 'static`?  Can we relax any/all of them?  It'd be nice to have it be the weakest requirements needed.  If it cannot be weakened by removing those traits, that is fine, but just checking.

CvvT: `'static` is required by `Box`

https://github.com/microsoft/litebox/blob/1e6a7906504dd819c8fd29e9850e883a78df971f/litebox_shim_linux/src/syscalls/process.rs#L550-L568

`Send` and `Sync` are required by `Arc` in multiple places for sharing across threads, e.g., https://github.com/microsoft/litebox/blob/1e6a7906504dd819c8fd29e9850e883a78df971f/litebox_shim_linux/src/syscalls/epoll.rs#L43

PR #663, opened 2026-02-11
Add 9p fs to LiteBox

CvvT:
This PR adds implementation of 9p fs that uses 9P2000.L protocol. For testing, I used [`diod`](https://github.com/chaos/diod) as the remote server. The implementation of 9p client was largely migrated from VSBox.

Note that different from other fs, this only support blocking operations as it relies on network. Not sure if it breaks any assumption that LiteBox has.

Currently it only allows issuing one call over network at a time and waiting for it to complete. We should allow concurrent requests. I will fix it in the next PR.

### DIFF
--- a/litebox/src/fs/nine_p/client.rs
+++ b/litebox/src/fs/nine_p/client.rs
@@ -413,7 +413,11 @@ impl<Platform: RawSyncPrimitivesProvider, T: Read + Write> Client<Platform, T> {
         )
     }
 
-    /// Read directory entries
+    /// Read directory entries starting at the given offset.
+    ///
+    /// The `offset` is an opaque cookie from the server (taken from
+    /// [`DirEntry::offset`](fcall::DirEntry::offset)); pass `0` to start from the beginning.
+    /// Use [`readdir_all`](Client::readdir_all) to read all entries.
     pub(super) fn readdir(
         &self,
         fid: fcall::Fid,
--- a/litebox/src/fs/nine_p/fcall.rs
+++ b/litebox/src/fs/nine_p/fcall.rs
@@ -1189,9 +1189,15 @@ fcall_types! {
 }
 
 /// Tagged 9P message
+///
+/// Every 9P message carries a `tag` chosen by the client to match requests
+/// with their responses. The special value [`NOTAG`] is reserved for
+/// `Tversion`/`Rversion` messages.
 #[derive(Clone, Debug)]
 pub(super) struct TaggedFcall<'a> {
+    /// Unique identifier chosen by the client to correlate a request with its response.
     pub(super) tag: u16,
+    /// The 9P message payload.
     pub(super) fcall: Fcall<'a>,
 }
 
--- a/litebox/src/fs/nine_p/mod.rs
+++ b/litebox/src/fs/nine_p/mod.rs
@@ -353,7 +353,7 @@ impl<Platform: sync::RawSyncPrimitivesProvider, T: transport::Read + transport::
         }
     }
 
-    /// Walk to the parent of a path and return the parent fid and the name
+    /// Walk to the parent of a path and return the parent fid and the name of the final component
     fn walk_to_parent<'a>(&self, path: &'a str) -> Result<(fcall::Fid, &'a str), Error> {
         let components: Vec<&str> = path
             .normalized_components()


litebox/src/fs/layered.rs
```diff
@@ -100,6 +100,7 @@ impl<Platform: sync::RawSyncPrimitivesProvider, Upper: super::FileSystem, Lower:
         match self.lower.file_status(path) {
             Ok(stat) => Ok(stat.file_type),
             Err(FileStatusError::ClosedFd) => unreachable!(),
+            Err(FileStatusError::Io) => Err(PathError::NoSuchFileOrDirectory),
```
wdcui: `Io` is a pretty general erorr name. Why would it imply `NoSuchFileOrDirectory`?

litebox/src/fs/errors.rs
```diff
@@ -25,6 +25,8 @@ pub enum OpenError {
     AlreadyExists,
     #[error("error when truncating: {0}")]
     TruncateError(#[from] TruncateError),
+    #[error("I/O error")]
+    Io,
```
wdcui, line 29: Should we interpret `Io` error as network failure instead of the file or directory not exist?

litebox/src/fs/nine_p/mod.rs
```diff
@@ -0,0 +1,925 @@
...
+        if components.is_empty() {
+            // Clone the root fid
+            self.client.clone_fid(self.root.1)
+        } else {
+            let (_, fid) = self.client.walk(self.root.1, &components)?;
+            Ok(fid)
+        }
+    }
+
+    /// Walk to the parent of a path and return the parent fid and the name
```
wdcui: Nit: "the name of the child"?

litebox/src/fs/nine_p/mod.rs
```diff
@@ -0,0 +1,925 @@
...
+
+    /// Convert getattr response to FileStatus
+    fn rgetattr_to_file_status(attr: &fcall::Rgetattr) -> super::FileStatus {
+        let file_type = Self::qid_type_to_file_type(attr.qid.typ);
+
+        if attr.valid.contains(fcall::GetattrMask::BASIC) {
+            super::FileStatus {
+                file_type,
+                mode: super::Mode::from_bits_truncate(attr.stat.mode),
+                size: usize::try_from(attr.stat.size).expect("file size exceeds usize"),
```
wdcui: `expect()` will cause litebox to crash, right? Should we avoid that?

litebox/src/fs/nine_p/mod.rs
```diff
@@ -0,0 +1,925 @@
...
+            .ok_or(super::errors::ReadError::ClosedFd)?;
+
+        let read_offset = match offset {
+            Some(o) => o,
+            None => current_offset,
+        };
+
+        let bytes_read = self.client.read(fid, read_offset as u64, buf)?;
+
+        // Update offset if not using explicit offset
```
wdcui, line 627: Shouldn't we always update the offset stored in the descriptor table?

litebox/src/fs/nine_p/client.rs
```diff
@@ -0,0 +1,571 @@
...
+use alloc::vec::Vec;
+use core::sync::atomic::{AtomicU16, Ordering};
+
+use crate::sync::{Mutex, RawSyncPrimitivesProvider};
+
+use super::Error;
+use super::fcall::{self, Fcall, FcallStr, GetattrMask, TaggedFcall};
+use super::transport::{self, Read, Write};
+
+/// ID generator for fids
```
wdcui, line 17: Sangho implemented similar but more advanced recycling for sessions ids in the optee shim. Maybe we can move this kind of code to some common crate?

litebox/src/fs/nine_p/client.rs
```diff
@@ -0,0 +1,571 @@
...
+        if ret.is_err() {
+            self.fids.free(new_fid);
+        }
+        ret
+    }
+
+    /// Walks the path from the given fid, handling paths longer than fcall::MAXWELEM by walking in chunks.
+    ///
+    /// Returns the qids for each path component and a new fid for the final location on success.
+    fn walk_chunked(
```
wdcui, line 248: Do we have tests for all these functionalities?

litebox/src/fs/nine_p/client.rs
```diff
@@ -0,0 +1,571 @@
...
+                    .map(fcall::DirEntry::into_owned)
+                    .collect()),
+                Fcall::Rlerror(e) => Err(Error::from(e)),
+                _ => Err(Error::InvalidResponse),
+            },
+        )
+    }
+
+    /// Read all directory entries
+    pub(super) fn readdir_all(
```
wdcui, line 442: What's the difference between `readdir` and `readdir_all`?

litebox/src/fs/nine_p/fcall.rs
```diff
@@ -0,0 +1,1921 @@
...
+    Rversion(Rversion<'a>),
+    Twalk(Twalk<'a>),
+    Rread(Rread<'a>),
+    Twrite(Twrite<'a>),
+}
+
+/// Tagged 9P message
+#[derive(Clone, Debug)]
+pub(super) struct TaggedFcall<'a> {
+    pub(super) tag: u16,
```
wdcui, line 1199: What is the `tag` for?

litebox/src/fs/nine_p/fcall.rs
```diff
@@ -0,0 +1,1921 @@
...
+        let TaggedFcall { tag, fcall } = self;
+
+        buf.clear();
+        buf.resize(4, 0); // Reserve space for size
+
+        // Encode the message directly to the buffer (appending after the size field)
+        encode_fcall(buf, tag, fcall)?;
+
+        // Write the size at the beginning
+        let size = u32::try_from(buf.len()).expect("buffer length exceeds u32");
```
wdcui: Should we return the error instead of panic?

litebox/src/fs/nine_p/fcall.rs
```diff
@@ -0,0 +1,1921 @@
...
+            return Err(super::Error::InvalidResponse);
+        }
+
+        let mut decoder = FcallDecoder { buf: &buf[4..] };
+        decoder.decode()
+    }
+}
+
+// ============================================================================
+// Encoding functions
```
wdcui: Can we simplify the encoding/decoding to serde of structs?

litebox/src/fs/nine_p/transport.rs
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//! 9P transport layer abstraction
+//!
+//! This module defines traits for reading and writing 9P protocol messages over
+//! an underlying transport (e.g., TCP socket, virtio-9p, etc.).
```
wdcui, line 7: Where is the network integration?

litebox/src/fs/nine_p/tests.rs
```diff
@@ -0,0 +1,675 @@
...
+use crate::platform::mock::MockPlatform;
+
+use super::transport;
+
+// ---------------------------------------------------------------------------
+// Transport adapter: implement litebox 9P transport traits for TcpStream
+// ---------------------------------------------------------------------------
+
+/// A wrapper around `TcpStream` that implements the litebox 9P transport traits.
+struct TcpTransport {
```
wdcui, line 20: Why do we have both implementations and tests in this file?

litebox/src/fs/nine_p/mod.rs
```diff
@@ -0,0 +1,925 @@
...
+//! network connection. The 9P protocol is a simple, message-based protocol originally designed
+//! for Plan 9 from Bell Labs. 9P2000.L is a Linux-specific variant that provides better
+//! compatibility with POSIX semantics.
+//!
+//! # Submodules
+//!
+//! The 9P implementation is split into several submodules:
+//! - `fcall` - Protocol message definitions and encoding/decoding
+//! - `transport` - Transport layer traits and message I/O
+//! - `client` - High-level 9P client for protocol operations
```
jaybosamiya-ms: Minor comment (I have not reviewed the code, just was looking through the docs): we might wish to skip this from the doc-strings and change it to a regular comment.  Anything in the doc-strings (here, module-level) effectively behaves like a spec. The details most of these submodules (which are not necessarily public) is not useful/needed to callers of this module.

wdcui (commented):
I went through the code and left some comments. Thank you for adding 9p support in LiteBox!

litebox/src/fs/nine_p/fcall.rs
CvvT: Good point! Encoding and decoding have a lot of repetition. `serde` has its own binary representation which may not conform to the 9p protocol. I asked agent to use macro to simplify the code.

litebox/src/fs/nine_p/tests.rs
CvvT: This uses std sockets for testing purposes only. All the following tests set up diod as remote server and thus require sockets for connection. This crate does not have network stack implementation. In the next PR, I will use linux shim to test 9p as well.

litebox/src/fs/nine_p/client.rs
CvvT: I asked agents to implement more tests to cover all functionalities.

litebox/src/fs/nine_p/client.rs
CvvT: Due to message size limit, one call may not be able to get all entries. `readdir` allows callers to specify an offset (usually it is just an index starting from 0) to get the remaining entries.

CvvT: I added some comments.

litebox/src/fs/nine_p/fcall.rs
CvvT: Concurrent requests may receive out-of-order responses. Each response should also have a tag that matches its request.

CvvT: I added some comments.

litebox/src/fs/nine_p/fcall.rs
CvvT: I agree. Updated all `expect` to return an error instead.

litebox/src/fs/nine_p/mod.rs
CvvT: Good point! I removed it.

litebox/src/fs/nine_p/mod.rs
CvvT: updated

litebox/src/fs/nine_p/mod.rs
CvvT: Right, updated.

litebox/src/fs/nine_p/mod.rs
CvvT: If the offset is provided by user, it means the file offset should remain unchanged.

litebox/src/fs/nine_p/transport.rs
CvvT: The network stack (e.g., TCP) is implemented in Linux shim. I will add it in the next PR.

litebox/src/fs/layered.rs
CvvT: Good catch. `PathError` does not have `Io`. I updated the return type.

litebox/src/fs/errors.rs
CvvT: Yes, I updated the code.

litebox/src/fs/nine_p/client.rs
CvvT: I asked agent to do it. I can submit a separate PR for it.

wdcui: Please create an issue to track this in case you don't plan to improve it right now.

litebox/src/fs/nine_p/client.rs
wdcui: Will you add these tests in this PR or a separate one?

CvvT: It is already included in this PR.

litebox/src/fs/nine_p/client.rs
CvvT: #684

wdcui:
I resolved all my comments. Please feel free to merge it. Thanks!

PR #666, opened 2026-02-18
Add support for non-contiguous physical page mapping

sangho2:
This PR introduces a vmap subsystem to the LVBS platform to map non-contiguous physical page frames into virtually contiguous addresses.

Changes
- Add `map_non_contiguous_phys_frames` which is similar to existing `map_phys_frame_range` except that this new function is for non-contiguous physical page frames.
- Reserve a 1 TB VA window for vmap and vunmap
- Maintain separate hash tables for tracking PA<->VA for VMAPs (`mm/vmap.rs`)
- Support cross-core TLB flush

### DIFF
--- a/litebox_platform_lvbs/src/host/per_cpu_variables.rs
+++ b/litebox_platform_lvbs/src/host/per_cpu_variables.rs
@@ -42,7 +42,7 @@ pub struct PerCpuVariables {
     pub vtl0_locked_regs: ControlRegMap,
     pub gdt: Option<&'static gdt::GdtWrapper>,
     pub tls: VirtAddr,
-    vp_index: u32,
+    pub vp_index: u32,
 }
 
 impl PerCpuVariables {
@@ -577,10 +577,13 @@ pub fn allocate_per_cpu_variables() {
     #[allow(clippy::needless_range_loop)]
     for i in 1..num_cores {
         let mut per_cpu_variables = Box::<PerCpuVariables>::new_uninit();
-        // Safety: `PerCpuVariables` is larger than the stack size, so we manually `memset` it to zero.
+        // Safety: `PerCpuVariables` is too large for the stack, so we zero-init
+        // via `write_bytes` then fix up `vp_index` to the `u32::MAX` sentinel
+        // before calling `assume_init`.
         let per_cpu_variables = unsafe {
             let ptr = per_cpu_variables.as_mut_ptr();
             ptr.write_bytes(0, 1);
+            (*ptr).vp_index = u32::MAX;
             per_cpu_variables.assume_init()
         };
         unsafe {


litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -365,6 +365,130 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
...
+                        fl.flush();
+                    }
+                }
+                Err(e) => {
+                    if mapped_count > 0 {
+                        Self::rollback_mapped_pages(
+                            &mut inner,
+                            Page::range_inclusive(
+                                start_page,
+                                start_page + (mapped_count as u64 - 1),
```
wdcui: why `mapped_count - 1`?

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -365,6 +365,130 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
...
+        inner: &mut MappedPageTable<'_, FrameMapping<M>>,
+        pages: x86_64::structures::paging::page::PageRangeInclusive<Size4KiB>,
+        allocator: &mut PageTableAllocator<M>,
+    ) {
+        for page in pages {
+            if let Ok((_frame, fl)) = inner.unmap(page)
+                && FLUSH_TLB
+            {
+                fl.flush();
+            }
```
wdcui: Should we log the error at least?

litebox_platform_lvbs/src/mm/vmap.rs
```diff
@@ -0,0 +1,366 @@
...
+    ///
+    /// The guard pages are reserved in the VA space but never mapped, so an
+    /// out-of-bounds access past the allocation triggers a page fault.
+    ///
+    /// First tries to find a suitable range in the free list, then falls back to
+    /// bump allocation.
+    ///
+    /// Returns `Some(VirtAddr)` with the starting virtual address on success,
+    /// or `None` if insufficient virtual address space is available.
+    fn allocate_va_range(&mut self, num_pages: usize) -> Option<VirtAddr> {
```
wdcui, line 93: Do we need to support address alignment?

litebox_platform_lvbs/src/mm/vmap.rs
```diff
@@ -0,0 +1,366 @@
...
+        if num_pages == 0 {
+            return None;
+        }
+
+        let total_pages = num_pages.checked_add(GUARD_PAGES)?;
+        let size = (total_pages as u64).checked_mul(PAGE_SIZE as u64)?;
+
+        // Try to find a suitable range in the free set (first-fit)
+        for range in self.free_set.iter() {
+            if range.end - range.start >= size {
```
wdcui: Here the range's size is measured by bytes, and other places a va range is measured by pages. It's a little confusing.

litebox_platform_lvbs/src/lib.rs
```diff
@@ -1195,72 +1196,119 @@ impl<Host: HostInterface, const ALIGN: usize> VmapManager<ALIGN> for LinuxKernel
...
+        {
+            let mut seen = hashbrown::HashSet::with_capacity(pages.len());
+            for page in pages {
+                if !seen.insert(page.as_usize()) {
+                    return Err(PhysPointerError::DuplicatePhysicalAddress(page.as_usize()));
+                }
+            }
+        }
+
+        if ALIGN != PAGE_SIZE {
```
wdcui, line 1223: Nit: check alignment before the more expensive loop?

litebox_platform_lvbs/src/mm/vmap.rs
```diff
@@ -0,0 +1,366 @@
...
+use spin::mutex::SpinMutex;
+use x86_64::VirtAddr;
+use x86_64::structures::paging::{PhysFrame, Size4KiB};
+
+use crate::mshv::vtl1_mem_layout::PAGE_SIZE;
+
+/// Start of the vmap virtual address region.
+/// This address is chosen to be within the 4-level paging canonical address space
+/// and not conflict with VTL1's direct-mapped physical memory.
+const VMAP_START: u64 = 0x6000_0000_0000;
```
wdcui: How do we prevent the page table code from using this va range?

wdcui (commented):
Over the code looks good to me. I left some comments below.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: `rollback_mapped_pages` internally calls `x86_64` crate's `clean_up_addr_range` which expects an inclusive range. we can use an exclusive range for `rollback_mapped_pages` if that is intuitive.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: You mean whether we rollbacked mappings? We can log the error, but `rollback_mapped_pages` is already within an error path.

litebox_platform_lvbs/src/mm/vmap.rs
sangho2: By default, we do use a global offset when we map physical page frames. Unless our target machine has an extreme amount of memory, there are no overlap. We can add extra check to the normal map function if needed.

litebox_platform_lvbs/src/mm/vmap.rs
sangho2: Functions in this module are called through `vmap` which has type-based strict alignment requirement.

litebox_platform_lvbs/src/mm/vmap.rs
sangho2: make sense. let me see whether I can change it.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -30,6 +31,49 @@ const FLUSH_TLB: bool = true;
...
+/// a security boundary, a stale TLB entry can allow access to a physical frame
+/// that has been freed and reallocated to a *different* process, breaking
+/// process isolation. This function should send IPIs to remote cores for a
+/// cross-core TLB shootdown (similar to Linux's `flush_tlb_others` /
+/// `native_flush_tlb_others`). On AMD CPUs with INVLPGB support, use the
+/// `invlpgb`/`tlbsync` instructions instead of IPIs for broadcast invalidation.
+/// On newer Intel CPUs, use Remote Action Requests (RAR) for the same purpose.
+///
+/// Note: OP-TEE TAs only unmap pages when tearing down the entire TA instance
+/// and its sessions, so multi-session TAs sharing a page table are not affected.
```
sangho2: #671

sangho2: Add support for cross-core TLB flush.

litebox_platform_lvbs/src/lib.rs
```diff
@@ -81,19 +82,53 @@ static CPU_MHZ: AtomicU64 = AtomicU64::new(0);
...
+pub(crate) const VMAP_END: usize = 0xFFFF_CFFF_FFFF_F000;
+
 /// Maximum virtual address (exclusive) for user-space allocations.
 /// This is set to (1 << 47) - PAGE_SIZE (upper limit of 4-level paging).
 const USER_ADDR_MAX: usize = 0x7FFF_FFFF_F000;
 
 /// Size of the user address space range.
-const USER_ADDR_RANGE_SIZE: usize = 0x1000_0000_0000; // 16 TiB
+/// TODO: Restricted to 4 GiB due to slow `cleanup_user_mappings`. Will fix this issue.
+const USER_ADDR_RANGE_SIZE: usize = 0x1_0000_0000; // 4 GiB
```
sangho2: #679

litebox_shim_optee/src/lib.rs
```diff
@@ -286,6 +286,22 @@ impl OpteeShim {
...
+    /// # Safety
+    ///
+    /// The caller must ensure that no references to the released memory regions
+    /// are held after this call.
+    pub unsafe fn release_user_mappings(&self) {
+        let release = |_r: core::ops::Range<usize>, _vm: litebox::mm::linux::VmFlags| true;
+        unsafe {
+            let _ = self.page_manager().release_memory(release);
+        }
+    }
```
sangho2, line 304: addressing #679

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
wdcui: Maybe add a comment about this so it's easier for a future reader to understand this line.

wdcui (approved):
LGTM. Thanks! Please feel free to merge it after addressing my last comment.

PR #669, opened 2026-02-19
Opportunistic CoW for platforms that support it

jaybosamiya-ms:
This PR improves LiteBox performance by **shaving off ~65% of our execution time** by introducing opportunistic CoW:
```
 _____
< MOO >
 -----
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

Concretely, this PR introduces a new platform interface to `PageManagementProvider` called `try_allocate_cow_pages`. This defaults to `UnsupportedByPlatform`, but any platform that has the ability to support copy-on-write page allocation can override this method in order to automatically obtain improved performance.

The improved performance itself comes from `mmap` in the Linux shim, where file-based mapping can be sped up.

This is not a trivial "pass through" since the files inside LiteBox need not correspond to files on the host (so even for Linux-on-Linux, there is work that must be done).  This motivates the actual design of `try_allocate_cow_pages` that I've chosen, which is closer to a memcpy-like interface, rather than anything file-oriented.  Platforms that have file-oriented support for it are welcome to keep those platform-specific details to themselves.  What this means is that `mmap` on the shim actually first looks up the file, then it checks to get a backing store memory of it; if one exists, then a CoW is attempted, at which point the platform itself does the lookup of the file (on Linux platform) and decides to make a CoW map of the page.

Currently, the biggest gains for us come from the actual ELF load time, since that is all that I have set up as supporting CoW for now (it was our biggest bottleneck; thus the 65% savings in execution time).  However, now that the core framework has been set up by this PR, future PRs should more easily be able to unlock performance wherever CoW might help.

There are some small bits I am not yet fully happy with the PR that I've marked with `TODO(jb)` etc., for example, it'd be good to store pre-loaded FDs rather than constantly opening/closing them.  However, that can/will be handled in a future PR.

### DIFF
--- a/litebox_shim_linux/src/syscalls/mm.rs
+++ b/litebox_shim_linux/src/syscalls/mm.rs
@@ -156,6 +156,11 @@ impl<FS: ShimFS> Task<FS> {
             perms
         };
 
+        // XXX: `try_allocate_cow_pages` and `register_existing_mapping` are not called under a
+        // unified lock, so there is a theoretical race if two threads concurrently attempt a
+        // fixed-address mapping with replacement at the same address. In practice this is benign:
+        // if a program races like this both threads will register the same mapping anyway. Updating
+        // to a begin/attempt/commit scheme could close this race window entirely.
         match <_ as PageManagementProvider<{ PAGE_SIZE }>>::try_allocate_cow_pages(
             litebox_platform_multiplex::platform(),
             suggested_addr.unwrap_or(0),


wdcui:
I think @CvvT should review this PR.

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -122,21 +134,23 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
...
+            data
         };
         (modes, data)
     };
     let tar_data: &'static [u8] = if let Some(tar_file) = cli_args.initial_files.as_ref() {
         if tar_file.extension().and_then(|x| x.to_str()) != Some("tar") {
             anyhow::bail!("Expected a .tar file, found {}", tar_file.display());
         }
-        mmapped_file_data(tar_file)?
+        mmapped_file(tar_file)?.data
```
CvvT, line 153: Do we need to consider files in the tar file?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -72,6 +75,124 @@ impl<FS: ShimFS> Task<FS> {
...
+            // Cannot fill full page
+            return None;
+        }
+
+        // TODO(jb): Do we ever need to do NoReplace?
+        let fixed_behavior = if flags.contains(MapFlags::MAP_FIXED) {
+            FixedAddressBehavior::Replace
+        } else {
+            FixedAddressBehavior::Hint
+        };
```
CvvT: Why not check `NoReplace`? The flag is controlled by user?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -72,6 +75,124 @@ impl<FS: ShimFS> Task<FS> {
...
+                unsafe {
+                    self.global.pm.register_existing_mapping(
+                        range,
+                        permissions,
+                        true,
+                        fixed_behavior == FixedAddressBehavior::Replace,
+                    )
+                }
+                .unwrap();
+                Some(Ok(ptr))
```
CvvT, line 186: Potential race issue? Should we call  `try_allocate_cow_pages` and `register_existing_mapping` in a single function that takes lock on `PageManager::vmem` so that other page allocation calls won't contend on the same address?

CvvT (commented):
Thanks for the change! I have a few questions.

litebox_runner_linux_userland/src/lib.rs
jaybosamiya-ms: The tar files don't actually easily support the type of mapping we want, so more work needs to be done there; I have explicitly left this to a future PR, since it is almost an orthogonal change to the key changes in this PR (and also needs further design effort).

litebox_shim_linux/src/syscalls/mm.rs
jaybosamiya-ms: I'd left in this TODO for some reason, I don't fully recall what the particular constraint here was, I'll look at it again shortly, thanks for pointing it out!

litebox_shim_linux/src/syscalls/mm.rs
jaybosamiya-ms: Hmm, I think you are correct in terms of _technically_ there is the possibility of a race but I am struggling to think of a scenario where it can be relevant.  As I understand it, it is only really relevant if two different threads try to race to set the fixed address with replacement, because otherwise the underlying host access will just either pick a diff address for us (if non-fixed) or will reject it (if non-replacing).  And if a program is trying to race with replacement, then the program _anyways_ will register the same mapping, right?

I will need to give a bit more thought on how to expose the relevant lock here, possibly some type of begin-attempt-commit type thing.

Thanks for noticing this!

CvvT: I agree. Probably won't be an issue in practice. It's fine to leave a comment here for now.

PR #673, opened 2026-02-19
Implement TimeProvider for LinuxKernel platform using monotonic clock

CvvT:
The kernel platform implements system time using the initial system time from the host + monotonic clock. The computed time may not be accurate but should be fine for now. For simplicity, I enabled `clock_gettime` in [snp driver](https://github.com/MSRSSP/sandbox_driver/commit/fbe55ece7f4c635eae33c7df495138e1e3b395c7), but the initial value could be passed though boot parameters (hence no need for another interface).

Also added a minor fix to simplify the futex `block_or_maybe_timeout` logic.

### DIFF
--- a/litebox_platform_linux_kernel/src/host/mock.rs
+++ b/litebox_platform_linux_kernel/src/host/mock.rs
@@ -106,7 +106,7 @@ impl HostInterface for MockHostInterface {
             syscalls::syscall2(
                 syscalls::Sysno::clock_gettime,
                 0, // CLOCK_REALTIME
-                &mut t as *mut _ as usize,
+                &raw mut t as usize,
             )
         };
         assert!(ret.is_ok(), "clock_gettime failed");


litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -275,13 +286,14 @@ impl Instant {
...
+        inner: core::time::Duration::ZERO,
+    };
 
-    fn duration_since(
-        &self,
-        _earlier: &Self,
-    ) -> Result<core::time::Duration, core::time::Duration> {
-        unimplemented!()
+    fn duration_since(&self, earlier: &Self) -> Result<core::time::Duration, core::time::Duration> {
+        self.inner
```
wdcui, line 294: Why is it okay to have both cases?

CvvT: You mean the error value? The contract for this interface is the following.

```
/// Returns the amount of time elapsed from an `earlier` point in time to this one. This is
/// fallible since the clock might have been adjusted backwards in time to before the earlier
/// point in time was measured; in such a case, it returns an `Err(_)` with the absolute
/// duration.
```

Issue #667, opened 2026-02-18
Refactor `ContinueOperation`


CvvT:
> We should probably reconsider the ContinueOperation as it is priarily used by userland before. I'm suggesting that we should have something like the following:
> pub enum ContinueOperation {
>     /// Resume to the previous execution
>     Resume,
>     /// Terminate this thread or the operation had failed (depending on the context)
>     ExitOrFail
> }
> 
> On kenrel platform, Resume does not necessarily return back to guest code, it may also return to kernel code.
> ExitOrFail means either exiting current thread for init and syscall or failing to handle the request for exception and interrupt (I don't have a better naming for it). This is because ExitThread does not make any sense for exception/interrupt handler in kernel mode.
> With this change, we don't need ResumeKernelPlatform and ExceptionFixup, i.e., ResumeKernelPlatform is merged into Resume and ExceptionFixup is merged into ExitOrFail. Of course, lvbs platform needs to be updated to support a single Resume. 

 _Originally posted by @CvvT in [#654](https://github.com/microsoft/litebox/pull/654/changes#r2824507036)_

PR #674, opened 2026-02-20
Fix: Simplify Shim's ContinueOperation

sangho2:
This PR simplifies Shim's `ContinueOperation`, resolving #667.

Since the caller works with a specific handler (syscall or exception) and its execution mode (guest or kernel), the caller does not need to fully rely on `ContinueOperation` to determine its next operation.


CvvT (approved):
LGTM, thanks!

PR #676, opened 2026-02-20
LVBS: Use VP index and separate CPU entry points

sangho2:
This PR make LVBS use VP index instead of CPUID-based APIC ID. This is needed to run LVBS on NUMA machines.

### DIFF
--- a/litebox_platform_lvbs/src/arch/x86/mod.rs
+++ b/litebox_platform_lvbs/src/arch/x86/mod.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+use litebox::utils::TruncateExt;
+
 pub mod gdt;
 pub mod instrs;
 pub mod interrupts;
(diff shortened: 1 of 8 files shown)


litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -142,7 +142,14 @@ pub fn mshv_vsm_boot_aps(cpu_online_mask_pfn: u64, boot_signal_pfn: u64) -> Resu
...
 
-    // Initialize VTL for each online CPU and update its boot signal byte
+    // Initialize VTL for each online CPU and update its boot signal byte.
+    // TODO: The cpumask provides Linux CPU IDs which may not match Hyper-V VP indices.
+    // Currently we assume cpu_id == vp_index, which holds for standard Hyper-V topologies
+    // but is not architecturally guaranteed. Two possible fixes:
+    //   1) Have VTL0 pass per-CPU hv_vp_index values so VTL1 can translate
+    //      Linux CPU IDs to Hyper-V VP indices.
+    //   2) Have VTL0 compute and pass an online VP-index mask (instead of a
+    //      conventional cpumask) so VTL1 can use VP indices directly.
```
sangho2, line 152: Issue #677

sangho2: @tgopinath-microsoft Looks like we no longer use the boot_signal page. In that sense, option 2 might be preferable.

litebox_platform_lvbs/src/arch/x86/mod.rs
```diff
@@ -19,16 +21,41 @@ pub(crate) use x86_64::{
...
+///
+/// # Panics
+///
+/// Panics if [`is_bsp`] has not been called yet (i.e., the BSP VP index
+/// has not been initialised).
+#[inline]
+pub fn bsp_vp_index() -> usize {
+    *BSP_VP_INDEX
+        .get()
+        .expect("BSP VP index not yet initialised")
```
wdcui, line 58: Should we replace expect with something else so the kernel won't panic?

litebox_runner_lvbs/src/main.rs
```diff
@@ -116,16 +116,29 @@ unsafe fn apply_relocations() {
...
+/// [`_ap_start`].
 #[expect(clippy::missing_safety_doc)]
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn _start() -> ! {
-    let core_id = get_core_id();
-    if core_id == 0 {
-        unsafe {
-            apply_relocations();
-        }
+    unsafe {
```
wdcui, line 125: We don't need to check if it's bsp?

wdcui (approved):
LGTM. I left some comments below.

litebox_platform_lvbs/src/arch/x86/mod.rs
sangho2: Let me clarify this. Basically, `bsp_vp_index` cannot be called before `BSP_VP_INDEX` is written unless there is a memory corruption of an attack.

litebox_runner_lvbs/src/main.rs
sangho2: Only the BSP uses this entry function. APs use `_ap_start`. This is enforced by `hv_vtl_populate_vp_context` that VTL0 cannot override.

wdcui:
@tgopinath-microsoft , it might be helpful to have you take a look at this PR. Thanks!

sangho2:
Draft for now.

PR #678, opened 2026-02-23
Fix integer overflow panics on guest-derived usize arithmetic (#3)

CvvT:
Fix integer overflow vulnerabilities in guest-derived arithmetic.

As mentioned in #668, I created an agentic workflow to scan the codebase to find and fix all potential integer overflows on a daily basis (see https://github.com/CvvT/litebox/actions/workflows/integer-overflow-scanner.lock.yml). So far it has submitted 3 issues and this PR fixes one of them. I have reviewed this one. This is a trivial PR as replacing existing `+` with `wrapping_add` is safe (in most if not all cases).

### DIFF
--- a/litebox_shim_linux/src/syscalls/mm.rs
+++ b/litebox_shim_linux/src/syscalls/mm.rs
@@ -17,7 +17,7 @@ use crate::Task;
 #[inline]
 fn align_up(addr: usize, align: usize) -> usize {
     debug_assert!(align.is_power_of_two());
-    (addr + align - 1) & !(align - 1)
+    addr.wrapping_add(align - 1) & !(align - 1)
 }
 
 #[expect(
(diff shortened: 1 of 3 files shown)


CvvT:
There might be more PR coming (https://github.com/CvvT/litebox/pull/6). I can wait until the agent no longer find more issues and submit one single PR, or submit them separately?

jaybosamiya-ms:
Glad to see that we can find and fix these up with the agentic workflows.  A couple of minor comments:
- might be good to collate the changes together as one single PR since the changes are all going to be of a very similar nature
- nit: we should probably use a better title for the PR since that is what gets used for the squash (along with the PR description), and having `(#3)` or similar from a different repo will set up incorrect links if we need to look at the history. If you really want to point at the other repo there is [syntax for that](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#issues-and-pull-requests) but I think in this case, dropping the `#3` is probably the correct move.
- some of the things found via #647 are of this nature, so just marking it as related

PR #682, opened 2026-02-24
LVBS: Add support for DEP, SMEP, and SMAP

sangho2:
This PR adds support for DEP, SMEP, and SMAP.

DEP and SMEP rely on page-table re-initialization with `NO_EXECUTE`.
All kernel memory pages are now `W^X`, or read-only.
Supporting SMAP requires adding a new function `with_user_memory_access` to
the `litebox` crate and revising `UserConstPtr` and `UserMutPtr`.

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -406,6 +406,7 @@ impl litebox::platform::common_providers::userspace_pointers::ValidateAccess
         }
     }
 
+    #[cfg(target_arch = "x86_64")]
     #[inline]
     fn with_user_memory_access<R>(f: impl FnOnce() -> R) -> R {
         // STAC: Set AC flag to temporarily allow supervisor access to user pages.


litebox_platform_lvbs/src/lib.rs
```diff
@@ -377,15 +377,58 @@ pub struct LinuxPunchthroughToken<'a, Host: HostInterface> {
...
+            None
+        }
+    }
+
+    #[inline]
+    fn with_user_memory_access<R>(f: impl FnOnce() -> R) -> R {
+        // STAC: Set AC flag to temporarily allow supervisor access to user pages.
+        // Safety: STAC is a privileged instruction that only modifies the AC flag
+        // in RFLAGS. It has no side effects beyond enabling SMAP bypass.
+        unsafe {
```
wdcui, line 415: This is architecture dependent but this function is not?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -431,40 +474,127 @@ impl<Host: HostInterface> PunchthroughProvider for LinuxKernel<Host> {
...
+        }
+
+        // Enable the NX (No-eXecute) bit in IA32_EFER before loading the new
+        // page table. Without EFER.NXE set, the CPU treats bit 63 (NX/XD) of PTEs
+        // as reserved; loading a CR3 whose page tables have that bit set would
+        // trigger a reserved-bit violation.
+        crate::arch::enable_dep();
+
+        // Switch to the new base page table.
+        // Safety: the new page table identity-maps the entire VTL1 memory range,
```
wdcui, line 561: Both the initial page table and the new one do identity-mapping? Do we need the new one to be identity-mapped?

litebox_platform_lvbs/src/lib.rs
sangho2: Let me add a feature gate to clarify architecture dependency.

litebox_platform_lvbs/src/lib.rs
sangho2: That is on the todo list. Let me create an issue.

Another reason of creating a new page table was to "shift" kernel address mapping (to negative canonical addresses while letting userspace fully own positive canonical addresses). However, this was nontrivial because the code is already running and the global allocator is already filled. We might need a bootloader-like separate binary or dual-mapping with stack-only memory operation.

wdcui:
I resolved my comments. Please feel free to merge this PR when you are ready.

Issue #684, opened 2026-02-25
Add ID recycling utility


CvvT:
Sangho implemented similar but more advanced recycling for sessions ids in the optee shim. Maybe we can move this kind of code to some common crate?

_Originally posted by @wdcui in https://github.com/microsoft/litebox/pull/663#discussion_r2801482013_

PR #686, opened 2026-02-25
Add ID recycling utility

CvvT:
As mentioned in #684, asked agent to move the bitmap-based ID generator to LiteBox to avoid duplication.

### DIFF
--- a/litebox/src/utils/id_pool.rs
+++ b/litebox/src/utils/id_pool.rs
@@ -19,7 +19,7 @@ use alloc::vec::Vec;
 ///
 /// - **Growable** ([`new`](Self::new) / [`with_max_capacity`](Self::with_max_capacity)):
 ///   bitmap starts empty and grows one word (64 IDs) at a time up to the
-///   specified cap (default [`DEFAULT_MAX_CAP`]).
+///   specified cap (default 65536).
 ///   [`allocate`](Self::allocate) returns `None` when all slots within the
 ///   cap are in use.
 /// - **Fixed-capacity** ([`with_capacity`](Self::with_capacity)): bitmap is
@@ -61,7 +61,7 @@ impl IdPool {
     /// Create an empty, growable pool.
     ///
     /// The bitmap starts empty and grows one word (64 IDs) at a time as
-    /// needed, up to [`DEFAULT_MAX_CAP`] IDs.
+    /// needed, up to `DEFAULT_MAX_CAP` (65536) IDs.
     #[must_use]
     pub const fn new() -> Self {
         Self::with_max_capacity(DEFAULT_MAX_CAP)


litebox/src/utils/id_pool.rs
```diff
@@ -0,0 +1,319 @@
...
+///
+/// ```
+/// use litebox::utils::id_pool::IdPool;
+///
+/// let mut pool = IdPool::new();
+/// let a = pool.allocate().unwrap(); // 0
+/// let b = pool.allocate().unwrap(); // 1
+/// pool.recycle(a);
+/// let c = pool.allocate().unwrap(); // 0 (recycled)
+/// ```
```
CvvT: Remove the example which is incorrect.

CvvT: Removed

litebox/src/utils/id_pool.rs
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        None
+    }
+
+    /// Grow the bitmap by one word and allocate the first ID from it.
+    fn grow(&mut self) -> Option<u32> {
+        let new_id = self.max_ids;
+        // Guard: bitmap.len() * 64 must fit in u32.
+        if self.bitmap.len() >= u32::MAX as usize / 64 {
```
sangho2: In the worst case, the size of this bitmap would be 512 MB. In practice, we might want to have a hybrid approach: growable with mem cap.

litebox/src/utils/id_pool.rs
```diff
@@ -0,0 +1,307 @@
...
+            return None;
+        }
+        self.bitmap.push(1); // mark bit 0 of the new word as in-use
+        // Safety of truncation: the guard above ensures bitmap.len() * 64 fits in u32.
+        #[allow(clippy::cast_possible_truncation)]
+        {
+            self.max_ids = (self.bitmap.len() as u32) * 64;
```
sangho2: `self.max_ids = self.max_ids + 64`?

litebox/src/utils/id_pool.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// For growable pools, extends the bitmap when exhausted.
+    ///
+    /// Returns `None` when all IDs within the pool's range are in use.
+    pub fn allocate(&mut self) -> Option<u32> {
+        let cap = self.max_ids;
+        if cap > 0 {
+            let start = if self.hint >= cap - 1 {
+                0
+            } else {
+                self.hint + 1
```
sangho2, line 119: minor: overflow?

litebox/src/utils/id_pool.rs
```diff
@@ -0,0 +1,307 @@
...
+                return Some(id);
+            }
+        }
+
+        if self.growable { self.grow() } else { None }
+    }
+
+    /// Mark an ID as free so it can be reused.
+    ///
+    /// IDs beyond the current capacity and already-free IDs are silently
```
sangho2: sounds like a bug. debug_assert might be helpful.

litebox_shim_optee/src/lib.rs
sangho2: With this new IdPool, 65536 might work now.

litebox/src/utils/id_pool.rs
CvvT: Good point, added

litebox/src/utils/id_pool.rs
CvvT: Make sense, added

litebox/src/utils/id_pool.rs
CvvT: After `push`, the length of the bitmap already incremented by one. Given we have a `cap` now, the calculation is different.

litebox/src/utils/id_pool.rs
CvvT: If `hint` is `u32::max`, it would not go to this else branch.

litebox_shim_optee/src/lib.rs
CvvT: Good catch, updated

PR #687, opened 2026-02-26
Expanded Python test + faster caching

jaybosamiya-ms:
This PR has two somewhat related changes I hit when running the Python test.  Some versions of Python (concretely, 3.12.1) require more files to be pulled in than were being pulled in before; the first commit in this PR fixes this.  The second commit speeds the cache up by just using mtime+size rather than spending a bunch of time on computing sha256.

### DIFF
--- a/litebox_runner_linux_userland/tests/cache.rs
+++ b/litebox_runner_linux_userland/tests/cache.rs
@@ -119,7 +119,7 @@ fn compute_file_summary(path: &Path) -> std::io::Result<String> {
         .duration_since(std::time::UNIX_EPOCH)
         .unwrap_or_default();
     let size = metadata.len();
-    Ok(format!("{}:{}", mtime.as_nanos(), size))
+    Ok(format!("{}:{}:{}", path.display(), mtime.as_nanos(), size))
 }
 
 fn compute_string_hash(content: &str) -> String {


jaybosamiya-ms:
Note: the sha256 change shifts one bottleneck to further along the pipeline for the particular environment I am running this in, but I will leave those changes to a separate PR when I get some more cycles.

litebox_runner_linux_userland/tests/cache.rs
```diff
@@ -112,11 +112,14 @@ impl CacheEntry {
...
-    hasher.update(&content);
-    Ok(format!("{:x}", hasher.finalize()))
+fn compute_file_summary(path: &Path) -> std::io::Result<String> {
+    let metadata = std::fs::metadata(path)?;
+    let mtime = metadata
+        .modified()?
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap_or_default();
+    let size = metadata.len();
+    Ok(format!("{}:{}", mtime.as_nanos(), size))
```
wdcui: We don't want to include the file path in the summary?

jaybosamiya-ms: Good point, does not really hurt to include the file path

Issue #685, opened 2026-02-25
Change kernel address mapping (negative canonical addresses)


sangho2:
Both the initial page table and the new one do identity-mapping? Do we need the new one to be identity-mapped?

_Originally posted by @wdcui in https://github.com/microsoft/litebox/pull/682#discussion_r2850532957_

PR #688, opened 2026-02-26
LVBS: Remap VTL1 kernel to high-canonical virtual address region

sangho2:
This PR remaps identity-mapped VTL1 memory to high-canonical virtual address region,
letting user and kernel own the low and high portions of virtual addresses, respectively.
This separation satisfies the common assumption (`addr[63]=0`->user; `addr[63]=1`->kernel),
which allows us to support recent/future hardware security features like LAM/UAI and LASS.
In addition, it separates regions for direct map, vmap, and kernel with guard gaps.

The implementation features two-phase page table setup and two-phase linker symbol
relocation. The first phase of each highly depends on the VTL0-provided early page table
layout and kernel ELF layout (i.e., it might break if they change). Also, the first phase only
uses kernel stack (kernel heap is not yet available because high-canonical virtual addresses
are not ready).

Issue: #685

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -88,7 +88,7 @@ pub const BASE_PAGE_TABLE_ID: usize = 0;
 //  0xFFFF_FFFF_FFFF_FFFF  ┌─────────────────────────────────┐
 //                         │ VTL1 kernel region (~30 TiB)    │
 //                         │ VA = PA + KERNEL_OFFSET         │
-//  0xFFFF_E200_0000_0000  ├─────────────────────────────────┤ ← KERNEL_START
+//  0xFFFF_E200_0000_0000  ├─────────────────────────────────┤ ← KERNEL_OFFSET
 //                         │ guard gap (1 TiB)               │
 //  0xFFFF_E0FF_FFFF_F000  ├─────────────────────────────────┤ ← VMAP_END
 //                         │ vmap region (32 TiB)            │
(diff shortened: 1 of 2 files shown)


litebox_platform_lvbs/src/mshv/hvcall_vp.rs
```diff
@@ -208,12 +208,12 @@ fn hvcall_enable_vp_vtl(
...
 
 unsafe extern "C" {
-    static _start: u8;
+    static _ap_start: u8;
 }
 
 #[inline]
 fn get_entry() -> u64 {
-    &raw const _start as u64
+    &raw const _ap_start as u64
```
wdcui, line 216: Why do we change from `_start` to `_ap_start`?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -1193,10 +1211,8 @@ pub trait HostInterface: 'static {
 }
 
 impl<Host: HostInterface, const ALIGN: usize> PageManagementProvider<ALIGN> for LinuxKernel<Host> {
-    // Use a high address for user space to separate from kernel identity-mapped memory.
-    // Kernel memory uses low addresses (identity mapped: VA == PA).
-    // User memory allocated via mmap uses high addresses (VA >= TASK_ADDR_MIN).
-    // This allows easy identification of user vs kernel pages during cleanup.
+    // User space occupies the low canonical half (0 .. 0x0000_7FFF_FFFF_FFFF).
```
wdcui, line 1215: Is it `F000`?

litebox_platform_lvbs/src/mshv/hvcall_vp.rs
sangho2: I tried to maintain unified `_start` for both BSP and APs, but it ended up with ugly code (redundancy, a jump instruction in the middle of a function). We can still maintain unified `_start` if needed (but it does have incompatibility with NUMA topology).

wdcui: We don't need to unify them. Cleaner is better. Thanks!

PR #689, opened 2026-02-27
Replace .unwrap() with proper error handling in run()

### LINKED ISSUE #650: Panic in "If it does actually panic, then ping the authors of LiteBox"
Trying to run `litebox_runner_linux_userland echo yay` on a stock desktop Ubuntu installation causes the following panic:

```
thread 'main' (41447) panicked at litebox_runner_linux_userland/src/lib.rs:118:48:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

Referring to this line: 

https://github.com/microsoft/litebox/blob/86e26df9909b8adb7276f6bc039c5fa2da2ebba7/litebox_runner_linux_userland/src/lib.rs#L118

The function it's contained in has this comment:

https://github.com/microsoft/litebox/blob/86e26df9909b8adb7276f6bc039c5fa2da2ebba7/litebox_runner_linux_userland/src/lib.rs#L97-L99

I could not find any documentation on the expected environment setup, so adding that and/or a better error message would be very welcome.

LEDazzio01:
## Summary

Replace panic-inducing `.unwrap()` calls in the `run()` function with proper error handling using `anyhow::Context`, so users get actionable error messages instead of panics when the environment is not set up correctly.

Fixes #650

## Problem

Running `litebox_runner_linux_userland echo yay` on a stock Ubuntu installation causes a panic:
```
thread 'main' panicked at litebox_runner_linux_userland/src/lib.rs:118:48:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

The function's own docstring acknowledges this:
> *"Can panic if any particulars of the environment are not set up as expected. If it does actually panic, then ping the authors of LiteBox."*

## Changes

- **Add `anyhow::Context` import** for `.with_context()` support
- **Replace `.unwrap()` on `std::path::absolute()`** (2 occurrences) with `.with_context()` providing the program name in the error
- **Replace `.unwrap()` on `path.metadata()`** with `.with_context()` explaining which path was inaccessible
- **Replace `.unwrap()` on `Mode::from_bits()`** with `.ok_or_else()` reporting the invalid mode bits
- **Update docstring** from `# Panics` to `# Errors` section reflecting the new behavior

## Scope

This PR intentionally focuses on **host-OS-facing `.unwrap()` calls only** — the ones that interact with the real filesystem and are likely to fail from environment issues (which is the scenario reported in #650). The remaining `.unwrap()` calls operate on the in-memory filesystem and are left untouched to keep this change focused and reviewable.

## Before / After

**Before:** A raw panic with a cryptic stack trace
```
thread 'main' panicked at litebox_runner_linux_userland/src/lib.rs:118:48:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

**After:** A clear, contextual error message
```
Error: Could not read metadata for '/path/to/program'. Ensure the path exists and is accessible.

Caused by:
    No such file or directory (os error 2)
```

### DIFF
--- a/litebox_runner_linux_userland/tests/error_handling.rs
+++ b/litebox_runner_linux_userland/tests/error_handling.rs
@@ -15,7 +15,7 @@ fn runner_binary() -> String {
 }
 
 /// Regression test for #650: running with a nonexistent program path should
-/// produce a clear error message instead of panicking.
+/// produce a clear, contextual error message instead of panicking.
 #[test]
 fn test_nonexistent_program_returns_error_not_panic() {
     let output = Command::new(runner_binary())
@@ -35,5 +35,13 @@ fn test_nonexistent_program_returns_error_not_panic() {
     assert!(
         !stderr.contains("panicked at"),
         "Runner panicked instead of returning an error.\nStderr: {stderr}"
+    );
+
+    // It SHOULD contain a contextual error message about the path
+    assert!(
+        stderr.contains("Could not resolve absolute path")
+            || stderr.contains("Could not read metadata"),
+        "Expected a contextual error message about path resolution or metadata, \
+         but got:\nStderr: {stderr}"
     );
 }


LEDazzio01:
Thanks for the thorough review, @copilot! All four comments have been addressed in commit f2f38c7:

1. **`from_bits_truncate`** — Switched to `Mode::from_bits_truncate()` as suggested, since `Mode` accepts all bits via `const _ = !0`.
2. **Regression test** — Added `tests/error_handling.rs` with a test that invokes the runner with a nonexistent path and asserts it returns an error (not a panic) with a contextual message. Just pushed a follow-up strengthening this test to also verify the error message content.
3. **Docstring accuracy** — Updated the docstring to include both `# Errors` (for the new error paths) and `# Panics` (for remaining internal panics from in-memory fs operations and syscall rewriting).
4. **Missed `.unwrap()`** — Applied `.with_context()` to the third `std::path::absolute()` call near the seccomp block as well.

LEDazzio01 (commented):
All four review comments have been addressed in the latest commit:

1. **`Mode::from_bits` → `from_bits_truncate`** — Applied as suggested, since `Mode` accepts all bits via `const _ = !0`.
2. **Regression test** — Added `tests/error_handling.rs` with `test_nonexistent_program_returns_error_not_panic()` that verifies the runner returns a contextual error (not a panic) for nonexistent program paths.
3. **Docstring accuracy** — Updated to include both `# Errors` (for the new contextual errors) and `# Panics` (for remaining internal panics from in-memory filesystem operations).
4. **Missed `.unwrap()` in seccomp block** — Replaced with `.with_context()` + `?` so it returns a proper error instead of panicking.

PR #690, opened 2026-02-27
refactor signal infrastructure and support SIGINT (ctrl-c)

CvvT:
This PR supports async signal SIGINT.

On Linux, the workflow is the following:

1. when boot, register SIGINT handler --> Ctrl-C arrives --> record signal in TLS and wake up a thread if it is being blocked
2. about to switch to guest (i.e., `prepare_to_run_guest`) --> check TLS and queue signals --> process signal
3. any blocking operations
    3.1. before commit to wait, `check_for_interrupt` would check TLS and queue signals --> process signal
    3.2. during wait, wake up and go to 3.1
    3.3. after waiting, got to 3.1

For shim to be able to retrieve pending signals from platform, add a new `SignalProvider` with the interface `fn take_pending_signals(&self, mut f: impl FnMut(litebox::shim::Signal))`. 

One tricky problem is that when a signal arrives, the running thread may be about to be blocked or blocked already (i.e., case 3.1 and 3.2), and we didn't interrupt a thread if it is running inside LiteBox code. In which case, we never have the chance to process that signal. It might be possible to examine the RIP of the thread in the signal handler and divert the execution to skip mutex wait call it if RIP is close to it (e.g., use assembly code to build a critical region), but this may not work on Windows (where we don't call syscall instruction directly). The current solution I have is based on the observation that for all blocking operations (e.g., futex wait, sleep, or any **interruptible wait**) boil down to `RawMutex::block_or_timeout` with the per-thread `WaitState` and the expected value `ThreadState::WAITING`. `WaitState` stores the thread status like `RUNNING_IN_HOST`, `WAITING`, `WOKEN`, and etc. Thus, in the signal handler, besides recording the signal, we can update the thread state from `WAITING` to `WOKEN` and then calls `futex wake`. If the thread is about to call futex wait, futex call would return immediately because of the change of the value. If the thread is being blocked, it would wake up and return. To access `WaitState` in the signal handler, I added two new interfaces `on_interruptible_wait_start` and `on_interruptible_wait_end` to `RawMutexProvider` so that platform can use `Waker` to wake up the blocking thread.

The signal handlers for SIGINT can be called on any threads. However, we don't want non-guest threads (e.g., background network worker, additional test thread spawned by `cargo test`) to be called with the signal. For background network worker, we can block those signals. For test thread, the signal handler would check if it has our custom TLS. If not, it blocks those signals and raise them again.

On Windows, the implementation is almost the same except that signal handler is called on a kernel thread (and thus cannot access litebox thread's TLS easily). I have to use a global variable to record all active threads.

I added some end-to-end C tests and Rust unit tests. To make Rust unit test work, I also introduced a new interface `fn run_test_thread<R>(f: impl FnOnce() -> R) -> R` to `ThreadProvider`. This is because spawning a guest thread (via `sys_clone`) and a regular thread in test (via std::thread::spawn) have different routes and require different setup. For example, while running in test, there is no guest code, and thus we don't need to do fs/gs swap (hence no setup for one of fs/gs). To make the code work for both guest thread and non-guest thread, we can set gs == fs before running the test thread.

### DIFF
--- a/litebox/src/platform/mod.rs
+++ b/litebox/src/platform/mod.rs
@@ -262,7 +262,7 @@ pub trait RawMutexProvider {
     /// This is a no-op by default.
     ///
     /// [`Waker::wake`]: crate::event::wait::Waker::wake
-    #[allow(unused_variables)]
+    #[expect(unused_variables)]
     fn update_waker(&self, waker: Option<crate::event::wait::Waker<Self>>)
     where
         Self: crate::sync::RawSyncPrimitivesProvider + Sized,


litebox/src/platform/mod.rs
```diff
@@ -95,6 +95,33 @@ pub trait ThreadProvider: RawPointerProvider {
...
+    /// thread registration needed for [`current_thread`](Self::current_thread)
+    /// and related functionality to work.
+    ///
+    /// This is intended for test threads that do not go through the normal
+    /// [`spawn_thread`](Self::spawn_thread) / guest entry path. The platform
+    /// sets up thread state before calling `f` and tears it down afterward.
+    ///
+    /// The default implementation simply calls `f()` with no additional setup.
+    /// Platforms that require explicit thread registration should override this.
+    fn run_test_thread<R>(f: impl FnOnce() -> R) -> R {
```
wdcui, line 110: Should we scope this function to tests only?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -926,10 +1086,62 @@ impl litebox::platform::ThreadProvider for LinuxUserland {
     fn interrupt_thread(&self, thread: &Self::ThreadHandle) {
         thread.interrupt();
     }
+
+    fn run_test_thread<R>(f: impl FnOnce() -> R) -> R {
```
wdcui, line 1151: Limit this to test only?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -764,34 +846,91 @@ thread_local! {
...
                 .expect("failed to clone current thread handle"),
             tls: SendConstPtr(tls),
-        }))));
+        }))))
+    }
+
+    /// Runs `f`, ensuring that [`CURRENT_THREAD_HANDLE`] is set while in the call to `f`.
+    fn run_with_handle<R>(tls: &TlsState, f: impl FnOnce() -> R) -> R {
+        // Safety: `tls_state` lives for the duration of this call.
+        unsafe { install_tls(tls) };
```
wdcui, line 884: Why do we install tls on demand?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -764,34 +846,91 @@ thread_local! {
...
-        let _guard = litebox::utils::defer(|| {
+        let _guard = litebox::utils::defer(move || {
             let current = CURRENT_THREAD_HANDLE.take().unwrap();
+            // Remove from the global registry.
+            ACTIVE_THREADS
+                .lock()
+                .unwrap()
+                .retain(|h| !Arc::ptr_eq(&h.0, &current.0));
             *current.0.lock().unwrap() = None;
+            let tls_index = TLS_INDEX.load(Ordering::Relaxed);
```
wdcui: Should we create a function to `uninstall_tls`?

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -764,34 +846,91 @@ thread_local! {
...
+        let bit: u32 = 1 << (signal.as_raw() - 1);
+
+        // Set the pending signal bit and wake the condvar in one lock scope.
+        {
+            let inner = self.0.lock().unwrap();
+            if let Some(inner) = inner.as_ref() {
+                // Safety: the TLS pointer is valid as long as the thread is
+                // alive, and we hold the thread handle lock.
+                let tls = unsafe { &*inner.tls.0 };
+                tls.pending_host_signals.fetch_or(bit, Ordering::SeqCst);
```
wdcui, line 920: What's the effect of this `fetch_or`?

wdcui (approved):
Overall, the code change looks good to me. I left some comments below.

wdcui (approved):
LGTM. I left some comments below.

litebox/src/platform/mod.rs
CvvT: I thought about it. Using `cfg(test)` does not work here because the tests reside in a different crate (litebox_shim_linux). Test functions cannot be used by a different crate. One possible solution is adding a new feature, but that adds complexity.

CvvT: Actually, we can use `cfg(debug_assertions)`.

litebox_platform_linux_userland/src/lib.rs
CvvT: Added `cfg(debug_assertions)`

litebox_platform_windows_userland/src/lib.rs
CvvT: Yes, updated.

litebox_platform_windows_userland/src/lib.rs
CvvT: All threads should install tls. Previously `install_tls` was called in the caller function, I moved it here because I want to reuse this function for test threads. So now threads including guest threads and test threads would install tls properly.

litebox_platform_windows_userland/src/lib.rs
CvvT: Performs a bitwise “or” operation on the current value and the argument val, and sets the new value to the result.

Returns the previous value.

litebox/src/shim.rs
```diff
@@ -133,3 +133,108 @@ impl Exception {
     /// #PF
     pub const PAGE_FAULT: Self = Self(14);
 }
+
+/// A signal number.
+///
+/// Signal numbers are 1-based and must be in the range 1–63.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct Signal(u32);
```
jaybosamiya-ms: Nit: store as a `u8`?

litebox/src/shim.rs
```diff
@@ -133,3 +133,108 @@ impl Exception {
...
+    }
+}
+
+/// A set of [`Signal`]s, stored as a 64-bit bitmask.
+///
+/// Bit `(signum - 1)` is set when signal `signum` is present in the set.
+/// Because signal numbers are 1-based and capped at 63, all 63 possible
+/// signals fit in a single `u64`.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SigSet(u64);
```
jaybosamiya-ms: If you are explicitly defining the internal details of how things are stored, you can just expose the value inside it via `pub struct SigSet(pub u64)` to remove `from_u64`/`as_u64`.  Alternatively, you should not be specifying how it is being _internally_ stored in the docs.

litebox/src/platform/mod.rs
```diff
@@ -95,6 +95,34 @@ pub trait ThreadProvider: RawPointerProvider {
...
+///
+/// Platforms can record signals (e.g., `SIGINT`) and the shim should call
+/// [`SignalProvider::take_pending_signals`] to consume them.
+pub trait SignalProvider {
+    /// Atomically take all pending asynchronous signals (e.g., SIGINT and SIGALRM)
+    /// for the current thread, passing each one to `f`.
+    ///
+    /// Platforms that support asynchronous signals should override this method.
+    #[allow(unused_variables, reason = "no-op by default")]
+    fn take_pending_signals(&self, f: impl FnMut(crate::shim::Signal)) {}
```
jaybosamiya-ms: I am _quite_ surprised by this provider.  There are some surprising inversions happening here, for example, it depends on `crate::shim` in its signature.

I don't think the `Signal` sits well in `shim`, since the `shim` module is meant for traits that the shim must implement but there isn't a trait being implemented there.

Also, "shim should call" implies that a shim that does not call stuff is incorrect.  This sort of requirement is _exactly_ what traits are for, and thus the way the design should be is to actually have the shim implement a trait that probably looks like the following:
```
pub trait ConsumeSignal {
  fn consume_signal(Signal);
}
```

Then the platform can always invoke this trait, and the shim can make a decision of it wants to buffer things or not.  If you need/want the "all in a group" behavior, then you can make the trait (instead of the above):
```
pub trait ConsumeSignal {
  fn consume_signals(&[Signal]);
}
```

litebox/src/platform/mod.rs
```diff
@@ -95,6 +95,34 @@ pub trait ThreadProvider: RawPointerProvider {
...
+/// Provider for consuming platform-originating signals.
+///
+/// Platforms can record signals (e.g., `SIGINT`) and the shim should call
+/// [`SignalProvider::take_pending_signals`] to consume them.
+pub trait SignalProvider {
+    /// Atomically take all pending asynchronous signals (e.g., SIGINT and SIGALRM)
+    /// for the current thread, passing each one to `f`.
+    ///
+    /// Platforms that support asynchronous signals should override this method.
+    #[allow(unused_variables, reason = "no-op by default")]
```
jaybosamiya-ms: Nit: prefer `expect` rather than `allow`

litebox/src/platform/mod.rs
```diff
@@ -220,6 +248,25 @@ where
 /// A provider of raw mutexes
 pub trait RawMutexProvider {
     type RawMutex: RawMutex;
+
+    /// Called when a thread enters an interruptible wait.
+    ///
+    /// The passed `waker` should live at least until the thread leaves the interruptible
+    /// wait (i.e., [`on_interruptible_wait_end`](Self::on_interruptible_wait_end) is called).
```
jaybosamiya-ms: This is a somewhat surprising requirement; we can simplify this by updating the signature in this to take a `Waker` rather than a `&Waker`, right?

litebox/src/platform/mod.rs
```diff
@@ -220,6 +248,25 @@ where
 /// A provider of raw mutexes
 pub trait RawMutexProvider {
     type RawMutex: RawMutex;
+
+    /// Called when a thread enters an interruptible wait.
```
jaybosamiya-ms: Called by whom?

litebox/src/platform/mod.rs
```diff
@@ -220,6 +248,25 @@ where
...
+    ///
+    /// The passed `waker` should live at least until the thread leaves the interruptible
+    /// wait (i.e., [`on_interruptible_wait_end`](Self::on_interruptible_wait_end) is called).
+    /// The platform can use the `waker` to wake up the thread while it is in the interruptible wait.
```
jaybosamiya-ms: "can use" -> "uses"?

litebox/src/platform/mod.rs
```diff
@@ -220,6 +248,25 @@ where
...
+    ///
+    /// This is a no-op by default.
+    #[allow(unused_variables)]
+    fn on_interruptible_wait_start(&self, waker: &crate::event::wait::Waker<Self>)
```
jaybosamiya-ms: The name of this function is somewhat weird.  It is not actually beginning the wait, right?  Maybe something like `register_waker` (and the other one should be `unregister_waker`)?

Also, what is the expected behavior if someone runs "start" multiple times in a row?

litebox/src/platform/mod.rs
```diff
@@ -220,6 +248,25 @@ where
...
+    where
+        Self: crate::sync::RawSyncPrimitivesProvider + Sized,
+    {
+    }
+
+    /// Called when a thread leaves an interruptible wait.
+    ///
+    /// This is a no-op by default.
+    fn on_interruptible_wait_end(&self) {}
```
jaybosamiya-ms: Can merge with the above one by taking in a `Option<Waker>` instead of a `Waker`; in that case, it is behaving more like `update_interruptible_waker`.

litebox_common_linux/src/signal/mod.rs
```diff
@@ -102,6 +102,17 @@ impl TryFrom<i32> for Signal {
...
     }
 }
+impl TryFrom<Signal> for litebox::shim::Signal {
+    type Error = Signal;
+
+    fn try_from(value: Signal) -> Result<Self, Self::Error> {
+        match value {
+            Signal::SIGINT => Ok(Self::SIGINT),
+            Signal::SIGALRM => Ok(Self::SIGALRM),
+            _ => Err(value),
```
jaybosamiya-ms: This is quite surprising, haven't you designed the `litebox::shim::Signal` literally to support the full set of Linux signals?  You've even hard-coded the "up to 64" and so on, so I see no reason why this does not support `SIGILL` etc converted over.

I do think that the long-term move is for `litebox::shim::Signal` to not have this 64 restriction, but if you are assuming that right now, then wouldn't the correct move just be `Signal::from_raw(...).ok_or(Err(...))`?

litebox/src/platform/mod.rs
CvvT: Fixed

litebox/src/shim.rs
CvvT: Fixed

litebox/src/platform/mod.rs
CvvT: Called by `WaitContext::start_wait`. I updated the doc.

litebox/src/platform/mod.rs
CvvT: Good point. `WaitContext` only has a reference to `Waker`, but it is fine to clone it.

litebox/src/platform/mod.rs
CvvT: Updated

litebox/src/platform/mod.rs
CvvT: I thought `on_start`, `on_stop` are common names for callbacks on change of some state. I don't have preference on this, and changed them to ~~`register_waker` and `unregister_waker`~~ `update_waker`.

CvvT: Probably it is fine to call "start" multiple times. It just updates a pointer in TLS, but I don't expect it to be used by anywhere except `WaitContext` (it's okay to be called at other places. The pointer would be overwritten anyway just before the wait). I added some comments saying this is called by `WaitContext`.

litebox/src/platform/mod.rs
CvvT: Good point, merged.

litebox/src/shim.rs
CvvT: I removed this `Signal`.

litebox/src/shim.rs
CvvT: I removed this `SigSet`.

litebox_common_linux/src/signal/mod.rs
CvvT: I removed this `Signal`

litebox/src/platform/mod.rs
CvvT: The challenge is that signal handler cannot call shim directly. For example, interruption in Linux userland works as follows:
    1. send interrupt signal to a thread
    2. set the flag `interrupt` to true (signal handler runs on signal stack and does not have access to shim which is stored on a different stack)
    3. if the thread is running in guest, update its RIP to execute `interrupt_callback` which in turn can call shim.
    4. if the thread is running in LiteBox, do nothing. After the thread finishes handling the syscall, it checks `interrupt` before switching to guest and calls into shim.

For signals, we do want to interrupt a thread running in LiteBox, e.g.,

```
pub fn wait_until(&self, mut ready: impl FnMut() -> bool) -> Result<(), WaitError> {
        assert_eq!(
            self.waker.0.state_for_assert(),
            ThreadState::RUNNING_IN_HOST
        );
        let _end_wait = crate::utils::defer(|| self.end_wait());
        loop {
            self.start_wait();
            if ready() {
                break Ok(());
            }
            self.commit_wait()?;
        }
    }
```

In this case, to break the loop, we need to call platform to consume any signals, not the other direction. We either add a new interface (e.g., `take_pending_signals`) or modify the existing one `RawMutex::block_or_timeout` (which is called inside `commit_wait`). The former one is cleaner.

CvvT:
@jaybosamiya-ms Are you going to resolve all the comments?

litebox/src/platform/mod.rs
```diff
@@ -220,6 +251,23 @@ where
...
+    ///
+    /// Called by `WaitContext::start_wait` with `Some(waker)` when the current thread
+    /// enters an interruptible wait, and by `WaitContext::end_wait` with
+    /// `None` when it leaves. The thread in an interruptible wait can be unblocked
+    /// by [`Waker::wake`].
+    ///
+    /// This is a no-op by default.
+    ///
+    /// [`Waker::wake`]: crate::event::wait::Waker::wake
+    #[allow(unused_variables)]
```
jaybosamiya-ms: nit: prefer `expect` over `allow`

litebox/src/platform/mod.rs
jaybosamiya-ms: Might be good to open an issue to track this thread, since I think cleaner design _may_ be possible but it'd be good to keep track of constraints; please resolve once opened, thanks!

jaybosamiya-ms:
I didn't realize you were waiting on me.  I've marked most as resolved, please feel free to merge this once you think it is ready, thanks!

litebox/src/platform/mod.rs
CvvT: Created #707

litebox/src/platform/mod.rs
CvvT: Good catch, fixed.

PR #692, opened 2026-02-28
support chdir

CvvT:
LiteBox's fs does not support changing current working directory. Add support for chdir in `litebox_shim_linux` and ensure always passing absolute path to LiteBox.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -246,7 +246,8 @@ impl<FS: ShimFS> Task<FS> {
             return Err(Errno::EINVAL);
         }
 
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -213,9 +246,9 @@ impl<FS: ShimFS> Task<FS> {
             return Err(Errno::EINVAL);
         }
 
-        let fs_path = FsPath::new(dirfd, pathname)?;
+        let fs_path = FsPath::new(dirfd, pathname, || self.fs.borrow().cwd.read().clone())?;
```
wdcui: Nit: should we be consistent with the code above: either define `get_cwd` closure or not?

CvvT: Sure, updated.

PR #693, opened 2026-02-28
Support alarm and SIGALRM

CvvT:
Add a new provider `TimerProvider` to create timer for alarm support.

For platforms that do not support timer, SIGALRM would only be sent when a syscall returns or a thread is about to be blocked (i.e., in `check_for_interrupt`).

### DIFF
--- a/litebox/src/platform/mod.rs
+++ b/litebox/src/platform/mod.rs
@@ -121,8 +121,9 @@ pub enum TimerCreationError {
 
 /// Timer support for proactive signal delivery.
 pub trait TimerProvider {
-    /// The platform-specific timer handle type.
+    /// The timer handle type.
     type TimerHandle: TimerHandle;
+    /// The signal type delivered by timers.
     type Signal;
 
     /// Create a new one-shot timer that delivers `signal` when it fires.


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -819,6 +819,112 @@ impl litebox::platform::ThreadProvider for WindowsUserland {
...
+unsafe extern "system" fn threadpool_timer_callback(
+    _instance: Win32_Threading::PTP_CALLBACK_INSTANCE,
+    context: *mut c_void,
+    _timer: Win32_Threading::PTP_TIMER,
+) {
+    // Safety: `context` points to the `TimerCallbackContext` owned by the
+    // `TimerHandle`. The handle's `Drop` impl waits for all in-flight
+    // callbacks before dropping the context, so this reference is valid.
+    let ctx = unsafe { &*context.cast::<TimerCallbackContext>() };
+    let thread = ACTIVE_THREADS.lock().unwrap().first().cloned();
```
wdcui, line 936: I got the following comment from GH Copilot. Is this a valid concern? If so, we may want to add a note about it the comments.

"The one thing to watch is that ACTIVE_THREADS.lock().unwrap().first() always picks the same thread. If you ever support multiple guest threads that each need independent timer delivery, you may want to associate the timer with a specific thread rather than broadcasting to the first one in the list. But for single-process/single-main-thread use, the current approach is fine."

wdcui (approved):
LGTM. I left a minor comment below.

litebox/src/platform/mod.rs
```diff
@@ -112,6 +112,33 @@ pub trait ThreadProvider: RawPointerProvider {
...
+pub trait TimerProvider {
+    /// The platform-specific timer handle type.
+    type TimerHandle: TimerHandle;
+
+    /// Whether this platform supports [`TimerProvider`] for proactive timer delivery.
+    const SUPPORTS_TIMER: bool = false;
+
+    /// Create a new one-shot timer that delivers `signal` when it fires.
+    fn create_timer(&self, signal: crate::shim::Signal) -> Self::TimerHandle;
+}
```
jaybosamiya-ms, line 137: Minor: 

I'm not sure `SUPPORTS_TIMER` is the ideal move, a possible alternative is to do
```
pub trait TimerProvider {
  type TimerHandle: TimerHandle; // Remind users to use `trivial_providers::UnsupportedTimerHandle` if not supported
  fn create_timer(&self, signal: Signal) -> Option<Self::TimerHandle> { None }
}
```

`UnsupportedTimerHandle` should be defined as `pub enum UnsupportedTimerHandle {}` which is equivalent to Rust `!`, which means that it can never actually be created; that way you are guaranteeing it via the type system.

litebox/src/platform/mod.rs
```diff
@@ -112,6 +112,33 @@ pub trait ThreadProvider: RawPointerProvider {
...
+
+/// A handle to a platform timer created by [`TimerProvider::create_timer`].
+///
+/// Dropping the handle **must** cancel any pending timer and ensure that the
+/// associated callback will not fire after the drop returns.
```
jaybosamiya-ms: "associated callback"?  It is unclear what this refers to.

Also who is in charge of cancelling?  As I understand it, it is saying that the platform cancels pending timers, yes?  If so, the writing style should just be `Dropping the handle cancels any pending timer` or similar, and it is the platform's job to uphold this contract.

litebox/src/platform/trivial_providers.rs
```diff
@@ -8,8 +8,18 @@
 
 use super::{
     Punchthrough, PunchthroughError, PunchthroughProvider, PunchthroughToken, RawConstPointer,
-    RawMutPointer,
+    RawMutPointer, TimerHandle,
 };
+
+/// A [`TimerHandle`] stub for platforms that have not yet implemented
+/// [`super::TimerProvider`]. All methods panic with `todo!()`.
+pub struct StubTimerHandle(());
```
jaybosamiya-ms: Mentioned elsewhere, define this as a `pub enum UnsupportedTimerHandle {}` instead of supporting its creation.

litebox/src/platform/mod.rs
jaybosamiya-ms: Re "associated callback", I think (if I understand correctly) this is making a long-range assumption that connects to the signals provider stuff.  That _should_ be documented if that's the design you want to go for.  Ideally however, we do not use a long-range assumption like that and just literally have a callback registration right here?

litebox_platform_windows_userland/src/lib.rs
CvvT: The signals originated from host are not targeting a specific thread. They are process targeted. Thus, we can pick any thread to record it. The signal would be recorded in `shared_pending` shared by all threads. The first thread that calls `process_signals` would consume it.

litebox/src/platform/mod.rs
CvvT: Good point! Returning `Option` is cleaner. I updated the code.

litebox/src/platform/mod.rs
CvvT: > "associated callback"? It is unclear what this refers to.
> 
> Also who is in charge of cancelling? As I understand it, it is saying that the platform cancels pending timers, yes? If so, the writing style should just be `Dropping the handle cancels any pending timer` or similar, and it is the platform's job to uphold this contract.

My previous intention was that platform needs to delete the timer (e.g., call `timer_delete` on Linux) if it is not used anymore (i.e., dropped). I added another interface `delete_timer` to make it clear. it should be fine to remove the comment since it is obvious now.

CvvT: > Re "associated callback", I think (if I understand correctly) this is making a long-range assumption that connects to the signals provider stuff. That _should_ be documented if that's the design you want to go for. Ideally however, we do not use a long-range assumption like that and just literally have a callback registration right here?

This provider is to set up a timer and fire a specific signal when it fires. Shim does not need to register a callback and support registering callback seems complicate.

litebox/src/platform/trivial_providers.rs
CvvT: Fixed, thanks!

litebox/src/platform/mod.rs
CvvT: > Good point! Returning `Option` is cleaner. I updated the code.

Actually I changed it to return a `Result`.

PR #695, opened 2026-03-02
LVBS: Refactor per-CPU variables

sangho2:
This PR refactors the LVBS platform's `PerCpuVariables`, improving its security, safety, and performance.

Changes:
* Seed initial heap to remove the static `PER_CPU_VARIABLE_ADDRESSES` array. `RefCell` is no longer needed. 
* Heap-allocate `PerCpuVariables` on demand when each AP firstly enters VTL1.
* Refactor `PerCpuVariables` struct for `asm` access and strict interior mutability (`Cell` and `UnsafeCell`).
* Add safe accessor methods.
* Clarify BSP/AP distinction.

### DIFF
--- a/litebox_platform_lvbs/src/arch/x86/mod.rs
+++ b/litebox_platform_lvbs/src/arch/x86/mod.rs
@@ -26,11 +26,7 @@ pub(crate) use x86_64::structures::paging::mapper::{MappedFrame, TranslateResult
 pub fn get_core_id() -> usize {
     const CPU_VERSION_INFO: u32 = 1;
 
-    #[allow(
-        unused_unsafe,
-        reason = "cpuid_count is safe on stable 1.94+ but unsafe on nightly 1.91"
-    )]
-    let result = unsafe { cpuid_count(CPU_VERSION_INFO, 0x0) };
+    let result = cpuid_count(CPU_VERSION_INFO, 0x0);
     let apic_id = (result.ebx >> 24) & 0xff;
 
     apic_id as usize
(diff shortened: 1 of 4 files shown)


litebox_platform_lvbs/src/mshv/hvcall_vp.rs
```diff
@@ -233,7 +229,14 @@ pub fn init_vtl_ap(core: u32) -> Result<u64, HypervCallError> {
     // has high-canonical mappings, so these are ready to use as-is for the
     // AP's initial VP context.
     let rip: u64 = get_entry();
-    let rsp = get_address_of_special_page(VTL1_KERNEL_STACK_PAGE) + PAGE_SIZE as u64 - 1;
+    // SAFETY: We dont support concurrent AP/VTL initialization and thus share
+    // the same stack pointer. If we plan to support concurrent initialization,
+    // we should provide seperate stack pointers for each AP (which might not
+    // scale if there are several 100s of APs).
```
wdcui: For performance reasons, would we need to initialize APs concurrently if there are too many of them?

wdcui (approved):
LGTM. Thanks!

litebox_platform_lvbs/src/mshv/hvcall_vp.rs
sangho2: We would have two options:
1) Spin-lock the shared boot stack page until a heap-allocated stack is ready (re-introduce https://github.com/microsoft/litebox/pull/695/changes/c42157c5bcbe4733edf2d12f413bf94cbaf50540).
2) Pre-allocate temporal stack pages for all possible vCPUs and reclaim them later.

I implemented 1 (c42157c) but not 2 because I hesitated to pre-allocate 100s of pages (e.g., Azure HBv5 instances support up to 368 vCPUs) and then reverted it just because VTL0 currently does not do concurrent AP/VTL1 initialization (effectively no-op). We can re-introduce it for future proofing.

sangho2: Reapply 1. Now each AP spin-acquires a boot stack page until it allocates/uses its own per-CPU stack.

sangho2:
I'll update Rust version for LVBS once #702 is merged.

PR #697, opened 2026-03-02
Fix rare panic inside iced-x86

jaybosamiya-ms:
A rare panic occurs inside iced-x86 ([example](https://github.com/microsoft/litebox/actions/runs/22598319383/job/65473935696)) due to an integer underflow when the bytes that are being decoded are near a 4GiB boundary.  There is already a fix available+merged upstream (https://github.com/icedland/iced/pull/697) but a new version has not been released for iced-x86 onto crates.io in ~2 years (see https://github.com/icedland/iced/issues/762 for issue requesting a new release semi-recently).

This PR fixes this by simply making sure we are not passing iced-x86 a 4GiB-boundary-crossing bytes.

### DIFF
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -311,17 +311,7 @@ fn hook_syscalls_in_section(
     dl_sysinfo_int80: Option<u64>,
     trampoline_data: &mut Vec<u8>,
 ) -> Result<()> {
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
We plan to potentially remove iced-x86 from our dependencies, esp since we might want future ARM support, but at least this PR should fix some of the rare CI failures we see from time to time.

PR #701, opened 2026-03-05
Connect 9p with Linux shim

CvvT:
Implement a TCP transport based on socket APIs from Linux shim for connecting to 9p server.

Instead of directly using `SockFd` for read/write operations, we save its internal `NetworkProxy` so that we don't need to acquire lock on file table. Otherwise, we have a deadlock issue where syscalls like read/write over 9p fs acquire lock on file table and then trigger read/write on the `SockFd` which requires lock on file table again.

### DIFF
--- a/litebox_shim_linux/src/transport.rs
+++ b/litebox_shim_linux/src/transport.rs
@@ -131,7 +131,8 @@ impl transport::Write for ShimTransport {
     }
 }
 
-#[cfg(not(target_os = "windows"))]
+// require network support
+#[cfg(target_os = "linux")]
 #[cfg(test)]
 mod tests {
     extern crate std;


litebox_shim_linux/src/transport.rs
```diff
@@ -0,0 +1,316 @@
...
+            Ipv4Addr::new(ip[0], ip[1], ip[2], ip[3]),
+            port,
+        ))
+    }
+
+    fn connect_9p(
+        task: &crate::Task<crate::DefaultFS>,
+        server: &DiodServer,
+    ) -> nine_p::FileSystem<crate::Platform, ShimTransport> {
+        let addr = socket_addr([10, 0, 0, 1], server.port);
```
wdcui, line 222: Which code sets up this server?

CvvT: The tests would set up diod:
https://github.com/microsoft/litebox/blob/b2f48b1c89d5c58ba8374e36525dffd4b3c23e00/litebox_shim_linux/src/transport.rs#L171-L181

PR #704, opened 2026-03-06
Add litebox_packager tool for self-contained ELF packaging

wdcui:
## Summary

Add `litebox_packager`, a CLI tool that packages Linux ELF programs into self-contained `.tar` bundles for execution under LiteBox. The tool automates the full workflow: dependency discovery, syscall rewriting, and archive creation.

### New: `litebox_packager` crate
- Discovers shared library dependencies via `ldd`
- Rewrites all ELF binaries using `litebox_syscall_rewriter`
- Produces a `.tar` with all files at their expected absolute paths (copies instead of symlinks, since the tar RO filesystem does not support them)
- Embeds `litebox_rtld_audit.so` at `lib/litebox_rtld_audit.so` (built at compile time, always without DEBUG)
- Supports `--include HOST:TAR` for extra files, `--no-rewrite` to skip specific files, and `--verbose` for diagnostics
- Fails on missing shared library dependencies with an actionable error message

### Runner improvements (`litebox_runner_linux_userland`)
- Add `--program-from-tar` flag to load the program binary from the tar archive instead of the host filesystem
- Runtime validation that `--program-from-tar` requires `--interception-backend=rewriter`
- Change default interception backend from `seccomp` to `rewriter`
- Improved help text: program path docs clarify host vs tar behavior
- Actionable hint when program not found on host but `--initial-files` is set
- Fix `ancestor_modes_and_users.last().unwrap()` panic  proper error with `ok_or_else`

### Windows runner simplification (`litebox_runner_linux_on_windows_userland`)
- Simplify to always load programs from a tar archive (`--initial-files` is now required)
- Remove `--program-from-tar`, `--rewrite-syscalls`, and `--insert-file` flags (no longer needed since all programs must come from tar)
- Remove host filesystem program loading path (ancestor dir creation, Windows-to-Unix path conversion)
- Always inject `LD_AUDIT=/lib/litebox_rtld_audit.so` via `set_load_filter` so rewritten binaries find the audit library automatically
- Remove unused `litebox_syscall_rewriter` and `windows-sys` dependencies
- Update dynamic test to include hooked binary in tar and use tar-internal path
- Only print platform system information in debug builds

### Build fixes (`litebox_runner_linux_userland/build.rs`)
- `env_remove("DEBUG")` in release builds to prevent inheriting stale env
- Force rebuild of `litebox_rtld_audit.so` on profile change

## Usage

```bash
# Package a program (Linux)
litebox_packager /usr/bin/ls -o ls.tar

# Run on Linux
litebox-runner-linux-userland -Z \
  --initial-files ls.tar \
  --program-from-tar \
  /usr/bin/ls [args...]

# Run on Windows
litebox_runner_linux_on_windows_userland.exe -Z \
  --initial-files ls.tar \
  /bin/ls [args...]
```

## Testing

- Manually verified with `ls` packaged and run end-to-end on both Linux and Windows.
- Windows runner tests pass (static and dynamic linked programs).

### DIFF
--- a/litebox_packager/src/lib.rs
+++ b/litebox_packager/src/lib.rs
@@ -159,7 +159,7 @@ pub fn run(args: CliArgs) -> anyhow::Result<()> {
         }
     }
 
-    // Include extra files.
+    // Include extra files (these files will not be rewritten).
     for inc in &includes {
         if !added_tar_paths.insert(inc.tar_path.clone()) {
             bail!(
--- a/litebox_runner_linux_on_windows_userland/src/lib.rs
+++ b/litebox_runner_linux_on_windows_userland/src/lib.rs
@@ -39,8 +39,7 @@ pub struct CliArgs {
     ///
     /// All ELF binaries should be pre-rewritten with the syscall rewriter
     /// (e.g., via `litebox-packager`).
-    #[arg(long = "initial-files", value_name = "PATH_TO_TAR", value_hint = clap::ValueHint::FilePath,
-          requires = "unstable", help_heading = "Unstable Options")]
+    #[arg(long = "initial-files", value_name = "PATH_TO_TAR", value_hint = clap::ValueHint::FilePath)]
     pub initial_files: PathBuf,
 }
 
--- a/litebox_runner_linux_on_windows_userland/tests/loader.rs
+++ b/litebox_runner_linux_on_windows_userland/tests/loader.rs
@@ -331,7 +331,6 @@ fn run_dynamic_linked_prog_with_rewriter(
 
     // Run litebox_runner_linux_on_windows_userland with the tar file
     let mut args = vec![
-        "--unstable",
         // Tell ld where to find the libraries.
         // See https://man7.org/linux/man-pages/man8/ld.so.8.html for how ld works.
         // Alternatively, we could add a `/etc/ld.so.cache` file to the rootfs.


litebox_packager/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+                    tar_path,
+                    data: rewritten.clone(),
+                    mode,
+                });
+            }
+        }
+    }
+
+    // Include extra files.
+    for inc in &includes {
```
sangho2, line 163: If we include extra ELF files, we need to rewrite them. Or we should clarify that `--include` is not for ELF flies.

litebox_packager/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+
+    /// Include extra files in the tar.
+    /// Format: HOST_PATH:TAR_PATH (split on the first colon, so the tar path
+    /// may contain colons but the host path must not).
+    #[arg(long = "include", value_name = "HOST_PATH:TAR_PATH")]
+    pub include: Vec<String>,
+
+    /// Skip rewriting specific files (by their absolute path on the host).
+    #[arg(long = "no-rewrite", value_name = "PATH")]
+    pub no_rewrite: Vec<PathBuf>,
```
sangho2, line 39: The purpose/usage of `no_rewrite` is a bit unclear. Does it expect that we run `ldd` in advance to figure out ELF files to exclude?

litebox_packager/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+        }
+        tar_entries.push(TarEntry {
+            tar_path: inc.tar_path.clone(),
+            data,
+            mode,
+        });
+    }
+
+    // Include the rtld audit library so the rewriter backend can load it.
+    #[cfg(target_arch = "x86_64")]
```
sangho2, line 190: Packaging on a 32-bit system is a bit unclear. This packager expects a rewriter backend. In that sense, it might not support 32-bit at all (due to a lack of 32-bit rtld_audit).

litebox_packager/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+#![cfg(target_os = "linux")]
+
+use anyhow::{Context, bail};
+use clap::Parser;
+use std::collections::{BTreeMap, BTreeSet};
+use std::os::unix::fs::MetadataExt as _;
+use std::path::{Path, PathBuf};
+use tar::{Builder, Header};
+
+/// Package Linux ELF programs for execution under LiteBox.
```
sangho2, line 15: We don't need to worry about this for now, but we might want to consider some version info to see whether rewritten ELF files are stale (LiteBox rewriter might be updated).

litebox_runner_linux_userland/src/lib.rs
sangho2: This `std::path::absolute()` might be incompatible with a tar bundle.

litebox_runner_linux_on_windows_userland/src/lib.rs
```diff
@@ -59,46 +35,13 @@ pub struct CliArgs {
...
-    #[arg(long = "insert-file", value_hint = clap::ValueHint::FilePath,
-          requires = "unstable", help_heading = "Unstable Options")]
-    pub insert_files: Vec<PathBuf>,
-    /// Pre-fill the files in this tar file into the initial file system state
+    /// Tar archive containing the program, its shared libraries, and litebox_rtld_audit.so.
+    ///
+    /// All ELF binaries should be pre-rewritten with the syscall rewriter
+    /// (e.g., via `litebox-packager`).
     #[arg(long = "initial-files", value_name = "PATH_TO_TAR", value_hint = clap::ValueHint::FilePath,
           requires = "unstable", help_heading = "Unstable Options")]
```
sangho2: `required = true`?

litebox_packager/src/lib.rs
wdcui: The rewriter has version info in it: LITEBOX0 is used for that purpose.

litebox_packager/src/lib.rs
wdcui: This just adds some flexibility. Yes, it applies to the list of ELF files found by ldd.

litebox_packager/src/lib.rs
wdcui: Updated the comment.

litebox_packager/src/lib.rs
wdcui: I think we should just drop the support for 32-bit at some point. Windows userland doesn't support 32bit either.

litebox_runner_linux_userland/src/lib.rs
wdcui: Good catch. Fixed.

litebox_runner_linux_on_windows_userland/src/lib.rs
wdcui: Fixed.

PR #706, opened 2026-03-09
Add OCI container image support to litebox_packager

sangho2:
Add a new `--oci-image` flag to `litebox_packager` that pulls OCI container images from registries (e.g., `docker.io/library/alpine:latest`), extracts the rootfs, rewrites ELF syscalls in parallel (`rayon`), and produces a tar for use with `litebox-runner-linux-userland --program-from-tar`.

This enables running OCI container images under LiteBox without manually extracting and preparing the filesystem.

```bash
# Prepare a Python script
cat > hello.py <<'EOF'
import os
v = os.environ.get('PYTHON_VERSION', 'MISSING')
print(f'Hello from Python {v} via LiteBox')
EOF

# Package a Python slim image
cargo run -p litebox_packager -- \
  --oci-image docker.io/library/python:3.12-slim \
  --include hello.py:/app/hello.py \
  -o python_slim.tar

# Run it with LiteBox
cargo run -p litebox_runner_linux_userland -- \
  -Z \
  --env LD_LIBRARY_PATH=/lib64:/lib32:/lib:/usr/lib/x86_64-linux-gnu:/usr/local/lib \
  --initial-files python_slim.tar \
  --program-from-tar \
  /bin/sh /litebox/config_and_run.sh python3 /app/hello.py
```


litebox_packager/src/lib.rs
```diff
@@ -121,49 +141,201 @@ pub fn run(args: CliArgs) -> anyhow::Result<()> {
...
+                data: extracted.config_json,
+                mode: 0o644,
+            });
+        } else {
+            eprintln!("warning: tar already contains {CONFIG_JSON_TAR_PATH}, skipping");
+        }
+    }
+
+    {
+        const CONFIG_AND_RUN_TAR_PATH: &str = "litebox/config_and_run.sh";
```
wdcui, line 288: Should we update .gitignore to ignore these files?

sangho2: `config.json` and `config_and_run.sh` are written in `tempfile` and then a tar file.

PR #709, opened 2026-03-10
Use 9p fs in SNP runner

CvvT:
Use 9p fs instead of tar fs for SNP runner to get around the size limit of tar file (we don't support allocating large chunk of memory in kernel yet ( > 4MB)). Also removed previously temporary solution `load_file_from_host` that requires a custom interface. 

Because 9p fs requires network, we cannot construct `GlobalState` with both fs and net at the same time. Moving `fs` from `GlobalState` to `FileState` to get around it.

Also implemented `SignalProvider` for SNP so that ctrl-c works (this requires changes in [sandbox_driver](https://github.com/MSRSSP/sandbox_driver/tree/weiteng/snp).

With this PR, we can run iperf3 successfully:

```
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec   100 MBytes   838 Mbits/sec    0    191 KBytes
[  5]   1.00-2.00   sec  99.4 MBytes   834 Mbits/sec    0    191 KBytes
[  5]   2.00-3.00   sec  99.9 MBytes   838 Mbits/sec    0    191 KBytes
[  5]   3.00-4.00   sec  99.3 MBytes   833 Mbits/sec    0    191 KBytes
[  5]   4.00-5.00   sec  99.6 MBytes   835 Mbits/sec    0    211 KBytes
[  5]   5.00-6.00   sec  99.3 MBytes   833 Mbits/sec    0    211 KBytes
[  5]   6.00-7.00   sec  99.9 MBytes   837 Mbits/sec    0    211 KBytes
[  5]   7.00-8.00   sec  99.3 MBytes   834 Mbits/sec    0    211 KBytes
[  5]   8.00-9.00   sec  99.4 MBytes   833 Mbits/sec    0    211 KBytes
[  5]   9.00-10.00  sec  99.3 MBytes   834 Mbits/sec    0    211 KBytes
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec   995 MBytes   835 Mbits/sec    0             sender
[  5]   0.00-10.01  sec   994 MBytes   833 Mbits/sec                  receiver
```

### DIFF
--- a/litebox_runner_linux_on_windows_userland/tests/common/mod.rs
+++ b/litebox_runner_linux_on_windows_userland/tests/common/mod.rs
@@ -10,7 +10,7 @@ use litebox_platform_multiplex::Platform;
 
 pub struct TestLauncher {
     platform: &'static Platform,
-    shim_builder: litebox_shim_linux::LinuxShimBuilder<litebox_shim_linux::DefaultFS>,
+    shim_builder: litebox_shim_linux::LinuxShimBuilder,
     fs: litebox_shim_linux::DefaultFS,
 }
 
@@ -75,7 +75,7 @@ impl TestLauncher {
         self.fs.close(&fd).unwrap();
     }
 
-    pub fn test_load_exec_common(mut self, executable_path: &str) {
+    pub fn test_load_exec_common(self, executable_path: &str) {
         let fs = std::sync::Arc::new(self.fs);
         let argv = vec![
             CString::new(executable_path).unwrap(),


litebox/src/fs/nine_p/mod.rs
```diff
@@ -551,6 +551,8 @@ impl<Platform: sync::RawSyncPrimitivesProvider, T: transport::Read + transport::
         flags: super::OFlags,
         mode: super::Mode,
     ) -> Result<FileFd<Platform, T>, super::errors::OpenError> {
+        // TODO: we don't support non-blocking, so ignore that flag instead of returning an error
+        let flags = flags - OFlags::NONBLOCK;
```
CvvT, line 555: Our current FS assumes every operation is nonblocking, which does not hold for 9p fs over network. To give the illusion that 9p fs is nonblocking, we use spin loop in the `litebox_shim_linux` that waits until an operation is done. This is a temporary workaround that makes it work for the SNP runner, in which case it might be fine because the remote 9p server supposes to run locally at VMPL0. We should not use it elsewhere.

PR #711, opened 2026-03-10
Add scripts to run Unixbench

CvvT:
Asked agent to implement python scripts to run UnixBench on both Linux and Windows. It uses the default parameters for each test (I ran UnixBench with `--verbose` and give the output to copilot). Currently LiteBox does not support `context1`, `spawn`, `shell1`, `shell8` because they require `fork`. Also, no parallel test yet.

This should help us identify the performance bottlenecks.

Current result on Linux (WSL2 on Windows):
```
=====================================================================================
Benchmark            Unit           Native      LiteBox    Ratio   Overhead
-------------------------------------------------------------------------------------
dhry2reg             lps       397470665.9  431779110.2   1.0863     -8.63%
whetstone-double     MWIPS          7214.8       7198.5   0.9977      0.23%
execl                lps           53721.7       8399.0   0.1563     84.37%
fstime               KBps        1976184.3    2617336.0   1.3244    -32.44%
fsbuffer             KBps         537851.7     720344.3   1.3393    -33.93%
fsdisk               KBps        5628540.3    7100348.3   1.2615    -26.15%
pipe                 lps        16532878.4   15899678.8   0.9617      3.83%
syscall              lps        18699262.0   12341697.2   0.6600     34.00%
context1             lps           89310.6          N/A      N/A        N/A
spawn                lps          106908.7          N/A      N/A        N/A
shell1               lpm            5151.3          N/A      N/A        N/A
shell8               lpm            1907.0          N/A      N/A        N/A
=====================================================================================

Geometric mean ratio (LiteBox/Native): 0.8369
Average overhead: 16.31%
```

On Windows (Unixbench does not run on Windows natively):
```
=====================================================================================
Benchmark            Unit           Native      LiteBox    Ratio   Overhead
-------------------------------------------------------------------------------------
dhry2reg             lps               N/A  365450433.6      N/A        N/A
whetstone-double     MWIPS             N/A       8290.6      N/A        N/A
execl                lps               N/A       5488.3      N/A        N/A
fstime               KBps              N/A    2130723.7      N/A        N/A
fsbuffer             KBps              N/A     612306.7      N/A        N/A
fsdisk               KBps              N/A    6503502.3      N/A        N/A
pipe                 lps               N/A   14014623.1      N/A        N/A
syscall              lps               N/A   12715909.4      N/A        N/A
context1                               N/A          N/A      N/A        N/A
spawn                                  N/A          N/A      N/A        N/A
shell1                                 N/A          N/A      N/A        N/A
shell8                                 N/A          N/A      N/A        N/A
=====================================================================================
```

### DIFF
--- a/dev_bench/unixbench/run_unixbench.py
+++ b/dev_bench/unixbench/run_unixbench.py
@@ -345,9 +345,7 @@ def _run_litebox_cmd(
     # Use a plain string (not Path) so Windows doesn't convert / to \.
     litebox_tmpdir = "/tmp" if bench.binary == "fstime" else None
     cmd += bench.args(duration, litebox_tmpdir)
-
-    display_args = ' '.join(bench.args(duration))
-    print(f"  Running: {Path(cmd[0]).name} ... {display_args}")
+    print(f"  Running: {' '.join(cmd)}")
     t0 = time.monotonic()
     # Use a shorter timeout for alarm-based benchmarks under LiteBox,
     # since if SIGALRM isn't delivered the process will hang forever.


jaybosamiya-ms:
Thanks Weiteng!  Seems like the results are quite promising.

Fwiw, since we already have [`dev_bench`](https://github.com/microsoft/litebox/tree/main/dev_bench) it would be nice if the benchmarks could be made a part of that. I've already put in effort when building it to aim towards something we can use longer-term to keep track of and manage regressions and so on.

No worries if integrating into that framework is too difficult rn though (since as I understand it, unixbench is "fixed time, measure work units" type benchmark, while the framework I've built is a "fixed work units, measure time"), but just making a note here that we should probably have a single way to run all our benchmarks.

Basically, if we merge some version of this without integrating with `dev_bench`, it would be helpful to open up an issue to track this work.  Please feel free to assign me to it and I'll (eventually) get to it.

CvvT:
> Thanks Weiteng! Seems like the results are quite promising.
> 
> Fwiw, since we already have [`dev_bench`](https://github.com/microsoft/litebox/tree/main/dev_bench) it would be nice if the benchmarks could be made a part of that. I've already put in effort when building it to aim towards something we can use longer-term to keep track of and manage regressions and so on.
> 
> No worries if integrating into that framework is too difficult rn though (since as I understand it, unixbench is "fixed time, measure work units" type benchmark, while the framework I've built is a "fixed work units, measure time"), but just making a note here that we should probably have a single way to run all our benchmarks.
> 
> Basically, if we merge some version of this without integrating with `dev_bench`, it would be helpful to open up an issue to track this work. Please feel free to assign me to it and I'll (eventually) get to it.

Created #718 to track it.

CvvT:
There is another popular benchmark (lmbench) I'd like to run, but it requires `fork` support.

wdcui (approved):
I didn't review the python code. For the small changes to the two `lib.rs` file, should we put them in a separate PR?

CvvT:
> I didn't review the python code. For the small changes to the two `lib.rs` file, should we put them in a separate PR?

I should be fine to skip the python code as it is just for testing.

The small change in Linux userland platform is just a style change (changing from `.global` to `.globl` to be consistent with other places). The other one is to add `\tmp` to in memory fs so that Unixbench can run. I don't think we need to separate them.

PR #713, opened 2026-03-11
fix: correct smoltcp time computation in Network::now()

wdcui:
Swap the arguments in `Network::now()` from `zero_time.duration_since(now)` to `now.duration_since(zero_time)`. The reversed subtraction always returned `Duration::ZERO`, so smoltcp permanently saw time=0 and no timer-dependent TCP behavior (delayed ACKs, retransmits, keep-alives) ever fired.

### DIFF
--- a/litebox/src/net/mod.rs
+++ b/litebox/src/net/mod.rs
@@ -685,8 +685,10 @@ where
             // microseconds is roughly 250 years. If a system has been up for that long, then it
             // deserves to panic.
             i64::try_from(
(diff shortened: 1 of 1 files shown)


CvvT:
Interesting to see whether agent can find this or similar bugs. I created another agentic workflow at my forked repo: https://github.com/CvvT/litebox/pull/20

Edit: asked copilot to scan this file and this is the first bug it reported.

PR #714, opened 2026-03-11
Support OP-TEE dynamic shared memory

athvu:
This PR adds support for OP-TEE SMC function 'CallWithRegdArg', which uses registered (dynamic) shared memory

### DIFF
--- a/litebox_common_optee/src/lib.rs
+++ b/litebox_common_optee/src/lib.rs
@@ -2025,7 +2025,7 @@ impl From<&OpteeSmcArgsPage> for OpteeSmcArgs {
 /// OP-TEE SMC call arguments.
 #[derive(Clone, Copy, Default, FromBytes)]
 pub struct OpteeSmcArgs {
-    pub args: [usize; Self::NUM_OPTEE_SMC_ARGS],
+    args: [usize; Self::NUM_OPTEE_SMC_ARGS],
 }
 
 impl OpteeSmcArgs {
--- a/litebox_shim_optee/src/msg_handler.rs
+++ b/litebox_shim_optee/src/msg_handler.rs
@@ -64,6 +64,47 @@ fn page_align_up(len: u64) -> u64 {
     len.next_multiple_of(PAGE_SIZE as u64)
 }
 
+fn parse_optee_msg_args(
+    blob: &[u8],
+    has_rpc_arg: bool,
+) -> Result<(Box<OpteeMsgArgs>, Option<Box<OpteeRpcArgs>>), OpteeSmcReturnCode> {
+    // Parse main header from the private buffer.
+    let main_header = OpteeMsgArgsHeader::read_from_prefix(blob)
+        .map_err(|_| OpteeSmcReturnCode::EBadAddr)?
+        .0;
+
+    // Validate num_params from the snapshot.
+    if main_header.num_params as usize > OpteeMsgArgs::MAX_ARG_PARAM_COUNT {
+        return Err(OpteeSmcReturnCode::EBadCmd);
+    }
+
+    let main_size = optee_msg_args_total_size(main_header.num_params);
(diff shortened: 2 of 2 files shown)


litebox_common_optee/src/lib.rs
```diff
@@ -2025,7 +2025,7 @@ impl From<&OpteeSmcArgsPage> for OpteeSmcArgs {
 /// OP-TEE SMC call arguments.
 #[derive(Clone, Copy, Default, FromBytes)]
 pub struct OpteeSmcArgs {
-    args: [usize; Self::NUM_OPTEE_SMC_ARGS],
+    pub args: [usize; Self::NUM_OPTEE_SMC_ARGS],
```
sangho2: does it require `pub`?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -185,13 +185,38 @@ pub fn handle_optee_smc_args(
...
+            })
+        }
+        OpteeSmcFunction::CallWithRegdArg => {
+            // `OpteeMsgArgs` is located at the offset specified in args[3] within the shared memory region pointed by args[1]:args[2].
+            let (shm_ref, offset) = smc.optee_regd_shm_ref_and_offset()?;
+            let shm_info = shm_ref_map()
+                .get(shm_ref)
+                .ok_or(OpteeSmcReturnCode::EBadAddr)?;
+            let page_index = (shm_info.page_offset + offset) / PAGE_SIZE;
+            let offset_in_page = (shm_info.page_offset + offset) % PAGE_SIZE;
```
sangho2: it might better to do `shm_info.page_offset.checked_add(offset).ok_or(OpteeSmcReturnCode::EBadAddr)?;` to deal with potential integer overflow attacks. Also, we don't have to compute this twice.

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -185,13 +185,38 @@ pub fn handle_optee_smc_args(
...
+            if page_index >= shm_info.page_addrs.len() {
+                return Err(OpteeSmcReturnCode::EBadAddr);
+            }
+            let msg_args_addr = shm_info.page_addrs[page_index].as_usize() + offset_in_page;
+
+            let (msg_args, rpc_args) = read_optee_msg_args_from_phys(msg_args_addr, true)?;
```
sangho2: Unlike `CallWithArg` and `CallWithRpcArg`, this `MsgArgs` passed through dynamic shared memory might be stored in non-contiguous pages. Let me open an issue to deal with this later.

praveen-pk: LiteBox has `read_data_from_shm` which reads from offset `0`. Could you add another method to read with an offset and use that method here?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -185,13 +185,38 @@ pub fn handle_optee_smc_args(
...
             let msg_args_addr: usize = msg_args_addr.truncate();
             let (msg_args, rpc_args) = read_optee_msg_args_from_phys(msg_args_addr, true)?;
-            Ok(OpteeSmcResult::CallWithArg { msg_args, rpc_args })
+            Ok(OpteeSmcResult::CallWithArg {
+                msg_args,
+                rpc_args,
+                msg_args_phys_addr: msg_args_addr as u64,
+            })
+        }
+        OpteeSmcFunction::CallWithRegdArg => {
```
praveen-pk, line 208: just curios, when is this method invoked?

athvu: On the VTL0 side, the optee_smc_do_call_with_arg() function will pass OPTEE_SMC_CALL_WITH_REGD_ARG if the tee shared memory was dynamically allocated

PR #738, opened 2026-04-02
Syscall rewriter improvement: API/error handling clean up, no-std support and bug fixes

wdcui:
This PR improves `litebox_syscall_rewriter` by cleaning up the public API, hardening error handling, adding `no_std` support and fixing several bugs.

### DIFF
--- a/litebox_syscall_rewriter/Cargo.toml
+++ b/litebox_syscall_rewriter/Cargo.toml
@@ -11,7 +11,6 @@ clap = ["dep:clap"]
 
 [dependencies]
 iced-x86 = { version = "1.21", default-features = false, features = ["no_std", "decoder", "encoder", "instr_info"] }
-litebox_util_log = { version = "0.1.0", path = "../litebox_util_log" }
 object = { version = "0.36.7", default-features = false, features = ["elf", "read_core"] }
 thiserror = { version = "2.0.6", default-features = false }
 zerocopy = { version = "0.8", default-features = false, features = ["derive"] }
--- a/litebox_syscall_rewriter/Cargo.toml
+++ b/litebox_syscall_rewriter/Cargo.toml
@@ -3,14 +3,23 @@ name = "litebox_syscall_rewriter"
 version = "0.1.0"
 edition = "2024"
 
+[features]
+default = ["std", "anyhow", "clap"]
+std = []
+
 [dependencies]
-anyhow = "1.0"
-clap = { version = "4.5.32", features = ["derive"] }
-iced-x86 = "1.21"
-memmap2 = "0.9"
-object = { version = "0.36.7", default-features = false, features = ["elf", "read", "std"] }
+iced-x86 = { version = "1.21", default-features = false, features = ["no_std", "decoder", "encoder", "instr_info"] }
+object = { version = "0.36.7", default-features = false, features = ["elf", "read_core"] }
 thiserror = { version = "2.0.6", default-features = false }
-zerocopy = { version = "0.8", features = ["derive"] }
+zerocopy = { version = "0.8", default-features = false, features = ["derive"] }
+
+# Binary-only dependencies
+anyhow = { version = "1.0", optional = true }
+clap = { version = "4.5.32", features = ["derive"], optional = true }
+
+[[bin]]
+name = "litebox_syscall_rewriter"
+required-features = ["std", "anyhow", "clap"]
 
 [lints]
 workspace = true
--- a/litebox_syscall_rewriter/Cargo.toml
+++ b/litebox_syscall_rewriter/Cargo.toml
@@ -11,6 +11,7 @@ clap = ["dep:clap"]
 
 [dependencies]
 iced-x86 = { version = "1.21", default-features = false, features = ["no_std", "decoder", "encoder", "instr_info"] }
+litebox_util_log = { version = "0.1.0", path = "../litebox_util_log" }
 object = { version = "0.36.7", default-features = false, features = ["elf", "read_core"] }
 thiserror = { version = "2.0.6", default-features = false }
 zerocopy = { version = "0.8", default-features = false, features = ["derive"] }
--- a/litebox_syscall_rewriter/Cargo.toml
+++ b/litebox_syscall_rewriter/Cargo.toml
@@ -6,6 +6,8 @@ edition = "2024"
 [features]
 default = ["std", "anyhow", "clap"]
 std = []
+anyhow = ["dep:anyhow"]
+clap = ["dep:clap"]
 
 [dependencies]
 iced-x86 = { version = "1.21", default-features = false, features = ["no_std", "decoder", "encoder", "instr_info"] }
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -58,6 +58,21 @@ type Result<T> = core::result::Result<T, Error>;
 
 const BUN_FOOTER_MARKER: &[u8] = b"\n---- Bun! ----\n";
 
+/// Log a standardized warning about unpatchable syscall instructions.
+///
+/// Call this after [`hook_syscalls_in_elf`] or [`patch_code_segment`] when the
+/// returned `skipped_addrs` list is non-empty.
+pub fn warn_skipped(path: &str, skipped_addrs: &[u64]) {
+    if !skipped_addrs.is_empty() {
+        litebox_util_log::warn!(
+            path:% = path,
+            count:% = skipped_addrs.len(),
+            addrs:? = skipped_addrs;
+            "unpatchable syscall instruction(s)"
+        );
+    }
+}
+
 /// The magic bytes used to identify the trampoline data.
 /// This is checked by the loader to verify that the trampoline is valid.
 pub const TRAMPOLINE_MAGIC: &[u8; 8] = b"LITEBOX0";
@@ -115,6 +130,13 @@ struct TextSectionInfo {
 /// addresses are syscall instructions that could not be patched because there
 /// is not enough space around the instruction (replaced with `icebp; hlt` so
 /// they trap instead of escaping to the host kernel).
+///
+/// Binaries that cannot or do not need to be patched (relocatable objects,
+/// non-ELF files, already-hooked binaries, binaries without executable
+/// sections or syscall instructions) are returned unchanged with an empty
+/// skipped list — these are not errors. Only genuinely broken inputs
+/// (corrupt ELF, unsupported executables like Bun, arithmetic overflow)
+/// produce `Err`.
 pub fn hook_syscalls_in_elf(
     input_binary: &[u8],
     trampoline: Option<u64>,
@@ -132,10 +154,7 @@ pub fn hook_syscalls_in_elf(
     if input_binary.len() >= 18 {
         let e_type = u16::from_le_bytes([input_binary[16], input_binary[17]]);
         if e_type == object::elf::ET_REL {
-            // ET_REL — relocatable object file
-            return Err(Error::UnsupportedObjectFile(
-                "relocatable object file".into(),
-            ));
+            return Ok((input_binary.to_vec(), Vec::new()));
         }
     }
 
@@ -159,7 +178,7 @@ pub fn hook_syscalls_in_elf(
         let arch = match file {
             object::File::Elf64(_) => Arch::X86_64,
             object::File::Elf32(_) => Arch::X86_32,
-            _ => return Err(Error::UnsupportedObjectFile("not an ELF file".into())),
+            _ => return Ok((input_binary.to_vec(), Vec::new())),
         };
 
         let dl_sysinfo_int80 = if arch == Arch::X86_32 {
@@ -168,10 +187,14 @@ pub fn hook_syscalls_in_elf(
             None
         };
 
-        let text_sections = text_sections(&file)?;
+        let text_sections = match text_sections(&file) {
+            Ok(sections) => sections,
+            Err(Error::NoTextSectionFound) => return Ok((input_binary.to_vec(), Vec::new())),
+            Err(e) => return Err(e),
+        };
 
         if is_already_hooked(&*buf, arch) {
-            return Err(Error::AlreadyHooked);
+            return Ok((input_binary.to_vec(), Vec::new()));
         }
 
         let control_transfer_targets = get_control_transfer_targets(arch, &*buf, &text_sections)?;
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -95,8 +95,8 @@ struct TextSectionInfo {
 /// The `trampoline` must be an absolute address if specified; if unspecified, it will be set to
 /// zeros, and it is the caller's decision to overwrite it at loading time.
 ///
-/// If it succeeds, it produces an executable with trampoline code appended at a page-aligned
-/// offset after the ELF file. The file layout is:
+/// If rewriting emits trampoline stubs, the returned executable has trampoline code appended at a
+/// page-aligned offset after the ELF file. The file layout is:
 /// `[original ELF][padding to page boundary][trampoline code][header]`
 ///
 /// The header at the end contains:
@@ -105,7 +105,9 @@ struct TextSectionInfo {
 /// - trampoline virtual address (8 bytes for 64-bit, 4 bytes for 32-bit)
 /// - trampoline size (8 bytes for 64-bit, 4 bytes for 32-bit)
 ///
-/// This layout allows loaders to read just the last 32/20 bytes to get the metadata.
+/// This layout allows loaders to read just the last 32/20 bytes to get the metadata. Even when
+/// no syscall instructions are patched, the rewriter still appends the header and the initial
+/// syscall-entry placeholder so the loader/audit path can tell the binary was processed.
 ///
 /// `skipped_addrs` receives the virtual addresses of any `syscall`
 /// instructions that could not be patched (replaced with `UD2` so they
@@ -175,7 +177,7 @@ pub fn hook_syscalls_in_elf(
 
         let control_transfer_targets = get_control_transfer_targets(arch, &*buf, &text_sections)?;
 
-        let trampoline_base_addr = find_addr_for_trampoline_code(&file);
+        let trampoline_base_addr = find_addr_for_trampoline_code(&file)?;
 
         let fork_to_vfork_patch = find_fork_vfork_patch(&file, &text_sections);
 
@@ -199,7 +201,6 @@ pub fn hook_syscalls_in_elf(
         let trampoline = u32::try_from(trampoline).map_err(|_| Error::TrampolineAddressTooLarge)?;
         trampoline_data.extend_from_slice(&trampoline.to_le_bytes());
     }
-
     // Patch syscalls in-place in buf
     for s in &text_sections {
         let section_data = section_slice_mut(buf, s)?;
@@ -222,16 +223,18 @@ pub fn hook_syscalls_in_elf(
     // Patch fork → vfork: overwrite the first bytes of __libc_fork with a
     // JMP to __libc_vfork. This prevents glibc's fork wrapper from running
     // post-fork handlers that corrupt shared state under vfork semantics.
-    if let Some((fork_file_offset, rel32)) = fork_to_vfork_patch {
+    if let Some((fork_file_offset, fork_patch_end, rel32)) = fork_to_vfork_patch {
         #[allow(clippy::cast_possible_truncation)]
         let off = fork_file_offset as usize;
-        if off + 5 <= buf.len() {
+        #[allow(clippy::cast_possible_truncation)]
+        let patch_end = fork_patch_end as usize;
+        if off + 5 <= buf.len() && patch_end <= buf.len() && off + 5 <= patch_end {
             buf[off] = 0xE9; // JMP rel32
             buf[off + 1..off + 5].copy_from_slice(&rel32.to_le_bytes());
         } else {
             return Err(Error::ParseError(format!(
-                "fork→vfork patch offset {off:#x} + 5 exceeds buffer length {}",
-                buf.len()
+                "fork→vfork patch range {off:#x}..{patch_end:#x} is invalid for buffer length {}",
+                buf.len(),
             )));
         }
     }
@@ -453,7 +456,11 @@ fn hook_syscalls_in_section(
         let replace_start = replace_start.unwrap();
         let replace_len = usize::try_from(replace_end - replace_start).unwrap();
 
-        let target_addr = trampoline_base_addr + trampoline_data.len() as u64;
+        let target_addr = checked_add_u64(
+            trampoline_base_addr,
+            trampoline_data.len() as u64,
+            "syscall trampoline target",
+        )?;
 
         // Copy the original instructions to the trampoline
         if replace_start < inst.ip() {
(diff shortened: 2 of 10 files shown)


litebox_syscall_rewriter/src/lib.rs
```diff
@@ -160,19 +210,33 @@ pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<u64>) -> Res
...
+    // post-fork handlers that corrupt shared state under vfork semantics.
+    if let Some((fork_file_offset, fork_patch_end, rel32)) = fork_to_vfork_patch {
+        #[allow(clippy::cast_possible_truncation)]
+        let off = fork_file_offset as usize;
+        #[allow(clippy::cast_possible_truncation)]
+        let patch_end = fork_patch_end as usize;
+        if off + 5 <= buf.len() && patch_end <= buf.len() && off + 5 <= patch_end {
+            buf[off] = 0xE9; // JMP rel32
+            buf[off + 1..off + 5].copy_from_slice(&rel32.to_le_bytes());
+        } else {
```
sangho2: Two potential issues:
1. Overwriting `endbr64` (if `fork_file_offset` is the starting address of `fork`)
2. No NOP-filling

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +473,412 @@ fn hook_syscalls_in_section(
...
+/// Find fork and vfork symbols in the ELF and compute the patch needed to
+/// redirect fork -> vfork. Returns `Some((fork_file_offset, jmp_rel32))` if
+/// both symbols are found, or `None` if this binary doesn't export fork.
+fn find_fork_vfork_patch(
+    file: &object::File<'_>,
+    text_sections: &[TextSectionInfo],
+) -> Option<(u64, u64, i32)> {
+    use object::ObjectSymbol as _;
+
+    let mut fork_vaddr = None;
```
sangho2: We might need to patch `_Fork` as well.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -94,8 +105,34 @@ struct TextSectionInfo {
...
+        return Err(Error::UnsupportedBunExecutable);
+    }
+
+    // Relocatable object files (.o) must not be patched: they are linker
+    // input, not executable code. Rewriting instructions or appending
+    // trampoline data would corrupt the object file for the linker.
+    // Check the ELF e_type field (bytes 16..18) before doing any work.
+    if input_binary.len() >= 18 {
+        let e_type = u16::from_le_bytes([input_binary[16], input_binary[17]]);
+        if e_type == 1 {
```
sangho2: nit. we could use `object::elf::ET_REL`.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -94,8 +105,34 @@ struct TextSectionInfo {
...
+) -> Result<Vec<u8>> {
+    if has_bun_footer_marker(input_binary) {
+        return Err(Error::UnsupportedBunExecutable);
+    }
+
+    // Relocatable object files (.o) must not be patched: they are linker
+    // input, not executable code. Rewriting instructions or appending
+    // trampoline data would corrupt the object file for the linker.
+    // Check the ELF e_type field (bytes 16..18) before doing any work.
+    if input_binary.len() >= 18 {
```
sangho2, line 147: why do we run this rewriter against `.o`?

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +473,412 @@ fn hook_syscalls_in_section(
...
             // Put jump back location into rcx.
-            let jmp_back_offset = i64::try_from(return_addr).unwrap()
-                - i64::try_from(trampoline_base_addr + trampoline_data.len() as u64 + 7).unwrap();
+            let jmp_back_base = checked_add_u64(
+                trampoline_base_addr,
+                trampoline_data.len() as u64,
+                "x86_64 trampoline jump-back base",
+            )?;
+            let jmp_back_base =
+                checked_add_u64(jmp_back_base, 7, "x86_64 trampoline jump-back base")?;
```
sangho2: since `trampoline_data.len()` can't be greater than `u64::MAX-7` (in general), we could combine these two.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +473,412 @@ fn hook_syscalls_in_section(
...
-            // So: disp32 = -(trampoline_data.len() + 4)
-            let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() + 4);
-            trampoline_data.extend_from_slice(&disp32.to_le_bytes());
+            // We want: RIP + disp32 = syscall_entry_addr
+            let entry_base = checked_add_u64(
+                trampoline_base_addr,
+                trampoline_data.len() as u64,
+                "x86_64 trampoline entry base",
+            )?;
+            let entry_base = checked_add_u64(entry_base, 4, "x86_64 trampoline entry base")?;
```
sangho2: similar

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +473,412 @@ fn hook_syscalls_in_section(
...
 
             // Add jmp back to original after syscall
-            let jmp_back_offset = i64::try_from(return_addr).unwrap()
-                - i64::try_from(trampoline_base_addr + trampoline_data.len() as u64 + 5).unwrap();
+            let jmp_back_base = checked_add_u64(
+                trampoline_base_addr,
+                trampoline_data.len() as u64,
+                "x86 trampoline jump-back base",
+            )?;
+            let jmp_back_base = checked_add_u64(jmp_back_base, 5, "x86 trampoline jump-back base")?;
```
sangho2: similar

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +473,412 @@ fn hook_syscalls_in_section(
...
+    };
+    for i in 0..e_phnum_usize {
+        let Some(entry_off) = new_start.checked_add(i.saturating_mul(e_phentsize_usize)) else {
+            break;
+        };
+        if entry_off + 16 > buf.len() {
+            break;
+        }
+        let p_type = u32::from_le_bytes(buf[entry_off..entry_off + 4].try_into().unwrap());
+        if p_type == 6 {
```
sangho2: `object::elf::PT_PHDR`

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +473,412 @@ fn hook_syscalls_in_section(
...
+        .checked_sub(i64::try_from(fork_vaddr).ok()? + 5)?;
+    let rel32 = i32::try_from(rel32).ok()?;
+
+    Some((fork_file_offset, fork_patch_end, rel32))
+}
+
+/// Check if the input binary has the Bun footer marker at the end.
+fn has_bun_footer_marker(input_binary: &[u8]) -> bool {
+    input_binary.len() >= BUN_FOOTER_MARKER.len()
+        && input_binary[input_binary.len() - BUN_FOOTER_MARKER.len()..] == *BUN_FOOTER_MARKER
```
sangho2: nits: `input_binary.ends_with(...)`

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -647,35 +1059,50 @@ fn hook_syscall_and_after(
...
-        // So: disp32 = -(trampoline_data.len() + 4)
-        let disp32 = -(i32::try_from(trampoline_data.len()).unwrap() + 4);
-        trampoline_data.extend_from_slice(&disp32.to_le_bytes());
+        // We want: RIP + disp32 = syscall_entry_addr
+        let entry_base = checked_add_u64(
+            trampoline_base_addr,
+            trampoline_data.len() as u64,
+            "x86_64 trampoline entry base",
+        )?;
+        let entry_base = checked_add_u64(entry_base, 4, "x86_64 trampoline entry base")?;
```
sangho2: similar

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -690,18 +1117,28 @@ fn hook_syscall_and_after(
...
 
     // Add jmp back to original after syscall
-    let jmp_back_offset = i64::try_from(replace_end).unwrap()
-        - i64::try_from(trampoline_base_addr + trampoline_data.len() as u64 + 5).unwrap();
+    let jmp_back_base = checked_add_u64(
+        trampoline_base_addr,
+        trampoline_data.len() as u64,
+        "trampoline jump-back base",
+    )?;
+    let jmp_back_base = checked_add_u64(jmp_back_base, 5, "trampoline jump-back base")?;
```
sangho2: similar

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -812,19 +1260,29 @@ fn hook_syscall_before_and_after(
...
     if need_jump_back {
         let return_addr = next_inst.next_ip();
-        let jmp_back_offset = i64::try_from(return_addr).unwrap()
-            - i64::try_from(trampoline_base_addr + trampoline_data.len() as u64 + 5).unwrap();
+        let jmp_back_base = checked_add_u64(
+            trampoline_base_addr,
+            trampoline_data.len() as u64,
+            "x86 trampoline jump-back base",
+        )?;
+        let jmp_back_base = checked_add_u64(jmp_back_base, 5, "x86 trampoline jump-back base")?;
```
sangho2: similar

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +473,412 @@ fn hook_syscalls_in_section(
...
+            // The PHDR segment size should match the phdr table; no change needed.
+        }
+    }
 }
 
-fn find_addr_for_trampoline_code(file: &object::File<'_>) -> u64 {
+/// Find fork and vfork symbols in the ELF and compute the patch needed to
+/// redirect fork -> vfork. Returns `Some((fork_file_offset, jmp_rel32))` if
+/// both symbols are found, or `None` if this binary doesn't export fork.
+fn find_fork_vfork_patch(
```
sangho2: better to check the prologue of `fork` to see whether it begins with `endbr64`. if it does, skip the first 4 bytes.

sangho2 (commented):
left some comments.

litebox_syscall_rewriter/src/lib.rs
wdcui: This is defense in depth. The broker right now just checks if it's an elf file so it may call the rewriter with a .o file.

litebox_syscall_rewriter/src/lib.rs
wdcui: Done. Thanks!

litebox_syscall_rewriter/src/lib.rs
wdcui: Good catch. handle it in the caller.

litebox_syscall_rewriter/src/lib.rs
wdcui: Good catch. Fixed.

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs
wdcui: I'm not sure about this one.

litebox_syscall_rewriter/src/lib.rs
wdcui: Done.

litebox_syscall_rewriter/src/lib.rs
wdcui: Done.

litebox_syscall_rewriter/src/lib.rs
wdcui: Done.

litebox_syscall_rewriter/src/lib.rs
wdcui: Done.

litebox_syscall_rewriter/Cargo.toml
```diff
@@ -3,14 +3,23 @@ name = "litebox_syscall_rewriter"
 version = "0.1.0"
 edition = "2024"
 
+[features]
+default = ["std", "anyhow", "clap"]
```
jaybosamiya-ms, line 7: Nit: we tend to use the explicit `dep:anyhow` rather than the implicit `anyhow` to pull in an optional dependency

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -29,6 +36,8 @@ pub enum Error {
     ParseError(String),
     #[error("unsupported executable")]
     UnsupportedObjectFile,
+    #[error("unsupported Bun-packaged executable")]
+    UnsupportedBunExecutable,
```
jaybosamiya-ms: I think overfitting on `bun` as part of the actual error type might be a step too far.  Minor tweak would be:
```rs
#[error("unsupported executable: {0}")]
UnsupportedExecutable(&'static str),
```

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -94,8 +105,34 @@ struct TextSectionInfo {
...
 ///
-/// This layout allows loaders to read just the last 32/20 bytes to get the metadata.
-pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<u64>) -> Result<Vec<u8>> {
+/// This layout allows loaders to read just the last 32/20 bytes to get the metadata. Even when
+/// no syscall instructions are patched, the rewriter still appends the header and the initial
+/// syscall-entry placeholder so the loader/audit path can tell the binary was processed.
+///
+/// `skipped_addrs` receives the virtual addresses of any `syscall`
+/// instructions that could not be patched (replaced with `UD2` so they
+/// trap instead of escaping to the host kernel).
```
jaybosamiya-ms: Could not be patched because?

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +487,421 @@ fn hook_syscalls_in_section(
...
+        .checked_sub(i64::try_from(fork_vaddr).ok()? + 5)?;
+    let rel32 = i32::try_from(rel32).ok()?;
+
+    Some((fork_file_offset, fork_func_end, rel32))
+}
+
+/// Check if the input binary has the Bun footer marker at the end.
+fn has_bun_footer_marker(input_binary: &[u8]) -> bool {
+    input_binary.ends_with(BUN_FOOTER_MARKER)
+}
```
jaybosamiya-ms: Nit: I see no reason for this to be a separate function wrt indirection, it replaces one line with one line, but the name of the function literally says _exactly_ the same thing too.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +487,421 @@ fn hook_syscalls_in_section(
...
+
+/// Replace an unpatchable syscall instruction with `UD2` (`0F 0B`) so that
+/// reaching it triggers SIGILL instead of silently escaping to the host kernel.
+///
+/// `syscall` (0F 05) and `int 0x80` (CD 80) are both 2 bytes — same size as
+/// `ud2`. For `call DWORD PTR gs:0x10` (7 bytes), the remaining 5 bytes are
+/// filled with NOPs.
+fn replace_with_ud2(section_data: &mut [u8], section_base_addr: u64, inst: &iced_x86::Instruction) {
```
jaybosamiya-ms: I wonder if picking a non-ud2 sequence might be better.  `ud2` _is_ actually common enough in practice in binaries to indicate `unreachable!()` and such.  If we pick something like (say) `int $0x00` (i.e., `cd 00`), then it is _also_ the same size, but does not have imply "unreachable" which might be better for future debugging?

Wtih some more thought: let's consider `icebp` (i.e., `f1`, aka `int1`) since it is a _strong_ indicator of "look here, interesting stuff!". It seems like that does not _actually_ trap on Linux at least (even though it would be nice).  _However_, combining that with `hlt` (i.e., `f4`) might be a _strong_ indicator, and fits into two bytes.  Specifically, `f1 f4` is guaranteed to trap in userspace, and is easier to check by a handler to see "oh, this is actually a _requested_ break, and not a spurious one".

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +487,421 @@ fn hook_syscalls_in_section(
...
+
+fn rel32_bytes(target: u64, base: u64, context: &'static str) -> Result<[u8; 4]> {
+    let disp = i128::from(target) - i128::from(base);
+    let disp = i32::try_from(disp).map_err(|_| {
+        Error::ParseError(format!(
+            "{context} displacement out of range: target {target:#x}, base {base:#x}"
+        ))
+    })?;
+    Ok(disp.to_le_bytes())
+}
```
jaybosamiya-ms, line 725: Nit: I'm not fully sure why these are `ParseError`.  These seem to be helpers that indeed do the right thing _for the current callers_ (afaict), but the names imply more general, at which point the errors would be actively annoying.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +487,421 @@ fn hook_syscalls_in_section(
...
+///
+/// # Returns
+///
+/// The trampoline stub bytes. The caller must copy them to
+/// `trampoline_write_vaddr`. Returns an empty `Vec` if no syscall
+/// instructions are found in `code`.
+///
+/// `skipped_addrs` receives the virtual addresses of any `syscall`
+/// instructions that could not be patched (replaced with `UD2` so they
+/// trap instead of escaping to the host kernel).
```
jaybosamiya-ms: Minor: I generally find arguments/returns sections in documentation to not be helpful, and just be unnecessary noise. In this case, some of the bits are useful, while others are noise.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +487,421 @@ fn hook_syscalls_in_section(
...
+pub fn patch_code_segment(
+    code: &mut [u8],
+    code_vaddr: u64,
+    trampoline_write_vaddr: u64,
+    syscall_entry_addr: u64,
+    skipped_addrs: &mut Vec<u64>,
+) -> Result<Vec<u8>> {
```
jaybosamiya-ms: `skipped_addrs` is used as a "return", so why is it an argument? Rust supports multiple returns, you can just return it normally, right?

jaybosamiya-ms (commented):
I have not done a deep dive into reviewing this PR, but I noticed a few things when looking through it.  I do think that the rewriter would benefit from a cleanup pass in the future, so for now, I do not want to block PRs to tweak it to get `no_std` support.

litebox_syscall_rewriter/src/lib.rs
sangho2: Both `libc_fork` and `_Fork` eventually invoke the `fork` syscall. The difference is whether they call `pthread_atfork` or not (mostly libc or other libraries' internal behaviors). Speaking of `pthread_atfork`, doing not call this function might result in library internal consistency break. https://docs.openssl.org/3.2/man3/OPENSSL_fork_prepare/ `libc_vfork` doesn't call atfork, so our rewriter might need to handle this.

wdcui: I will remove the fork->vfork patching. I think we should just work with the `fork` syscall (when we get to the support for multi processes and vfork/fork).

litebox_syscall_rewriter/Cargo.toml
wdcui: Fixed.

litebox_syscall_rewriter/src/lib.rs
wdcui: Good idea!

litebox_syscall_rewriter/src/lib.rs
wdcui: Updated.

litebox_syscall_rewriter/src/lib.rs
wdcui: I agree.

litebox_syscall_rewriter/src/lib.rs
wdcui: I don't fully understand your comment here. Did you mean to replace `ParseError` with something like `PatchError`?

jaybosamiya-ms: It is a bit of a nit, it is that `checked_sub_u64` (for example) does not indicate anything about it being parsing related, but it doing a `ParseError` is surprising, that's all.

It is fine to leave it as-is.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +451,315 @@ fn hook_syscalls_in_section(
...
+
+/// Replace an unpatchable syscall instruction with `ICEBP; HLT` (`F1 F4`) so
+/// that reaching it traps instead of silently escaping to the host kernel.
+///
+/// We avoid `UD2` (`0F 0B`) because it commonly appears in binaries to mark
+/// `unreachable!()` paths.  The `ICEBP; HLT` sequence is a strong, distinctive
+/// indicator of "this syscall was intentionally poisoned" — `ICEBP` alone does
+/// not trap on Linux in userspace, but `HLT` does (SIGILL in ring 3), and the
+/// `F1` prefix makes it easy for a signal handler to distinguish an
+/// intentional break from a spurious one.
```
jaybosamiya-ms: This comment is clearly incorrect.

```console
$ cat x.c
int main(void) {
    __asm__ volatile("hlt");
    return 0;
}

$ gcc -o x x.c -g

$ gdb ./x
Reading symbols from ./x...
(gdb) r
Starting program: /tmp/tmp.2026-04-07/15-35-06/polished-lumpsucker/x
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
main () at x.c:2
2               __asm__ volatile("hlt");
```

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -387,87 +451,315 @@ fn hook_syscalls_in_section(
...
+            // PT_PHDR — update p_offset to match new location
+            let p_offset_off = entry_off + 8;
+            let old_p_offset =
+                u64::from_le_bytes(buf[p_offset_off..p_offset_off + 8].try_into().unwrap());
+            if old_p_offset == e_phoff {
+                let new_p_offset = (old_p_offset + padding as u64).to_le_bytes();
+                buf[p_offset_off..p_offset_off + 8].copy_from_slice(&new_p_offset);
+            }
+            // The PHDR segment size should match the phdr table; no change needed.
+        }
```
sangho2, line 677: currently, this code does not update `PT_PHDR.p_vaddr`/`PT_PHDR.p_paddr`.

wdcui: Fixed.

litebox_syscall_rewriter/src/lib.rs

litebox_syscall_rewriter/src/lib.rs
wdcui: Now there is an AddressOverflow error.

litebox_syscall_rewriter/src/lib.rs
wdcui: It's removed.

jaybosamiya-ms (commented):
Top level comment: before merging, please do manually rewrite the PR description to a human-written description that accurately conveys the changes in this PR, rather than an LLM-written summary.  This description will become part of the repo history, so it is important to keep it clean.  For example, the description currently says "Removed fork-to-vfork ... code." but `find_fork_vfork_patch` does not exist on `main`, and was a function introduced by _and_ removed by the commits in this PR, so from the perspective of the PR diff, is completely irrelevant, and thus would be incorrect to claim in the description that it was removed by this PR.  Otherwise, if/when we need to deal with fork related stuff, this commit would lead to confusion and wasted time/effort.

Btw, I haven't done a deep review of this PR again after the changes, but I see that the comments I had from before seem to be ok, so I've marked them as resolved.

wdcui:
Ah, the current PR summary was updated by the agent without me knowing it. :-( I had a clean version at some point. Will update it.

jaybosamiya-ms (approved):
The new one-sentence description looks good to me.  Since this is primarily in the rewriter (which is not part of our TCB) I haven't done a deep review of the latest version, but I do not wish to block it.  @wdcui if you are happy with it, please feel free to merge.

wdcui:
Thank you, @jaybosamiya-ms.

@sangho2 and @CvvT , I will wait for your approvals before I merge this PR.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -94,8 +120,37 @@ struct TextSectionInfo {
...
 /// - trampoline size (8 bytes for 64-bit, 4 bytes for 32-bit)
 ///
-/// This layout allows loaders to read just the last 32/20 bytes to get the metadata.
+/// This layout allows loaders to read just the last 32/20 bytes to get the metadata. Even when
+/// no syscall instructions are patched, the rewriter still appends the header and the initial
+/// syscall-entry placeholder so the loader/audit path can tell the binary was processed.
+///
+/// Returns the rewritten binary. Binaries that cannot or do not need to be
+/// patched (relocatable objects, non-ELF files, already-hooked binaries,
+/// binaries without executable sections or syscall instructions) are returned
```
sangho2, line 129: What is the difference between this "binary without syscall instructions" and the above "no syscall instructions are patched?" Changed vs unchanged is a bit confusing.

sangho2 (approved):
Looks reasonable to me. Left a minor comment on documentation.

litebox_syscall_rewriter/src/lib.rs
wdcui: "no syscall instructions are patched" means "there is no syscall instruction in the binary". I udpated the comment.

PR #739, opened 2026-04-02
Runtime ELF patching, trampoline format changes, and rtld_audit removal

wdcui:
## Summary
- **Runtime ELF patching**: The shim's mmap hook now patches syscall instructions in executable segments on the fly, and the loader patches the main binary at load time when it lacks a trampoline.
- **Trampoline format changes**: Redzone reservation (128-byte SUB RSP), R11-based restart instead of LEA RCX, and RIP-relative instruction re-encoding in the rewriter.
- **rtld_audit removal**: The audit library build, injection, and LD_AUDIT environment setup are fully removed. Runtime patching replaces the audit library's role.


litebox_common_linux/src/loader.rs
```diff
@@ -567,6 +579,24 @@ pub trait ReadAt {
...
 
+impl ReadAt for &[u8] {
+    type Error = Errno;
+
+    fn read_at(&mut self, offset: u64, buf: &mut [u8]) -> Result<(), Self::Error> {
+        let offset: usize = offset.truncate();
+        let end = offset.checked_add(buf.len()).ok_or(Errno::ENODATA)?;
+        if end > self.len() {
+            return Err(Errno::ENODATA);
+        }
```
sangho2: `ENODATA` here is a bit weird. Something like `EFAULT`, `EINVAL`, `EOVERFLOW`, or `ENOBUFS` might be better.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -456,22 +488,94 @@ fn hook_syscalls_in_section(
...
+
+            // Put the address of the original JMP (call-site) into R11 so
+            // that SA_RESTART can rewind ctx.rip to re-enter the trampoline.
+            // The real `syscall` instruction clobbers R11 with RFLAGS, so
+            // this register is free from the guest's perspective.
+            // LEA R11, [RIP + disp32] = 4C 8D 1D <disp32>
+            let r11_disp = i64::try_from(replace_start).unwrap()
+                - i64::try_from(trampoline_base_addr + trampoline_data.len() as u64 + 7).unwrap();
+            trampoline_data.extend_from_slice(&[0x4C, 0x8D, 0x1D]); // LEA R11, [RIP + disp32]
+            trampoline_data.extend_from_slice(&(i32::try_from(r11_disp).unwrap().to_le_bytes()));
```
sangho2: `r11` is already so busy because the platform is using it as a scratch register.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -537,6 +537,8 @@ core::arch::global_asm!(
     "
     .section .tbss
     .align 8
+saved_r11:
+    .quad 0
```
sangho2: `saved_r11` is a bit misleading. better to explicitly specify what we are saving. `r11` is special because it stores `rflags` (will be clobbered always). Now, both rewriter and platform are trying to repurpose it.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -678,7 +703,7 @@ syscall_callback:
     push    r8          // pt_regs->r8
     push    r9          // pt_regs->r9
     push    r10         // pt_regs->r10
-    push    [rsp + 88]  // pt_regs->r11 = rflags
+    push    QWORD PTR gs:saved_r11@tpoff // pt_regs->r11 (syscall call-site from rewriter)
```
sangho2: Currently, `saved_r11` doesn't store `rflags`. Better to read the current (guest) rflags and store it to `r11`.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -581,7 +594,8 @@ syscall_callback:
     push    r8          // pt_regs->r8
     push    r9          // pt_regs->r9
     push    r10         // pt_regs->r10
-    push    [rsp + 88]  // pt_regs->r11 = rflags
+    mov     r10, gs:[0x28]          // recover guest R11 saved at entry
+    push    r10         // pt_regs->r11 = guest R11 (restart addr from rewriter)
```
sangho2: similarly, this no longer stores `rflags`, which is different from `PtRegs`'s expectation.

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -148,6 +152,79 @@ impl<FS: ShimFS> litebox_common_linux::loader::MapMemory for ElfFile<'_, FS> {
...
+            let end = core::cmp::min(offset + len, self.data.len());
+            let src = &self.data[offset..end];
+            let dest = MutPtr::<u8>::from_usize(address);
+            dest.copy_from_slice(0, src).ok_or(Errno::EFAULT)?;
+        }
+
+        // Set final permissions if different from the writable mapping above.
+        if !prot.write || prot.execute {
+            self.inner.protect(address, len, prot)?;
+        }
```
sangho2: This function never unmaps pages when these operations fail. Does it assume that the caller will unmap pages if there is an error?

litebox_common_linux/src/loader.rs
sangho2, line 239: The meaning/outcome of this function has been completely swapped (`Ok(())` <-> `Err(...)`), so it can break other callers. For example, `litebox_shim_optee` also calls this function but like  `parsed.parse_trampoline(&mut &file, task.global.platform.get_syscall_entry_point())?;` because it assumes that calling this function against a non-patched ELF (e.g., the LVBS scenario) returns `Ok()`.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -549,19 +549,32 @@ unsafe extern "C-unwind" fn run_thread_arch(thread_ctx: &mut ThreadContext, tls_
...
+    // All other registers hold guest state.
+    .globl  syscall_callback_redzone
+syscall_callback_redzone:
+    // Save guest R11 (restart address from rewriter trampoline) into
+    // TEB.ArbitraryUserPointer (gs:[0x28]) before the TLS index lookup
+    // clobbers R11.  This slot is per-thread and the window is very
+    // narrow: only ~20 instructions of inline asm with no API calls,
+    // no Rust code, and no DLL activity, so the ntdll loader (which
+    // also uses this slot for debugger communication) cannot interfere.
+    mov     gs:[0x28], r11
```
sangho2, line 571: Is it okay not to save/restore the original value of `gs:[0x28]`?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +409,474 @@ impl<FS: ShimFS> Task<FS> {
...
+        // Make the code segment writable for in-place patching.
+        if self
+            .sys_mprotect(
+                mapped_addr,
+                len,
+                ProtFlags::PROT_READ | ProtFlags::PROT_WRITE,
+            )
+            .is_err()
+        {
+            return true;
```
sangho2, line 773: We might need to write protect the trampoline before this return instruction (and some others).

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +409,474 @@ impl<FS: ShimFS> Task<FS> {
...
+    /// Returns (is_pre_patched, file_offset, vaddr, trampoline_size).
+    fn check_trampoline_magic(&self, fd: i32) -> (bool, u64, u64, u64) {
+        let Ok(stat) = self.sys_fstat(fd) else {
+            return (false, 0, 0, 0);
+        };
+        let file_size = stat.st_size;
+        if file_size < 32 {
+            return (false, 0, 0, 0);
+        }
+        let mut tail = [0u8; 32];
```
sangho2: nits. this is incompatible with 32-bit ELF.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +409,474 @@ impl<FS: ShimFS> Task<FS> {
...
+        fd: i32,
+        offset: usize,
+        syscall_entry: usize,
+    ) -> bool {
+        // Initialize patch state if this is the first mmap for this fd.
+        if offset == 0 {
+            self.init_elf_patch_state(fd, mapped_addr.as_usize());
+        }
+
+        let mut cache = self.global.elf_patch_cache.lock();
```
sangho2, line 597: wonder whether this heavy lock is acceptable.

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -536,6 +536,10 @@ impl<FS: ShimFS> Task<FS> {
 
     /// Handle syscall `close`
     pub(crate) fn sys_close(&self, fd: i32) -> Result<(), Errno> {
+        // Finalize any in-progress ELF patching for this fd (mprotect
+        // trampoline RW→RX) before closing the descriptor.
+        self.finalize_elf_patch(fd);
```
sangho2, line 575: nits. `fd` based trampoline management might be incompatible with `dup2/3`. Not so sure whether a program dup ELF files though.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +409,474 @@ impl<FS: ShimFS> Task<FS> {
...
+        let mut phdrs_buf = alloc::vec![0u8; phdrs_size];
+        match self.sys_read(fd, &mut phdrs_buf, Some(e_phoff)) {
+            Ok(n) if n == phdrs_buf.len() => {}
+            _ => return,
+        }
+
+        // Find highest PT_LOAD end (p_vaddr + p_memsz)
+        let mut max_load_end: u64 = 0;
+        for i in 0..e_phnum {
+            let ph = &phdrs_buf[i * e_phentsize..][..e_phentsize];
```
sangho2, line 473: we need to check whether `e_phentsize` contains a valid value.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -456,22 +488,94 @@ fn hook_syscalls_in_section(
...
+            // LEA RSP, [RSP - 0x80] = 48 8D 64 24 80
+            trampoline_data.extend_from_slice(&[0x48, 0x8D, 0x64, 0x24, 0x80]);
+
+            // Put the address of the original JMP (call-site) into R11 so
+            // that SA_RESTART can rewind ctx.rip to re-enter the trampoline.
+            // The real `syscall` instruction clobbers R11 with RFLAGS, so
+            // this register is free from the guest's perspective.
+            // LEA R11, [RIP + disp32] = 4C 8D 1D <disp32>
+            let r11_disp = i64::try_from(replace_start).unwrap()
+                - i64::try_from(trampoline_base_addr + trampoline_data.len() as u64 + 7).unwrap();
```
sangho2: nits. we might want to use checked_add/sub.

litebox_common_linux/src/loader.rs
wdcui: Changed to EINVAL.

litebox_syscall_rewriter/src/lib.rs
wdcui: Update the comment to explain the contract.

litebox_platform_linux_userland/src/lib.rs
wdcui: renamed it to `saved_restart_addr`.

litebox_platform_linux_userland/src/lib.rs
wdcui: Fixed.

litebox_platform_windows_userland/src/lib.rs
wdcui: Fixed.

litebox_shim_linux/src/loader/elf.rs
wdcui: Added a TODO.

wdcui: Create a new issue #762 to track it.

litebox_common_linux/src/loader.rs
wdcui: Fixed the callers of this function. Please help check if the fix is correct.

litebox_platform_windows_userland/src/lib.rs
wdcui: I think so. I didn't see a better way to solve this problem.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: I think we don't expect a process to load executables concurrently?

litebox_shim_linux/src/syscalls/file.rs
wdcui: fixed it in sys_dup. please help check if it's correct.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Added a check.

litebox_syscall_rewriter/src/lib.rs

litebox_common_linux/src/loader.rs
sangho2: checked the OP-TEE shim side fix. seems fine. We might need design discussion on `get_syscall_entry_point()` and overall syscall rewriting later.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +409,502 @@ impl<FS: ShimFS> Task<FS> {
...
+            let p_type = u32::from_le_bytes(ph[0..4].try_into().unwrap());
+            if p_type != 1 {
+                // PT_LOAD = 1
+                continue;
+            }
+            let p_vaddr = u64::from_le_bytes(ph[16..24].try_into().unwrap());
+            let p_memsz = u64::from_le_bytes(ph[40..48].try_into().unwrap());
+            let end = p_vaddr + p_memsz;
```
sangho2: nits. checked_add.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +409,502 @@ impl<FS: ShimFS> Task<FS> {
...
+        let e_type = u16::from_le_bytes(ehdr_buf[16..18].try_into().unwrap());
+
+        // Validate e_phentsize: must be at least sizeof(Elf64_Phdr) = 56 bytes,
+        // otherwise the field accesses (e.g. ph[40..48] for p_memsz) will panic.
+        if e_phentsize < 56 {
+            return;
+        }
+
+        // Read program headers to find max PT_LOAD end
+        let phdrs_size = e_phentsize * e_phnum;
```
sangho2: nits. checked_mul.

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -549,19 +554,37 @@ unsafe extern "C-unwind" fn run_thread_arch(thread_ctx: &mut ThreadContext, tls_
...
     // Get the TLS state from the TLS slot and clear the in-guest flag.
     mov     r11d, DWORD PTR [rip + {TLS_INDEX}]
     mov     r11, QWORD PTR gs:[r11 * 8 + TEB_TLS_SLOTS_OFFSET]
     mov     BYTE PTR [r11 + {IS_IN_GUEST}], 0
-    // Set rsp to the top of the guest context.
+    // Recover the restart address from the TEB slot and store it in TLS.
+    // We use SCRATCH as a temporary since all guest GPRs must be preserved
+    // and RSP modifications would break the stack pointer recovery below.
+    push    QWORD PTR gs:[0x28]
```
sangho2, line 579: stack overflow?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -165,15 +248,134 @@ pub(crate) struct ElfLoader<'a, FS: ShimFS> {
...
+            .suppress_elf_runtime_patch
+            .set(self.patched_data.is_some() || self.parsed.has_trampoline());
+        let result = if let Some(ref data) = self.patched_data {
+            let mut mapper = PatchedMapper {
+                inner: &mut self.file,
+                data,
+            };
+            self.parsed.load(&mut mapper, &mut &*platform)
+        } else {
+            self.parsed.load(&mut self.file, &mut &*platform)
```
sangho2: if this load fails, `self.file.task.suppress_elf_runtime_patch.set(false)` is never executed.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +409,502 @@ impl<FS: ShimFS> Task<FS> {
...
+                let tramp_addr = state.trampoline_addr;
+                let tramp_len = align_up(state.trampoline_file_size, PAGE_SIZE);
+
+                // Allocate RW region at the trampoline address.
+                let alloc_result = self
+                    .do_mmap_anonymous(
+                        Some(tramp_addr),
+                        tramp_len,
+                        ProtFlags::PROT_READ | ProtFlags::PROT_WRITE,
+                        MapFlags::MAP_ANONYMOUS | MapFlags::MAP_PRIVATE | MapFlags::MAP_FIXED,
```
sangho2: `MAP_FIXED_NOREPLACE` might be better.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -211,12 +222,62 @@ pub fn hook_syscalls_in_elf(
...
+            };
+            out.extend_from_slice(header.as_bytes());
+        }
+        return Ok((out, skipped_addrs));
+    }
+
+    // Patch fork → vfork: overwrite the first bytes of __libc_fork with a
+    // JMP to __libc_vfork. This prevents glibc's fork wrapper from running
+    // post-fork handlers that corrupt shared state under vfork semantics.
+    if let Some((fork_file_offset, fork_patch_end, rel32)) = fork_to_vfork_patch {
```
sangho2: This patch is introduced again?

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -195,7 +195,13 @@ impl<'a> FileAndParsed<'a> {
         let file = ElfFileInMemory::new(task, elf_buf);
         let mut parsed = litebox_common_linux::loader::ElfParsedFile::parse(&mut &file)
             .map_err(ElfLoaderError::ParseError)?;
-        parsed.parse_trampoline(&mut &file, task.global.platform.get_syscall_entry_point())?;
+        match parsed.parse_trampoline(&mut &file, task.global.platform.get_syscall_entry_point()) {
+            Ok(()) | Err(litebox_common_linux::loader::ElfParseError::UnpatchedBinary) => {
+                // Unpatched binary is expected in the LVBS scenario where not
+                // all binaries are rewritten. Proceed without a trampoline.
+            }
```
sangho2: Better to have TODO or open an issue to track this.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -638,6 +780,86 @@ fn fixup_phdr_alignment(buf: &mut [u8]) {
     }
 }
 
+/// Find fork and vfork symbols in the ELF and compute the patch needed to
+/// redirect fork -> vfork. Returns `Some((fork_file_offset, jmp_rel32))` if
```
sangho2: This function now returns 3-tuple.

litebox_packager/build.rs
jaybosamiya-ms, line 1: Now that the entire body of this has been removed, I don't think this file is needed anymore, right?

dev_bench/unixbench/prepare_unixbench.py
```diff
@@ -61,8 +61,8 @@ def prepare_benchmark(
     """
     Prepare a single benchmark using litebox_packager.
 
-    The packager discovers dependencies, rewrites all ELFs, and creates a tar
-    (including litebox_rtld_audit.so).  The rewritten main binary is extracted
+    The packager discovers dependencies, rewrites all ELFs, and creates a tar.
+    The rewritten main binary is extracted
     from the tar and placed alongside it.
```
jaybosamiya-ms: Nit: reflow text

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -2047,7 +2077,24 @@ impl ThreadContext<'_> {
...
+        // When the seccomp/systrap backend is active, syscall instructions are
+        // trapped via SIGSYS — no binary rewriting needed.
+        #[cfg(feature = "systrap_backend")]
+        if self
+            .seccomp_interception_enabled
+            .load(std::sync::atomic::Ordering::SeqCst)
+        {
+            return 0;
+        }
+
```
jaybosamiya-ms: I'm surprised by this conditional here. Do we _need_ to give a 0 here?  If it is not needed, it can just be kept at the non-tweaked default behavior, no?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -2714,7 +2761,12 @@ unsafe fn interrupt_signal_handler(
     // FUTURE: handle trampoline code, too. This is somewhat less important
     // because it's probably fine for the shim to observe a guest context that
     // is inside the trampoline.
-    if ip == syscall_callback as *const () as usize {
+    #[cfg(target_arch = "x86")]
+    let is_at_syscall_callback = ip == syscall_callback as *const () as usize;
+    #[cfg(target_arch = "x86_64")]
+    let is_at_syscall_callback = ip == syscall_callback_redzone as *const () as usize
+        || ip == syscall_callback as *const () as usize;
+    if is_at_syscall_callback {
```
jaybosamiya-ms: Seems somewhat repetitive, and also somewhat surprising.  `get_syscall_entry_point` only uses `redzone` variant, but this one checks both.  Wouldn't it be correct to just do a `ip == get_syscall_entry_point()`?

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -306,34 +303,10 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
...
-                    );
-                    fs.close(&fd)
-                        .expect("Failed to close /lib/litebox_rtld_audit.so");
-                });
-            }
-            InterceptionBackend::Seccomp => {
-                // No need to include rtld_audit.so for seccomp backend
-            }
+            InterceptionBackend::Rewriter | InterceptionBackend::Seccomp => {}
         }
```
jaybosamiya-ms: Superfluous code left in

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -396,7 +368,7 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
     match cli_args.interception_backend {
         InterceptionBackend::Seccomp => platform.enable_seccomp_based_syscall_interception(),
         InterceptionBackend::Rewriter => {
-            REQUIRE_RTLD_AUDIT.store(true, core::sync::atomic::Ordering::SeqCst);
+            // Runtime patching is handled by the shim's mmap hook — nothing to do here.
```
jaybosamiya-ms: I don't see why we need to promise here that it is run by the shim's mmap hook.  The "correct" thing to do is to have it be handled by the platform.  Specifically, the correct move is to actually trigger it upon an `allocate_pages`/`update_permissions`, and not introduce new stuff on the shim side.

litebox_shim_linux/src/loader/elf.rs
jaybosamiya-ms, line 1: Afaict, the shim should not be doing the rewriting, that should be the platform's job.  The shim is in charge of the actual handling of syscalls, the platform is in charge of making sure that syscalls actually show up to the shim.  allocate_pages/update_permissions from the platform traits are probably the correct places to hook in for the platform to trigger things.  If you want to keep shared pieces together across platforms, then litebox_common_linux is probably where the common bits should sit.  The actual invocation of the rewriter is not a shim-level concern, imho.

jaybosamiya-ms: Due to this above comment, I have not done a proper review of the exact code in this file, since I think it should (fully, or at least mostly) be left untouched by this PR.  I will look at the migrated code after we're done updating it.

litebox_shim_linux/src/syscalls/file.rs
jaybosamiya-ms, line 1: Similar to the elf.rs thing, nothing about elf patching should really be showing up here, right?

litebox_shim_linux/src/syscalls/mm.rs
jaybosamiya-ms, line 1: Similar to elf.rs

litebox_shim_optee/src/loader/elf.rs
jaybosamiya-ms: Out of curiosity: how is OPTEE-on-Linux supposed to work for unpatched binaries?

litebox_shim_linux/Cargo.toml
```diff
@@ -16,6 +16,7 @@ syscalls = { version = "0.6", default-features = false }
 seq-macro = "0.3"
 ringbuf = { version = "0.4.8", default-features = false, features = ["alloc"] }
 zerocopy = { version = "0.8", default-features = false, features = ["derive"] }
+litebox_syscall_rewriter = { version = "0.1.0", path = "../litebox_syscall_rewriter", default-features = false }
```
jaybosamiya-ms, line 20: See comment at elf.rs, the shim itself should not need to even be aware of the syscall rewriter.

dev_bench/unixbench/prepare_unixbench.py
wdcui: Done.

litebox_platform_linux_userland/src/lib.rs
wdcui: It's not easy to access LinuxUserland.get_syscall_entry_point() inside an extern "C" function, right?

litebox_runner_linux_userland/src/lib.rs
wdcui: It's gone now?

litebox_shim_linux/src/loader/elf.rs
wdcui: moved it to the load_mapped() function.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Agreed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: fixed

litebox_shim_optee/src/loader/elf.rs
wdcui: I reverted the change here.

litebox_syscall_rewriter/src/lib.rs
wdcui: the old code is removed.

litebox_syscall_rewriter/src/lib.rs
wdcui: the old code is removed.

wdcui:
This PR is split into 4 smaller PRs.

PR #740, opened 2026-04-02
Fix `fcntl(F_DUPFD)` to make it comply with `min_fd`

sangho2:
This PR fixes a bug of the current `fcntl(F_DUPFD)` implementation. `fcntl(F_DUPFD)` should return an `fd` greater than or equal to the `arg` (`min_fd`) but it doesn't.


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -1939,13 +1936,17 @@ impl<FS: ShimFS> Task<FS> {
...
+            match target {
+                DupFdRequest::Exact(target) => {
+                    if !rds.fd_into_specific_raw_integer(fd, target) {
+                        return Err(Errno::EBADF);
+                    }
+                    Ok(target)
+                }
+                DupFdRequest::LowestAvailable => Ok(rds.fd_into_raw_integer(fd)),
+                DupFdRequest::LowestAtOrAbove(min_fd) => {
+                    Ok(rds.fd_into_raw_integer_at_or_above(fd, min_fd))
```
CvvT: Weidong has a similar fix but do not require adding a new interface to the LiteBox:

https://github.com/microsoft/litebox/blob/b0bca9f685294a60f19a2a4dca0b77e1d9f9f5b1/litebox_shim_linux/src/syscalls/file.rs#L4362-L4369

sangho2: I did have a similar fix but there is a potential TOCTOU race.

CvvT: why is that? `rds` is locked at the top.

sangho2: Found that my old one and Weidong's one are similar but different. Revised this PR.
Draft for now because this PR is not urgent.

CvvT (approved):
LGTM, thanks! Although I found one issue with the `NOFILE` limit, it is not related to this PR. I will try to fix it in a separate PR.

PR #741, opened 2026-04-03
Cross-platform packager: OCI symlink resolution, rewrite-include, rtld_audit removal

wdcui:
## Summary

Make `litebox_packager` compile and work on non-Linux hosts (primarily Windows) for the OCI image packaging path.

- Remove crate-level `#![cfg(target_os = "linux")]` gate; gate only ldd-based host mode behind `#[cfg(target_os = "linux")]`
- Add cross-platform `file_mode()` helper (`MetadataExt::mode()` on Unix, permission-based heuristic elsewhere)
- **Cross-platform OCI symlink handling**: Track symlinks in-memory (`DeferredSymlink`) instead of creating OS symlinks (which require special privileges on Windows). Resolve symlink chains through an in-memory map and materialize as file copies or directory placeholders
- Add `--rewrite-include` CLI flag for dlopen'd libraries (e.g., NSS modules) that aren't discovered by the automatic dependency scan
- Add OCI platform resolver to always pull `linux/amd64` images regardless of host platform
- Change Bun-packaged executable detection from a warning to a hard error (these can't be rewritten)
- Switch tar format from GNU to USTAR (required by the no_std tar reader)
- Normalize `\` to `/` in tar entry paths for Windows compatibility
- Delete the no-op `build.rs` (rtld_audit fully removed from packager)

## Bug fixes (not in feature branch)

- **Whiteout symlink pruning**: OCI opaque whiteouts (`.wh..wh..opq`) and regular whiteouts (`.wh.<name>`) removed files from disk but didn't prune corresponding entries from the in-memory `symlinks` vec, causing `materialize_symlinks()` to resurrect deleted symlinks
- **Degenerate symlink target resolution**: `normalize_path()` could return an empty `PathBuf` for targets with excess `..` segments, causing `resolve_symlink_in_rootfs()` to match the rootfs directory itself

## Stack

- PR #738 (merged): Syscall rewriter interface changes
- PR #739: Runtime ELF patching + rtld_audit removal
- **This PR**: Cross-platform packager

## Test results

- 219 tests: 204 passed, 15 failed (all TUN/EPERM — no device permissions), 34 skipped
- fmt ✅, clippy ✅


litebox_packager/src/oci.rs
```diff
@@ -426,20 +555,244 @@ fn extract_tar<R: Read>(reader: R, rootfs: &Path) -> anyhow::Result<()> {
...
+/// filesystem (no symlink resolution, no existence checks). Strips any root
+/// component so the result is always a relative path.
+fn normalize_path(path: &Path) -> PathBuf {
+    let mut result = Vec::new();
+    for component in path.components() {
+        match component {
+            std::path::Component::ParentDir => {
+                result.pop();
+            }
+            std::path::Component::CurDir | std::path::Component::RootDir => {}
```
sangho2: We might need `std::path::Component::Prefix(_) => {}` for Windows hosts.

litebox_packager/src/lib.rs
```diff
@@ -1,21 +1,37 @@
...
+}
+
+#[cfg(not(unix))]
+fn file_mode(metadata: &std::fs::Metadata) -> u32 {
+    if metadata.permissions().readonly() {
+        0o644
+    } else {
+        0o755
+    }
+}
```
jaybosamiya-ms: This doc string will not correctly apply to the two functions.  My suggestion is to put the `#[cfg(...)]`  bits _inside_ the function body.

litebox_packager/src/lib.rs
```diff
@@ -54,6 +70,14 @@ pub struct CliArgs {
     #[arg(long = "include", value_name = "HOST_PATH:TAR_PATH")]
     pub include: Vec<String>,
 
+    /// Include extra ELF files in the tar **with** syscall rewriting.
+    /// Use this for shared libraries that are loaded at runtime via `dlopen`
+    /// (e.g., NSS modules like `libnss_dns.so.2`) and therefore not discovered
+    /// by the automatic dependency scan.
+    /// Format: HOST_PATH:TAR_PATH (same as `--include`).
+    #[arg(long = "rewrite-include", value_name = "HOST_PATH:TAR_PATH")]
+    pub rewrite_include: Vec<String>,
```
jaybosamiya-ms: Why not just auto-detect whether it is an ELF or not?  Don't all ELFs need to be rewritten?  Why should the user manually need to say "we need to rewrite this one"?

litebox_packager/src/lib.rs
```diff
@@ -378,20 +476,23 @@ fn finalize_tar(
 }
 
 // ---------------------------------------------------------------------------
-// Dependency discovery (via ldd)
+// Dependency discovery (via ldd) — Linux only
```
jaybosamiya-ms: Nit: this change seems to be re-documenting literally what is in the literal next line, no?

litebox_packager/src/lib.rs
```diff
@@ -593,6 +695,11 @@ fn rewrite_elf(data: &[u8], path: &Path, verbose: bool) -> anyhow::Result<Vec<u8
             }
             Ok(data.to_vec())
         }
+        Err(litebox_syscall_rewriter::Error::UnsupportedBunExecutable) => bail!(
+            "{} is a Bun-packaged executable and cannot be packaged as-is: \
+             tar-loaded programs must already contain LiteBox syscall trampolines",
+            path.display()
+        ),
```
jaybosamiya-ms: See bun packaging error comment on the other PR (#738)

litebox_packager/src/lib.rs
```diff
@@ -646,3 +747,17 @@ fn build_tar(entries: &[TarEntry], output: &Path) -> anyhow::Result<()> {
...
+    use std::path::Path;
+
+    #[test]
+    fn rewrite_elf_skips_non_elf_files() {
+        // Non-ELF data should be returned unmodified.
+        let data = b"#!/bin/sh\necho hello\n";
+        let result = rewrite_elf(data, Path::new("/tmp/script.sh"), false).unwrap();
+        assert_eq!(result, data);
+    }
+}
```
jaybosamiya-ms: This test already shows that you know how to check if ELF or not, and you are doing skipping; clearly `--rewrite-include` is not needed as a separate command then, right?

litebox_packager/src/oci.rs
```diff
@@ -426,20 +555,244 @@ fn extract_tar<R: Read>(reader: R, rootfs: &Path) -> anyhow::Result<()> {
...
+///
+/// On Windows, `Path::is_absolute()` requires a drive letter, so Unix-style
+/// paths like `/lib/foo` are not detected as absolute. This helper checks
+/// the raw string instead.
+fn is_unix_absolute(path: &Path) -> bool {
+    path.as_os_str()
+        .to_str()
+        .is_some_and(|s| s.starts_with('/'))
+        || path.is_absolute()
+}
```
jaybosamiya-ms, line 628: Why is there an `|| path.is_absolute()` here?

litebox_packager/src/oci.rs
```diff
@@ -642,4 +993,245 @@ mod tests {
...
+    #[test]
+    fn lookup_mode_defaults_to_644_when_nothing_available() {
+        let permissions = HashMap::new();
+        let mode = lookup_mode(
+            Path::new("nonexistent"),
+            Path::new("/no/such/file"),
+            &permissions,
+        );
+        assert_eq!(mode, 0o644);
+    }
```
jaybosamiya-ms: This is an absolutely massive number of tests that I see no good reason to exist (at least not as separate tests).  Are you sure these are all needed, or were they just generated by the LLM because the LLM wants to generate lots of tests?

jaybosamiya-ms (commented):
I haven't done a deep review of this PR, since it does not seem to require an in-depth review afaict.  It is also an overlong PR.  Nonetheless, I've added a few inline comments.

Also top-level comment: please simplify the PR description to make sure we don't have overlong descriptions showing up in the squashed history.

Also, maybe a better title is just "Cross-platform packager"?

litebox_packager/src/lib.rs
wdcui: This function is removed.

litebox_packager/src/oci.rs
wdcui: fixed.

litebox_packager/src/oci.rs
wdcui: fixed.

litebox_packager/src/oci.rs
wdcui: removed the trivial ones and consolidated the symlink tests into two.

litebox_packager/src/lib.rs
wdcui: removed.

litebox_packager/src/lib.rs
wdcui: removed.

litebox_packager/src/lib.rs
wdcui: removed.

litebox_packager/src/lib.rs
wdcui: Unified.

litebox_packager/src/lib.rs
```diff
@@ -627,7 +653,7 @@ fn build_tar(entries: &[TarEntry], output: &Path) -> anyhow::Result<()> {
     let mut builder = Builder::new(file);
 
     for entry in entries {
-        let mut header = Header::new_gnu();
+        let mut header = Header::new_ustar();
```
sangho2, line 656: Do we need to use USTAR? It does support up to 255-byte path lengths by default.

litebox_packager/src/oci.rs
```diff
@@ -383,20 +476,50 @@ fn extract_tar<R: Read>(reader: R, rootfs: &Path) -> anyhow::Result<()> {
...
+                .link_name()?
+                .context("symlink entry has no link name")?
+                .into_owned();
+            // A later layer may override this symlink, so remove any stale
+            // entry with the same rel_path.
+            symlinks.retain(|s| s.rel_path != path);
+            symlinks.push(DeferredSymlink {
+                rel_path: path.clone(),
+                link_target,
+            });
```
sangho2, line 510: A symlink can be overridden by not only a symlink but also any regular file/directory.

litebox_packager/src/lib.rs
```diff
@@ -194,7 +219,47 @@ pub fn run(args: CliArgs) -> anyhow::Result<()> {
...
+        }
+        let data = std::fs::read(&inc.host_path)
+            .with_context(|| format!("failed to read included file {}", inc.host_path.display()))?;
+        let mode = {
+            use std::os::unix::fs::MetadataExt as _;
+            std::fs::metadata(&inc.host_path)
+                .map(|m| m.mode())
+                .unwrap_or(0o755)
+        };
+        let rewritten = rewrite_elf(&data, &inc.host_path, args.verbose)?;
```
sangho2, line 247: This ignores `--no-rewrite`.

litebox_packager/src/lib.rs
```diff
@@ -194,7 +219,47 @@ pub fn run(args: CliArgs) -> anyhow::Result<()> {
...
+                inc.tar_path
+            );
+        }
+        let data = std::fs::read(&inc.host_path)
+            .with_context(|| format!("failed to read included file {}", inc.host_path.display()))?;
+        let mode = {
+            use std::os::unix::fs::MetadataExt as _;
+            std::fs::metadata(&inc.host_path)
+                .map(|m| m.mode())
+                .unwrap_or(0o755)
```
sangho2, line 245: `lookup_mode` in `oci.rs` uses `0o644` as a default value.

litebox_packager/src/oci.rs
```diff
@@ -383,20 +476,50 @@ fn extract_tar<R: Read>(reader: R, rootfs: &Path) -> anyhow::Result<()> {
...
+        // Unix-style relative paths reliably, so we resolve them ourselves
+        // after all layers are extracted.
+        if entry_type == tar::EntryType::Symlink {
+            let link_target = entry
+                .link_name()?
+                .context("symlink entry has no link name")?
+                .into_owned();
+            // A later layer may override this symlink, so remove any stale
+            // entry with the same rel_path.
+            symlinks.retain(|s| s.rel_path != path);
```
sangho2, line 506: nits. this linearly scans all symlinks.

litebox_packager/src/oci.rs
```diff
@@ -96,6 +102,17 @@ pub fn pull_and_extract(image_ref: &str, verbose: bool) -> anyhow::Result<Extrac
...
+            platform_resolver: Some(Box::new(|entries| {
+                entries
+                    .iter()
+                    .find(|entry| {
+                        entry.platform.as_ref().is_some_and(|p| {
+                            p.os == "linux".into() && p.architecture == "amd64".into()
+                        })
+                    })
+                    .map(|e| e.digest.clone())
+            })),
```
jaybosamiya-ms, line 115: Shouldn't you pull linux image, but the arch should match the actual host you are running on? Otherwise, running this on an M-series MacBook might be quite surprising, no?

litebox_packager/src/oci.rs
```diff
@@ -426,20 +555,241 @@ fn extract_tar<R: Read>(reader: R, rootfs: &Path) -> anyhow::Result<()> {
...
+/// all files under the target directory are duplicated under the symlink's
+/// path prefix so that paths like `lib64/ld-linux-x86-64.so.2` exist in the tar.
 ///
-/// **Directory symlinks** (e.g., `/lib64` → `/usr/lib64`) are expanded: all
-/// files under the target directory are duplicated under the symlink's path
-/// prefix so that paths like `/lib64/ld-linux-x86-64.so.2` exist in the tar.
-pub fn scan_rootfs(rootfs: &Path, verbose: bool) -> anyhow::Result<RootfsFileMap> {
+/// `permissions` provides Unix permission modes captured from tar headers
+/// during extraction, so permission bits are accurate on non-Unix hosts.
+#[allow(clippy::implicit_hasher)]
```
jaybosamiya-ms, line 763: Nit: prefer expect over allow

jaybosamiya-ms (commented):
A couple of minor comments.

Also, reminder for earlier top-level comment https://github.com/microsoft/litebox/pull/741#pullrequestreview-4072230637

litebox_packager/src/oci.rs
sangho2: fyi. this commit does a similar thing: https://github.com/microsoft/litebox/commit/4e1eee7c6e5552c37567189ea47548f2fdc8e70c

wdcui:
This PR is replaced by PR #815

PR #742, opened 2026-04-03
Suppress release-mode panics that ldelf/TA can trigger

sangho2:
This PR suppresses some release-mode kernel panics (e.g., panic, todo, expect, unwrap, ...) that ldelf/TA can trigger.

### DIFF
--- a/litebox_shim_optee/src/syscalls/cryp.rs
+++ b/litebox_shim_optee/src/syscalls/cryp.rs
@@ -119,16 +119,16 @@ impl Task {
             if let Some(handle) = cryp_state.get_object_handle(false)
                 && tee_obj_map.exists(handle)
             {
-                if cfg!(debug_assertions) {
-                    todo!("support two-key algorithms");
-                }
+                #[cfg(debug_assertions)]
+                todo!("support two-key algorithms");
+                #[cfg(not(debug_assertions))]
                 return Err(TeeResult::NotSupported);
             }
 
             let Some(cipher) = create_cipher(cryp_state.algorithm(), key, iv) else {
-                if cfg!(debug_assertions) {
-                    todo!("implement algorithm {}", cryp_state.algorithm() as u32);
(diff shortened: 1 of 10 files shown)


litebox_shim_optee/src/syscalls/cryp.rs
```diff
@@ -119,11 +119,17 @@ impl Task {
             if let Some(handle) = cryp_state.get_object_handle(false)
                 && tee_obj_map.exists(handle)
             {
-                todo!("support two-key algorithms");
+                if cfg!(debug_assertions) {
```
wdcui: Nit: Is #[cfg(debug_assertions)] better than cfg!(debug_assertions)?

litebox_shim_optee/src/syscalls/cryp.rs
```diff
@@ -180,9 +196,6 @@ impl Task {
                 }
                 *dst_len = src_slice.len();
             }
-            if last_block {
-                todo!("support algorithms which have a certain finalization logic");
-            }
             Ok(())
```
wdcui: Do we need an else branch above instead of always returning `Ok(())`?

wdcui (approved):
Thank you for the quality improvement.

litebox_shim_optee/src/syscalls/cryp.rs
sangho2: that will end up with
```rust
#[cfg(debug_assertions)]
todo!("support two-key algorithms");
#[cfg(not(debug_assertions))]
return Err(TeeResult::NotSupported);
```
to satisfy clippy.

litebox_shim_optee/src/syscalls/cryp.rs
sangho2: an else branch might be helpful to catch a bug. let me add it.

PR #743, opened 2026-04-03
Foundation types: platform traits, event extensions, sync changes, AnyMap clone

wdcui:
## Summary

Layer 0 of the core `litebox` crate changes for multi-process support — platform traits, event extensions, sync changes, and AnyMap clone support.

### Platform traits
- **`AddressSpaceProvider`** trait (new file) for per-process address space management
- **`RawMessageProvider`** for direct guest–broker byte channels (bypasses IP stack)
- **`StdioProvider` expansion**: terminal attributes, window size, nonblocking stdin, cancel, host TTY device info
- New error variants: `SendError::Io`, `ReceiveError::ProtocolError/Eof`, `StdioReadError::WouldBlock`
- `SystemInfoProvider::current_processor_number()`
- `Provider` supertrait now requires `+ RawMessageProvider + AddressSpaceProvider`

### Event / sync
- `IOPollable::needs_host_poll()` and `should_block_read()` default methods
- `Observer::prune_dead_observers()` — GCs stale `Weak` refs on register
- `Waker::ptr_eq()` utility
- `FutexManager::wait/wake` — `address_space_id` parameter for multi-process futex isolation
- `trace_fs` feature gate for `RawSyncPrimitivesProvider`

### Utilities
- `AnyMap`: `Clone` bound on `insert`, type-erased `Clone` impl for the map
- `PageManagementProvider::allocate_pages`: `noreserve` parameter
- `copy_pages`: fix chunk-length bug (was using `old_range.len()` per iteration instead of `min(remaining, ALIGN)`)

### Caller updates
- All platform crates (linux_userland, linux_kernel, LVBS): `RawMessageProvider` + `AddressSpaceProvider` impls, `noreserve` param
- `litebox_shim_linux`: `Clone` derives for metadata types, `address_space_id` in futex calls

## Stack
- PR 1c: #738 (syscall rewriter)
- PR 1a+1b: #739 (runtime ELF patching)
- **PR 2: #741 (cross-platform packager) ← base**
- **PR 3: this PR**


litebox/src/utilities/anymap.rs
jaybosamiya-ms, line 1: This is an unnecessarily complex change.  The `insert` here forces any types that exist in the `AnyMap` to _also_ be `Clone` (which is fine), but that means that the storage could just declare `+ Clone`, and not need to do all this song and dance about type-erased clone functions.  Is there an actual reason we cannot just declare `storage` as `HashMap<TypeId, Box<dyn Any + Send + Sync + Clone>>`?

litebox/src/event/observer.rs
jaybosamiya-ms, line 1: These changes should be factored out into a separate PR, they are orthogonal to the rest of the changes here.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+//! Address-space management types and traits for multi-process support.
+//!
+//! The [`AddressSpaceProvider`] trait is an **optional** South interface that
+//! platforms implement to manage per-process address spaces. On kernel platforms
+//! (e.g., LVBS) each process gets its own page table; on userland platforms the
+//! single host address space is partitioned into non-overlapping VA sub-ranges.
+//!
+//! See the multi-process design document at
+//! `docs/multi-process-single-address-space.md` §4 for the full
+//! rationale.
```
jaybosamiya-ms: Please don't keep misleading docs (e.g., `multi-process-single-address-space.md`).

Also,  currently this doc string is not telling me anything at all about multi process support.  Why is there documentation about `AddressSpaceProvider` being optional _here_, why not literally at `AddressSpaceProvider`?

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+pub enum ForkedAddressSpace<Id> {
+    /// Kernel platforms: independent copy-on-write copy with the full address
+    /// range. The child has its own page tables; CoW faults are resolved by
+    /// the platform.
+    Independent(Id),
+    /// Userland platforms: a new VA-range partition is assigned to the child.
+    /// Parent memory is shared (vfork-style semantics); the shim is
+    /// responsible for copying pages as needed.
+    SharedWithParent(Id),
+}
```
jaybosamiya-ms, line 27: I'm not convinced that asking both the shim and the platform to separately handle things is the right design.  Ideally, if every shim needs to be copy pages, then that should be provided by the core, not handled at the shim.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[error("no address space slots available")]
+    NoSpace,
+    /// The given address-space ID is not valid (already destroyed, never
+    /// created, etc.).
+    #[error("invalid address space id")]
+    InvalidId,
+    /// The platform does not support this operation.
+    #[error("operation not supported by this platform")]
+    NotSupported,
+}
```
jaybosamiya-ms, line 43: Nit: generally speaking for error types, we've only maintained `#[error(...)]` and no doc strings, because duplicating / keeping almost-equiv text is not valuable.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+///
+/// This is an **optional** trait — platforms that do not yet support
+/// multi-process may leave all methods at the default (which returns
+/// [`AddressSpaceError::NotSupported`]).
+///
+/// # Associated Type
+///
+/// `AddressSpaceId` is an opaque, lightweight handle that identifies one
+/// address space. It must be `Copy + Eq + Send + Sync` so it can be stored
+/// inside `ProcessContext` and passed across threads.
```
jaybosamiya-ms: Please do not document the associated type at the trait, document it at the associated type itself.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+    type AddressSpaceId: Copy + Eq + Send + Sync + core::fmt::Debug;
+
+    /// Create a new, empty address space.
+    ///
+    /// * Kernel platforms: allocate a top-level page table (e.g., PML4) and
+    ///   copy kernel entries.
+    /// * Userland platforms: claim a free VA-range partition.
+    fn create_address_space(&self) -> Result<Self::AddressSpaceId, AddressSpaceError> {
+        Err(AddressSpaceError::NotSupported)
+    }
```
jaybosamiya-ms, line 66: Generally speaking, prefer non-default methods _unless_ it is clearly obvious that forgetting one of the methods would be ok.  In the current design, is it ok for leaving one part as _accidentally_ not-supported, while other parts are supported by a platform?

Basically, force a platform to mark themselves as `NotSupported` if they don't want to support things, don't actually automatically set to `NotSupported` unless there is good design rationale for a _particular method_ being optional.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    /// Destroy an address space, releasing all associated resources.
+    ///
+    /// After this call, `id` is invalid and must not be reused.
+    fn destroy_address_space(&self, id: Self::AddressSpaceId) -> Result<(), AddressSpaceError> {
+        let _ = id;
+        Err(AddressSpaceError::NotSupported)
+    }
```
jaybosamiya-ms, line 74: If you don't use `Copy`/`Clone` for `AddressSpaceId` then you can enforce things in the type system, rather than needing to do it through documentation.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+    ///   copy.
+    /// * [`SharedWithParent`](ForkedAddressSpace::SharedWithParent) — userland:
+    ///   new VA partition, parent pages shared.
+    fn fork_address_space(
+        &self,
+        parent: Self::AddressSpaceId,
+    ) -> Result<ForkedAddressSpace<Self::AddressSpaceId>, AddressSpaceError> {
+        let _ = parent;
+        Err(AddressSpaceError::NotSupported)
+    }
```
jaybosamiya-ms, line 90: Why is this documentation re-stating whatever was already stated at `ForkedAddressSpace`?  Unnecessary duplication, please remove.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    /// Make `id` the active address space for the current CPU / thread.
+    ///
+    /// * Kernel: switch CR3 (or equivalent).
+    /// * Userland: typically a no-op.
```
jaybosamiya-ms: What does "typically a no-op" mean? When would it not be a no-op?

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+        let _ = id;
+        Err(AddressSpaceError::NotSupported)
+    }
+
+    /// Execute `f` with the given address space active, then restore the
+    /// previously active address space.
+    ///
+    /// On kernel platforms implementations **must** restore the prior address
+    /// space even if `f` panics (use a guard / RAII pattern). Userland
+    /// platforms may keep this as a thin wrapper.
```
jaybosamiya-ms: A better way to handle this would be to have two functions, one that the platform implements, and one that is provided _by the trait_ with the latter doing the correct restoration pattern.

Also, please document what happens on `abort` scenarios, where it is impossible to actually capture the rewind.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,151 @@
...
+    ///
+    /// When `true`, the shim eagerly copies all writable guest pages before
+    /// spawning the vfork child and restores them after the child execs or
+    /// exits. When `false` (the default), the shim marks writable pages
+    /// read-only and lazily snapshots individual pages on first write fault.
+    ///
+    /// Platforms where the exception/fault handler shares the guest address
+    /// space (e.g., Windows userland) must set this to `true` because a
+    /// CoW fault inside the handler itself would be fatal.
+    const EAGER_COW_FOR_VFORK: bool = false;
```
jaybosamiya-ms: vfork is a Linux/POSIX-ism, the documentation/writing here should not be referring to vfork.

litebox/src/platform/address_space.rs
jaybosamiya-ms, line 1: Overall, I would suggest writing a design rationale for this platform trait, since I am not convinced this is actually a general enough design.  The way it is currently set up, it seems to be setting up a whole bunch of hardcodes that a platform just sets up and the shim needs to be aware of, rather than having a coherent/consistent view.

litebox/src/platform/mock.rs
```diff
@@ -210,7 +218,7 @@ impl IPInterfaceProvider for MockPlatform {
 
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub(crate) struct MockInstant {
-    time: u64,
+    pub(crate) time: u64,
```
jaybosamiya-ms, line 221: Why are these `pub(crate)`?

litebox/src/platform/mod.rs
```diff
@@ -43,10 +47,12 @@ macro_rules! log_println {
 pub trait Provider:
     RawMutexProvider
     + IPInterfaceProvider
+    + RawMessageProvider
     + TimeProvider
     + PunchthroughProvider
     + DebugLogProvider
     + RawPointerProvider
+    + AddressSpaceProvider
```
jaybosamiya-ms, line 55: Minor: we plan to remove the _bare_ `Provider` super-trait, but nonetheless, adding it here contradicts the claim that `AddressSpacePRovider` is an optional trait.

litebox/src/platform/mod.rs
```diff
@@ -385,14 +391,48 @@ pub trait IPInterfaceProvider {
 /// A non-exhaustive list of errors that can be thrown by [`IPInterfaceProvider::send_ip_packet`].
 #[derive(Error, Debug)]
 #[non_exhaustive]
-pub enum SendError {}
+pub enum SendError {
+    /// The underlying device returned an I/O error. The packet was not sent.
+    #[error("I/O error on send: errno {0}")]
+    Io(i32),
+}
```
jaybosamiya-ms, line 401: i32 is hardcoding POSIX/Linux-ism here, please have actual proper error codes

litebox/src/platform/mod.rs
```diff
@@ -385,14 +391,48 @@ pub trait IPInterfaceProvider {
 /// A non-exhaustive list of errors that can be thrown by [`IPInterfaceProvider::send_ip_packet`].
```
jaybosamiya-ms: If you are using `SendError` for more than just `IPInterfaceProvider::send_ip_packet`, this doc string should be updated.

litebox/src/platform/mod.rs
```diff
@@ -385,14 +391,48 @@ pub trait IPInterfaceProvider {
...
     #[error("Receive operation would block")]
     WouldBlock,
+    #[error("IPC protocol error: oversized frame")]
+    ProtocolError,
+    #[error("Channel closed (EOF)")]
+    Eof,
+}
+
+/// A raw byte-stream channel for direct message passing between the guest and
+/// the broker (bypassing the IP network stack).
```
jaybosamiya-ms: As far as the shared core is concerned, it should not even know of things like "broker".

litebox/src/platform/mod.rs
```diff
@@ -385,14 +391,48 @@ pub trait IPInterfaceProvider {
...
+///
+/// When available, this provides a fast path for protocols like 9P that would
+/// otherwise pay the overhead of traversing two smoltcp stacks.
+///
+/// The default implementation returns [`ReceiveError::WouldBlock`] /
+/// [`SendError::Io`], indicating the channel is not available.  Platforms that
+/// support direct messaging override these methods.
```
jaybosamiya-ms: similar

litebox/src/platform/mod.rs
```diff
@@ -385,14 +391,48 @@ pub trait IPInterfaceProvider {
...
+pub trait RawMessageProvider {
+    /// Send bytes to the broker over the raw channel.
+    ///
+    /// Returns `Ok(n)` with the number of bytes sent, or an error.
+    fn send_raw_message(&self, _data: &[u8]) -> Result<usize, SendError> {
+        Err(SendError::Io(0))
```
jaybosamiya-ms: Why `Io(0)`? What does the `0` mean?

litebox/src/platform/mod.rs
```diff
@@ -385,14 +391,48 @@ pub trait IPInterfaceProvider {
...
+    }
+
+    /// Receive bytes from the broker over the raw channel.
+    ///
+    /// Returns `Ok(n)` with the number of bytes read into `buf`, or
+    /// [`ReceiveError::WouldBlock`] if no data is available yet.
+    fn recv_raw_message(&self, _buf: &mut [u8]) -> Result<usize, ReceiveError> {
+        Err(ReceiveError::WouldBlock)
+    }
```
jaybosamiya-ms, line 438: Shouldn't it be "unsupported" or similar?

litebox/src/platform/mod.rs
```diff
@@ -641,16 +683,244 @@ pub enum StdioStream {
...
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum StdioIoctlError {
+    /// The fd is not a terminal (ENOTTY).
+    #[error("not a terminal")]
+    NotATerminal,
+    /// The operation failed with an OS error code (errno on Linux, mapped
+    /// equivalent on other platforms).
+    #[error("ioctl failed: {0}")]
+    OsError(i32),
```
jaybosamiya-ms, line 699: Again, overly specific Linux-isms, not ok

litebox/src/platform/mod.rs
```diff
@@ -641,16 +683,244 @@ pub enum StdioStream {
     Stderr = 2,
 }
 
+/// A non-exhaustive list of errors from terminal operations on [`StdioProvider`].
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum StdioIoctlError {
+    /// The fd is not a terminal (ENOTTY).
```
jaybosamiya-ms: Why `ENOTTY` in the doc string?

litebox/src/platform/mod.rs
```diff
@@ -641,16 +683,244 @@ pub enum StdioStream {
...
+    OsError(i32),
+}
+
+/// Platform-agnostic terminal attributes, mirroring the fields of Linux
+/// `struct termios`.
+///
+/// The guest always runs Linux binaries and speaks the Linux terminal ABI.
+/// Platform implementations fill this struct using their native APIs (e.g.,
+/// direct ioctl forwarding on Linux, `GetConsoleMode`/`SetConsoleMode` on
+/// Windows).
```
jaybosamiya-ms: This doc string is just straight up wrong.  The guest does not always run Linux binaries.

litebox/src/platform/mod.rs
```diff
@@ -641,16 +683,244 @@ pub enum StdioStream {
...
+    pub c_oflag: u32,
+    /// Control mode flags.
+    pub c_cflag: u32,
+    /// Local mode flags.
+    pub c_lflag: u32,
+    /// Line discipline (typically `0` for `N_TTY`).
+    pub c_line: u8,
+    /// Control characters.
+    pub c_cc: [u8; 19],
+}
```
jaybosamiya-ms, line 723: Please do not define hardcoded Linux-isms like this unless there is _very_ good reason to.  For example, what does `c_cc` even mean?  What do the flags mean?  How would a non-Linux shim actually even use these?

litebox/src/platform/mod.rs
```diff
@@ -641,16 +683,244 @@ pub enum StdioStream {
...
+
+/// Platform-agnostic terminal attributes, mirroring the fields of Linux
+/// `struct termios`.
+///
+/// The guest always runs Linux binaries and speaks the Linux terminal ABI.
+/// Platform implementations fill this struct using their native APIs (e.g.,
+/// direct ioctl forwarding on Linux, `GetConsoleMode`/`SetConsoleMode` on
+/// Windows).
+#[derive(Debug, Clone)]
+pub struct TerminalAttributes {
```
jaybosamiya-ms, line 710: This should not be sitting in the platform module.

litebox/src/platform/mod.rs
```diff
@@ -641,16 +683,244 @@ pub enum StdioStream {
...
+            c_oflag: 0x0005, // OPOST | ONLCR
+            c_cflag: 0x04bf, // CS8 | CREAD | CLOCAL | B38400
+            c_lflag: 0x8a3b, // ECHO | ECHOE | ECHOK | ISIG | ICANON | IEXTEN | ECHOCTL | ECHOKE
+            c_line: 0,       // N_TTY
+            c_cc: [
+                0x03, 0x1c, 0x7f, 0x15, 0x04, 0x00, 0x01, 0x00, 0x11, 0x13, 0x1a, 0xff, 0x12, 0x0f,
+                0x17, 0x16, 0xff, 0x00, 0x00,
+            ],
+        }
+    }
```
jaybosamiya-ms, line 749: If you really want to make a Linux-ism (I'm not convinced it should be here, it should instead be sitting in the Linux shim), please name it `new_default_linux` or something like that.

litebox/src/platform/mod.rs
```diff
@@ -641,16 +683,244 @@ pub enum StdioStream {
...
+    pub c_line: u8,
+    /// Control characters.
+    pub c_cc: [u8; 19],
+}
+
+// Terminal attribute flag constants.
+const TERMATTR_ECHO: u32 = 0x0008;
+const TERMATTR_ICRNL: u32 = 0x0100;
+const TERMATTR_OPOST: u32 = 0x0001;
+const TERMATTR_ONLCR: u32 = 0x0004;
```
jaybosamiya-ms, line 729: Hardcoded linux-isms?

litebox/src/platform/mod.rs
```diff
@@ -641,16 +683,244 @@ pub enum StdioStream {
...
+    /// Returns `true` if the `ICRNL` input flag is set.
+    pub fn icrnl_enabled(&self) -> bool {
+        self.c_iflag & TERMATTR_ICRNL != 0
+    }
+
+    /// Returns `true` if output post-processing with newline translation
+    /// (`OPOST | ONLCR`) is enabled.
+    pub fn onlcr_enabled(&self) -> bool {
+        (self.c_oflag & TERMATTR_OPOST != 0) && (self.c_oflag & TERMATTR_ONLCR != 0)
+    }
```
jaybosamiya-ms, line 765: Why are these pub functions necessary?  The particular flags are `pub` already, so these are just utility functions?  Seems bad to have these here.

litebox/src/platform/mod.rs
```diff
@@ -641,16 +683,244 @@ pub enum StdioStream {
...
+    /// Apply immediately (Linux `TCSETS`).
+    Now,
+    /// Drain output first, then apply (Linux `TCSETSW`).
+    AfterDrain,
+    /// Drain output first, flush pending input, then apply (Linux `TCSETSF`).
+    AfterDrainFlushInput,
+}
+
 /// A provider of standard input/output functionality.
 pub trait StdioProvider {
```
jaybosamiya-ms, line 794: All the tweaks here are making it _incredibly_ Linux-specific.  At minimum, I would suggest a design doc that compares Linux and Windows on these, and suggests a better middle ground, rather than just becoming a thin Linux veneer.

litebox/src/platform/mod.rs
```diff
@@ -666,6 +936,14 @@ pub trait SystemInfoProvider {
...
     /// if the platform does not support or provide a VDSO.
     fn get_vdso_address(&self) -> Option<usize>;
+
+    /// Returns the current processor number exposed to guest compatibility features.
+    ///
+    /// Platforms that do not expose a stable processor identifier, or that
+    /// virtualize CPU topology, may return `0`.
+    fn current_processor_number(&self) -> u32 {
+        0
+    }
```
jaybosamiya-ms: I'm not sure I understand what "guest compatibility features" means.

litebox/src/sync/futex.rs
```diff
@@ -25,8 +25,13 @@ use thiserror::Error;
...
 ///
-/// Note: currently, this only supports "private" futexes, since it assumes only a single process.
-/// In the future, this may be expanded to support multi-process futexes.
+/// Supports both private and shared futexes. On userland platforms where all
+/// processes share a single flat virtual address space (with non-overlapping VA
+/// partitions), both private and shared futexes naturally work with
+/// `address_space_id = 0` because the same virtual address always refers to
+/// the same backing memory. On kernel platforms, the `address_space_id`
+/// discriminator prevents false aliasing between processes that might have
+/// overlapping virtual address ranges.
```
jaybosamiya-ms: This includes internal details (such as `address_space_id`) inside a public doc string.

litebox/src/sync/futex.rs
```diff
@@ -41,6 +46,10 @@ const HASH_TABLE_ENTRIES: usize = 256;
 
 struct FutexEntry<Platform: RawSyncPrimitivesProvider> {
     addr: usize,
+    /// Discriminator to prevent false aliasing across address spaces.
+    /// On userland (non-overlapping VA partitions) this is always 0.
+    /// On kernel platforms each process passes its AddressSpaceId as u64.
+    address_space_id: u64,
```
jaybosamiya-ms: Do we need to know what it is in userspace/kernel space _at this level_?  The code here does not care whether it is 0 or not, right?

litebox/src/sync/futex.rs
```diff
@@ -80,22 +96,29 @@ impl<Platform: RawSyncPrimitivesProvider + RawPointerProvider + TimeProvider>
     /// If `bitset` is `Some`, then the waiter is only woken if the wake call's
     /// `bitset` has a non-zero intersection with the waiter's mask. Specifying
     /// `None` is equivalent to setting all bits in the mask.
+    ///
+    /// `address_space_id` is an opaque discriminator that distinguishes futexes
+    /// at the same virtual address in different address spaces. On userland
+    /// platforms (non-overlapping VA partitions) pass `0`. On kernel platforms
+    /// pass the process's `AddressSpaceId` converted to `u64`.
```
jaybosamiya-ms: How is an `AddressSpaceId` converted to a `u64`? Isn't it supposed to be opaque?  Why not just use an opaque identifier?

litebox/src/sync/futex.rs
```diff
@@ -131,23 +154,30 @@ impl<Platform: RawSyncPrimitivesProvider + RawPointerProvider + TimeProvider>
     /// (subject to the `num_to_wake` limit). If `bitset` is `None`, then all
     /// waiters are eligible to be woken.
     ///
+    /// `address_space_id` must match the value passed to the corresponding
+    /// [`wait`](Self::wait) call.
+    ///
```
jaybosamiya-ms, line 153: What happens if it does not should be documented

litebox/src/sync/mod.rs
jaybosamiya-ms, line 1: I'm very confused by this change, why is this needed?

jaybosamiya-ms (changes requested):
This is an overlarge PR that conflates a lot of distinct concepts together, and should be broken down.  Nonetheless, I've left comments throughout.  I do not think in the current state (i.e., as one big PR) this is merge-able, or indeed even that a proper review can be done, since a bunch of the pieces require deeper design analysis, while some parts are easy enough to merge on their own.  By properly splitting it up, we might actually be able to move forward without it becoming bogged down for a long time.

PR #751, opened 2026-04-07
Prune dead observers on registration

wdcui:
Prune stale `Weak` references from the observer list every time a new observer is registered, preventing unbounded growth when observers are dropped without explicit removal.

### DIFF
--- a/litebox/src/event/observer.rs
+++ b/litebox/src/event/observer.rs
@@ -93,9 +93,28 @@ impl<E, F: EventsFilter<E>, Platform: RawSyncPrimitivesProvider> Subject<E, F, P
         }
     }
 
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms (commented):
Generally looks fine to me, adding @CvvT as a reviewer due to greater familiarity with the event system for final approval.

Minor nit: we don't need the `# Summary` in the PR description, or that it was split from #743.  It also does not need to document that it adds a unit test.  I'll update the PR description, but just making a note.

CvvT (approved):
This PR looks good to me, but I am not sure whether this is necessary. Did you encounter any issues because of too many stale observers?

PR #752, opened 2026-04-07
Add Clone support to AnyMap via type-erased clone functions

wdcui:
## Summary

- Add `Clone` impl for `AnyMap` using type-erased clone function pointers (since `Clone` is not object-safe, `Box<dyn Any + Clone>` is invalid Rust).
- Require `Clone` bound on all types stored in `AnyMap` (`insert()` and `set_entry_metadata()`/`set_fd_metadata()`).
- Add `#[derive(Clone)]` to shim types stored as fd metadata: `StdioStatusFlags`, `PipeStatusFlags`, `SocketOptions`, `SocketOFlags`, `SocketProxy`.

Split from #743.

### DIFF
--- a/litebox/src/fd/mod.rs
+++ b/litebox/src/fd/mod.rs
@@ -477,7 +477,7 @@ impl<Platform: RawSyncPrimitivesProvider> Descriptors<Platform> {
     ) -> Option<T>
     where
         Subsystem: FdEnabledSubsystem,
-        T: core::any::Any + Send + Sync,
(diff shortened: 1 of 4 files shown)


jaybosamiya-ms (changes requested):
The design suggested in this PR feels somewhat brittle to me.  Thus, I've implemented a stricter variant in #771 that fully supersedes this PR.  No changes requested here, just keeping it open so we have a better understanding of the whole experiment at once, rather than missing this.

PR #753, opened 2026-04-07
Add needs_host_poll() and should_block_read() to IOPollable

wdcui:
## Summary

- Add `needs_host_poll()` default method to `IOPollable` — returns `false` by default; overridden by IO objects that require the host to perform the actual poll (e.g., host-backed stdin).
- Add `should_block_read()` default method to `IOPollable` — returns `true` by default; overridden by IO objects where reads should never block even when the fd is in blocking mode.
- Add `Waker::ptr_eq()` utility for comparing waker identity without waking.

Split from #743.

### DIFF
--- a/litebox/src/event/mod.rs
+++ b/litebox/src/event/mod.rs
@@ -48,6 +48,29 @@ pub trait IOPollable {
     /// calls are what notify observers. This particular function itself however _may_ be used to
     /// essentially get "the current status" of events for the system.
     fn check_io_events(&self) -> Events;
+
+    /// Returns `true` if this pollable cannot deliver asynchronous observer
+    /// notifications (e.g. host-backed stdin where the host has no callback
+    /// mechanism). Callers should use periodic polling instead of blocking
+    /// indefinitely on observer wakeups.
+    ///
+    /// Defaults to `false` (async notifications work). This is safe for all
+    /// existing implementors; callers that use this method arrive in subsequent
+    /// stacked PRs.
+    fn needs_host_poll(&self) -> bool {
+        false
+    }
+
+    /// Returns `true` if reads on this pollable should block when no data is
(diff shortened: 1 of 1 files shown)


litebox/src/event/mod.rs
```diff
@@ -48,6 +48,29 @@ pub trait IOPollable {
...
+    /// Returns `true` if reads on this pollable should block when no data is
+    /// available. Returns `false` for pollables whose callers perform
+    /// asynchronous readiness checks and expect a "would block" indication
+    /// immediately (e.g. PTY master side).
+    ///
+    /// Defaults to `true` (blocking reads). This is safe for all existing
+    /// implementors; callers arrive in subsequent stacked PRs.
+    fn should_block_read(&self) -> bool {
+        true
+    }
```
jaybosamiya-ms, line 73: A few comments here:
- documentation comments should not talk about "existing implementors" and such; they are meant to be long-term objects in the codebase
- similarly, callers and other stacked PRs should not be in the docs
- Also see top-level comment for this review

jaybosamiya-ms (changes requested):
Without the context of _why_ these are needed, which I would expect to see in the PR description _at least_, it is unclear what these are adding.  Neither the implementors nor the callers of this show up, so without context, this is simply increasing the interface size without clear indication of _why_.

Also, I am not convinced _at least as is_ this is a good design.  It is almost never the case that a boolean-returning design is the right move.  In this particular case, maybe it is `register_observer` that should return a `#[must_use]` object that forces you to poll for the `needs_host_poll` state?  I'm not even sure what the `should_block_read` is supposed to do here.

Importantly, _without_ the context of the users/implementors of this interface it is not clear to me this is a good interface.  And indeed since they are just boolean-returning things, they are essentially acting like config flags, which is more-often-than-not a bad interface design.

PR #754, opened 2026-04-07
Add noreserve parameter to allocate_pages and fix copy_pages chunk length

wdcui:
## Summary

- Add `noreserve: bool` parameter to `PageManagementProvider::allocate_pages()`. When true, Linux userland uses `MAP_NORESERVE` for sparse memory reservations. Kernel, LVBS, and Windows platforms accept but ignore it. All current callers pass `false` (no behavior change).
- Fix a bug in the default `remap_pages()` copy loop: `to_owned_slice()` was using `old_range.len()` (the entire range) instead of per-chunk length, reading beyond each chunk boundary.

Split from #743.

### DIFF
--- a/litebox/src/mm/linux.rs
+++ b/litebox/src/mm/linux.rs
@@ -509,6 +509,7 @@ impl<Platform: PageManagementProvider<ALIGN> + 'static, const ALIGN: usize> Vmem
                 MemoryRegionPermissions::from_bits(permissions).unwrap(),
                 vma.flags.contains(VmFlags::VM_GROWSDOWN),
                 populate_pages_immediately,
+                false,
                 platform_fixed_address_behavior,
             )
             .map_err(|err| match err {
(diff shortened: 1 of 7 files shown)


jaybosamiya-ms (changes requested):
The description needs to be updated to document what this might be used for.

Is it primarily to provide NORESERVE to a Linux shim?  If so, I don't think this PR is a good one to merge.  This is because it can cause some programs to seem like they work on Linux platform, and then break when you migrate it to a different platform.  _If_ the shim-level NORESERVE is the question, why not just squash it there, and force reservation?  Are there actual programs that break if you do it through forced reservation?  What is the plan to handle those programs on a Linux-on-Windows or similar scenario?

Also, minor nit: the number of boolean parameters to this particular function has now gone up once more; this means that every callee of it is (yet again) slightly harder to read.  Does not need to be fixed in this PR, just making a note.

PR #755, opened 2026-04-07
Misc cleanup: pub(crate) mock fields, log_println allow, current_processor_number, stdin guard

wdcui:
## Summary

- Make `MockPlatform` fields `pub(crate)` instead of fully public, since the mock is only used within the crate's test infrastructure.
- Add `#[allow(unused_imports)]` for `log_println!` macro import (used conditionally).
- Add `current_processor_number()` to `SystemInfoProvider` with a default returning `0`, for `getcpu`-family syscall emulation.
- Add a `Drop` guard to `MockPlatform::read_stdin()` to clear the stdin buffer on return.

Split from #743.

### DIFF
--- a/litebox/src/platform/mock.rs
+++ b/litebox/src/platform/mock.rs
@@ -210,7 +210,7 @@ impl IPInterfaceProvider for MockPlatform {
 
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub(crate) struct MockInstant {
-    time: u64,
+    pub(crate) time: u64,
 }
 
 impl Instant for MockInstant {
@@ -230,7 +230,7 @@ impl Instant for MockInstant {
 }
 
 pub(crate) struct MockSystemTime {
-    time: u64,
+    pub(crate) time: u64,
 }
 
 impl SystemTime for MockSystemTime {
@@ -290,6 +290,9 @@ impl RawPointerProvider for MockPlatform {
 
 impl StdioProvider for MockPlatform {
     fn read_from_stdin(&self, buf: &mut [u8]) -> Result<usize, StdioReadError> {
+        if buf.is_empty() {
+            return Ok(0);
+        }
         let Some(front) = self.stdin_queue.write().unwrap().pop_front() else {
             return Err(StdioReadError::Closed);
         };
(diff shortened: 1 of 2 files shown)


litebox/src/platform/mock.rs
```diff
@@ -230,7 +230,7 @@ impl Instant for MockInstant {
 }
 
 pub(crate) struct MockSystemTime {
-    time: u64,
+    pub(crate) time: u64,
```
jaybosamiya-ms, line 233: This is actually making them from private into crate-public, which contradicts the claim in the PR description.  I don't understand why this change is done/needed.

litebox/src/platform/mod.rs
```diff
@@ -23,11 +23,13 @@ pub use page_mgmt::PageManagementProvider;
 #[macro_export]
 macro_rules! log_println {
     ($platform:expr, $s:expr) => {{
+        #[allow(unused_imports)]
         use $crate::platform::DebugLogProvider as _;
         $platform.debug_log_print($s);
     }};
     ($platform:expr, $($tt:tt)*) => {{
         use core::fmt::Write as _;
+        #[allow(unused_imports)]
```
jaybosamiya-ms, line 32: I'm not sure I understand why these `unused_imports` are being marked as `allow` here.  Is there a use case that actually breaks if you don't put the `allow`?

This part of the change is fine to keep btw, no issue _specifically_ due to the exact scenario it is in (i.e., a macro in a logging utility), but more broadly, introducing `allow`s is a code smell.

litebox/src/platform/mod.rs
```diff
@@ -666,6 +668,16 @@ pub trait SystemInfoProvider {
...
+
+    /// Returns the current processor number, used to emulate `getcpu`-family
+    /// syscalls and related VDSO interfaces.
+    ///
+    /// Platforms that do not expose a stable processor identifier, or that
+    /// virtualize CPU topology, may return `0`. Callers arrive in subsequent
+    /// stacked PRs.
+    fn current_processor_number(&self) -> u32 {
+        0
+    }
```
jaybosamiya-ms, line 680: If we are allowing platforms to just give 0, are there scenarios where it matters that we are accurate?  Why can't the shim just assume 0 always?  Otherwise, this will lead to making the interface wider unnecessarily.

Also, same comment as in other PRs, doc strings should not mention "subsequent stacked PRs" and such.

jaybosamiya-ms (changes requested):

PR #756, opened 2026-04-07
Add RawMessageProvider trait and expand SendError/ReceiveError variants

wdcui:
## Summary

- Add `RawMessageProvider` trait for direct guest-host byte channels that bypass the IP stack. Default impl returns `Err(Unavailable)`.
- Add `Provider` supertrait bound `+ RawMessageProvider`.
- Expand `SendError` with `Io(i32)` and `Unavailable` variants; expand `ReceiveError` with `ProtocolError` and `Eof` variants. Update doc strings for shared usage across `IPInterfaceProvider` and `RawMessageProvider`.
- Add stub impls in all platform crates (linux kernel, linux userland, LVBS, Windows userland) and the mock.

Split from #743.

### DIFF
--- a/litebox/src/platform/mod.rs
+++ b/litebox/src/platform/mod.rs
@@ -43,6 +43,7 @@ macro_rules! log_println {
 pub trait Provider:
     RawMutexProvider
     + IPInterfaceProvider
+    + RawMessageProvider
     + TimeProvider
     + PunchthroughProvider
     + DebugLogProvider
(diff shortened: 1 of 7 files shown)


litebox/src/platform/mod.rs
```diff
@@ -382,17 +383,54 @@ pub trait IPInterfaceProvider {
...
 
-/// A non-exhaustive list of errors that can be thrown by [`IPInterfaceProvider::send_ip_packet`].
+/// Errors from send operations on [`IPInterfaceProvider`] and [`RawMessageProvider`].
 #[derive(Error, Debug)]
 #[non_exhaustive]
-pub enum SendError {}
+pub enum SendError {
+    /// The underlying device returned an I/O error. The packet was not sent.
+    #[error("I/O error on send: errno {0}")]
+    Io(i32),
```
jaybosamiya-ms, line 392: As mentioned on #743, hardcoding an i32 here is wrong; pick actually relevant errors

litebox/src/platform/mod.rs
```diff
@@ -382,17 +383,54 @@ pub trait IPInterfaceProvider {
...
+    ProtocolError,
+    #[error("Channel closed (EOF)")]
+    Eof,
+}
+
+/// A raw byte-stream channel for direct message passing between the guest and
+/// the host (bypassing the IP network stack).
+///
+/// When available, this provides a fast path for protocols like 9P that would
+/// otherwise pay the overhead of traversing two smoltcp stacks.
```
jaybosamiya-ms, line 414: Nit: "network stacks" not "smoltcp stacks". Doc strings should not promise unnecessary internal details.

litebox/src/platform/mod.rs
```diff
@@ -382,17 +383,54 @@ pub trait IPInterfaceProvider {
...
+///
+/// The default implementation returns [`ReceiveError::WouldBlock`] /
+/// [`SendError::Unavailable`], indicating the channel is not available.
```
jaybosamiya-ms, line 417: I think the default should be `Unavailable` for _both_, because otherwise a shim has no way to know if it tries to recv _first_ whether it is just that something has not shown up (so it should try again) or whether it is never going to become available.

jaybosamiya-ms (changes requested):
Currently, neither the implementations on the platform sides, nor the implementations on the shim sides of how this interface will be used have shown up.  So it is a little hard to actually confirm if this is the right design.  As a concrete alternative design, maybe `IpInterfaceProvider` should be changed to be a `StreamProvider` that has send/recv but takes an extra `StreamType::IP` or `StreamType::Raw`?  I am not sure whether the current design or _that_ design is the nicer one, and it is hard to tell without seeing more. 

Nonetheless _for this interface_ this cleanup can be done in the future too, so I don't see any particular reason to block it on figuring the above bits up.  

I would like to see the nicer version of the i32 though, since that is unacceptable to merge, so currently marking it as "request changes" but it is quite close to an "approve" right now.

PR #757, opened 2026-04-07
Expand StdioProvider with terminal types and nonblocking stdin

wdcui:
## Summary

- Add `StdioReadError::WouldBlock` variant for nonblocking stdin reads.
- Add `StdioIoctlError` enum for terminal operation failures.
- Add platform-agnostic `TerminalAttributes` and `WindowSize` structs.
- Add `HostTtyDeviceInfo` struct for host terminal device metadata.
- Expand `StdioProvider` with methods: `get_terminal_attributes`, `set_terminal_attributes`, `get_window_size`, `set_window_size`, `read_stdin_nonblocking`, `cancel_stdin_read`, `host_tty_device_info`.
- Update `fs/devices.rs` to handle `StdioReadError::WouldBlock`.

Split from #743.

### DIFF
--- a/litebox/src/platform/mod.rs
+++ b/litebox/src/platform/mod.rs
@@ -611,6 +611,8 @@ where
 pub enum StdioReadError {
     #[error("input stream has been closed")]
     Closed,
+    #[error("input would block")]
+    WouldBlock,
 }
 
 /// A non-exhaustive list of errors that can be thrown by [`StdioProvider::write_to`].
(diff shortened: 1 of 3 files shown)


litebox/src/platform/mod.rs
```diff
@@ -641,16 +643,230 @@ pub enum StdioStream {
     Stderr = 2,
 }
 
+/// Errors from terminal operations on [`StdioProvider`].
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum StdioIoctlError {
```
jaybosamiya-ms, line 649: Nit: Consider `StdioTerminalError` instead

litebox/src/platform/mod.rs
```diff
@@ -641,16 +643,230 @@ pub enum StdioStream {
...
+    /// The stream is not a terminal.
+    #[error("not a terminal")]
+    NotATerminal,
+    /// The operation failed with an OS error code (errno on Linux, mapped
+    /// equivalent on other platforms).
+    #[error("ioctl failed: {0}")]
+    OsError(i32),
```
jaybosamiya-ms, line 656: This i32 overfits to Linux, needs a better generalized error mapping

litebox/src/platform/mod.rs
```diff
@@ -641,16 +643,230 @@ pub enum StdioStream {
...
+    pub c_oflag: u32,
+    /// Control mode flags.
+    pub c_cflag: u32,
+    /// Local mode flags.
+    pub c_lflag: u32,
+    /// Line discipline (typically `0` for `N_TTY`).
+    pub c_line: u8,
+    /// Control characters.
+    pub c_cc: [u8; 19],
+}
```
jaybosamiya-ms, line 680: I'm not _totally_ against this, but I need to see the Windows userland platform implementation of this to be convinced that this is actually fine.

As it currently stands, it is _extremely_ Linux centric (e.g., the flags literally hardcode the Linux ABI).

Also, "typically" begs the question of "when not"?  Also, what is `N_TTY` in the context of a platform-agnostic thing?

litebox/src/platform/mod.rs
```diff
@@ -641,16 +643,230 @@ pub enum StdioStream {
...
+            c_oflag: 0x0005, // OPOST | ONLCR
+            c_cflag: 0x04bf, // CS8 | CREAD | CLOCAL | B38400
+            c_lflag: 0x8a3b, // ECHO | ECHOE | ECHOK | ISIG | ICANON | IEXTEN | ECHOCTL | ECHOKE
+            c_line: 0,       // N_TTY
+            c_cc: [
+                0x03, 0x1c, 0x7f, 0x15, 0x04, 0x00, 0x01, 0x00, 0x11, 0x13, 0x1a, 0xff, 0x12, 0x0f,
+                0x17, 0x16, 0xff, 0x00, 0x00,
+            ],
+        }
+    }
```
jaybosamiya-ms, line 706: All this is hardcoding a _lot_ of Linux-ism.

litebox/src/platform/mod.rs
```diff
@@ -641,16 +643,230 @@ pub enum StdioStream {
...
+    /// Returns `true` if the `ICRNL` input flag is set.
+    pub fn icrnl_enabled(&self) -> bool {
+        self.c_iflag & TERMATTR_ICRNL != 0
+    }
+
+    /// Returns `true` if output post-processing with newline translation
+    /// (`OPOST | ONLCR`) is enabled.
+    pub fn onlcr_enabled(&self) -> bool {
+        (self.c_oflag & TERMATTR_OPOST != 0) && (self.c_oflag & TERMATTR_ONLCR != 0)
+    }
```
jaybosamiya-ms, line 722: Why are these functions necessary? Can't a shim just look at the public fields?

litebox/src/platform/mod.rs
```diff
@@ -641,16 +643,230 @@ pub enum StdioStream {
...
+}
+
+/// Platform-agnostic terminal window size.
```
jaybosamiya-ms, line 725: Why is this mentioning "_Platform-agnostic_ terminal window size" (emphasis mine).  This seems like the agent was told to remove "Linux" and make it platform agnostic, and it literally did that.  Unnecessary/superflous text should just be removed entirely from documentation.

litebox/src/platform/mod.rs
```diff
@@ -641,16 +643,230 @@ pub enum StdioStream {
...
+#[derive(Debug, Clone, Copy)]
+pub struct WindowSize {
+    /// Number of rows (height in characters).
+    pub rows: u16,
+    /// Number of columns (width in characters).
+    pub cols: u16,
+    /// Horizontal size in pixels (informational, often zero).
+    pub xpixel: u16,
+    /// Vertical size in pixels (informational, often zero).
+    pub ypixel: u16,
```
jaybosamiya-ms, line 735: If it is informational and often zero, why not just remove it?

jaybosamiya-ms (changes requested):
I think this PR needs the specific platform side implementations to actually figure out whether the interface defined here is good enough or not.  It is currently _drastically_ expanding out the platform interface with a very Linux-y view, which makes me worry that the cross-platform nature becomes harder.  At minimum, looking at the relative sizes of implementations of these interfaces for the Linux platform and the Windows platform should help.

I am also not convinced that merging this in directly with `StdioProvider` is the right move.  One possibility is to have `StdioProvider` have an associated type called `Terminal` which implements a `Terminal` trait that has all these bits inside it; this trait would not have any defaults set, _but_ `litebox::platform::trivial_providers` or such can provide a `NotATerminal: Terminal` type that platforms can just pick up if they don't want to support terminals.  That would also improve long term stability, since we would not be _accidentally_ regressing a fully terminal-supported platform to a partial-supported one if/when the interface evolves.

PR #758, opened 2026-04-07
Add AddressSpaceProvider trait for multi-process address space management

wdcui:
## Summary

- Add `AddressSpaceProvider` trait (new file `platform/address_space.rs`) for per-process address space management: `create_address_space`, `switch_address_space`, `destroy_address_space`, `EAGER_COW_ON_FORK` constant.
- Add `Provider` supertrait bound `+ AddressSpaceProvider`.
- Add stub impls in all platform crates (linux kernel, linux userland, LVBS, Windows userland) and the mock.

Stacked on #756 (RawMessageProvider) due to adjacent impl blocks in platform crates.

Split from #743.

### DIFF
--- a/litebox/src/platform/address_space.rs
+++ b/litebox/src/platform/address_space.rs
@@ -1,0 +1,137 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//! Address-space management types and traits for multi-process support.
+//!
+//! The [`AddressSpaceProvider`] trait is an **optional** South interface that
+//! platforms implement to manage per-process address spaces. Platforms may use
+//! separate page tables, VA-range partitioning, or other techniques to isolate
+//! address spaces.
+
+use core::ops::Range;
+use thiserror::Error;
+
+/// The result of forking an address space.
+///
+/// The variant tells the caller what kind of copy was created so it can adjust
+/// its behavior (e.g., whether to copy page contents or share them).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ForkedAddressSpace<Id> {
+    /// Independent copy-on-write copy with the full address range. The child
+    /// has its own backing structures; CoW faults are resolved by the
+    /// platform.
+    Independent(Id),
+    /// A new VA-range partition is assigned to the child. Parent memory is
+    /// shared; the shim is responsible for copying pages as needed.
+    SharedWithParent(Id),
+}
+
+/// Errors that can occur during address-space operations.
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum AddressSpaceError {
+    /// No free address-space slots or VA ranges available.
+    #[error("no address space slots available")]
+    NoSpace,
+    /// The given address-space ID is not valid (already destroyed, never
+    /// created, etc.).
+    #[error("invalid address space id")]
+    InvalidId,
+    /// The platform does not support this operation.
+    #[error("operation not supported by this platform")]
+    NotSupported,
+}
+
+/// A provider for managing per-process address spaces.
+///
+/// This is an **optional** trait — platforms that do not yet support
+/// multi-process may leave all methods at the default (which returns
+/// [`AddressSpaceError::NotSupported`]).
+///
+/// # Associated Type
+///
+/// `AddressSpaceId` is an opaque, lightweight handle that identifies one
+/// address space. It must be `Copy + Eq + Send + Sync` so it can be stored
+/// inside process contexts and passed across threads.
+pub trait AddressSpaceProvider {
+    /// Opaque identifier for an address space.
+    type AddressSpaceId: Copy + Eq + Send + Sync + core::fmt::Debug;
+
+    /// Create a new, empty address space.
+    ///
+    /// The platform allocates whatever backing structures are needed for the
+    /// new address space.
+    fn create_address_space(&self) -> Result<Self::AddressSpaceId, AddressSpaceError> {
+        Err(AddressSpaceError::NotSupported)
+    }
+
+    /// Destroy an address space, releasing all associated resources.
+    ///
+    /// After this call, `id` is invalid and must not be reused.
+    fn destroy_address_space(&self, id: Self::AddressSpaceId) -> Result<(), AddressSpaceError> {
+        let _ = id;
+        Err(AddressSpaceError::NotSupported)
+    }
+
+    /// Fork an address space from `parent`.
+    ///
+    /// Returns a [`ForkedAddressSpace`] indicating what kind of fork was
+    /// performed:
+    ///
+    /// * [`Independent`](ForkedAddressSpace::Independent) — full CoW copy.
+    /// * [`SharedWithParent`](ForkedAddressSpace::SharedWithParent) — new VA
+    ///   partition, parent pages shared.
+    fn fork_address_space(
+        &self,
+        parent: Self::AddressSpaceId,
+    ) -> Result<ForkedAddressSpace<Self::AddressSpaceId>, AddressSpaceError> {
+        let _ = parent;
+        Err(AddressSpaceError::NotSupported)
+    }
+
+    /// Make `id` the active address space for the current CPU / thread.
+    fn activate_address_space(&self, id: Self::AddressSpaceId) -> Result<(), AddressSpaceError> {
+        let _ = id;
+        Err(AddressSpaceError::NotSupported)
+    }
+
+    /// Execute `f` with the given address space active, then restore the
+    /// previously active address space.
+    ///
+    /// Implementations **must** restore the prior address space even if `f`
+    /// panics (use a guard / RAII pattern).
+    ///
+    /// The default returns [`AddressSpaceError::NotSupported`]. Platforms that
+    /// implement [`activate_address_space`](Self::activate_address_space) should
+    /// also override this method with a proper save/restore sequence.
+    fn with_address_space<R>(
+        &self,
+        id: Self::AddressSpaceId,
+        f: impl FnOnce() -> R,
+    ) -> Result<R, AddressSpaceError> {
+        let _ = (id, f);
+        Err(AddressSpaceError::NotSupported)
+    }
+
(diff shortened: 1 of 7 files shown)


litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,137 @@
...
+/// its behavior (e.g., whether to copy page contents or share them).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ForkedAddressSpace<Id> {
+    /// Independent copy-on-write copy with the full address range. The child
+    /// has its own backing structures; CoW faults are resolved by the
+    /// platform.
+    Independent(Id),
+    /// A new VA-range partition is assigned to the child. Parent memory is
+    /// shared; the shim is responsible for copying pages as needed.
+    SharedWithParent(Id),
```
jaybosamiya-ms, line 26: When are pages expected to be copied by the shim?  Why not have the platform itself copy the pages over if so?  Alternatively, even for "SharedWithParent", it should be possible to often set up some copy-on-first-access behavior by the platform, right?

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// The platform does not support this operation.
+    #[error("operation not supported by this platform")]
+    NotSupported,
+}
+
+/// A provider for managing per-process address spaces.
+///
+/// This is an **optional** trait — platforms that do not yet support
+/// multi-process may leave all methods at the default (which returns
+/// [`AddressSpaceError::NotSupported`]).
```
jaybosamiya-ms, line 49: I don't like the design of a full trait filled with methods with `NotSupported`.  In one of the other PRs, I commented a better design, which is to have most of the methods be required methods, but also provide a trivial impl that can be reused

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,137 @@
...
+///
+/// # Associated Type
+///
+/// `AddressSpaceId` is an opaque, lightweight handle that identifies one
+/// address space. It must be `Copy + Eq + Send + Sync` so it can be stored
+/// inside process contexts and passed across threads.
```
jaybosamiya-ms, line 55: Doc strings for associated types should be at the associated type, not at the top level of the trait

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,137 @@
...
+pub trait AddressSpaceProvider {
+    /// Opaque identifier for an address space.
+    type AddressSpaceId: Copy + Eq + Send + Sync + core::fmt::Debug;
```
jaybosamiya-ms, line 58: This is in conflict with #759 which needs a concrete `u64` for its opaque identifier, whereas this one defines it at being truly opaque.  Not sure what the intention is.

I am also not convinced that ` + core::fmt::Debug` is good to have in the associated type, but this is not a strong belief of mine; it is more of a question about "do we need to do this in more places"?

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,137 @@
...
+        Err(AddressSpaceError::NotSupported)
+    }
+
+    /// Destroy an address space, releasing all associated resources.
+    ///
+    /// After this call, `id` is invalid and must not be reused.
+    fn destroy_address_space(&self, id: Self::AddressSpaceId) -> Result<(), AddressSpaceError> {
+        let _ = id;
+        Err(AddressSpaceError::NotSupported)
+    }
```
jaybosamiya-ms, line 74: If `AddreessSpaceId` is not maintained as a `Copy` but instead is a true opaque linear token, then we wouldn't need to do this through documentation, and can instead rely on the Rust compiler and type system to guarantee that we are using it correctly.  That would be a more long-term easier to maintain abstraction.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    /// Fork an address space from `parent`.
+    ///
+    /// Returns a [`ForkedAddressSpace`] indicating what kind of fork was
+    /// performed:
+    ///
+    /// * [`Independent`](ForkedAddressSpace::Independent) — full CoW copy.
+    /// * [`SharedWithParent`](ForkedAddressSpace::SharedWithParent) — new VA
+    ///   partition, parent pages shared.
```
jaybosamiya-ms, line 83: No point re-documenting what is documented at the enum.  This leads to long-term drift if things change.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn fork_address_space(
+        &self,
+        parent: Self::AddressSpaceId,
+    ) -> Result<ForkedAddressSpace<Self::AddressSpaceId>, AddressSpaceError> {
+        let _ = parent;
+        Err(AddressSpaceError::NotSupported)
+    }
+
+    /// Make `id` the active address space for the current CPU / thread.
+    fn activate_address_space(&self, id: Self::AddressSpaceId) -> Result<(), AddressSpaceError> {
```
jaybosamiya-ms, line 93: As part of the "linear" move, functions like this would take a ref to the addressspaceid instead of the object

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// implement [`activate_address_space`](Self::activate_address_space) should
+    /// also override this method with a proper save/restore sequence.
+    fn with_address_space<R>(
+        &self,
+        id: Self::AddressSpaceId,
+        f: impl FnOnce() -> R,
+    ) -> Result<R, AddressSpaceError> {
+        let _ = (id, f);
+        Err(AddressSpaceError::NotSupported)
+    }
```
jaybosamiya-ms, line 114: As I'd spent time reviewing on the non-split https://github.com/microsoft/litebox/pull/743#discussion_r3042304071:
> A better way to handle this would be to have two functions, one that the platform implements, and one that is provided by the trait with the latter doing the correct restoration pattern.
>
> Also, please document what happens on abort scenarios, where it is impossible to actually capture the rewind.

litebox/src/platform/address_space.rs
```diff
@@ -0,0 +1,137 @@
...
+    ///
+    /// When `true`, the shim eagerly copies all writable guest pages before
+    /// spawning the forked child and restores them after the child execs or
+    /// exits. When `false` (the default), the shim marks writable pages
+    /// read-only and lazily snapshots individual pages on first write fault.
+    ///
+    /// Platforms where the exception/fault handler shares the guest address
+    /// space must set this to `true` because a CoW fault inside the handler
+    /// itself would be fatal.
+    const EAGER_COW_ON_FORK: bool = false;
```
jaybosamiya-ms, line 127: I'm not sure this should be a shim-level task, right?

Also, generally speaking, bool-based config flags are often a smell of bad abstraction.

In this case, as far as I can tell, the issue is wrt exception/fault handler: why can't the platform just do an eager copy of the exception/fault handler page, and _then_ everything can just work through normal CoW faults, right?  That would also lead to better performance than forcing literally everything to become eager just to handle one page?  The platform already should know at least a little bit about the exception handling behavior, because of our userpointer design (which we have not finished finalizing the _fully_ nice version, but nonetheless, should already iirc have the necessary functionality exposable directly there).

litebox/src/platform/mock.rs
```diff
@@ -61,6 +61,12 @@ impl Provider for MockPlatform {}
 
 impl RawMessageProvider for MockPlatform {}
 
+impl AddressSpaceProvider for MockPlatform {
+    // All methods default to `Err(NotSupported)`, which is correct for the
+    // mock platform (single-process only).
+    type AddressSpaceId = u32;
+}
```
jaybosamiya-ms, line 68: If it is all unsupported, using `core::convert::Infallible` or similar is the better move, no?

litebox/src/platform/mod.rs
```diff
@@ -18,6 +19,7 @@ use either::Either;
 use thiserror::Error;
 use zerocopy::{FromBytes, IntoBytes};
 
+pub use address_space::*;
```
jaybosamiya-ms, line 22: Prefer not using wildcard imports

litebox_platform_lvbs/src/lib.rs
```diff
@@ -1353,6 +1353,12 @@ impl<Host: HostInterface> litebox::platform::SystemInfoProvider for LinuxKernel<
 
 impl<Host: HostInterface> litebox::platform::RawMessageProvider for LinuxKernel<Host> {}
 
+impl<Host: HostInterface> litebox::platform::AddressSpaceProvider for LinuxKernel<Host> {
+    // All methods default to `Err(NotSupported)` — real implementation comes
+    // when LVBS multi-process (separate page tables) is added.
+    type AddressSpaceId = u32;
+}
```
jaybosamiya-ms, line 1360: If the real implementation comes later, then those bits should still be marked as `unimplemented!()` or similar, not be silently marked as `NotSupported`.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -447,6 +447,10 @@ impl litebox::platform::Provider for LinuxUserland {}
 
 impl litebox::platform::RawMessageProvider for LinuxUserland {}
 
+impl litebox::platform::AddressSpaceProvider for LinuxUserland {
+    type AddressSpaceId = u32;
+}
```
jaybosamiya-ms, line 452: Do we not plan to support multi process on Linux userland?  Why is there a comment on LVBS but not here, for example?  Similarly for other platforms.

jaybosamiya-ms (commented):
I've added a few comments.  There were other comments in #743 that I'd added to this bit of the code before it was split, and some might be handled, others don't seem to have been handled.  I am going to ignore the comments that I'd spent time writing in #743 and have only looked at the design of the current one.

Top level comment is that there is a lot of coupling between specific shim implementations and specific platform implementations happening here.  We should be able to clean this up through a nicer design, imho, such that all the bits are actually sitting primarily at the platform side, and the shim should mostly be unaware of needing to handle CoW or whether there are "shared" or "independent".

More detailed comments inline.

PR #760, opened 2026-04-07
Add trace_fs feature gate for conditional DebugLogProvider bound

wdcui:
## Summary

- Add `trace_fs` feature to `litebox/Cargo.toml`.
- Gate the `DebugLogProvider` bound on `RawSyncPrimitivesProvider` behind `#[cfg(feature = "trace_fs")]`, so platforms that don't need filesystem tracing can avoid the dependency.
- Add explanatory comment for why the feature gate exists.

Split from #743.

### DIFF
--- a/litebox/Cargo.toml
+++ b/litebox/Cargo.toml
@@ -31,6 +31,7 @@ windows-sys = { version = "0.60.2", features = [
 
 [features]
 lock_tracing = ["dep:arrayvec", "spin/mutex"]
+trace_fs = []
 panic_on_unclosed_fd_drop = []
 enforce_singleton_litebox_instance = []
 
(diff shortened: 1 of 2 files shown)


jaybosamiya-ms (changes requested):
I'm not sure that the explanation here for `trace_fs` makes sense.  What is happening with the file system tracing?  I see nothing in the file system implementations as part of this PR, so there is literally no context to this flag.  

Adding new conditional-compilation flags increases testing/development/etc. complexity, so I am gonna say no to merging this PR _unless_ very clear/strong evidence is shown for why it is needed.

My best guess is that this was part of some local testing code of whatever you were debugging and the agent split it out without realizing that it should actually remove the local testing code?

PR #763, opened 2026-04-10
Harden VTL memory protection function

sangho2:
This PR hardens the VTL memory protection function(s). Currently, the VTL memory protection function doesn't check whether the given physical memory range belongs to VTL0 or VTL1. This potentially lets the adversary tamper with the VTL1's memory protection.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -96,8 +96,10 @@ pub(crate) fn init(is_bsp: bool) {
             };
             let start = PhysAddr::new(start);
             if protect_vtl1_physical_memory_range(PhysFrame::range(
-                PhysFrame::containing_address(start),
-                PhysFrame::containing_address(end),
+                PhysFrame::from_start_address(start)
+                    .expect("VTL1 memory start address is not page-aligned"),
+                PhysFrame::from_start_address(end)
+                    .expect("VTL1 memory end address is not page-aligned"),
             ))
             .is_err()
             {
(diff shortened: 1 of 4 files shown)


litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1306,14 +1303,19 @@ fn copy_heki_pages_from_vtl0(pa: u64, nranges: u64) -> Option<Vec<HekiPage>> {
     Some(heki_pages)
 }
 
-/// This function protects a physical memory range. It is a safe wrapper for `hv_modify_vtl_protection_mask`.
-/// `phys_frame_range` specifies the physical frame range to protect
-/// `mem_attr` specifies the memory attributes to be applied to the range
+/// This function protects a VTL0 physical memory range from VTL0 by restricting its access permissions
```
sangho2: "protects a VTL0 physical memory range from VTL0" is a bit misleading. Revise it.

litebox_platform_lvbs/src/mshv/vsm.rs
wdcui, line 89: Does get_vtl1_memory_info() guarantee to return page aligned `start` and `size`?

wdcui: Are all HEKI kexec/trampoline physical ranges guaranteed to be page-aligned?

sangho2: Yes. These `start` and `size` are from early boot and they are trusted. If they are malformed, no way to securely boot this VM. And we do use `containing_address`.

sangho2: `HekiPage` data structure does have an `is_valid` method for various sanity checks.

wdcui (approved):
LGTM. Thanks!

PR #764, opened 2026-04-10
Harden HEKI's kernel text patch fuction

sangho2:
This PR hardens HEKI's kernel text patch function. In particular, the current implementation (`validate_text_poke_bp_batch`) allows a malicious VTL0 kernel to write `0xcc` to some write-protected kernel memory addresses through VTL1. This PR adds an explicit address check.


wdcui (approved):
LGTM. Thanks!

PR #769, opened 2026-04-13
Fix missing TLB flush

sangho2:
This PR fixes missing TLB flushes. In particular, it covers cases like cross-core *kernel-page* not-present -> present or PDE widening, which result in faults because we do not maintain VM areas for kernel address space such that a page fault handler cannot properly update the PTEs. This PR simply flushes corresponding TLB entries.

### DIFF
--- a/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
+++ b/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
@@ -241,13 +241,14 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
         let end: Page<Size4KiB> = Page::from_start_address(VirtAddr::new(old_range.end as u64))
             .or(Err(page_mgmt::RemapError::Unaligned))?;
 
-        // Note: TLB entries for the old addresses are batch-flushed after all pages
-        // are remapped, consistent with the Linux kernel's approach.
+        // Note: TLB entries for both old and new addresses are batch-flushed
+        // after all pages are remapped, consistent with the Linux kernel's approach.
         // Note this implementation is slow as each page requires three full page table walks.
         // If we have N pages, it will be 3N times slower.
         let mut allocator = PageTableAllocator::<M>::new();
         let mut inner = self.inner.lock();
         let flush_start = start;
+        let new_flush_start = new_start;
         while start < end {
             match inner.translate(start.start_address()) {
                 TranslateResult::Mapped {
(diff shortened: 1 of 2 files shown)


CvvT:
> This PR fixes missing TLB flushes. In particular, it covers not-present -> present cases which suffer from corner cases like speculative page table walks and more.

If it's not-present --> present, my understanding is that in worst case CPU would trigger page fault. Once it's resolved, TLB would be updated. So, we don't need to flush TLB proactively.

sangho2:
> > This PR fixes missing TLB flushes. In particular, it covers not-present -> present cases which suffer from corner cases like speculative page table walks and more.
> 
> If it's not-present --> present, my understanding is that in worst case CPU would trigger page fault. Once it's resolved, TLB would be updated. So, we don't need to flush TLB proactively.

There are some corner cases that I can reproduce and that's why I wrote this PR :)

CvvT:
> > > This PR fixes missing TLB flushes. In particular, it covers not-present -> present cases which suffer from corner cases like speculative page table walks and more.
> > 
> > 
> > If it's not-present --> present, my understanding is that in worst case CPU would trigger page fault. Once it's resolved, TLB would be updated. So, we don't need to flush TLB proactively.
> 
> There are some corner cases that I can reproduce and that's why I wrote this PR :)

What is the corner case?

sangho2:
> > > > This PR fixes missing TLB flushes. In particular, it covers not-present -> present cases which suffer from corner cases like speculative page table walks and more.
> > > 
> > > 
> > > If it's not-present --> present, my understanding is that in worst case CPU would trigger page fault. Once it's resolved, TLB would be updated. So, we don't need to flush TLB proactively.
> > 
> > 
> > There are some corner cases that I can reproduce and that's why I wrote this PR :)
> 
> What is the corner case?

This is about a kernel-level page fault. we don't have demand paging / page fault handler for it.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -295,6 +296,7 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
         // Flush old (unmapped) addresses — other cores may hold stale entries.
         let page_count = (end.start_address() - flush_start.start_address()) / Size4KiB::SIZE;
         flush_tlb_range(flush_start, page_count.truncate());
+        flush_tlb_range(new_flush_start, page_count.truncate());
```
CvvT: Is `remap_pages` only used to remap user space memory or can be also used to remap kernel space memory in lvbs case? If it's the former case, we may avoid this extra flush (but I guess it's safer to flush the TLB anyway).

CvvT (approved):
Overall LGTM, left one question.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: You're right. Let me take a look this PR again to see whether we can skip some TLB flushes.

PR #770, opened 2026-04-13
support sendmsg for unix socket

CvvT:
Support calling `sendmsg` on unix socket.


jaybosamiya-ms (approved):
LGTM, thanks Weiteng!

jaybosamiya-ms:
Seems like GitHub Actions might be getting stuck right now waiting for a runner?  We should re-attempt the queue later on
image

PR #771, opened 2026-04-13
Make `AnyMap` clonable

jaybosamiya-ms:
This PR changes LiteBox's internal `AnyMap` implementation to support `Clone`.  To do this, we set up some safe abstractions that guarantee `Clone` after which the correct coercions let us maintain the rest.


jaybosamiya-ms:
This PR supersedes #752.  I implemented this one from scratch because I did not particularly find the approach that was taken there to be particularly clean.

wdcui (approved):
LGTM. Thanks.

PR #772, opened 2026-04-14
Add support for setting up platform root key for LVBS

sangho2:
This PR adds support for setting up the Platform Root Key for the LVBS platform, which can be used to derive other keys like per-TA keys.


sangho2:
@jaybosamiya-ms I've added `PlatformRootKeyProvider` to `litebox`, which is based on the discussion we had in the past.

sangho2:
@tgopinath-microsoft This PR introduces a VTL function/call for setting a platform root key (`mshv_vsm_set_platform_root_key`). The function (and overall PR) is straightforward, but it requires your or your team's input because it defines a new ABI.

sangho2:
Once this PR is merged, I'll rebase #643 .

jaybosamiya-ms:
Thanks @sangho2!  For visibility, Sangho and I discussed this PR offline, and I've instead introduced #781 as the alternate interface to consider, which balances both the OP-TEE requirements as well as more distrustful platforms.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -916,6 +918,34 @@ fn mshv_vsm_allocate_ringbuffer_memory(phys_addr: u64, size: usize) -> Result<i6
...
 
+/// This function sets the platform root key by copying key data from VTL0.
+///
+/// - `key_pa`: The physical address (VTL0) of the platform root key.
+/// - `key_size`: The size of the platform root key.
+///
+/// This function assumes that the caller stores the key data in a single or
+/// contiguous physical memory page(s). If the caller cannot ensure this,
+/// we should make this function use `HekiPage`.
+fn mshv_vsm_set_platform_root_key(key_pa: u64, key_size: u64) -> Result<i64, VsmError> {
```
tgopinath-microsoft: Do you think we will have the key size to be something other than 8 bytes ? If no, may be just need the key_pa and we do not need the size argument ? If you think we should extend the root key to be more than 8 bytes, then maybe we can have a an array of 8 bytes for key_pa and then specify the size in key_size

sangho2: I think the key should be at least 32 bytes (16 bytes might be okay but a bit weak)? If this something about we don't need to support various key size options, I think it is okay to only support a 32-byte root key.

tgopinath-microsoft: Ah I somehow mistook key_pa as key. ok then let us just support a 32 byte key and do away with the size

sangho2: sounds good. I'll revise the function signature.

sangho2: revised.

litebox_platform_lvbs/src/mshv/error.rs
```diff
@@ -217,7 +220,8 @@ impl From<VsmError> for Errno {
             | VsmError::SymbolNameInvalidUtf8
             | VsmError::SymbolNameNoTerminator
             | VsmError::CertificateDerLengthInvalid { .. }
-            | VsmError::CertificateParseFailed => Errno::EINVAL,
+            | VsmError::CertificateParseFailed
+            | VsmError::PlatformRootKeyInvalid => Errno::EINVAL,
```
tgopinath-microsoft: PlatformRootKeyInvalid  is not used anywhere ??

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -894,6 +896,28 @@ fn mshv_vsm_allocate_ringbuffer_memory(phys_addr: u64, size: usize) -> Result<i6
...
+
+    let key_pa = PhysAddr::try_new(key_pa).map_err(|_| VsmError::InvalidPhysicalAddress)?;
+
+    let mut keybuf = [0u8; PRK_LEN];
+    if unsafe { crate::platform_low().copy_slice_from_vtl0_phys(key_pa, &mut keybuf) } {
+        set_platform_root_key(&keybuf);
+        Ok(0)
+    } else {
+        Err(VsmError::Vtl0CopyFailed)
+    }
```
tgopinath-microsoft, line 919: In either success or failure case,  can you zeroize first 32 bytes or the page at key_pa ? This is defense in depth

tgopinath-microsoft: Sam with keybuf. Please zeroize the content after use

litebox_platform_lvbs/src/host/lvbs_impl.rs
```diff
@@ -114,6 +114,27 @@ impl litebox::platform::CrngProvider for LvbsLinuxKernel {
...
+/// from hardware or a boot nonce.
+///
+/// # Panics
+/// Panics if `key` length does not match `PRK_LEN`.
+pub fn set_platform_root_key(key: &[u8]) {
+    assert_eq!(key.len(), PRK_LEN, "Platform Root Key length mismatch");
+    PRK_ONCE.call_once(|| {
+        let mut prk = [0u8; PRK_LEN];
+        prk.copy_from_slice(key);
+        prk
```
tgopinath-microsoft: Zeroize prk ?

tgopinath-microsoft:
Thanks Sangho for updating the PR. In general I am ok with the approach here. How do you envision to use PRK ?

litebox_platform_lvbs/src/mshv/error.rs
sangho2: good catch. this was used when we check input key length. let me remove it.

sangho2:
> Thanks Sangho for updating the PR. In general I am ok with the approach here. How do you envision to use PRK ?

#780 does use PRK to derive a TA unique key.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Good point. Let me use the `zeroize` crate.

tgopinath-microsoft: Thanks for doing this. One more step will be to actually 0 the physical memory off the key material without relying on VTL0 to do that. But we can take that up later

tgopinath-microsoft (approved):

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Since we trust VTL0 at this point, this would be just something about whether a caller or a callee wipes out the material. Anyhow, VTL0 should zeroize all buffers it uses for interacting with TPM as well.

tgopinath-microsoft: Yes. VTL0 does that. I think we should still do it here for defense in depth

PR #773, opened 2026-04-14
Enable logger for userland Linux runner

### LINKED ISSUE #171: Add tracing support

jaybosamiya-ms:
This PR is (effectively) a re-spin of #558.  It resolves #171.

The previous PR (#558) introduced `tracing` directly.  In contrast here, we are using the #665 facade inside `litebox` itself.

This means that by-default `litebox` uses the lighter-weight `log`, which is more suitable for embedded-ish kernel projects.  Runners that wish to have the heavier-weight `tracing` can flip the config flag (this is all primarily due to the work put into #665, this PR is mostly just realizing it).

At the runner, this PR uses `tracing_subscriber` directly and flips the config flag, which gives pretty output :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1450,6 +1450,7 @@ dependencies = [
  "buddy_system_allocator",
  "either",
  "hashbrown",
+ "litebox_util_log",
  "rangemap",
  "ringbuf",
  "slabmalloc",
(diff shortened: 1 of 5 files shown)


CvvT (approved):
LGTM, thanks!

PR #775, opened 2026-04-14
Add subscribers for the various runners

jaybosamiya-ms:
Similar to how #773 introduced loggers/subscribers to the Linux userland runner, this PR introduces them to the other runners.


CvvT (approved):
LGTM, thanks! @sangho2 you may want to take a look at the lvbs part.

sangho2:
It seems that `log::set_max_level(log::LevelFilter::Trace);` generates too many logs for kernel platforms, resulting in soft lockup cpu# stuck.

PR #776, opened 2026-04-14
Finish migration to `litebox_util_log` + remove `DebugLogProvider`

jaybosamiya-ms:
This PR removes `DebugLogProvider` entirely by switching all the direct and indirect uses of it for logging to the `litebox_util_log` crate introduced in #665.


jaybosamiya-ms:
Most of this is quite straightforward, but the repeated `:% = format_args!(...)` pattern makes me think I should update `litebox_util_log` to support `:x` or similar to easily get the hex output, rather than needing this pattern.  I'm not doing that in this particular PR though.

CvvT (approved):
LGTM, thanks!

PR #777, opened 2026-04-15
Support recvmsg

CvvT:
Support syscall [`recvmsg`](https://man7.org/linux/man-pages/man3/recvmsg.3p.html).

Some flags like `MSG_PEEK` is not supported yet but can be added when needed.

Also fix a message boundary issue for datagram unix socket, i.e., one read should only read one message.

### DIFF
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -1545,8 +1545,7 @@ impl<FS: ShimFS> Task<FS> {
         let Ok(sockfd) = u32::try_from(fd) else {
             return Err(Errno::EBADF);
         };
-        let msg =
-            ConstPtr::<litebox_common_linux::UserMsgHdr<Platform>>::from_usize(msg_ptr.as_usize())
+        let msg = msg_ptr
                 .read_at_offset(0)
                 .ok_or(Errno::EFAULT)?;
 
@@ -1557,7 +1556,7 @@ impl<FS: ShimFS> Task<FS> {
         let msg_controllen = msg.msg_controllen;
 
         if msg_controllen != 0 {
-            unimplemented!("ancillary data is not supported");
+            log_unsupported!("ancillary data is not supported");
         }
         if msg_iovlen == 0 || msg_iovlen > 1024 {
             return Err(Errno::EINVAL);
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -1470,15 +1470,29 @@ impl<FS: ShimFS> Task<FS> {
                 None
             },
         )?;
-        buf.copy_from_slice(0, &recv_buf[..size.min(recv_buf.len())])
+        let capped_size = size.min(recv_buf.len());
+        buf.copy_from_slice(0, &recv_buf[..capped_size])
             .ok_or(Errno::EFAULT)?;
         if let Some(src_addr) = source_addr
             && let Some(sock_ptr) = addr
         {
             write_sockaddr_to_user(src_addr, sock_ptr, addrlen)?;
         }
-        Ok(size)
+
+        if flags.contains(ReceiveFlags::TRUNC) {
+            // the actual message size
+            Ok(size)
+        } else {
+            // the number of bytes copied
+            Ok(capped_size)
+        }
     }
+    /// Receive data from a socket.
+    ///
+    /// `source_addr` can be provided to receive the source address if available.
+    ///
+    /// On success, returns the number of bytes received. Note that for datagram sockets,
+    /// this may be larger than the provided buffer length as the excessive data will be truncated.
     fn do_recvfrom(
         &self,
         sockfd: u32,
@@ -1609,7 +1623,13 @@ impl<FS: ShimFS> Task<FS> {
             offset += chunk;
         }
 
-        let total_received = size;
+        let total_received = if flags.contains(ReceiveFlags::TRUNC) {
+            // the actual message size
+            size
+        } else {
+            // the number of bytes copied
+            size.min(total_iov_capacity)
+        };
 
         // Write back source address if requested.
         if want_source {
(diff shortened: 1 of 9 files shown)


litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -1527,17 +1541,128 @@ impl<FS: ShimFS> Task<FS> {
...
+                Some(&mut source_addr)
+            } else {
+                None
+            },
+        )?;
+
+        // Set MSG_TRUNC if the received message was larger than the total buffer.
+        if size > total_iov_capacity {
+            ret_flags |= ReceiveFlags::TRUNC;
+        }
```
sangho2, line 1609: > High: recvmsg does not set output MSG_TRUNC for truncated datagrams.
> this calls do_recvfrom without forcing MSG_TRUNC, then sets msg_flags only when size > total_iov_capacity. For datagram sockets, do_recvfrom returns the copied length unless the caller passed input MSG_TRUNC, so a too-large datagram received without input MSG_TRUNC returns the copied byte count but leaves msg_flags empty. Linux recvmsg should set msg_flags |= MSG_TRUNC when the datagram was larger than the supplied iovecs.

It sounds like `size` cannot be greater than `total_iov_capacity` for datagrams received without `MSG_TRUNC` even if there was truncation.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -1527,17 +1541,128 @@ impl<FS: ShimFS> Task<FS> {
...
+        // Compute total buffer capacity across all non-empty iovecs, capped at MAX_LEN.
+        let total_iov_capacity: usize = iovs
+            .iter()
+            .map(|iov| iov.iov_len)
+            .fold(0usize, usize::saturating_add)
+            .min(MAX_LEN);
+
+        if total_iov_capacity == 0 {
+            return Ok(0);
+        }
```
sangho2: > High: zero-length recvmsg is incorrect.
> This rejects msg_iovlen == 0 with EINVAL, and returns 0 immediately when total iovec capacity is zero. Linux permits zero-length receives. For datagram sockets, a zero-length receive can still dequeue one pending datagram and should update output fields such as msg_flags, msg_namelen, and msg_controllen.

It sounds like LiteBox might not dequeue pending zero-length datagram?

litebox_shim_linux/src/syscalls/net.rs
CvvT: Yes, it is checked at

https://github.com/microsoft/litebox/blob/5441699b423bb0c14ca50a6089482758aa87244a/litebox_shim_linux/src/syscalls/net.rs#L1631-L1637

litebox_shim_linux/src/syscalls/net.rs
CvvT: Right, I added a test to confirm that's Linux's behavior. It is fixed now.

wdcui (commented):
LGTM. Thanks!

wdcui (approved):

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -1522,17 +1536,124 @@ impl<FS: ShimFS> Task<FS> {
...
+                Some(&mut source_addr)
+            } else {
+                None
+            },
+        )?;
+
+        // Set MSG_TRUNC if the received message was larger than the total buffer.
+        if size > total_iov_capacity {
+            ret_flags |= ReceiveFlags::TRUNC;
+        }
```
CvvT, line 1609: > Correctness bug — MSG_TRUNC not set for truncated INET datagrams
  litebox_shim_linux/src/syscalls/net.rs:1600-1603. The unix path now reports the full message size, so the size > total_iov_capacity check works. The INET path (DatagramSocketChannel::try_read) already caps return to min(buf.len(), data.len()), so the comparison
  can never trigger and MSG_TRUNC is never set for inet UDP. Suggested: always pass ReceiveFlags::TRUNC to do_recvfrom so the proxy reports the real datagram size, then derive output flags from the comparison.

I only fixed unix datagram socket previously. I updated udp socket as well to return the actual size of the message instead of the length of the copied data. Tests are also updated to cover udp.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -1522,17 +1536,124 @@ impl<FS: ShimFS> Task<FS> {
...
+
+        // Copy fields out of the packed struct to avoid unaligned references.
+        let msg_name = msg.msg_name;
+        let msg_iov = msg.msg_iov;
+        let msg_iovlen = msg.msg_iovlen;
+        let msg_controllen = msg.msg_controllen;
+
+        if msg_controllen != 0 {
+            log_unsupported!("ancillary data is not supported");
+        }
```
CvvT, line 1573: > Correctness — msg_controllen never zeroed on return
  net.rs:1565-1567. Linux always writes back msg_controllen (to the cmsg bytes written, 0 otherwise). The shim only logs and proceeds. Same for MSG_CTRUNC: if ancillary data was requested but not delivered, the kernel signals truncation; the shim silently drops the request.

We don't support ancillary data or control message yet. For now, just set `msg_controllen` to zero.

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -1522,17 +1537,137 @@ impl<FS: ShimFS> Task<FS> {
...
+            .fold(0usize, usize::saturating_add)
+            .min(MAX_LEN);
+
+        // Perform a single recv into a contiguous buffer.
+        let want_source = msg_name.as_usize() != 0;
+        let mut source_addr = None;
+        let mut ret_flags = ReceiveFlags::empty();
+
+        // Heap-allocate the recv buffer to avoid stack overflow for large iovecs.
+        let mut buffer = alloc::vec![0u8; total_iov_capacity];
```
CvvT: > Correctness — MAX_LEN = 65536 silently caps user iov capacity
  net.rs:1552, 1574-1579. For unix-datagram messages between 64 KiB and sum(iov_len), the code sets MSG_TRUNC even though everything would fit. Either remove the cap, return EMSGSIZE, or document it.

Copilot suggested to use `try_reserve_exact` which returns an error when OOM.

PR #778, opened 2026-04-15
Support syscall rmdir

CvvT:
support syscall `rmdir`, which is equivalent to `unlinkat` with the flag `AT_REMOVEDIR`


jaybosamiya-ms (approved):
LGTM, thanks Weiteng!

PR #779, opened 2026-04-15
Fix eventfd write panic on short buffer

CvvT:
a minor bug fix


jaybosamiya-ms (approved):
LGTM, thanks Weiteng!

PR #780, opened 2026-04-15
OP-TEE TA unique key derivation

sangho2:
This PR implements OP-TEE's `PTA_SYSTEM_DERIVE_TA_UNIQUE_KEY` (https://github.com/OP-TEE/optee_os/blob/06c4e95e469c9c89e9ba4a6915d1be7bb8ea6fbc/core/pta/system.c#L62) for the OP-TEE shim.

### DIFF
--- a/litebox_runner_optee_on_linux_userland/src/lib.rs
+++ b/litebox_runner_optee_on_linux_userland/src/lib.rs
@@ -85,6 +85,8 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
     let _litebox = shim_builder.litebox();
     let shim = shim_builder.build();
 
+    platform.initialize_boot_specific_kdf_support();
+
     if cli_args.command_sequence.is_empty() {
         run_ta_with_default_commands(&shim, ldelf_data.as_slice(), prog_data.as_slice());
     } else {
--- a/litebox_shim_optee/src/syscalls/pta.rs
+++ b/litebox_shim_optee/src/syscalls/pta.rs
@@ -5,12 +5,19 @@
 //! the functions of built-in TAs.
 
 use crate::{Task, UserConstPtr, UserMutPtr};
-use litebox::{
-    platform::{RawConstPointer as _, RawMutPointer as _},
-    utils::TruncateExt,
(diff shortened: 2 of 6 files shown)


sangho2:
This PR and #828 do have some common code. Minor fix is needed once both (or one of these) are approved.

litebox_shim_optee/src/syscalls/pta.rs
```diff
@@ -141,4 +106,116 @@ impl Task {
...
+        let extra_data = if extra_data_size == 0 {
+            Vec::new().into_boxed_slice()
+        } else {
+            let extra_data_ptr = UserConstPtr::<u8>::from_usize(extra_data_addr.truncate());
+            extra_data_ptr
+                .to_owned_slice(extra_data_size)
+                .ok_or(TeeResult::BadParameters)?
+        };
+
+        // In LiteBox for LVBS, TA's address space is always in the secure world.
```
wdcui: Is this comment needed? Isn't it that TA always runs in the secure world?

sangho2: I'll extend it. Basically, this is for convincing human or AI reviewers: the original OP-TEE implementation does have secure-versus-normal world address checks (because a malicious or buggy TA could provide a normal-world pointer for storing a derived key) that LiteBox doesn't need due to strict pointer abstraction.

litebox_runner_optee_on_linux_userland/src/lib.rs
```diff
@@ -85,6 +85,8 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
     let _litebox = shim_builder.litebox();
     let shim = shim_builder.build();
 
+    platform.initialize_boot_specific_kdf_support();
```
wdcui, line 88: I don't see this function implemented in the LVBS platform?

sangho2: we don't. Only the Linux userland platform has a default KDF. This is intended #781

wdcui (approved):
LGTM. Thanks.

PR #781, opened 2026-04-16
Introduce device-specific derived key support

jaybosamiya-ms:
This PR adds support for derived keys through a `DerivedKeyProvider` platform trait.  This design manages the tradeoffs between some requirements for OP-TEE and strictness of wrapping.

The strictest version of this design looks similar to a TPM, in that the platform would never expose the root key to anything else, only ingesting context and outputting a derived key.  However, this design does not fit some constraints required by the OP-TEE shim, which (for reasons) wants to specify its own key derivation function.  The naive way to support this is through the platform revealing the device-specific key directly, but this (unsurprisingly) can lead to proliferation of key usage, rather than keeping it strictly confined.  Indeed, direct exposing of such keys cannot even be supported by some platforms (e.g., TPM).

As a middle ground, the platform can choose to use a shim-provided KDF if it wishes to (if it makes sense for some shim/platform pairs) while for other more-distrustful platforms can choose to do their own KDF in addition to the shim-provided KDF, or even to ignore the shim-provided KDF entirely.  Effectively, the platform is the true arbiter of how the secret gets protected.

This PR introduces this interface, as well as implements it for the Linux userland platform to demonstrate one concrete implementation of it.

### DIFF
--- a/litebox/src/platform/mod.rs
+++ b/litebox/src/platform/mod.rs
@@ -754,3 +754,59 @@ pub trait CrngProvider {
     /// failures.
     fn fill_bytes_crng(&self, buf: &mut [u8]);
 }
+
+/// Provider of derived device-specific keys.
+///
+/// Some shims need support for deriving keys that survives past reboots (for example, to support
+/// secure storage). Such keys are derived from some device-specific secret (called `root_key`) and
+/// some `context`, and a key derivation function (KDF).
+///
+/// Platforms might differ drastically on their own notion of device-specific secrets, and what
+/// "reboot-surviving" means. Some platforms might have a real never-revealed never-modified root
+/// key (e.g., TPMs), while others might have maintain a persistent key across LiteBox invocations,
+/// but that persistent-key might be re-initialized to a new value every "real" reboot, etc.
+///
+/// Concretely, no shim can depend _directly_ on the existence of a device-specific secret. However,
+/// interestingly, for cryptographically strong KDFs where you do not know the root key,
+/// `KDF(root_key, context)` is indistinguishable from `KDF(KDF'(root_key, context'), context)`, etc.
+/// Thus, while some shims might be particular about their choice of KDF, and platforms might be
+/// particular about their choice of KDFs, they can mutually-distrustingly just simply run two KDFs
+/// if needed. For performance reasons however, this is not ideal to be forced always, and thus this
+/// specific provider supports a model that allows for more pragmatic choices, while making sure
+/// that the platform has final say on the total strictness of the root key (since it is what
+/// finally owns the root key).
+pub trait DerivedKeyProvider {
+    /// Derive a new key using the `shim_kdf` (if provided) and the current context
+    /// (`params.context`), and place it into `params.output`.
+    ///
+    /// The platform is allowed to completely ignore the provided `shim_kdf` and use its own KDF
+    /// instead if if chooses to; alternatively, some platforms might not have their own KDFs, and
+    /// only run if the shim provides a KDF.
+    ///
+    /// The `shim_kdf` is a `fn` not a `Fn`/`FnMut`/`FnOnce` in order to incentivize usage of pure
+    /// functions.
+    fn derive_key(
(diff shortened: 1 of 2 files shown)


litebox/src/platform/mod.rs
```diff
@@ -754,3 +754,59 @@ pub trait CrngProvider {
...
+/// if needed. For performance reasons however, this is not ideal to be forced always, and thus this
+/// specific provider supports a model that allows for more pragmatic choices, while making sure
+/// that the platform has final say on the total strictness of the root key (since it is what
+/// finally owns the root key).
+pub trait DerivedKeyProvider {
+    /// Derive a new key using the `shim_kdf` (if provided) and the current context
+    /// (`params.context`), and place it into `params.output`.
+    ///
+    /// The platform is allowed to completely ignore the provided `shim_kdf` and use its own KDF
+    /// instead if if chooses to; alternatively, some platforms might not have their own KDFs, and
```
sangho2: nits: if if -> if it

litebox/src/platform/mod.rs
```diff
@@ -754,3 +754,59 @@ pub trait CrngProvider {
...
+
+/// Provider of derived device-specific keys.
+///
+/// Some shims need support for deriving keys that survives past reboots (for example, to support
+/// secure storage). Such keys are derived from some device-specific secret (called `root_key`) and
+/// some `context`, and a key derivation function (KDF).
+///
+/// Platforms might differ drastically on their own notion of device-specific secrets, and what
+/// "reboot-surviving" means. Some platforms might have a real never-revealed never-modified root
+/// key (e.g., TPMs), while others might have maintain a persistent key across LiteBox invocations,
```
sangho2: nits: might have maintain

litebox/src/platform/mod.rs
```diff
@@ -754,3 +754,59 @@ pub trait CrngProvider {
...
+    /// input context is provided.
+    pub context: &'a [u8],
+    /// The output of the KDF produces a key of the exact length of the buffer provided as a
+    /// parameter.
+    pub output: &'a mut [u8],
+}
+
+#[derive(Debug, Error)]
+/// Errors that might be returned upon attempting to derive a key.
+pub enum DerivedKeyError {
```
sangho2: The KDFs of platform and shim use might have different security strengths (lengths, entropy, ...). The platform might want to return an error if something is not satisfied.

sangho2 (approved):
Looks good to me. Thanks!

There are some minor comments.

litebox/src/platform/mod.rs
jaybosamiya-ms: Good catch, fixed, thanks!

litebox/src/platform/mod.rs
jaybosamiya-ms: Good catch, fixed, thanks!

litebox/src/platform/mod.rs
jaybosamiya-ms: Thanks, this is an interesting point.

The only real disagreement between the shim and the platform that I can think of are things where the shim is being more picky than it should be.  For this particular scenario, I introduced a `ShimKDFError` term that the shim can generically return.

I don't actually think the platform can/would be too picky.  In terms of length, KDFs are key stretching routines in a way sense, so it can "always" make more to fit whatever output lengths are required.  Roughly, `KDF_{l1+l2}(secret, context)` is indistinguishable from `KDF_{l1}(secret, context + '1') + KDF_{l2}(secret, context + '2')`, and thus even if the _host_ only supports some small sized outputs, the platform can always support longer sized outputs, acting as a good abstraction.

jaybosamiya-ms:
CI failures due to new Rust version released today which introduces new clippy lints, I'll open a PR to fix things soon.  Those are unrelated to the changes in this PR.  

Edit: #782

jaybosamiya-ms:
@sangho2 please resolve the comment if the tweak to support shim-level errors looks good to you, and it'll auto-merge, thanks!

PR #783, opened 2026-04-17
fix lseek/getdirent64 for dir

CvvT:
`lseek` should also work on directory fd, though `offset` of a directory is opaque to user. The common use case should be resetting offset to zero. Setting offset to arbitrary value is a undefined behavior.

Also fix `sys_getdirent64` to return correct error code.


jaybosamiya-ms (approved):
LGTM, thanks Weiteng!

PR #784, opened 2026-04-17
Fix a use-after-free condition of single-instance TAs

sangho2:
This PR fixes a use-after-free condition of single-instance TAs where a CPU core attempts to open a session with a single-instance TA with a cached instance in memory while another is tearing down it (due to session close, TA crash/panic, ...). This PR gets rid of such condition by re-checking whether a cached instance is valid or not. Also, this PR removes a fast path for single-instance TA opening which is racy with less meaningful performance benefit.

### DIFF
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -520,14 +520,9 @@ fn handle_open_session(
                     // Evict the zombie. Analog of OP-TEE's `maybe_release_ta_ctx`
                     // removing the dead ctx from `tee_ctxes`.
                     session_manager().remove_single_instance_if_same(&ta_uuid, &existing);
-                    write_msg_args_to_normal_world(
-                        msg_args,
-                        msg_args_phys_addr,
-                        TeeResult::TargetDead,
-                        None,
-                        None,
-                        Some(&ta_req_info),
-                    )?;
+                    msg_args.ret = TeeResult::TargetDead;
+                    msg_args.ret_origin = TeeOrigin::Tee;
+                    write_non_ta_msg_args_to_normal_world(msg_args, msg_args_phys_addr)?;
                     Ok(())
                 }
             }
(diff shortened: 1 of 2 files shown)


litebox_shim_optee/src/session.rs
```diff
@@ -115,8 +115,14 @@ pub struct TaInstance {
...
     pub loaded_program: alloc::boxed::Box<LoadedProgram>,
-    /// The task page table ID associated with this TA instance.
+    /// The task page table ID associated with this TA instance. Valid only
+    /// while `dead == false`.
     pub task_page_table_id: usize,
+    /// Set when the TA panics or its last session closes. Any lock holder
+    /// should first check whether `dead == true` and if it is, bail without
+    /// touching `task_page_table_id`. `shim` and `loaded_program` remain
+    /// valid until the last `Arc` is dropped.
+    pub dead: bool,
```
wdcui: Let's change `dead` to a friendly name such as `closed` or `unloaded` or something else. Also, do we need atomic access to this field?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -780,20 +796,22 @@ fn open_session_new_instance(
...
             ldelf_return_code,
             None, // No session ID on failure
             None,
             Some(ta_req_info),
-        )?;
+        );
 
+        // Safety: We are about to tear down this TA instance;
+        // no references to user-space memory will be held afterwards.
+        unsafe { teardown_ta_page_table(&shim, task_pt_id) };
```
wdcui, line 820: Do we need to mark the instance as "dead" here?

wdcui:
I left some comments above and shared with you offline the review comments by Claude.

litebox_runner_lvbs/src/lib.rs
sangho2: TaInstance is not yet created at this point.

litebox_shim_optee/src/session.rs
sangho2: let me use `closed` or something else. This data structure is behind mutex.

wdcui (approved):
LGTM. Thanks.

PR #785, opened 2026-04-17
Remove systrap backend

CvvT:
Remove the seccomp-based syscall interception backend. We may still want to use seccomp to confine the syscalls we allow to call, but will add it in a separate PR.


litebox_runner_linux_userland/src/lib.rs
```diff
@@ -87,8 +87,6 @@ pub struct CliArgs {
 #[non_exhaustive]
 #[derive(Debug, Clone, clap::ValueEnum)]
 pub enum InterceptionBackend {
-    /// Use seccomp-based syscall interception
-    Seccomp,
     /// Depend purely on rewriten syscalls to intercept them
     Rewriter,
 }
```
jaybosamiya-ms: Minor: if this is down to just one value in the enum, we should remove it, right?

litebox_runner_optee_on_linux_userland/src/lib.rs
```diff
@@ -49,8 +49,6 @@ pub struct CliArgs {
 #[non_exhaustive]
 #[derive(Debug, Clone, clap::ValueEnum)]
 pub enum InterceptionBackend {
```
jaybosamiya-ms: same

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -2082,16 +2004,13 @@ impl litebox::platform::SystemInfoProvider for LinuxUserland {
...
-        }
-        self.vdso_address
+        // Enabling VDSO on x86 causes glibc to not set a restorer in signal
+        // handlers, which we do not currently support. Disable VDSO for
+        // now.
+        //
+        // TODO: implement VDSO in the shim, don't try to pass through the
+        // platform VDSO.
+        None
     }
```
jaybosamiya-ms, line 2014: Maybe this function should also be removed?  Seems like all the VDSO stuff is removed in this PR, so might as well remove this too, no?

jaybosamiya-ms (approved):
Overall looks good to me, thanks Weiteng! Just a couple of minor comments to do a bit more cleanup.

litebox_platform_linux_userland/src/lib.rs
CvvT: The trait does not have a default implementation for this function. I'd also like to keep the comment, so just leave it for now.

litebox_runner_linux_userland/src/lib.rs
CvvT: Sure, removed now. All tests are also updated.

litebox_runner_optee_on_linux_userland/src/lib.rs
CvvT: Updated.

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: Sounds reasonable, I missed that this was part of the trait, we can do the trait cleanup later in that case.

PR #786, opened 2026-04-17
Support mknodat

CvvT:
Support [mknodat](https://linux.die.net/man/2/mknodat) but only regular file is allowed. Others like char and block files return `EPERM` for now.

This syscall also exposed some issues:

1. When we create a file, we also check the file permissions against `mode`. The fix is to skip the check if `open` creates a new file.
2. `flags.contains(OFlags::RDONLY)` is always true because `RDONLY` is zero.
3. Empty pathname is invalid for most syscalls.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -168,6 +168,9 @@ impl<FS: ShimFS> Task<FS> {
     /// Resolve a path against the current working directory.
     fn resolve_path(&self, path: impl path::Arg) -> Result<CString, Errno> {
         let path_str = path.as_rust_str().map_err(|_| Errno::EINVAL)?;
+        if path_str.is_empty() {
+            return Err(Errno::ENOENT);
+        }
         if path_str.starts_with('/') {
             CString::new(path_str.to_string()).map_err(|_| Errno::EINVAL)
         } else {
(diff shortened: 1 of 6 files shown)


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -258,6 +258,45 @@ impl<FS: ShimFS> Task<FS> {
...
+        let file_type = if file_type == 0 {
+            // zero translates to S_IFREG
+            InodeType::File
+        } else {
+            InodeType::try_from(file_type).map_err(|_| Errno::EINVAL)?
+        };
+        match file_type {
+            InodeType::File => {
+                let mode = Mode::from_bits_truncate(mode_and_type & !FILE_TYPE_MASK);
+                let fd = self.sys_openat(
```
sangho2: > - mknod("/x", S_IFREG | 0000, 0) should create /x and return success, but this path creates the file and then can fail permission checks while leaving the file behind. It can also fail with EMFILE when the process is at its fd limit, even though mknod should not allocate a file descriptor.
> - mknod("", S_IFREG | 0644, 0) should fail with ENOENT, but with AT_FDCWD this resolves to the current directory and can report EEXIST; with a real dirfd it goes through the unsupported FsPath::Fd path and reports EINVAL.

sangho2 (approved):
Looks good to me. It seems that there are some corner cases.

litebox_shim_linux/src/syscalls/file.rs
CvvT: The root cause of the first issue was that when we create a file, we also check the file permissions against `mode`. The fix is to skip the check if `open` creates a new file.

CvvT: Another subtle issue found by agent was that `flags.contains(OFlags::RDONLY)` is always true because `RDONLY` is zero. Fix it in this PR as well.

CvvT: Regarding the `EMFILE` file, the fix is to refactor `sys_openat` so that we can open a file without creating a fd.

CvvT: The second issue on empty pathname also applies to other syscalls like unlinkat/openat/... Fix all of them as well.

PR #787, opened 2026-04-20
Use COM1 as serial port on Azure

praveen-pk:
Azure VMs do not support COM2 port. So set the default serial port to COM1.

If built with `--devbox` (the newly introduced) feature, LiteBox will use COM2 as the serial port.


sangho2 (approved):
Thanks! Looks good to me.

I have no preference, but we might want to decide the default configuration: Azure Hyper-V (COM1) or Windows Hyper-V (COM2). @tgopinath-microsoft ?

praveen-pk:
Make sense. As Azure will be the most used target, I am inclined to make Azure the default and Hyper-V a build time config.

tgopinath-microsoft (approved):
I discussed with Praveen and COM1 (port of Azure VMs) are now the default

PR #788, opened 2026-04-20
Fix: Adjust max log level for LVBS

sangho2:
This PR adjusts the max log level for LVBS. In particular, `log::LevelFilter::Trace` generates too many logs for LVBS, resulting in VTL0 kernel panic due to CPU stuck (the VTL0 kernel cannot interrupt the VTL1 kernel.)

### DIFF
--- a/litebox_runner_lvbs/src/main.rs
+++ b/litebox_runner_lvbs/src/main.rs
@@ -449,7 +449,7 @@ unsafe extern "C" fn kernel_main(is_bsp: bool) -> ! {
         #[cfg(debug_assertions)]
         log::set_max_level(log::LevelFilter::Debug);
         #[cfg(not(debug_assertions))]
-        log::set_max_level(log::LevelFilter::Info);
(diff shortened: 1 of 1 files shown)


wdcui:
Is this a mitigation or the root cause?

sangho2:
> Is this a mitigation or the root cause?

Good question. I don't know how many trace points we have across LiteBox. Anyhow, `log::LevelFilter::Trace` attempts to print out a message for each trace point, resulting in CPU stuck (for long-running tasks like module verification, kernel text patching). Userland platform (and SNP platform) would be fine because there are preemptions. LVBS doesn't have such a luxury.

jaybosamiya-ms (approved):
Thanks Sangho!  I think `info` (or even `warn`, if we want to be even more silent) as default max level for release mode is the correct move.

sangho2:
> Thanks Sangho! I think `info` (or even `warn`, if we want to be even more silent) as default max level for release mode is the correct move.

`warn` sounds good. Let me use it.

PR #790, opened 2026-04-21
Shebang support

CvvT:
Add shebang support for `execve` and the initial program to run.

To make the test run, I changed some `unimplemented` and `todo` to log + returning `EINVAL`.

### DIFF
--- a/litebox/src/fd/mod.rs
+++ b/litebox/src/fd/mod.rs
@@ -624,7 +624,7 @@ impl RawDescriptorStorage {
         // If this assertion failure is hit in practice, we might need to be more defensive via the
         // HashMap, rather than just silently allow big growth
         assert!(
-            raw_fd < self.stored_fds.len() + 100,
+            raw_fd < self.stored_fds.len() + 256,
             "explicit upper bound restriction for now; see implementation details"
         );
         if self.stored_fds.get(raw_fd).is_some_and(Option::is_some) {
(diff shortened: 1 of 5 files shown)


jaybosamiya-ms:
This code is more verbose than it needs to be (with clear indications of LLM-generation).  Afaict, it does seem to do what it says, but it is doing it in a very manual way (e.g., why maintain `depth` is a mutable variable that is incremented by 1 each loop with a manual check; that is _exactly_ what for loops can do `for _ in 0..SHEBANG_MAX_RECURSION` and outside the loop a single `Err(Errno::ELOOP)`).  

I'll let you decide if this is good enough or not since this is a primarily a shim-level change _and_ is quite an isolated feature

CvvT:
> This code is more verbose than it needs to be (with clear indications of LLM-generation). Afaict, it does seem to do what it says, but it is doing it in a very manual way (e.g., why maintain `depth` is a mutable variable that is incremented by 1 each loop with a manual check; that is _exactly_ what for loops can do `for _ in 0..SHEBANG_MAX_RECURSION` and outside the loop a single `Err(Errno::ELOOP)`).
> 
> I'll let you decide if this is good enough or not since this is a primarily a shim-level change _and_ is quite an isolated feature

Good catch! I didn't notice it. I guess I was only focusing on correctness. I hope now it is in a good shape. Thanks!

jaybosamiya-ms (approved):
This specific code path is largely insulated, so no need to block the PR, correctness for this is the primary goal, and I think it satisfies that well enough.  Thanks for making the updates!

PR #791, opened 2026-04-21
Implement SIGPIPE delivery on EPIPE

CvvT:
handle EPIPE error for various syscalls like `write` and `sendto`.

Also fixed the partial send issue where we incorrectly return an error code.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -22,7 +22,7 @@ use litebox_common_linux::{
 };
 use litebox_platform_multiplex::Platform;
 
-use crate::{ConstPtr, GlobalState, MutPtr, ShimFS, Task, syscalls::signal::siginfo_kill};
+use crate::{ConstPtr, GlobalState, MutPtr, ShimFS, Task, syscalls::signal};
 use core::sync::atomic::{AtomicUsize, Ordering};
 
 /// Task state shared by `CLONE_FS`.
@@ -504,7 +504,7 @@ impl<FS: ShimFS> Task<FS> {
             )
             .flatten();
         if let Err(Errno::EPIPE) = res {
-            self.send_signal(Signal::SIGPIPE, siginfo_kill(Signal::SIGPIPE));
+            self.send_signal(Signal::SIGPIPE, signal::siginfo_kill(Signal::SIGPIPE));
         }
         res
     }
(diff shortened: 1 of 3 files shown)


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -18,11 +18,11 @@ use litebox::{
 };
 use litebox_common_linux::{
     AtFlags, EfdFlags, EpollCreateFlags, FcntlArg, FileDescriptorFlags, FileStat, IoReadVec,
-    IoWriteVec, IoctlArg, TimeParam, errno::Errno,
+    IoWriteVec, IoctlArg, TimeParam, errno::Errno, signal::Signal,
 };
 use litebox_platform_multiplex::Platform;
 
-use crate::{ConstPtr, GlobalState, MutPtr, ShimFS, Task};
+use crate::{ConstPtr, GlobalState, MutPtr, ShimFS, Task, syscalls::signal::siginfo_kill};
```
jaybosamiya-ms: By convention, `use` declarations for types/macros/modules are fine, but we make sure not to do so for functions, in order to match the conventions most mature Rust projects use.

CvvT: I see. Fixed now, thanks!

litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -1445,7 +1445,13 @@ impl<FS: ShimFS> Task<FS> {
...
                 Ok(total_sent)
             },
-        )
+        );
+        if let Err(Errno::EPIPE) = res
+            && !flags.contains(SendFlags::NOSIGNAL)
+        {
+            self.send_signal(Signal::SIGPIPE, signal::siginfo_kill(Signal::SIGPIPE));
+        }
+        res
```
CvvT: > Partial send now becomes process termination
  litebox_shim_linux/src/syscalls/net.rs:1424-1444 (sys_sendmsg iovec loop) and litebox_shim_linux/src/syscalls/file.rs:621 (write_to_iovec used by sys_writev). The per-iovec loops propagate errors with ?. If iovec N succeeds and iovec N+1 hits EPIPE, Linux returns the partial byte count with no error and no signal. Here the accumulated count is discarded, EPIPE leaks to the caller, and — new in this PR — SIGPIPE is raised (default disposition: terminate). A pre-existing short-count bug is now upgraded to a fatal one. Suggested fix: in those loops, on Err(e) return Ok(total_sent) if total_sent > 0, else Err(e).

Fixed now.

wdcui (approved):
LGTM. Thanks.

PR #794, opened 2026-04-21
Further reduce unsafe from shim via zerocopy

jaybosamiya-ms:
This PR removes some more of the `unsafe` from our shim.  With this, the last of the maybe-uninit from our shim is also gone.


CvvT (approved):
LGTM, thanks!

PR #795, opened 2026-04-22
More unsafe cleanup

jaybosamiya-ms:


jaybosamiya-ms:
Opened separately from #794 since it is a slightly different kind of unsafe cleanup (still using zerocopy for parts, but mostly is just eliminating unnecessary casts).

CvvT (approved):
LGTM, thanks!

PR #796, opened 2026-04-22
Use checked arithmetic against VTL0-controllable integers

sangho2:
This PR uses checked arithmetic against VTL0-controllable integers to avoid potential attacks via integer overflows or underflows.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -87,15 +87,18 @@ pub(crate) fn init(is_bsp: bool) {
 
     if is_bsp {
         if let Ok((start, size)) = get_vtl1_memory_info() {
-            let end = start
-                .checked_add(size)
-                .ok_or(VsmError::IntegerOverflow)
-                .expect("VTL1 memory range overflow while protecting startup memory");
+            let Some(end) = start.checked_add(size) else {
+                panic!("Failed to protect VTL1 memory");
+            };
             debug_serial_println!("VSM: Protect GPAs from {:#x} to {:#x}", start, end);
+            let Ok(end) = PhysAddr::try_new(end) else {
+                panic!("Failed to protect VTL1 memory");
+            };
+            let start = PhysAddr::new(start);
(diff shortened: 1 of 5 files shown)


litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -904,11 +908,14 @@ fn apply_vtl0_text_patch(heki_patch: HekiPatch) -> Result<(), VsmError> {
...
 fn mshv_vsm_allocate_ringbuffer_memory(phys_addr: u64, size: usize) -> Result<i64, VsmError> {
+    let end = phys_addr
+        .checked_add(size as u64)
+        .ok_or(VsmError::IntegerOverflow)?;
     set_ringbuffer(PhysAddr::new(phys_addr), size);
     protect_physical_memory_range(
         PhysFrame::range(
             PhysFrame::containing_address(PhysAddr::new(phys_addr)),
-            PhysFrame::containing_address(PhysAddr::new(phys_addr + (size as u64))),
+            PhysFrame::containing_address(PhysAddr::new(end)),
```
CvvT: > [major] litebox_platform_lvbs/src/mshv/vsm.rs:909-910 (also 84, 914)                                                                                                             
  The new checked_add guards u64 wrap, but PhysAddr::new(phys_addr) and PhysAddr::new(end) still panic when VTL0 supplies a value with bits 52-63 set — the same panic-DoS hole the
   PR is meant to close.                                                                                                                                                           
  Use PhysAddr::try_new(...).map_err(|_| VsmError::InvalidPhysicalAddress)? on every PA constructed from VTL0 input (start, phys_addr, end at the three sites).

CvvT (approved):
Overall LGTM; agent found one potential issue.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Thanks! Technically, this PR doesn't focus on DoS since VTL0 can always DoS VTL1 (as an untrusted hypervisor does). Anyhow, I'll use `try_new()`.

PR #798, opened 2026-04-22
Fix futex test

CvvT:
The intention of the test was that all waiters are blocked and waiting for a wake up and then the main thread wake all of them up. However, it only uses `sleep` to ensure all waiters have called `futex` with wait operation, which may not hold, resulting in some waiters return immediately without being blocked. The fix is simple; some waiters may return `ImmediatelyWokenBecauseValueMismatch` and we just need to account for it.


jaybosamiya-ms (approved):
LGTM, thanks Weiteng!

PR #800, opened 2026-04-22
Remove 32-bit x86 support

jaybosamiya-ms:
We've decided that the payoff for maintaining 32-bit x86 support is not high enough to shoulder the ongoing maintenance burden of it.  Thus, this PR rips out 32-bit x86 support.

I've made sure to maintain `#[cfg(target_arch = "x86_64")]` and such in places where we want to continue documenting that we are making x86-64-specific decisions, so that adding support for arm64 and such is easier in the future.

There are still some x86-references (e.g., due to the rewriter), but I've chosen not to touch that due to some ongoing PRs in those areas.  A future pass can clean out the remaining references to 32-bit.


jaybosamiya-ms:
Note: the CI will _currently_ mark `Build and Test (32-bit)` as expected-but-not-completed until we change the repository settings, I'll do that closer to merging time, to try to reduce the timing window of impact of the CI change on other PRs.

litebox_common_linux/src/lib.rs
```diff
@@ -2543,32 +2385,13 @@ impl<Platform: litebox::platform::RawPointerProvider> SyscallRequest<Platform> {
             }
             // TODO: fcntl64 is identical to fcntl except certain commands (e.g., `F_OFD_SETLK`)
             // that we don't support yet.
```
CvvT: The comment can be removed as well.

CvvT (approved):
LGTM, thanks for making the effort.

litebox_common_linux/src/lib.rs
jaybosamiya-ms: Good catch, thanks! Removed.

PR #801, opened 2026-04-23
Fix dup

CvvT:
Looks like `dup` and `fcntl` with `F_DUPFD` return different error codes when the specified fd is larger than the process file limit.

Also fix a race issue in `dup` where the specified fd needs to to be closed and then reused, which requires performing `close` and `insert` in an atomic manner. Otherwise, in between these two operations, the fd might be taken by another thread.


wdcui (approved):
LGTM. Thanks.

PR #805, opened 2026-04-24
Enable seccomp filter

### LINKED ISSUE #793: Add secccomp enforcement for linux userland
We should update the linux userland runner to enable seccomp enforcement after the initial loading.

CvvT:
Address #793 to enable seccomp filter for Linux userland.

In release mode, disallowed syscalls would return `EINVAL` directly while in debug mode, they are also logged in addition to returning `EINVAL`.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -2394,4 +2394,18 @@ mod tests {
             prev = page.end;
         }
     }
+
+    #[test]
+    fn test_seccomp_filter() {
+        let _platform = LinuxUserland::new(None);
+        LinuxUserland::enable_seccomp_filter();
+        let pathname = "/tmp/test_seccomp";
+        let res = unsafe {
+            syscalls::syscall2(syscalls::Sysno::mkdir, pathname.as_ptr() as usize, 0o755)
+        };
+        assert!(
+            res.is_err() && res.unwrap_err() == syscalls::Errno::EINVAL,
+            "mkdir should be blocked by seccomp filter"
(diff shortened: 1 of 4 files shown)


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -412,6 +412,91 @@ impl LinuxUserland {
...
+        ];
+        let rule_map: std::collections::BTreeMap<i64, Vec<SeccompRule>> =
+            rules.into_iter().collect();
+        let filter = SeccompFilter::new(
+            rule_map,
+            // In debug builds, log violations instead of silently returning an error so that
+            // it won't fail silently during development (which may hard to debug) and we can
+            // tell there are missing seccomp rules to be added by comparing debug and release runs.
+            if cfg!(debug_assertions) {
+                SeccompAction::Log
```
jaybosamiya-ms: Minor suggestion: we can use [`SCMP_ACT_TRAP`](https://www.man7.org/linux/man-pages/man3/seccomp_init.3.html#:~:text=seccomp%20filter%20rules.-,SCMP_ACT_TRAP,-The%20thread%20will) to get a SIGSYS with SYS_SECCOMP that we can use to actually log via `litebox_util_log::error!(...)` which might be nicer than needing to diff debug/release runs

CvvT: Good point! One minor issue: `litebox_util_log::error!` is not async-signal-safe, so I use raw syscall `write` instead.

wdcui (approved):
LGTM. Thanks.

PR #806, opened 2026-04-24
Add arch-specific provider; remove PunchthroughProvider

jaybosamiya-ms:
This PR removes the last vestiges of the `PunchthroughProvider`, moving out architecture-specific register management into its own trait.  For now, I've only added support for x86-64's fsbase/gsbase, with platforms deciding which ones they'd like to support and which ones they do not want to support.  Currently, the kernel platforms explicitly do an `unimplemented!` for the gsbase (see https://github.com/microsoft/litebox/pull/806#discussion_r3210873538).

### DIFF
--- a/litebox_common_linux/src/errno/mod.rs
+++ b/litebox_common_linux/src/errno/mod.rs
@@ -615,8 +615,8 @@ impl From<litebox::platform::ArchSpecificError> for Errno {
             litebox::platform::ArchSpecificError::RegisterUnsupported => {
                 // Reaching here means that a shim is attempting to easily run on a platform that
                 // fully disallows it. There is no reasonable handling here, so we panic.
-                // XXX: should this be ENOSYS?
-                panic!()
+                // XXX: should this be ENOSYS or EINVAL?
+                unimplemented!()
             }
             litebox::platform::ArchSpecificError::RegisterReserved => Errno::EINVAL,
             _ => unimplemented!(),
--- a/litebox_common_linux/src/errno/mod.rs
+++ b/litebox_common_linux/src/errno/mod.rs
@@ -607,3 +607,19 @@ impl From<litebox::fs::errors::TruncateError> for Errno {
         }
     }
 }
+
+#[cfg(target_arch = "x86_64")]
+impl From<litebox::platform::ArchSpecificError> for Errno {
+    fn from(value: litebox::platform::ArchSpecificError) -> Self {
+        match value {
+            litebox::platform::ArchSpecificError::RegisterUnsupported => {
+                // Reaching here means that a shim is attempting to easily run on a platform that
+                // fully disallows it. There is no reasonable handling here, so we panic.
+                // XXX: should this be ENOSYS?
+                panic!()
+            }
+            litebox::platform::ArchSpecificError::RegisterReserved => Errno::EINVAL,
+            _ => unimplemented!(),
+        }
+    }
+}
--- a/litebox_platform_linux_kernel/src/lib.rs
+++ b/litebox_platform_linux_kernel/src/lib.rs
@@ -13,10 +13,10 @@ use litebox::mm::linux::PageRange;
 use litebox::platform::RawPointerProvider;
 use litebox::platform::page_mgmt::FixedAddressBehavior;
 use litebox::platform::{
-    IPInterfaceProvider, ImmediatelyWokenUp, PageManagementProvider, Provider, Punchthrough,
-    PunchthroughProvider, PunchthroughToken, RawMutexProvider, TimeProvider, UnblockedOrTimedOut,
+    ArchSpecificError, ArchSpecificProvider, ArchSpecificRegister, IPInterfaceProvider,
+    ImmediatelyWokenUp, PageManagementProvider, Provider, RawMutexProvider, TimeProvider,
+    UnblockedOrTimedOut,
 };
-use litebox_common_linux::PunchthroughSyscall;
 use litebox_common_linux::errno::Errno;
 
 extern crate alloc;
@@ -54,34 +54,6 @@ impl<Host: HostInterface> core::fmt::Debug for LinuxKernel<Host> {
     }
 }
 
-pub struct LinuxPunchthroughToken<'a, Host: HostInterface> {
-    punchthrough: PunchthroughSyscall<'a, LinuxKernel<Host>>,
-    host: core::marker::PhantomData<Host>,
-}
-
-impl<'a, Host: HostInterface> PunchthroughToken for LinuxPunchthroughToken<'a, Host> {
-    type Punchthrough = PunchthroughSyscall<'a, LinuxKernel<Host>>;
-
-    fn execute(
-        self,
-    ) -> Result<
-        <Self::Punchthrough as Punchthrough>::ReturnSuccess,
-        litebox::platform::PunchthroughError<<Self::Punchthrough as Punchthrough>::ReturnFailure>,
-    > {
-        let r = match self.punchthrough {
-            PunchthroughSyscall::SetFsBase { addr } => {
-                unsafe { litebox_common_linux::wrfsbase(addr) };
-                Ok(0)
-            }
-            PunchthroughSyscall::GetFsBase => Ok(unsafe { litebox_common_linux::rdfsbase() }),
-        };
-        match r {
-            Ok(v) => Ok(v),
-            Err(e) => Err(litebox::platform::PunchthroughError::Failure(e)),
-        }
-    }
-}
-
 impl<Host: HostInterface> Provider for LinuxKernel<Host> {}
 
 // TODO: implement pointer validation to ensure the pointers are in user space.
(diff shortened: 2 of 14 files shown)


litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -99,17 +71,34 @@ impl<Host: HostInterface> RawPointerProvider for LinuxKernel<Host> {
...
+        reg: &ArchSpecificRegister,
+        val: usize,
+    ) -> Result<(), ArchSpecificError> {
+        match reg {
+            ArchSpecificRegister::FsBase => {
+                unsafe { litebox_common_linux::wrfsbase(val) };
+                Ok(())
+            }
+            ArchSpecificRegister::GsBase => {
+                unsafe { litebox_common_linux::wrgsbase(val) };
```
CvvT: In kernel mode, we call `swapgs` at the entry and exit of a syscall. At this point, userspace gsbase is stored in MSR (see https://elixir.bootlin.com/linux/v5.19.17/source/arch/x86/kernel/process_64.c#L763).

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -99,17 +71,34 @@ impl<Host: HostInterface> RawPointerProvider for LinuxKernel<Host> {
...
-    ) -> Option<Self::PunchthroughToken<'a>> {
-        Some(LinuxPunchthroughToken {
-            punchthrough,
-            host: core::marker::PhantomData,
-        })
+        reg: &ArchSpecificRegister,
+    ) -> Result<usize, ArchSpecificError> {
+        match reg {
+            ArchSpecificRegister::FsBase => Ok(unsafe { litebox_common_linux::rdfsbase() }),
+            ArchSpecificRegister::GsBase => Ok(unsafe { litebox_common_linux::rdgsbase() }),
```
CvvT: Similar

litebox_platform_lvbs/src/lib.rs
```diff
@@ -464,40 +460,34 @@ impl<Host: HostInterface> RawPointerProvider for LinuxKernel<Host> {
...
+                unsafe { litebox_common_linux::wrfsbase(val) };
+                Ok(())
             }
-            PunchthroughSyscall::GetFsBase => Ok(unsafe { litebox_common_linux::rdfsbase() }),
-        };
-        match r {
-            Ok(v) => Ok(v),
-            Err(e) => Err(litebox::platform::PunchthroughError::Failure(e)),
+            ArchSpecificRegister::GsBase => {
+                unsafe { litebox_common_linux::wrgsbase(val) };
```
CvvT: probably same

litebox_platform_lvbs/src/lib.rs
```diff
@@ -464,40 +460,34 @@ impl<Host: HostInterface> RawPointerProvider for LinuxKernel<Host> {
...
-    ) -> Option<Self::PunchthroughToken<'a>> {
-        Some(LinuxPunchthroughToken {
-            punchthrough,
-            host: core::marker::PhantomData,
-        })
+        reg: &ArchSpecificRegister,
+    ) -> Result<usize, ArchSpecificError> {
+        match reg {
+            ArchSpecificRegister::FsBase => Ok(unsafe { litebox_common_linux::rdfsbase() }),
+            ArchSpecificRegister::GsBase => Ok(unsafe { litebox_common_linux::rdgsbase() }),
```
CvvT: same

litebox_platform_linux_kernel/src/lib.rs
CvvT: Note the issue was found by GPT-5.5

litebox_common_linux/src/errno/mod.rs
```diff
@@ -579,3 +579,19 @@ impl From<litebox::fs::errors::TruncateError> for Errno {
...
+
+#[cfg(target_arch = "x86_64")]
+impl From<litebox::platform::ArchSpecificError> for Errno {
+    fn from(value: litebox::platform::ArchSpecificError) -> Self {
+        match value {
+            litebox::platform::ArchSpecificError::RegisterUnsupported => {
+                // Reaching here means that a shim is attempting to easily run on a platform that
+                // fully disallows it. There is no reasonable handling here, so we panic.
+                // XXX: should this be ENOSYS?
+                panic!()
```
wdcui: I wouldn't expect an error code conversion function to panic?

litebox_platform_linux_kernel/src/lib.rs
wdcui: But it depends on when/where this function is called, right? If the caller is after `swapgs` and it actually knows that it wants to access which register, this function is correct, right?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -1193,40 +1191,44 @@ impl litebox::platform::SystemTime for SystemTime {
...
+            litebox::platform::ArchSpecificRegister::FsBase => {
+                set_guest_fsbase(val);
+                Ok(())
             }
-            #[cfg(target_arch = "x86_64")]
-            PunchthroughSyscall::GetFsBase => Ok(get_guest_fsbase()),
+            litebox::platform::ArchSpecificRegister::GsBase => {
+                // GS base is used internally by this platform to hold the host
+                // TLS base across the guest/host fs-gs swap, so it is not
+                // directly programmable by the guest.
```
wdcui, line 1304: Why would you expect the guest to call this platform trait?

wdcui (commented):
I left some comments below. Will have some offline discussion with you.

litebox_platform_linux_kernel/src/lib.rs
wdcui: Ah, I see the issue. When this kernel platform code executes, if we do wrfsbase or wrgsbase, it actually directly affects the current gsbase used by the kernel. Is my understanding correct?

sangho2: we need `wrmsr(IA32_KERNEL_GS_BASE, val)` here to change the "user" GSBASE here due to `swapgs`.

jaybosamiya-ms: Thanks for catching the issue.  Previously, the kernel mode side of the platform did not do anything with the gsbase, so it was probably wrong of me to introduce that in parallel with the refactor.  It would be confusing to have that be part of this change, so for now, I'm just replacing it with a `unimplemented!()` with a pointer to this comment thread, so that we can fix it up separately later (and not block this PR).

litebox_platform_linux_userland/src/lib.rs
jaybosamiya-ms: Thanks!  The guest program may invoke `SYS_arch_prctl` syscall, which the shim needs to handle, and that _eventually_ invokes the platform trait.  I'm adding some more comments towards this at the trait itself.

litebox_platform_linux_kernel/src/lib.rs
jaybosamiya-ms: similarly punted

litebox_platform_lvbs/src/lib.rs
jaybosamiya-ms: similarly punted

litebox_platform_lvbs/src/lib.rs
jaybosamiya-ms: similarly punted

litebox_common_linux/src/errno/mod.rs
jaybosamiya-ms: Indeed, it ideally should not panic.  However, there does not seem to be any valid handling I am able to think of, so in some sense, no option but to panic and then figure it out when a real application hits it.  I'll change it to `unimplemented!()` for now to denote that we need to handle it better later.

jaybosamiya-ms:
This PR has been in limbo for a little while. I've now added some more clarity to the docs, and switched a couple of things to `unimplemented!` where it did not make sense for this PR to be blocked on that (the gsbase stuff were changes that I'd introduced to show more of how things might be used, but the exact implementation is not critical to this refactoring, so I've punted that for later).  Hopefully once we can merge this one in, we can fill out the `unimplemented!`s separately if/when we actually hit them in practice.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -99,17 +71,37 @@ impl<Host: HostInterface> RawPointerProvider for LinuxKernel<Host> {
...
+        val: usize,
+    ) -> Result<(), ArchSpecificError> {
+        match reg {
+            ArchSpecificRegister::FsBase => {
+                unsafe { litebox_common_linux::wrfsbase(val) };
+                Ok(())
+            }
+            ArchSpecificRegister::GsBase => {
+                // See https://github.com/microsoft/litebox/pull/806#discussion_r3210873538
+                unimplemented!()
```
wdcui, line 87: Why not return the `RegisterUnsupported` or `RegisterReserved` error?

wdcui (approved):
LGTM. I left one comment below. Thanks!

litebox_platform_linux_kernel/src/lib.rs
jaybosamiya-ms: In some sense those errors are intended to capture _intention_ for the platform.  `RegisterReserved` == this register will not be supported by this platform, do not ask us for this one.  `RegisterUnsupported` == this register is _currently_ unsupported, but that might be a transient reason.  The `unimplemented!` is actually even noisier signal that is intended to say "hey, if anyone uses this, we need to pay attention and actually fix this up with a real implementation, we believe it is possible to do, just haven't taken the time to do it yet"

PR #807, opened 2026-04-24
Fix mremap

CvvT:
This PR fixes three issues in `mremap` found by AI.

1. `old_size` and `new_size` are rounded down to page boundary instead of up (Linux expects round-up).
2.  The default impl of `remap_pages` copies wrong chunk size, reading out-of-bounds on multi-page remaps.
3. Fix overlap check in kernel platform.

### DIFF
--- a/litebox_common_linux/src/mm.rs
+++ b/litebox_common_linux/src/mm.rs
@@ -20,12 +20,6 @@ fn align_up(addr: usize, align: usize) -> usize {
     (addr + align - 1) & !(align - 1)
 }
 
-#[inline]
-fn align_down(addr: usize, align: usize) -> usize {
-    debug_assert!(align.is_power_of_two());
-    addr & !(align - 1)
-}
-
 pub fn do_mmap<
     Platform: litebox::platform::RawPointerProvider
         + litebox::sync::RawSyncPrimitivesProvider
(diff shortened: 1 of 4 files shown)


CvvT:
The second issue is also fixed in #754, but I feel it might be better to include it in this PR.

litebox_platform_linux_kernel/src/lib.rs
```diff
@@ -459,7 +459,7 @@ impl<Host: HostInterface, const ALIGN: usize> PageManagementProvider<ALIGN> for
             .ok_or(litebox::platform::page_mgmt::RemapError::Unaligned)?;
         let new_range = PageRange::new(new_range.start, new_range.end)
             .ok_or(litebox::platform::page_mgmt::RemapError::Unaligned)?;
-        if old_range.start.max(new_range.start) <= old_range.end.min(new_range.end) {
+        if old_range.start.max(new_range.start) < old_range.end.min(new_range.end) {
```
sangho2, line 462: `litebox_platform_lvbs/src/lib.rs:1269` does have the same problem. It would be better to fix these two together.

litebox_common_linux/src/mm.rs
```diff
@@ -184,8 +178,8 @@ pub fn sys_mremap<
         return Err(Errno::EINVAL);
     }
 
-    let old_size = align_down(old_size, PAGE_SIZE);
-    let new_size = align_down(new_size, PAGE_SIZE);
+    let old_size = align_up(old_size, PAGE_SIZE);
+    let new_size = align_up(new_size, PAGE_SIZE);
```
sangho2: `checked_next_multiple_of()` might be better?

sangho2 (approved):

litebox_common_linux/src/mm.rs
CvvT: Good point! Fixed.

litebox_platform_linux_kernel/src/lib.rs
CvvT: Sure, fixed now. Thanks!

wdcui (approved):
LGTM. Thanks.

PR #808, opened 2026-04-24
Map physical ringbuffer memory range in VTL1

karissanchez:
Map the ringbuffer physical memory range in VTL1 and write to the ringbuffer using memcpy_fallible with the VTL1 virtual address. This will be faster than copying a slice to VTL0 physical memory for each ringbuffer write.


sangho2:
I have a question before reviewing this PR. Do we need to improve the logging performance? Having this persistent, write-only mapping is probably fine, but it is incompatible with our VTL0 memory abstraction, requiring another memory abstraction or exception.

sangho2:
Let's track this one with Issue #822.

PR #809, opened 2026-04-25
Fix wake on Windows

CvvT:
@sangho2 found out a subtle issue in `RwLock` with help of GPT-5.5.

Agent implemented a test to reproduce the issue (see `test_rwlock_readers_not_starved_after_writer_handoff`) and the scenario for the bug is the following:

    1. W1 holds the write lock.
    2. Reader R blocks, setting `READERS_WAITING`.
    3. Writer W2 blocks, setting `WRITERS_WAITING` (and internal `other_writers_waiting`).
    4. W1 unlocks → `wake_writer_or_readers` sees both flags, wakes W2, skips readers.
    5. W2 acquires the lock **with the `WRITERS_WAITING` bit** (from `other_writers_waiting`).
    6. W2 unlocks → `wake_writer_or_readers` sees both flags again, tries to wake a writer.

On Windows `wake_one` always returns `true` even though no writer is blocked (because Windows does not tell how may waiters were woken up and our previous implementation just returns the number of waiters requested), so it returns early **without waking readers**. Readers are stuck forever.

Looking at how [std handles Windows](https://android.googlesource.com/toolchain/rustc/+/HEAD/library/std/src/sys/pal/windows/futex.rs#87), it turns out it always return `false` (meaning no waiters were woken up). This fix just follows that.

### DIFF
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -1355,8 +1355,8 @@ impl litebox::platform::RawMutex for RawMutex {
         }
 
         // For windows, the OS kernel does not tell us how many threads were actually woken up,
-        // so we just return `n`
-        n as usize
+        // so we just return zero
+        0
     }
 
     fn block(&self, val: u32) -> Result<(), ImmediatelyWokenUp> {
--- a/litebox_platform_windows_userland/src/lib.rs
+++ b/litebox_platform_windows_userland/src/lib.rs
@@ -1355,7 +1355,7 @@ impl litebox::platform::RawMutex for RawMutex {
         }
 
         // For windows, the OS kernel does not tell us how many threads were actually woken up,
-        // so we just return zero
+        // so we return zero to indicate that the count is unknown.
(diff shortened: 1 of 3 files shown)


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -1355,8 +1355,8 @@ impl litebox::platform::RawMutex for RawMutex {
         }
 
         // For windows, the OS kernel does not tell us how many threads were actually woken up,
-        // so we just return `n`
-        n as usize
+        // so we just return zero
+        0
```
sangho2: The comment for `wake_many` in the platform trait needs to be updated accordingly.

sangho2 (approved):

litebox_platform_windows_userland/src/lib.rs
CvvT: Good catch! Fixed now, thanks!

wdcui (approved):
LGTM. Thanks.

PR #810, opened 2026-04-25
Runtime ELF patching and rtld_audit removal

wdcui:
This PR replace the rtld_audit based syscall interception with runtime ELF patching during mmap. When a PROT_EXEC segment is mapped, the shim now patches syscall instructions in-place and places trampoline stubs in a dynamically-allocated region near the code.


wdcui:

sangho2:
Agents have found a bunch of vulnerable code paths that do not rewrite executable pages and more. We should fix all of them and perhaps should merge #805 before this PR. I add them below to ease agent-driven code patching. I'll "manually" add review later.

- do_mmap_file() only fails the mapping if maybe_patch_exec_segment() returns false (litebox_shim_linux/src/syscalls/mm.rs:129-140). The runtime patching path logs and returns true for trampoline allocation failure, too-far trampoline, entry write failure, mprotect failures, read/writeback failures, and rewriter errors (:692-724, :747-770, :773-889). That leaves the executable mapping live with original syscall bytes.
- patch_code_segment() can mutate code_buf by replacing unpatchable syscalls with traps and return empty stubs plus skipped addresses (litebox_syscall_rewriter/src/lib.rs:448-452, :753-765). The Linux mmap caller only writes code_buf back inside Ok(stubs) if !stubs.is_empty() (litebox_shim_linux/src/syscalls/mm.rs:810-864), while Ok(_) does nothing (:878-880). Trap-only rewrites are therefore dropped. In mixed cases, the trap bytes are written back because stubs are non-empty, but skipped sites are only logged and the mapping still succeeds.
- Runtime rewriting is gated on the original mmap prot containing PROT_EXEC (litebox_shim_linux/src/syscalls/mm.rs:118-134). mprotect dispatch goes straight to sys_mprotect (litebox_shim_linux/src/lib.rs:641-642), which delegates directly to common mprotect (litebox_shim_linux/src/syscalls/mm.rs:375-381). Since mappings get VM_MAYEXEC by default, a non-exec file mapping can later become exec without rewrite.
- AOT rewriting filters actual executable text sections (litebox_syscall_rewriter/src/lib.rs:183-194, :219-230, :282-312). Runtime rewriting passes the whole mapped executable range to patch_code_segment() (litebox_shim_linux/src/syscalls/mm.rs:773-795) and writes the whole buffer back (:863-864). That can patch bytes in executable PT_LOAD padding/data that AOT would not touch.
-  For unpatched binaries, ElfParsedFile::load() only bumps info.brk (litebox_common_linux/src/loader.rs:487-491); it does not map or reserve a VMA. The Linux loader loads the interpreter before set_initial_brk(info.brk (litebox_shim_linux/src/loader/elf.rs:240-250), and ET_DYN reservation only reserves the ELF load span, not the extra runtime-trampoline gap (litebox_common_linux/src/loader.rs:398-405). Nuance: brk expansion checks overlaps (litebox/src/mm/mod.rs:343-356), so it should not silently map over an interpreter VMA above the current break. But because the unreserved gap is below the artificially bumped initial break, later mappings can occupy it, runtime trampoline allocation can collide/fail, and brk shrink can treat that gap as heap-owned and unmap mappings inside it (:323-340).
- Untrusted trampoline trailer disables runtime rewriting. parse_trampoline() treats a trailer with TRAMPOLINE_MAGIC and trampoline_size == 0 as success, interpreting it as “rewriter checked this binary and found no syscall instructions,” without authenticating that claim (litebox_common_linux/src/loader.rs:270-306). The Linux loader accepts that result and proceeds (litebox_shim_linux/src/loader/elf.rs:182-190), while has_trampoline() remains false so it only reserves runtime space (:200-209). Later, the mmap-time runtime hook independently calls check_trampoline_magic(), which only checks the final 32-byte magic and returns pre_patched = true even for trampoline_size == 0 (litebox_shim_linux/src/syscalls/mm.rs:541-561). maybe_patch_exec_segment() then takes the state.pre_patched branch, skips mapping when size is zero, and returns true without scanning or rewriting the executable segment (:599-664). A malicious unpatched ELF can therefore append a fake LITEBOX0 trailer with zero size and keep raw syscall instructions executable.
- OP-TEE accepts UnpatchedBinary (litebox_shim_optee/src/loader/elf.rs:205-208), maps ELF bytes through anonymous mappings plus copy (:107-145), and OP-TEE mmap rejects file-backed mappings (litebox_shim_optee/src/syscalls/mm.rs:78-88). The Linux runtime mmap rewriter is not present there, so the comment that mmap-time patching will handle it is wrong. Caveat: on LVBS, raw syscall is still routed through the platform syscall callback, but OP-TEE-on-Linux-userland has the same raw-host-syscall concern as the Linux userland shim.
- Remapping the same file offset leaves fresh unpatched code (litebox_shim_linux/src/syscalls/mm.rs:730). patched_offsets is keyed only by file offset. A second mmap() of the same executable file offset creates new bytes from the original file, but the rewriter skips it as "already patched." That gives another direct raw-syscall mapping.
- ET_DYN base address inference is unreliable (litebox_shim_linux/src/syscalls/mm.rs:142, litebox_shim_linux/src/syscalls/mm.rs:501, litebox_shim_linux/src/syscalls/mm.rs:587). The runtime state uses the first observed mapping address as the ELF base. If the first mapping is not the offset-0/base mapping, trampoline addresses are computed incorrectly, causing skipped rewriting, failed jumps, or wrong mappings.
- Trampoline metadata is less validated in the mmap path than in the loader (litebox_shim_linux/src/syscalls/mm.rs:558, litebox_shim_linux/src/syscalls/mm.rs:603, litebox_shim_linux/src/syscalls/mm.rs:610). The mmap path does not validate alignment, bounds, file_offset + size == header_offset, overflow, or sane address ranges before MAP_FIXED. This can clobber guest mappings or amplify the forged-header bypass.

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -195,9 +195,37 @@ impl<'a> FileAndParsed<'a> {
...
+    ) -> Result<litebox_common_linux::loader::MappingInfo, ElfLoaderError> {
+        let syscall_entry_point = self.file.task.global.platform.get_syscall_entry_point();
+        let reserve = if syscall_entry_point != 0 && !self.parsed.has_trampoline() {
+            Some(litebox::mm::linux::DEFAULT_RESERVED_SPACE_SIZE)
+        } else {
+            None
+        };
+        let result = self.parsed.load(&mut self.file, &mut &*platform, reserve);
+        Ok(result?)
+    }
```
sangho2: OP-TEE shim doesn't (and will not?) support runtime ELF patching. In that sense, it is better not to use `load_mapped`.

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -214,10 +242,7 @@ impl<'a> ElfLoader<'a> {
         }
         let task = self.main.file.task;
         let global = &task.global;
-        let ldelf_info = self
-            .main
-            .parsed
-            .load(&mut self.main.file, &mut &*global.platform)?;
+        let ldelf_info = self.main.load_mapped(global.platform)?;
```
sangho2: same. better to revert?

CvvT:
@sangho2 and I discussed about it and have an alternative idea: how about we hook `open` to rewrite all ELF files and then `mmap` only needs to load the trampoline code? It may avoid the issues like the `syscall` instruction happens to across two `mmap` regions (unlikely though) and then is missed by the rewriter.

One potential issue with this design is that if we are going to use broker for fs, the changes to the ELF files are reflected to the host as well (unless we have some sort of overlay fs).

wdcui:
I thought about this before. What if I just open an elf file for binary analysis?

sangho2:
> I thought about this before. What if I just open an elf file for binary analysis?

mmap does have more corner cases than open. To tackle all of them (or to avoid regression to due to a lack of rtld), we might need to add a bunch of new code.

litebox_shim_optee/src/loader/elf.rs
wdcui: Fixed.

litebox_shim_optee/src/loader/elf.rs
wdcui: Fixed.

wdcui:
> * do_mmap_file() only fails the mapping if maybe_patch_exec_segment() returns false (litebox_shim_linux/src/syscalls/mm.rs:129-140). The runtime patching path logs and returns true for trampoline allocation failure, too-far trampoline, entry write failure, mprotect failures, read/writeback failures, and rewriter errors (:692-724, :747-770, :773-889). That leaves the executable mapping live with original syscall bytes.
> * patch_code_segment() can mutate code_buf by replacing unpatchable syscalls with traps and return empty stubs plus skipped addresses (litebox_syscall_rewriter/src/lib.rs:448-452, :753-765). The Linux mmap caller only writes code_buf back inside Ok(stubs) if !stubs.is_empty() (litebox_shim_linux/src/syscalls/mm.rs:810-864), while Ok(_) does nothing (:878-880). Trap-only rewrites are therefore dropped. In mixed cases, the trap bytes are written back because stubs are non-empty, but skipped sites are only logged and the mapping still succeeds.
> * Runtime rewriting is gated on the original mmap prot containing PROT_EXEC (litebox_shim_linux/src/syscalls/mm.rs:118-134). mprotect dispatch goes straight to sys_mprotect (litebox_shim_linux/src/lib.rs:641-642), which delegates directly to common mprotect (litebox_shim_linux/src/syscalls/mm.rs:375-381). Since mappings get VM_MAYEXEC by default, a non-exec file mapping can later become exec without rewrite.
> * AOT rewriting filters actual executable text sections (litebox_syscall_rewriter/src/lib.rs:183-194, :219-230, :282-312). Runtime rewriting passes the whole mapped executable range to patch_code_segment() (litebox_shim_linux/src/syscalls/mm.rs:773-795) and writes the whole buffer back (:863-864). That can patch bytes in executable PT_LOAD padding/data that AOT would not touch.
> * For unpatched binaries, ElfParsedFile::load() only bumps info.brk (litebox_common_linux/src/loader.rs:487-491); it does not map or reserve a VMA. The Linux loader loads the interpreter before set_initial_brk(info.brk (litebox_shim_linux/src/loader/elf.rs:240-250), and ET_DYN reservation only reserves the ELF load span, not the extra runtime-trampoline gap (litebox_common_linux/src/loader.rs:398-405). Nuance: brk expansion checks overlaps (litebox/src/mm/mod.rs:343-356), so it should not silently map over an interpreter VMA above the current break. But because the unreserved gap is below the artificially bumped initial break, later mappings can occupy it, runtime trampoline allocation can collide/fail, and brk shrink can treat that gap as heap-owned and unmap mappings inside it (:323-340).
> * Untrusted trampoline trailer disables runtime rewriting. parse_trampoline() treats a trailer with TRAMPOLINE_MAGIC and trampoline_size == 0 as success, interpreting it as “rewriter checked this binary and found no syscall instructions,” without authenticating that claim (litebox_common_linux/src/loader.rs:270-306). The Linux loader accepts that result and proceeds (litebox_shim_linux/src/loader/elf.rs:182-190), while has_trampoline() remains false so it only reserves runtime space (:200-209). Later, the mmap-time runtime hook independently calls check_trampoline_magic(), which only checks the final 32-byte magic and returns pre_patched = true even for trampoline_size == 0 (litebox_shim_linux/src/syscalls/mm.rs:541-561). maybe_patch_exec_segment() then takes the state.pre_patched branch, skips mapping when size is zero, and returns true without scanning or rewriting the executable segment (:599-664). A malicious unpatched ELF can therefore append a fake LITEBOX0 trailer with zero size and keep raw syscall instructions executable.
> * OP-TEE accepts UnpatchedBinary (litebox_shim_optee/src/loader/elf.rs:205-208), maps ELF bytes through anonymous mappings plus copy (:107-145), and OP-TEE mmap rejects file-backed mappings (litebox_shim_optee/src/syscalls/mm.rs:78-88). The Linux runtime mmap rewriter is not present there, so the comment that mmap-time patching will handle it is wrong. Caveat: on LVBS, raw syscall is still routed through the platform syscall callback, but OP-TEE-on-Linux-userland has the same raw-host-syscall concern as the Linux userland shim.
> * Remapping the same file offset leaves fresh unpatched code (litebox_shim_linux/src/syscalls/mm.rs:730). patched_offsets is keyed only by file offset. A second mmap() of the same executable file offset creates new bytes from the original file, but the rewriter skips it as "already patched." That gives another direct raw-syscall mapping.
> * ET_DYN base address inference is unreliable (litebox_shim_linux/src/syscalls/mm.rs:142, litebox_shim_linux/src/syscalls/mm.rs:501, litebox_shim_linux/src/syscalls/mm.rs:587). The runtime state uses the first observed mapping address as the ELF base. If the first mapping is not the offset-0/base mapping, trampoline addresses are computed incorrectly, causing skipped rewriting, failed jumps, or wrong mappings.
> * Trampoline metadata is less validated in the mmap path than in the loader (litebox_shim_linux/src/syscalls/mm.rs:558, litebox_shim_linux/src/syscalls/mm.rs:603, litebox_shim_linux/src/syscalls/mm.rs:610). The mmap path does not validate alignment, bounds, file_offset + size == header_offset, overflow, or sane address ranges before MAP_FIXED. This can clobber guest mappings or amplify the forged-header bypass.

I worked an agent and fixed some of the issues. Here is a summary.

Context: Syscall rewriting is for reliability, NOT security. Security is enforced by kernel-level monitors (seccomp). Adversarial binary scenarios are out of scope.

================================================================================

#1 — Runtime patching returns true on failures, leaving unpatched exec mappings (Medium)

Status: No action.

do_mmap_file() only fails the mapping if maybe_patch_exec_segment() returns false. The runtime patching path logs and returns true for: trampoline allocation failure, too-far trampoline, entry write failure, mprotect failures, read/writeback failures, and rewriter errors. This leaves the executable mapping live with original syscall bytes.

Rationale: Returning false would fail the mmap (OutOfMemory), breaking legitimate programs that happen to have unmappable trampoline regions. Since rewriting is for reliability not security, letting the program continue with unintercepted syscalls is the less-harmful failure mode. The two options are:
  1. Fail the mmap → kills the program
  2. Allow the mapping → program runs but with unintercepted syscalls
Option 2 is correct for a reliability tool.

================================================================================

#2 — Trap-only rewrites dropped when stubs.is_empty() (High)

Status: FIXED (commit 71436220).

patch_code_segment() can mutate code_buf by replacing unpatchable syscalls with ICEBP;HLT traps and return empty stubs. The Ok(stubs) if !stubs.is_empty() arm wrote code_buf back, but the Ok(_) arm did nothing — trap bytes were silently discarded.

Fix: The Ok(_) arm now writes back code_buf if it differs from the original code, then falls through to restore RX protections.

================================================================================

#3 — mprotect bypass: non-exec → exec without rewriting (High)

Status: No action in this PR. Worth filing as follow-up issue.

mprotect dispatch (lib.rs:641-643) goes straight to sys_mprotect with no special handling. Since mappings get VM_MAYEXEC by default, a non-exec file mapping can later become exec without rewrite.

Rationale: Fixing this requires intercepting mprotect, tracking file-backed VMAs, and running the rewriter at mprotect time — a significant architectural change beyond the scope of this PR. Only exploitable by adversarial code (deliberately mapping non-exec then flipping to exec), which is out of scope for a reliability tool.

================================================================================

#4 — Runtime rewriting patches entire mapped exec range, not just .text (Low)

Status: No action.

AOT rewriting filters actual executable text sections. Runtime rewriting passes the whole mapped executable range to patch_code_segment() and writes the whole buffer back, potentially patching bytes in executable PT_LOAD padding/data that AOT would not touch.

Rationale: The rewriter scans for actual 0F 05 (syscall) opcodes, so it only patches real syscall instructions. False positives (0F 05 appearing as data constants within executable PT_LOAD segments) are vanishingly rare in practice, and the consequence (a corrupted data constant) is benign compared to leaving a real syscall unpatched.

================================================================================

#5 — Unpatched binary brk/VMA reservation gap (Medium)

Status: No action in this PR. Worth filing as follow-up issue.

For unpatched binaries, load() bumps info.brk but doesn't map or reserve a VMA for the trampoline gap. The unreserved gap below the artificially bumped initial break can be occupied by later mappings, causing runtime trampoline allocation to collide/fail. brk shrink could also treat that gap as heap-owned and unmap mappings inside it.

Rationale: Runtime trampoline allocation uses MAP_FIXED_NOREPLACE with fallback, so collisions result in degraded patching (comment #1 applies — patching silently doesn't happen but the program continues). The brk-shrink-unmapping scenario is theoretical — real programs don't shrink brk into this region. Low risk but worth tracking.

================================================================================

#6 — Fake LITEBOX0 trailer bypasses runtime rewriting (Critical per reviewer)

Status: No action.

A malicious unpatched ELF can append a fake LITEBOX0 trailer with trampoline_size=0. parse_trampoline() treats this as "rewriter checked this binary and found no syscall instructions." check_trampoline_magic() returns pre_patched=true even for size=0. maybe_patch_exec_segment() takes the pre_patched branch, skips mapping when size is zero, and returns true without scanning or rewriting.

Rationale: This is the clearest out-of-scope item. Rewriting is for reliability, not security. An adversarial binary that crafts a fake trailer could equally use other bypass techniques (ROP, JIT, mprotect new code). Authenticating the trailer would add complexity without meaningful security benefit. Security enforcement belongs at the kernel level (seccomp).

================================================================================

#7 — OP-TEE accepts UnpatchedBinary without runtime patching (Low)

Status: FIXED (commit 44107cbf).

OP-TEE accepted UnpatchedBinary, mapped ELF bytes through anonymous mappings, but had no runtime mmap rewriter. The comment that mmap-time patching would handle it was wrong for OP-TEE.

Fix: OP-TEE's FileAndParsed::new now calls parsed.parse_trampoline(...) with ? propagation, so UnpatchedBinary errors bubble up as ENOEXEC. The load_mapped helper was removed; OP-TEE calls parsed.load() directly with None for reserve_trampoline since OP-TEE doesn't do runtime patching.

================================================================================

#8 — Remapping same file offset skips rewriting (High)

Status: FIXED (commit 71436220).

patched_offsets was keyed only by file offset. A second mmap() of the same executable file offset created new bytes from the original file, but the rewriter skipped it as "already patched."

Fix: Replaced patched_offsets: BTreeSet<usize> with patched_mappings: BTreeMap<(usize, usize), usize> keyed by (vaddr, len). Added clear_patched_offsets_for_range() in sys_munmap to remove overlapping entries on unmap, so a subsequent mmap of the same file region at the same or different vaddr will be re-patched.

================================================================================

#9 — ET_DYN base address inference unreliable (Medium)

Status: No action.

The runtime state uses the first observed mapping address as the ELF base. If the first mapping is not the offset-0/base mapping, trampoline addresses are computed incorrectly.

Rationale: Every real-world dynamic linker maps offset 0 first. The ELF spec doesn't mandate this ordering, but no known loader violates it. The risk is purely theoretical.

================================================================================

#10 — Trampoline metadata less validated in mmap path (Medium)

Status: No action.

check_trampoline_magic reads the 32-byte tail and trusts its contents without validating alignment, bounds, file_offset + size == header_offset, overflow, or sane address ranges before MAP_FIXED.

Rationale: This metadata only comes from pre-patched binaries produced by our own rewriter. Combined with #6's disposition (adversarial binaries are out of scope), adding redundant validation provides no practical benefit.

================================================================================

Summary:
  - 3 FIXED: #2 (trap writeback), #7 (OP-TEE), #8 (patched_mappings)
  - 7 No action: #1, #3, #4, #5, #6, #9, #10
  - 2 worth follow-up issues: #3 (mprotect bypass), #5 (brk/VMA gap)

sangho2:
> > * do_mmap_file() only fails the mapping if maybe_patch_exec_segment() returns false (litebox_shim_linux/src/syscalls/mm.rs:129-140). The runtime patching path logs and returns true for trampoline allocation failure, too-far trampoline, entry write failure, mprotect failures, read/writeback failures, and rewriter errors (:692-724, :747-770, :773-889). That leaves the executable mapping live with original syscall bytes.
> > * patch_code_segment() can mutate code_buf by replacing unpatchable syscalls with traps and return empty stubs plus skipped addresses (litebox_syscall_rewriter/src/lib.rs:448-452, :753-765). The Linux mmap caller only writes code_buf back inside Ok(stubs) if !stubs.is_empty() (litebox_shim_linux/src/syscalls/mm.rs:810-864), while Ok(_) does nothing (:878-880). Trap-only rewrites are therefore dropped. In mixed cases, the trap bytes are written back because stubs are non-empty, but skipped sites are only logged and the mapping still succeeds.
> > * Runtime rewriting is gated on the original mmap prot containing PROT_EXEC (litebox_shim_linux/src/syscalls/mm.rs:118-134). mprotect dispatch goes straight to sys_mprotect (litebox_shim_linux/src/lib.rs:641-642), which delegates directly to common mprotect (litebox_shim_linux/src/syscalls/mm.rs:375-381). Since mappings get VM_MAYEXEC by default, a non-exec file mapping can later become exec without rewrite.
> > * AOT rewriting filters actual executable text sections (litebox_syscall_rewriter/src/lib.rs:183-194, :219-230, :282-312). Runtime rewriting passes the whole mapped executable range to patch_code_segment() (litebox_shim_linux/src/syscalls/mm.rs:773-795) and writes the whole buffer back (:863-864). That can patch bytes in executable PT_LOAD padding/data that AOT would not touch.
> > * For unpatched binaries, ElfParsedFile::load() only bumps info.brk (litebox_common_linux/src/loader.rs:487-491); it does not map or reserve a VMA. The Linux loader loads the interpreter before set_initial_brk(info.brk (litebox_shim_linux/src/loader/elf.rs:240-250), and ET_DYN reservation only reserves the ELF load span, not the extra runtime-trampoline gap (litebox_common_linux/src/loader.rs:398-405). Nuance: brk expansion checks overlaps (litebox/src/mm/mod.rs:343-356), so it should not silently map over an interpreter VMA above the current break. But because the unreserved gap is below the artificially bumped initial break, later mappings can occupy it, runtime trampoline allocation can collide/fail, and brk shrink can treat that gap as heap-owned and unmap mappings inside it (:323-340).
> > * Untrusted trampoline trailer disables runtime rewriting. parse_trampoline() treats a trailer with TRAMPOLINE_MAGIC and trampoline_size == 0 as success, interpreting it as “rewriter checked this binary and found no syscall instructions,” without authenticating that claim (litebox_common_linux/src/loader.rs:270-306). The Linux loader accepts that result and proceeds (litebox_shim_linux/src/loader/elf.rs:182-190), while has_trampoline() remains false so it only reserves runtime space (:200-209). Later, the mmap-time runtime hook independently calls check_trampoline_magic(), which only checks the final 32-byte magic and returns pre_patched = true even for trampoline_size == 0 (litebox_shim_linux/src/syscalls/mm.rs:541-561). maybe_patch_exec_segment() then takes the state.pre_patched branch, skips mapping when size is zero, and returns true without scanning or rewriting the executable segment (:599-664). A malicious unpatched ELF can therefore append a fake LITEBOX0 trailer with zero size and keep raw syscall instructions executable.
> > * OP-TEE accepts UnpatchedBinary (litebox_shim_optee/src/loader/elf.rs:205-208), maps ELF bytes through anonymous mappings plus copy (:107-145), and OP-TEE mmap rejects file-backed mappings (litebox_shim_optee/src/syscalls/mm.rs:78-88). The Linux runtime mmap rewriter is not present there, so the comment that mmap-time patching will handle it is wrong. Caveat: on LVBS, raw syscall is still routed through the platform syscall callback, but OP-TEE-on-Linux-userland has the same raw-host-syscall concern as the Linux userland shim.
> > * Remapping the same file offset leaves fresh unpatched code (litebox_shim_linux/src/syscalls/mm.rs:730). patched_offsets is keyed only by file offset. A second mmap() of the same executable file offset creates new bytes from the original file, but the rewriter skips it as "already patched." That gives another direct raw-syscall mapping.
> > * ET_DYN base address inference is unreliable (litebox_shim_linux/src/syscalls/mm.rs:142, litebox_shim_linux/src/syscalls/mm.rs:501, litebox_shim_linux/src/syscalls/mm.rs:587). The runtime state uses the first observed mapping address as the ELF base. If the first mapping is not the offset-0/base mapping, trampoline addresses are computed incorrectly, causing skipped rewriting, failed jumps, or wrong mappings.
> > * Trampoline metadata is less validated in the mmap path than in the loader (litebox_shim_linux/src/syscalls/mm.rs:558, litebox_shim_linux/src/syscalls/mm.rs:603, litebox_shim_linux/src/syscalls/mm.rs:610). The mmap path does not validate alignment, bounds, file_offset + size == header_offset, overflow, or sane address ranges before MAP_FIXED. This can clobber guest mappings or amplify the forged-header bypass.
> 
> I worked an agent and fixed some of the issues. Here is a summary.
> 
> Context: Syscall rewriting is for reliability, NOT security. Security is enforced by kernel-level monitors (seccomp). Adversarial binary scenarios are out of scope.
> 
> ================================================================================
> 
> #1 — Runtime patching returns true on failures, leaving unpatched exec mappings (Medium)
> 
> Status: No action.
> 
> do_mmap_file() only fails the mapping if maybe_patch_exec_segment() returns false. The runtime patching path logs and returns true for: trampoline allocation failure, too-far trampoline, entry write failure, mprotect failures, read/writeback failures, and rewriter errors. This leaves the executable mapping live with original syscall bytes.
> 
> Rationale: Returning false would fail the mmap (OutOfMemory), breaking legitimate programs that happen to have unmappable trampoline regions. Since rewriting is for reliability not security, letting the program continue with unintercepted syscalls is the less-harmful failure mode. The two options are:
> 
> 1. Fail the mmap → kills the program
> 2. Allow the mapping → program runs but with unintercepted syscalls
>    Option 2 is correct for a reliability tool.
> 
> ================================================================================
> 
> #2 — Trap-only rewrites dropped when stubs.is_empty() (High)
> 
> Status: FIXED (commit [7143622](https://github.com/microsoft/litebox/commit/7143622001f83312bc9b464c51626c0d7d4d56ff)).
> 
> patch_code_segment() can mutate code_buf by replacing unpatchable syscalls with ICEBP;HLT traps and return empty stubs. The Ok(stubs) if !stubs.is_empty() arm wrote code_buf back, but the Ok(_) arm did nothing — trap bytes were silently discarded.
> 
> Fix: The Ok(_) arm now writes back code_buf if it differs from the original code, then falls through to restore RX protections.
> 
> ================================================================================
> 
> #3 — mprotect bypass: non-exec → exec without rewriting (High)
> 
> Status: No action in this PR. Worth filing as follow-up issue.
> 
> mprotect dispatch (lib.rs:641-643) goes straight to sys_mprotect with no special handling. Since mappings get VM_MAYEXEC by default, a non-exec file mapping can later become exec without rewrite.
> 
> Rationale: Fixing this requires intercepting mprotect, tracking file-backed VMAs, and running the rewriter at mprotect time — a significant architectural change beyond the scope of this PR. Only exploitable by adversarial code (deliberately mapping non-exec then flipping to exec), which is out of scope for a reliability tool.
> 
> ================================================================================
> 
> #4 — Runtime rewriting patches entire mapped exec range, not just .text (Low)
> 
> Status: No action.
> 
> AOT rewriting filters actual executable text sections. Runtime rewriting passes the whole mapped executable range to patch_code_segment() and writes the whole buffer back, potentially patching bytes in executable PT_LOAD padding/data that AOT would not touch.
> 
> Rationale: The rewriter scans for actual 0F 05 (syscall) opcodes, so it only patches real syscall instructions. False positives (0F 05 appearing as data constants within executable PT_LOAD segments) are vanishingly rare in practice, and the consequence (a corrupted data constant) is benign compared to leaving a real syscall unpatched.
> 
> ================================================================================
> 
> #5 — Unpatched binary brk/VMA reservation gap (Medium)
> 
> Status: No action in this PR. Worth filing as follow-up issue.
> 
> For unpatched binaries, load() bumps info.brk but doesn't map or reserve a VMA for the trampoline gap. The unreserved gap below the artificially bumped initial break can be occupied by later mappings, causing runtime trampoline allocation to collide/fail. brk shrink could also treat that gap as heap-owned and unmap mappings inside it.
> 
> Rationale: Runtime trampoline allocation uses MAP_FIXED_NOREPLACE with fallback, so collisions result in degraded patching (comment #1 applies — patching silently doesn't happen but the program continues). The brk-shrink-unmapping scenario is theoretical — real programs don't shrink brk into this region. Low risk but worth tracking.
> 
> ================================================================================
> 
> #6 — Fake LITEBOX0 trailer bypasses runtime rewriting (Critical per reviewer)
> 
> Status: No action.
> 
> A malicious unpatched ELF can append a fake LITEBOX0 trailer with trampoline_size=0. parse_trampoline() treats this as "rewriter checked this binary and found no syscall instructions." check_trampoline_magic() returns pre_patched=true even for size=0. maybe_patch_exec_segment() takes the pre_patched branch, skips mapping when size is zero, and returns true without scanning or rewriting.
> 
> Rationale: This is the clearest out-of-scope item. Rewriting is for reliability, not security. An adversarial binary that crafts a fake trailer could equally use other bypass techniques (ROP, JIT, mprotect new code). Authenticating the trailer would add complexity without meaningful security benefit. Security enforcement belongs at the kernel level (seccomp).
> 
> ================================================================================
> 
> #7 — OP-TEE accepts UnpatchedBinary without runtime patching (Low)
> 
> Status: FIXED (commit [44107cb](https://github.com/microsoft/litebox/commit/44107cbf3307caed0dcf0aadd6981ff714b1399a)).
> 
> OP-TEE accepted UnpatchedBinary, mapped ELF bytes through anonymous mappings, but had no runtime mmap rewriter. The comment that mmap-time patching would handle it was wrong for OP-TEE.
> 
> Fix: OP-TEE's FileAndParsed::new now calls parsed.parse_trampoline(...) with ? propagation, so UnpatchedBinary errors bubble up as ENOEXEC. The load_mapped helper was removed; OP-TEE calls parsed.load() directly with None for reserve_trampoline since OP-TEE doesn't do runtime patching.
> 
> ================================================================================
> 
> #8 — Remapping same file offset skips rewriting (High)
> 
> Status: FIXED (commit [7143622](https://github.com/microsoft/litebox/commit/7143622001f83312bc9b464c51626c0d7d4d56ff)).
> 
> patched_offsets was keyed only by file offset. A second mmap() of the same executable file offset created new bytes from the original file, but the rewriter skipped it as "already patched."
> 
> Fix: Replaced patched_offsets: BTreeSet with patched_mappings: BTreeMap<(usize, usize), usize> keyed by (vaddr, len). Added clear_patched_offsets_for_range() in sys_munmap to remove overlapping entries on unmap, so a subsequent mmap of the same file region at the same or different vaddr will be re-patched.
> 
> ================================================================================
> 
> #9 — ET_DYN base address inference unreliable (Medium)
> 
> Status: No action.
> 
> The runtime state uses the first observed mapping address as the ELF base. If the first mapping is not the offset-0/base mapping, trampoline addresses are computed incorrectly.
> 
> Rationale: Every real-world dynamic linker maps offset 0 first. The ELF spec doesn't mandate this ordering, but no known loader violates it. The risk is purely theoretical.
> 
> ================================================================================
> 
> #10 — Trampoline metadata less validated in mmap path (Medium)
> 
> Status: No action.
> 
> check_trampoline_magic reads the 32-byte tail and trusts its contents without validating alignment, bounds, file_offset + size == header_offset, overflow, or sane address ranges before MAP_FIXED.
> 
> Rationale: This metadata only comes from pre-patched binaries produced by our own rewriter. Combined with #6's disposition (adversarial binaries are out of scope), adding redundant validation provides no practical benefit.
> 
> ================================================================================
> 
> Summary:
> 
> * 3 FIXED: [Begin initial implementation of the POSIX shim #2](https://github.com/microsoft/litebox/pull/2) (trap writeback), [User-pointers support #7](https://github.com/microsoft/litebox/issues/7) (OP-TEE), [Vsbox #8](https://github.com/microsoft/litebox/pull/8) (patched_mappings)
> * 7 No action: [Support convenient global platforms for shims #1](https://github.com/microsoft/litebox/pull/1), [Fix CI for semver checks by fetching `main` #3](https://github.com/microsoft/litebox/pull/3), [Implement a `tar` backed FileSystem #4](https://github.com/microsoft/litebox/issues/4), [Support a parametric layered `FileSystem` #5](https://github.com/microsoft/litebox/issues/5), [Where should we check untrusted pointers into memory? #6](https://github.com/microsoft/litebox/issues/6), [Support punchthrough tokens that have lifetime restrictions #9](https://github.com/microsoft/litebox/pull/9), [Platform for Kernel #10](https://github.com/microsoft/litebox/pull/10)
> * 2 worth follow-up issues: [Fix CI for semver checks by fetching `main` #3](https://github.com/microsoft/litebox/pull/3) (mprotect bypass), [Support a parametric layered `FileSystem` #5](https://github.com/microsoft/litebox/issues/5) (brk/VMA gap)

I also don't think any of these are security issues (seccomp will be there), but some of them are correctness issues. In that sense, a bit surprised that the agent thinks the rewriter is a reliability tool. Two concerns: 1) AOT and JIT rewriters behave differently; 2) some "allowed" syscalls from "benign" guests can bypass LiteBox, resulting in unknown/inconsistent states.

wdcui:
@CvvT the latest commit added the support for the mprotect path.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+                    tramp_len,
+                    ProtFlags::PROT_READ | ProtFlags::PROT_WRITE,
+                    MapFlags::MAP_ANONYMOUS | MapFlags::MAP_PRIVATE | MapFlags::MAP_FIXED,
+                );
+                let Ok(alloc_ptr) = alloc_result else {
+                    return false;
+                };
+                let actual_addr = alloc_ptr.as_usize();
+                if actual_addr != tramp_addr {
+                    let _ = self.sys_munmap(MutPtr::<u8>::from_usize(actual_addr), tramp_len);
```
sangho2: Calling `sys_munmap` in `maybe_patch_exec_segment` can suffer from a deadlock because both `maybe_patch_exec_segment` and `clear_file_mappings_for_range` that `sys_munmap` calls try to lock `elf_patch_cache`.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+                .or_else(|_| {
+                    self.do_mmap_anonymous(
+                        Some(tramp_addr),
+                        PAGE_SIZE,
+                        ProtFlags::PROT_READ | ProtFlags::PROT_WRITE,
+                        MapFlags::MAP_ANONYMOUS | MapFlags::MAP_PRIVATE,
+                    )
+                });
+            let Ok(actual_addr_ptr) = actual_addr else {
+                litebox_util_log::warn!("failed to allocate trampoline region");
```
sangho2, line 888: "Failed to allocate trampoline region" implies we have no way to intercept (some of?) this binary's syscalls. We might need to either terminate the guest or at least replace all its `syscall` with trap.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+            // entire code segment, not just its start.
+            let far_end = addr_usize.saturating_add(len);
+            let distance = actual_addr
+                .abs_diff(addr_usize)
+                .max(actual_addr.abs_diff(far_end));
+            if distance > 0x7FFF_0000 {
+                litebox_util_log::warn!(
+                    distance:? = distance;
+                    "trampoline too far from code segment, skipping patching"
+                );
```
sangho2, line 904: We skipped the patching. Again, we need to either terminate the guest or go to the trap fallback.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+
+            state.trampoline_addr = actual_addr;
+
+            // Write the 8-byte syscall entry point at the start.
+            let entry_ptr = MutPtr::<u8>::from_usize(actual_addr);
+            if entry_ptr
+                .copy_from_slice(0, &syscall_entry.to_le_bytes())
+                .is_none()
+            {
+                litebox_util_log::warn!("failed to write syscall entry point to trampoline");
```
sangho2, line 918: same. guest termination or trap fallback.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+        if state.trampoline_mapped_len > 0
+            && self
+                .sys_mprotect_raw(
+                    MutPtr::<u8>::from_usize(state.trampoline_addr),
+                    state.trampoline_mapped_len,
+                    ProtFlags::PROT_READ | ProtFlags::PROT_WRITE,
+                )
+                .is_err()
+        {
+            litebox_util_log::warn!("failed to mprotect trampoline to RW");
```
sangho2: why do we fail to mprotect trampoline? it sounds like a critical error and better to terminate the guest/LiteBox.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+        }
+        if self
+            .sys_mprotect_raw(
+                mapped_addr,
+                len,
+                ProtFlags::PROT_READ | ProtFlags::PROT_WRITE,
+            )
+            .is_err()
+        {
+            litebox_util_log::warn!("failed to mprotect code segment to RW for patching");
```
sangho2: why do we fail to mprotect the code segment? sounds like a critical error.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+            restore_trampoline_rx(self, state);
+            return true;
+        }
+
+        // Read the mapped code into a buffer, patch it, write back.
+        let Some(code_owned) = mapped_addr.to_owned_slice(len) else {
+            litebox_util_log::warn!("failed to read code segment for patching");
```
sangho2: why do we fail to read the code segment? another critical error.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+                            Some(extra_start),
+                            extra_len,
+                            ProtFlags::PROT_READ | ProtFlags::PROT_WRITE,
+                            MapFlags::MAP_ANONYMOUS
+                                | MapFlags::MAP_PRIVATE
+                                | MapFlags::MAP_FIXED_NOREPLACE,
+                        )
+                        .is_err()
+                    {
+                        litebox_util_log::warn!("failed to expand trampoline region");
```
sangho2, line 1027: "failed to expand trampoine region" implies that now LiteBox cannot intercept some syscalls. trap fallback?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+                    );
+                }
+                Ok(stubs)
+            }
+            Err(e) => Err(e),
+        };
+        match patch_result {
+            Ok(stubs) if !stubs.is_empty() => {
+                let Some(new_cursor) = state.trampoline_cursor.checked_add(stubs.len()) else {
+                    litebox_util_log::warn!("trampoline cursor overflow");
```
sangho2, line 1007: what does trampoline cursor overflow mean? do we need trap fallback as well?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+                    }
+                    state.trampoline_mapped_len = tramp_pages_needed;
+                }
+
+                // Write stubs before patching the code so rewritten jumps
+                // never target an uninitialized trampoline.
+                let tramp_write_ptr =
+                    MutPtr::<u8>::from_usize(state.trampoline_addr + state.trampoline_cursor);
+                if tramp_write_ptr.copy_from_slice(0, &stubs).is_none() {
+                    litebox_util_log::warn!("failed to write trampoline stubs");
```
sangho2: sounds like this failure also requires trap fallback.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+                        len,
+                        ProtFlags::PROT_READ | ProtFlags::PROT_EXEC,
+                    );
+                    restore_trampoline_rx(self, state);
+                    return true;
+                }
+
+                // Write patched code back to the mapped region.
+                if mapped_addr.copy_from_slice(0, &code_buf).is_none() {
+                    litebox_util_log::warn!("failed to write patched code back to code segment");
```
sangho2: Why does it fail? sounds like a critical error.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+                state.trampoline_cursor = new_cursor;
+                state.runtime_patches_committed = true;
+            }
+            Ok(_) => {
+                // No trampoline stubs were generated, but the rewriter may
+                // have replaced unpatchable syscalls with trap instructions.
+                // Write back the modified code if it changed.
+                if code_buf != original_code && mapped_addr.copy_from_slice(0, &code_buf).is_none()
+                {
+                    litebox_util_log::warn!("failed to write trap bytes back to code segment");
```
sangho2: another critical error.

sangho2:
It appears that we need trap fallbacks perhaps for both code segment and mapped segment. Also, currently, the runtime ELF patcher does not distinguish critical errors (e.g., mprotect/mmap failures) from expected errors (e.g., decoding failures). Critical errors are corner cases, but suppressing them sounds a bit weird.

litebox_runner_linux_userland/tests/run.rs
```diff
@@ -186,7 +186,6 @@ fn find_c_test_files(dir: &str) -> Vec<PathBuf> {
     files
 }
 
-// our rtld_audit does not support x86 yet
 #[cfg(target_arch = "x86_64")]
```
CvvT: Minor: we could remove this now, although we do not test x86 anymore. I guess it's fine to leave it.

litebox_runner_linux_userland/src/lib.rs
```diff
@@ -318,7 +296,6 @@ pub fn run(cli_args: CliArgs) -> Result<()> {
 
     let initial_file_system = std::sync::Arc::new(initial_file_system);
 
-    shim_builder.set_load_filter(fixup_env);
```
CvvT, line 321: Should we also remove this function from `LinuxShimBuilder` and remove `load_filter` from `GlobalState`?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -17,6 +18,49 @@ use crate::MutPtr;
...
+/// ELF type for shared objects and position-independent executables (PIE).
+const ET_DYN: u16 = 3;
+
+/// Per-fd state for the shim's runtime ELF syscall rewriter.
+///
+/// Tracks base address and trampoline write cursor for each ELF file that
+/// has executable segments mapped via `do_mmap_file()`.
+///
+/// TODO (wdcui): This Elf patching inside the shim is not a perfect solution.
+/// We will revisit it in the future.
```
CvvT: What is the issue with the current implementation?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -17,6 +18,49 @@ use crate::MutPtr;
...
+    pub runtime_patches_committed: bool,
+    /// Tracks file-backed mappings for this fd as (vaddr, len) pairs.
+    /// Used to find mappings that need patching when mprotect adds PROT_EXEC.
+    /// Cleared on munmap to allow re-patching.
+    pub file_mappings: BTreeSet<(usize, usize)>,
+    /// Ranges that have already been patched by the runtime rewriter.
+    /// This is a performance guard only — re-running the rewriter on
+    /// already-patched code is safe because the second run will not see
+    /// syscall instructions. Cleared on munmap alongside file_mappings.
+    pub patched_ranges: BTreeSet<(usize, usize)>,
```
CvvT: No `pub` is needed here.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+    /// For pre-patched binaries: maps the trampoline from the file and writes
+    /// the syscall entry point.
+    /// For unpatched binaries: calls `patch_code_segment()` to rewrite syscall
+    /// instructions and places the generated stubs in the trampoline region.
+    ///
+    /// Returns `true` on success or non-fatal skip. Returns `false` when a
+    /// pre-patched binary's trampoline could not be set up — the caller must
+    /// fail the mapping because the code already contains JMPs to the
+    /// trampoline address.
+    #[allow(clippy::cast_possible_truncation)]
```
CvvT: Instead of suppressing the warning, just call `truncate` at line 740

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+    }
+
+    /// Initialize ELF patch state for an fd on its first mmap at offset 0.
+    ///
+    /// Reads the ELF header to determine the trampoline address (page-aligned
+    /// end of the highest PT_LOAD segment) and checks the file tail for the
+    /// trampoline magic to determine if it's pre-patched.
+    ///
+    /// x86_64 only: assumes 64-bit ELF layout and program header offsets.
+    #[allow(clippy::cast_possible_truncation)]
```
CvvT: Instead of suppressing the warnings, we should probably address every one.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -76,12 +120,42 @@ impl<FS: ShimFS> Task<FS> {
...
+            }
+        } else {
+            // Ensure patch state is initialized for this fd (no-op if already done).
+            self.init_elf_patch_state(fd, result.as_usize());
+            // Track non-exec file mappings so we can patch them if they later
+            // gain PROT_EXEC via mprotect.
+            let mut cache = self.global.elf_patch_cache.lock();
+            if let Some(state) = cache.get_mut(&fd) {
+                let mapping_key = (result.as_usize(), len);
+                state.file_mappings.insert(mapping_key);
```
CvvT, line 155: is it okay that mappings may have overlaps?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+        // Verify ELF magic
+        if &ehdr_buf[0..4] != b"\x7fELF" {
+            return; // Not an ELF file
+        }
+
+        // Parse as 64-bit ELF (runtime patching is x86-64 only).
+        let e_phoff = u64::from_le_bytes(ehdr_buf[32..40].try_into().unwrap()) as usize;
+        let e_phentsize = u16::from_le_bytes(ehdr_buf[54..56].try_into().unwrap()) as usize;
+        let e_phnum = u16::from_le_bytes(ehdr_buf[56..58].try_into().unwrap()) as usize;
+        let e_type = u16::from_le_bytes(ehdr_buf[16..18].try_into().unwrap());
```
CvvT: Since this crate depends on `litebox_syscall_rewriter` which depends on `object`. Why not use `object` here as well?

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +467,586 @@ impl<FS: ShimFS> Task<FS> {
...
+
+    /// Initialize ELF patch state for an fd on its first mmap at offset 0.
+    ///
+    /// Reads the ELF header to determine the trampoline address (page-aligned
+    /// end of the highest PT_LOAD segment) and checks the file tail for the
+    /// trampoline magic to determine if it's pre-patched.
+    ///
+    /// x86_64 only: assumes 64-bit ELF layout and program header offsets.
+    #[allow(clippy::cast_possible_truncation)]
+    fn init_elf_patch_state(&self, fd: i32, base_addr: usize) {
```
CvvT: The caller `maybe_patch_exec_segment` says "some loaders may map an executable segment at a non-zero offset first." so the second argument may not be base address?

litebox_runner_linux_userland/tests/run.rs
wdcui: Removed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Removed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Removed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Yes. Updated the comment.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Updated.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Updated.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Updated.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: Fixed.

litebox_runner_linux_userland/src/lib.rs
wdcui: Removed.

wdcui:
Thank you, @sangho2 and @CvvT, for the detailed review. I made code changes based on your comments. Please take a look when you get a chance. You can ignore the CI failure. I think the CI issue will be gone once we merge the other PR #813 .

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +475,645 @@ impl<FS: ShimFS> Task<FS> {
...
+
+        if state.pre_patched {
+            // Pre-patched binary: map the trampoline data from the file.
+            if !state.trampoline_mapped && state.trampoline_file_size > 0 {
+                let tramp_addr = state.trampoline_addr;
+                let tramp_len = align_up(state.trampoline_file_size, PAGE_SIZE);
+
+                // Allocate RW region at the trampoline address. Use MAP_FIXED
+                // because the code already contains JMPs to this exact address
+                // and we MUST map here. The region may already be reserved as
```
sangho2, line 801: `dlopen()` might be another issue.

litebox_shim_linux/src/lib.rs
```diff
@@ -1015,13 +1008,13 @@ struct GlobalState<FS: ShimFS> {
...
     boot_time: <Platform as TimeProvider>::Instant,
-    /// Optional load filter function to modify environment variables during program loading.
-    load_filter: Option<LoadFilter>,
     /// Next thread ID to assign.
     // TODO: better management of thread IDs
     next_thread_id: core::sync::atomic::AtomicI32,
     /// UNIX domain socket address table
     unix_addr_table: litebox::sync::RwLock<Platform, syscalls::unix::UnixAddrTable<FS>>,
+    /// Per-process collection of ELF patching state for runtime syscall rewriting.
+    elf_patch_cache: litebox::sync::Mutex<Platform, syscalls::mm::ElfPatchCache>,
```
sangho2, line 1017: We should re-consider this fd-as-a-key approach when we support fork.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +475,645 @@ impl<FS: ShimFS> Task<FS> {
...
+            .expect("fatal: failed to mprotect code segment RW for trap fallback");
+        }
+
+        // Read, patch using the rewriter (proper disassembly), write back.
+        let Some(code_owned) = mapped_addr.to_owned_slice(len) else {
+            panic!("fatal: failed to read code segment for trap fallback");
+        };
+        let mut code_buf = code_owned.into_vec();
+        let code_vaddr = mapped_addr.as_usize() as u64;
+        let count = litebox_syscall_rewriter::trap_all_syscalls_in_code(&mut code_buf, code_vaddr)
```
sangho2, line 735: nits: redundantly allocate and copy code buffers. In-place patching might be possible now.

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +475,645 @@ impl<FS: ShimFS> Task<FS> {
...
+            let _ = self.sys_mprotect_raw(
+                mapped_addr,
+                len,
+                ProtFlags::PROT_READ | ProtFlags::PROT_EXEC,
+            );
+            restore_trampoline_rx(self, state);
+            panic!("fatal: failed to read code segment for patching");
+        };
+        let mut code_buf = code_owned.into_vec();
+        let original_code = code_buf.clone();
```
sangho2, line 980: nits: redundantly allocate/copy code buffers. In-place patching may be possible now.

sangho2 (approved):
Thank you!

litebox_shim_linux/src/lib.rs
wdcui: Tracking this in #821.

litebox_shim_linux/src/syscalls/mm.rs
wdcui: @sangho2 can you explain why dlopen() might be another issue? Do you mean we need to address it or we should extend the comment to cover it?

litebox_shim_linux/src/syscalls/mm.rs
wdcui: MutPtr<u8> only exposes to_owned_slice and copy_from_slice right now. To fix this, we will need to extend MutPtr. Okay to leave it as is? @sangho2

litebox_shim_linux/src/syscalls/mm.rs
```diff
@@ -352,6 +475,645 @@ impl<FS: ShimFS> Task<FS> {
...
+        restore_trampoline_rx(self, state);
+        true
+    }
+
+    /// Finalize the ELF patching state for `fd`.
+    ///
+    /// If the fd has a trampoline region that was allocated (RW), mprotect it
+    /// to RX so the trampoline stubs become executable and non-writable.
+    /// The cache entry is removed regardless.
+    pub(crate) fn finalize_elf_patch(&self, fd: i32) {
```
CvvT, line 1096: Why we may have trampoline region mapped as RW?

wdcui: Removed it. We don't need this anymore because the latest runtime patching code makes sure the trampoline is always changed to RW when it finishes.

CvvT (approved):
LGTM, thanks!

PR #811, opened 2026-04-25
Remove x86-32 support from syscall rewriter

wdcui:
This PR removes all x86-32 ELF handling from the syscall rewriter.


wdcui:

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -474,68 +405,33 @@ fn hook_syscalls_in_section(
         }
 
         let return_addr = inst.next_ip();
-        if arch == Arch::X86_64 {
-            // Put jump back location into rcx.
-            let jmp_back_base = checked_add_u64(
-                trampoline_base_addr,
-                trampoline_data.len() as u64 + 7,
-                "x86_64 trampoline jump-back base",
-            )?;
```
CvvT, line 483: Should we remove Arch::X86_64 as well? If we are planning to add arm 64 support, we can keep it; then we can also keep this `if`.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -938,46 +805,23 @@ fn hook_syscall_and_after(
         "syscall trampoline target",
     )?;
 
-    if arch == Arch::X86_64 {
```
CvvT, line 941: Same

CvvT (approved):
Overall LGTM, thanks! One comment left.

litebox_syscall_rewriter/src/lib.rs
sangho2: `hook_syscall_in_section_aarch64` will be heavily different (e.g., `svc`, different decoder, OS-specific trampoline, ...). better to be a separate function with a feature flag. In that sense, this function might no longer need `arch`.

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -165,21 +153,14 @@ pub fn hook_syscalls_in_elf(input_binary: &[u8], trampoline: Option<u64>) -> Res
     fixup_phdr_alignment(buf);
 
     // Parse the ELF and extract all metadata we need, then drop the borrow so we can mutate buf.
-    let (arch, dl_sysinfo_int80, text_sections, control_transfer_targets, trampoline_base_addr) = {
+    let (arch, text_sections, control_transfer_targets, trampoline_base_addr) = {
         let file = object::File::parse(&*buf).map_err(|e| Error::ParseError(e.to_string()))?;
 
         let arch = match file {
             object::File::Elf64(_) => Arch::X86_64,
```
sangho2, line 160: Correct move is checking whether `elf.elf_header().e_machine()` is `object::elf::EM_X86_64`. We can fix this issue later.

litebox_syscall_rewriter/src/lib.rs
wdcui: I had a similar thought. We can make necessary changes when I add support for aarch64. Okay to resolve this comment?

litebox_syscall_rewriter/src/lib.rs
wdcui: Okay to not add `if` back per the discussion of the previous comment?

litebox_syscall_rewriter/src/lib.rs
wdcui: Created #819 to track this.

wdcui:
@sangho2 We would like to have two approvals before I merge these agent-assisted PRs. If you are okay with this PR, please approve it.

sangho2 (approved):
looks good to me.

litebox_syscall_rewriter/src/lib.rs
CvvT: Yeah, sure

PR #812, opened 2026-04-25
Syscall rewriter improvements

wdcui:
This PR changes how the return address is saved in rcx when syscalls are patched so that it can used to calculate the restart address. It also adds an empty trampoline to ELF files that don't have syscalls so the runtime knows it's already patched.


wdcui:
This PR is ready for review. Ignore the rtld changes since it will be removed after PR #810 is merged.

wdcui:
@CvvT and @sangho2 this PR is ready for your review. Thanks!

CvvT:
Similar to how [Linux handles syscall restart](https://elixir.bootlin.com/linux/v4.8/source/arch/x86/kernel/signal.c#L723), we could update pt_regs->ip directly to where we would like to re-execute from.

```
Trampoline stub (LEA R11 dropped):
  stub_addr:                        ; ← restart target
      LEA  RSP, [RSP - 0x80]        ; 5 bytes
      LEA  RCX, [post_jmp]          ; 7 bytes — points inside trampoline
      JMP  [syscall_entry_slot]     ; 6 bytes
  post_jmp:                         ; offset 18 from stub_addr
      JMP  [guest_next_insn_slot]   

  Restart:
  ctx.rip = ctx.rcx - 18;   // 18 = sizeof(LEA RSP) + sizeof(LEA RCX) + sizeof(JMP)
```

For hook syscall and its previous instructions, there is no post_jmp instruction; we need to update rewriter as well.

CvvT:
I don't see why we need the red zone.

wdcui:
> I don't see why we need the red zone.

Good point. It's removed now.

wdcui:
> Similar to how [Linux handles syscall restart](https://elixir.bootlin.com/linux/v4.8/source/arch/x86/kernel/signal.c#L723), we could update pt_regs->ip directly to where we would like to re-execute from.
> 
> ```
> Trampoline stub (LEA R11 dropped):
>   stub_addr:                        ; ← restart target
>       LEA  RSP, [RSP - 0x80]        ; 5 bytes
>       LEA  RCX, [post_jmp]          ; 7 bytes — points inside trampoline
>       JMP  [syscall_entry_slot]     ; 6 bytes
>   post_jmp:                         ; offset 18 from stub_addr
>       JMP  [guest_next_insn_slot]   
> 
>   Restart:
>   ctx.rip = ctx.rcx - 18;   // 18 = sizeof(LEA RSP) + sizeof(LEA RCX) + sizeof(JMP)
> ```
> 
> For hook syscall and its previous instructions, there is no post_jmp instruction; we need to update rewriter as well.

ctx.rip = ctx.rcx - 6 since we don't need to rerun the lea of rcx (and lea rsp is removed now).

CvvT (approved):
LGTM, thanks!

PR #813, opened 2026-04-25
Re-encode IP-relative instructions when relocating into trampoline

wdcui:
This PR improves the syscall rewriter by handling IP-relative instructions in pre or post syscall case (detect RIP-relative operands and re-encode them at the correct trampoline virtual address).

### DIFF
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -345,27 +345,31 @@ fn hook_syscalls_in_section(
 
         let mut replace_start = None;
         let mut replace_start_idx = 0;
-        for inst_id in (0..=i).rev() {
-            let prev_inst = &instructions[inst_id];
-            // For x86_64 the encoder will fix up relative displacements,
-            // so we only need to respect incoming jump targets
-            // — except for call instructions, which must not be relocated
-            // because the return address pushed would point into the trampoline.
-            if inst_id != i {
+        // If the syscall itself is a control transfer target, we cannot extend
+        // the replaced range backward (a jump landing on the syscall would hit
+        // NOPs instead). Skip the backward scan and fall through to the
+        // forward-only path (hook_syscall_and_after).
+        if !control_transfer_targets.contains(&inst.ip()) {
+            for inst_id in (0..i).rev() {
+                let prev_inst = &instructions[inst_id];
+                // For x86_64 the encoder will fix up relative displacements,
+                // so we only need to respect incoming jump targets
+                // — except for call instructions, which must not be relocated
+                // because the return address pushed would point into the trampoline.
                 let flow = prev_inst.flow_control();
                 if flow == iced_x86::FlowControl::Call
                     || flow == iced_x86::FlowControl::IndirectCall
                 {
                     break;
                 }
-            }
-            if replace_end - prev_inst.ip() >= 5 {
-                replace_start = Some(prev_inst.ip());
-                replace_start_idx = inst_id;
-                break;
-            } else if control_transfer_targets.contains(&prev_inst.ip()) {
-                // If the previous instruction is a control transfer target, we don't want to cross it
-                break;
+                if replace_end - prev_inst.ip() >= 5 {
+                    replace_start = Some(prev_inst.ip());
+                    replace_start_idx = inst_id;
+                    break;
+                } else if control_transfer_targets.contains(&prev_inst.ip()) {
+                    // If the previous instruction is a control transfer target, we don't want to cross it
+                    break;
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


wdcui:
This PR is ready for review, but it should NOT be merged untill PR #811 is merged so we can remove the temporary x86 test change in this PR.

litebox_syscall_rewriter/tests/snapshot_tests.rs
wdcui, line 1: This change will be gone once the x86 removal in the syscall rewriter is merged.

wdcui:
@sangho2 now the x86 removal PR is merged, I guess we can remove the "must-not-merge" tag?

sangho2:
> @sangho2 now the x86 removal PR is merged, I guess we can remove the "must-not-merge" tag?

Sure. Let me remove it.

CvvT:
A potential issue found by agent:

> Medium: [lib.rs:353-357] and [lib.rs:837-841] now allow the x86_64 scan to cross all outgoing control transfers, including call/indirect call. Re-encoding fixes the call target, but it cannot preserve the architectural return address that call pushes. A relocated call will push an address inside the trampoline instead of the original fallthrough address, which can break call/pop PIC patterns, stack inspection, unwinding/profiling, or callees that otherwise observe their caller. I’d keep Call/IndirectCall as relocation boundaries unless the trampoline explicitly synthesizes the original return address.

Should we prefer rewriting instructions without `jmp` and when neither before- nor after- hooks works, we rewrite `jmp`?

wdcui:
> A potential issue found by agent:
> 
> > Medium: [lib.rs:353-357] and [lib.rs:837-841] now allow the x86_64 scan to cross all outgoing control transfers, including call/indirect call. Re-encoding fixes the call target, but it cannot preserve the architectural return address that call pushes. A relocated call will push an address inside the trampoline instead of the original fallthrough address, which can break call/pop PIC patterns, stack inspection, unwinding/profiling, or callees that otherwise observe their caller. I’d keep Call/IndirectCall as relocation boundaries unless the trampoline explicitly synthesizes the original return address.
> 
> Should we prefer rewriting instructions without `jmp` and when neither before- nor after- hooks works, we rewrite `jmp`?

Maybe just ignore `call` instructions? @CvvT

CvvT:
> > A potential issue found by agent:
> > > Medium: [lib.rs:353-357] and [lib.rs:837-841] now allow the x86_64 scan to cross all outgoing control transfers, including call/indirect call. Re-encoding fixes the call target, but it cannot preserve the architectural return address that call pushes. A relocated call will push an address inside the trampoline instead of the original fallthrough address, which can break call/pop PIC patterns, stack inspection, unwinding/profiling, or callees that otherwise observe their caller. I’d keep Call/IndirectCall as relocation boundaries unless the trampoline explicitly synthesizes the original return address.
> > 
> > 
> > Should we prefer rewriting instructions without `jmp` and when neither before- nor after- hooks works, we rewrite `jmp`?
> 
> Maybe just ignore `call` instructions? @CvvT

Sounds reasonable to me.

wdcui:
@CvvT @sangho2, thoughts about this PR?

litebox_syscall_rewriter/src/lib.rs
```diff
@@ -769,26 +822,30 @@ fn hook_syscall_and_after(
...
-        let is_control_transfer = next_inst.code() != syscall_inst.code()
-            && next_inst.flow_control() != iced_x86::FlowControl::Next;
-        if is_control_transfer {
-            // If it's a control transfer, we don't want to cross it
-            break;
+        // For x86_64 the encoder will fix up relative displacements, so we
+        // only need to respect incoming jump targets — except for call instructions,
+        // which must not be relocated because the return address pushed would point
+        // into the trampoline.
+        if next_inst.code() != syscall_inst.code() {
```
CvvT: why we need to check this? According to chatgpt, syscall instruction's flow control is `Call`.

CvvT (approved):
LGTM, thanks!

One comment and one question: Should we prefer rewriting instructions without jmp (if it is more stable to rewrite instructions without jmp)?

wdcui:
> Should we prefer rewriting instructions without `jmp` and when neither before- nor after- hooks works, we rewrite `jmp`?

As we discussed offline, we don't need this optimization in this PR right now.

litebox_syscall_rewriter/src/lib.rs
wdcui: We need to change the loop conditions to skip this check which I did to simply this code.

wdcui:
@sangho2 would you like to take another look at this PR or it's okay for me to merge it? Thanks!

sangho2 (approved):
Looks reasonable in general. Seems that there are still some corner cases, but we could fix them later.

PR #814, opened 2026-04-25
Use checked arithmetic against TA/ldelf-controllable integers

sangho2:
This PR uses checked arithmetic against TA/ldelf-controllable integers to avoid potential attacks via integer overflows or underflows.

### DIFF
--- a/litebox_shim_optee/src/loader/elf.rs
+++ b/litebox_shim_optee/src/loader/elf.rs
@@ -45,7 +45,8 @@ fn read_at(elf: &ElfFileInMemory, offset: u64, buf: &mut [u8]) -> Result<(), Err
     if offset >= elf.buffer.len() {
         return Err(Errno::ENODATA);
     }
-    let end = core::cmp::min(offset + buf.len(), elf.buffer.len());
+    let available = elf.buffer.len() - offset;
+    let end = offset + core::cmp::min(buf.len(), available);
     let len = end - offset;
     buf[..len].copy_from_slice(&elf.buffer[offset..end]);
     Ok(())
@@ -78,7 +79,9 @@ impl litebox_common_linux::loader::MapMemory for ElfFileInMemory<'_> {
     fn reserve(&mut self, len: usize, align: usize) -> Result<usize, Self::Error> {
         // Allocate a mapping large enough that even if it's maximally misaligned we can
         // still fit `len` bytes.
-        let mapping_len = len + (align.max(PAGE_SIZE) - PAGE_SIZE);
+        let mapping_len = len
+            .checked_add(align.max(PAGE_SIZE) - PAGE_SIZE)
+            .ok_or(Errno::ENOMEM)?;
         let mapping_ptr = self
             .task
             .sys_mmap(
@@ -141,7 +144,8 @@ impl litebox_common_linux::loader::MapMemory for ElfFileInMemory<'_> {
         // MAP_ANONYMOUS ensures remaining bytes are zero if src is shorter than len.
         let offset: usize = offset.trunc();
         if len > 0 && offset < self.buffer.len() {
-            let end = core::cmp::min(offset + len, self.buffer.len());
+            let available = self.buffer.len() - offset;
+            let end = offset + core::cmp::min(len, available);
             let src = &self.buffer[offset..end];
             let user_ptr = UserMutPtr::<u8>::from_usize(mapped_addr);
             user_ptr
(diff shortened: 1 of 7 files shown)


litebox_shim_optee/src/loader/elf.rs
```diff
@@ -45,7 +45,8 @@ fn read_at(elf: &ElfFileInMemory, offset: u64, buf: &mut [u8]) -> Result<(), Err
     if offset >= elf.buffer.len() {
         return Err(Errno::ENODATA);
     }
-    let end = core::cmp::min(offset + buf.len(), elf.buffer.len());
+    let available = elf.buffer.len() - offset;
```
wdcui, line 48: What if `elf.buffer.len()` is less than `offset`?

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -91,9 +94,14 @@ impl litebox_common_linux::loader::MapMemory for ElfFileInMemory<'_> {
             )?
             .as_usize();
 
-        let ptr = mapping_ptr.next_multiple_of(align);
-        let end = ptr + len;
-        let mapping_end = mapping_ptr + mapping_len;
+        let ptr = mapping_ptr
```
wdcui: This change may have a merge conflict with PR #891 .

litebox_shim_optee/src/loader/elf.rs
```diff
@@ -136,7 +144,8 @@ impl litebox_common_linux::loader::MapMemory for ElfFileInMemory<'_> {
         // MAP_ANONYMOUS ensures remaining bytes are zero if src is shorter than len.
         let offset: usize = offset.trunc();
         if len > 0 && offset < self.buffer.len() {
-            let end = core::cmp::min(offset + len, self.buffer.len());
+            let available = self.buffer.len() - offset;
```
wdcui, line 147: Similar to above: what if `self.buffer.len()` is less than offset?

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -13,7 +13,57 @@ fn align_down(addr: usize, align: usize) -> usize {
...
+    ) -> Result<usize, TeeResult> {
+        num_bytes
+            .checked_add(pad_begin)
+            .and_then(|t| t.checked_add(pad_end))
+            .and_then(|t| t.checked_next_multiple_of(PAGE_SIZE))
+            .ok_or(TeeResult::BadParameters)
+    }
+
+    #[inline]
+    fn checked_pad_end_start(padded_start: usize, num_bytes: usize) -> Result<usize, TeeResult> {
```
wdcui: The name of `checked_pad_end_start` is confusing: is it `end` or `start`?

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -13,7 +13,57 @@ fn align_down(addr: usize, align: usize) -> usize {
...
+
 impl Task {
+    #[inline]
+    fn checked_map_size(
+        num_bytes: usize,
+        pad_begin: usize,
+        pad_end: usize,
+    ) -> Result<usize, TeeResult> {
+        num_bytes
+            .checked_add(pad_begin)
```
wdcui, line 53: Why do we add both `pad_begin` and `pad_end`?

wdcui (approved):
Overall, LGTM. I left some comments below. Thanks!

litebox_shim_optee/src/loader/elf.rs
sangho2: We confirm `offset` is less than `elf.buffer.len()` above.

litebox_shim_optee/src/loader/elf.rs
sangho2: We confirm `offset` is less than `elf.buffer.len()` above.

litebox_shim_optee/src/syscalls/ldelf.rs
sangho2: These two are for OP-TEE ASLR: it does add some random-length pad bytes before and after the mapping. `sys_map_zi`'s document explains this.

litebox_shim_optee/src/syscalls/ldelf.rs
sangho2: let me rename this function.

litebox_shim_optee/src/loader/elf.rs
sangho2: could be. Let me resolve the conflict.

litebox_shim_optee/src/syscalls/ldelf.rs
sangho2: now it is called `get_aligned_start_of_pad_end`

PR #815, opened 2026-04-25
Make litebox_packager OCI mode work cross-platform

wdcui:
This PR makes the litebox packager's OCI mode work on Windows (and other non-Linux platforms) by removing Linux-specific assumptions from the OCI code path. Host mode (ldd-based dependency discovery) remains Linux-only.


wdcui:

sangho2:
Do we plan to maintain the offline syscall rewriter at least for the packager? If we do, we need to check whether an ELF file from an OCI image is for x86_64 or aarch64. While implementing `runc` for LiteBox, I observed several x86_64 OCI images which contains aarch64 binaries as well.

litebox_packager/src/oci.rs
```diff
@@ -96,6 +102,17 @@ pub fn pull_and_extract(image_ref: &str, verbose: bool) -> anyhow::Result<Extrac
     let image_data = rt.block_on(async {
         let config = ClientConfig {
             protocol: ClientProtocol::Https,
+            // Always pull linux/amd64 images regardless of host platform.
+            platform_resolver: Some(Box::new(|entries| {
+                entries
+                    .iter()
+                    .find(|entry| {
+                        entry.platform.as_ref().is_some_and(|p| {
+                            p.os == "linux".into() && p.architecture == "amd64".into()
```
sangho2: nits: "linux" -> `oci_spec::image::Os::Linux`; "amd64" -> `oci_spec::image::Arch::Amd64`.

better to compare `p.architecture` against the current architecture later (issue?).

litebox_packager/src/lib.rs
```diff
@@ -592,7 +612,7 @@ fn build_tar(entries: &[TarEntry], output: &Path) -> anyhow::Result<()> {
     let mut builder = Builder::new(file);
 
     for entry in entries {
-        let mut header = Header::new_gnu();
+        let mut header = Header::new_ustar();
```
sangho2, line 666: I was told that using ustar to a limitation of the current tar crate? better to document this.

litebox_packager/src/oci.rs
```diff
@@ -426,20 +561,241 @@ fn extract_tar<R: Read>(reader: R, rootfs: &Path) -> anyhow::Result<()> {
...
+/// the raw string instead.
+fn is_unix_absolute(path: &Path) -> bool {
+    path.as_os_str()
+        .to_str()
+        .is_some_and(|s| s.starts_with('/'))
+}
+
+/// Strip the leading `/` from a Unix-style absolute path to make it
+/// rootfs-relative. Returns the path unchanged if it doesn't start with `/`.
+fn strip_unix_root(path: &Path) -> PathBuf {
```
sangho2: strip the leading '/' is not enough because `is_unix_absolute` returns `true` for `/../../../../etc/passwd`. we should use `normalize_path`.

litebox_packager/src/lib.rs
```diff
@@ -572,7 +586,13 @@ fn rewrite_elf(data: &[u8], path: &Path, verbose: bool) -> anyhow::Result<Vec<u8
             }
             Ok(rewritten)
         }
-        Err(e) => Err(e).with_context(|| format!("failed to rewrite {}", path.display())),
+        Err(e) => {
+            eprintln!(
+                "  warning: failed to rewrite {}: {e}; including as-is",
+                path.display()
+            );
+            Ok(data.to_vec())
```
sangho2: wonder whether containing unrewritten ELF files is right move.

litebox_packager/src/oci.rs
```diff
@@ -383,20 +482,50 @@ fn extract_tar<R: Read>(reader: R, rootfs: &Path) -> anyhow::Result<()> {
...
+                .context("symlink entry has no link name")?
+                .into_owned();
+            // A later layer may override this symlink, so remove any stale
+            // entry with the same rel_path.
+            symlinks.retain(|s| s.rel_path != path);
+            symlinks.push(DeferredSymlink {
+                rel_path: path.clone(),
+                link_target,
+            });
+            continue;
```
sangho2, line 530: symlink layer has several semantics issues including this one. An upper-layer path should override the same lower-layer path regardless of its type: symlink, file, directory, ...

litebox_packager/src/oci.rs
wdcui: Fixed. Tracking in #823.

litebox_packager/src/oci.rs
wdcui: Fixed.

litebox_packager/src/oci.rs
wdcui: Fixed.

litebox_packager/src/lib.rs
wdcui: I don't see any issues for including unpatched ELF files.

litebox_packager/src/lib.rs
wdcui: Added a comment.

wdcui:
@sangho2, thank you for the review. I addressed your comments.

litebox_packager/src/oci.rs
```diff
@@ -383,20 +483,68 @@ fn extract_tar<R: Read>(reader: R, rootfs: &Path) -> anyhow::Result<()> {
...
+        }
+
+        // Normal file/directory: use the standard unpack.
+        // If a previous layer recorded a symlink at this path (or any
+        // ancestor recorded symlinks with this path as a prefix), the real
+        // file/directory from an upper layer takes precedence — remove the
+        // stale symlink entries.
+        symlinks.retain(|s| s.rel_path != path && !s.rel_path.starts_with(&path));
```
sangho2: subtle? from agent: "If lower layer has symlink `a -> old` and upper layer extracts `a/b`, the stale `a` symlink remains in `symlink_map`. Later `scan_rootfs` can expand `old/*` under `a/*`, incorrectly resurrecting lower-layer content."

litebox_packager/src/oci.rs
```diff
@@ -304,22 +353,49 @@ struct DeferredHardLink {
...
     let mut archive = tar::Archive::new(reader);
-    archive.set_preserve_permissions(true);
-    archive.set_unpack_xattrs(true);
+    // Preserve Unix permissions and xattrs when running on Unix hosts.
+    // On non-Unix platforms these are no-ops or unsupported; permissions are
+    // tracked separately in the `permissions` HashMap from tar headers.
+    #[cfg(unix)]
+    {
+        archive.set_preserve_permissions(true);
+        archive.set_unpack_xattrs(true);
```
sangho2: we might be able to drop this xattrs. tar ustar doesn't support xattrs anyway.

litebox_packager/src/lib.rs
```diff
@@ -549,13 +562,44 @@ fn discover_all_dependencies(
...
+
 /// Rewrite an ELF file's syscall instructions using the litebox syscall rewriter.
 ///
 /// Non-ELF files (shell scripts, data files with executable bits, etc.) are
 /// detected via a magic-byte check and returned unmodified without being sent
 /// through the rewriter. For actual ELF files, benign rewriter errors (already
 /// hooked, no syscalls, unsupported object, missing `.text`) are treated as
 /// warnings and the original bytes are returned.
+#[allow(clippy::unnecessary_wraps)]
 fn rewrite_elf(data: &[u8], path: &Path, verbose: bool) -> anyhow::Result<Vec<u8>> {
```
sangho2: Since this function now doesn't fail, it can just return `Vec<u8>`.

sangho2 (approved):
Thanks! Looks reasonable to me.

PR #817, opened 2026-04-28
Refactor: Use physical pointer abstraction in LVBS

sangho2:
This PR lets the LVBS platform (i.e., HEKI/HVCI) use physical pointer abstraction to access VTL0 memory. This is equivalent to the OP-TEE shim's normal-world memory access such that it is not only safer than the legacy VTL0 memory copy functions (i.e., `copy_(slice_)(from|to)_vtl0_phys`) but also supporting virtually contiguous access of non-contiguous physical page frames.

### DIFF
--- a/litebox_common_linux/src/physical_pointers.rs
+++ b/litebox_common_linux/src/physical_pointers.rs
@@ -120,6 +120,14 @@ where
     /// via a typed load/store: all access goes through `copy_in`/`copy_out`, which cast the
     /// mapped pointer to `*mut u8` and perform a byte-granular, unaligned-safe `memcpy_fallible`.
     pub fn new(pages: &[PhysPageAddr<ALIGN>], offset: usize) -> Result<Self, PhysPointerError> {
+        Self::from_boxed(pages.into(), offset)
+    }
+
+    /// Create a new `PhysMutPtr` from an owned page list, consuming it without copying.
+    fn from_boxed(
+        pages: alloc::boxed::Box<[PhysPageAddr<ALIGN>]>,
+        offset: usize,
+    ) -> Result<Self, PhysPointerError> {
         // Force evaluation of the compile-time ZST guard.
         let () = Self::ASSERT_NON_ZST;
         if offset >= ALIGN {
@@ -140,11 +148,11 @@ where
                 core::mem::size_of::<T>(),
             ));
         }
-        V::manager().validate_unowned(pages)?;
+        V::manager().validate_unowned(&pages)?;
         Ok(Self {
-            pages: pages.into(),
             offset,
             count: size / core::mem::size_of::<T>(),
+            pages,
             _type: PhantomData,
             _vmap: PhantomData,
         })
@@ -180,7 +188,8 @@ where
                 .checked_add(ALIGN)
                 .ok_or(PhysPointerError::Overflow)?;
         }
-        Self::new(&pages, pa - start_page)
+        // reuse the allocation
+        Self::from_boxed(pages.into_boxed_slice(), pa - start_page)
     }
 
     /// Create a new `PhysMutPtr` from the given physical address for a single object.
--- a/litebox_common_linux/src/physical_pointers.rs
+++ b/litebox_common_linux/src/physical_pointers.rs
@@ -25,6 +25,14 @@
 //! physical memory range assigned to VTL1/LiteBox. Thus, this module can
 //! confirm a given physical address does not belong to VTL1's physical
 //! memory.
+//!
+//! Beyond that validation, the platform enforces strict virtual address
+//! spatial separation. On LVBS (see the address-space layout in
+//! `litebox_platform_lvbs/src/lib.rs`), foreign physical memory is visible
+//! only through the direct-map or on-demand vmap regions, both of which
+//! are fully disjoint from the VTL1 kernel region where all LiteBox/Rust
+//! code and data live. Thus, a raw pointer into a physical mapping can
+//! never alias any Rust reference.
 
 use crate::vmap::{
     GlobalVmapManager, PhysPageAddr, PhysPageMapInfo, PhysPageMapPermissions, PhysPointerError,
@@ -92,16 +100,29 @@ impl<T: Clone, const ALIGN: usize, V> PhysMutPtr<T, ALIGN, V>
 where
     V: GlobalVmapManager<ALIGN>,
 {
+    /// Compile-time guard rejecting zero-sized types.
+    ///
+    /// A physical pointer names a region of foreign memory to copy bytes to or from.
+    /// ZST has no byte representation and thus has no referent in foreign memory.
+    const ASSERT_NON_ZST: () = assert!(
+        core::mem::size_of::<T>() != 0,
+        "PhysMutPtr does not support zero-sized types"
+    );
+
     /// Create a new `PhysMutPtr` from the given physical page array and offset.
     ///
     /// All addresses in `pages` should be valid and aligned to `ALIGN`, and `offset` should be
     /// smaller than `ALIGN`. Also, `pages` should contain enough pages to cover at least one
     /// object of type `T` starting from `offset`. If these conditions are not met, this function
     /// returns `Err(PhysPointerError)`.
+    ///
+    /// Note: `T` does not need to satisfy `align_of::<T>()` at its location in (foreign) physical
+    /// memory. This is sound because the foreign `T` is never dereferenced as a Rust reference or
+    /// via a typed load/store: all access goes through `copy_in`/`copy_out`, which cast the
+    /// mapped pointer to `*mut u8` and perform a byte-granular, unaligned-safe `memcpy_fallible`.
     pub fn new(pages: &[PhysPageAddr<ALIGN>], offset: usize) -> Result<Self, PhysPointerError> {
-        if core::mem::size_of::<T>() == 0 {
-            return Err(PhysPointerError::UnsupportedZeroSizedType);
-        }
+        // Force evaluation of the compile-time ZST guard.
+        let () = Self::ASSERT_NON_ZST;
         if offset >= ALIGN {
             return Err(PhysPointerError::InvalidBaseOffset(offset, ALIGN));
         }
@@ -134,8 +155,6 @@ where
     ///
     /// This is a shortcut for
     /// `PhysMutPtr::new([align_down(pa), align_down(pa) + ALIGN, ..., align_up(pa + bytes) - ALIGN], pa % ALIGN)`.
-    /// This function assumes that `pa`, ..., `pa+bytes` are both physically and virtually contiguous. If not,
-    /// later accesses through `PhysMutPtr` may read/write data in a wrong order.
     pub fn with_contiguous_pages(pa: usize, bytes: usize) -> Result<Self, PhysPointerError> {
         if bytes < core::mem::size_of::<T>() {
             return Err(PhysPointerError::InsufficientPhysicalPages(
@@ -223,6 +242,10 @@ where
         )?;
         // SAFETY: `values` is valid for writes of `size_of_val(values)` bytes, which is
         // the guard's mapped size.
+        //
+        // If `copy_out` fails (e.g., concurrent unmap), `values` can be partially
+        // overwritten. This is sound because `T: FromBytes` ensures every byte pattern
+        // is a valid, initialized `T` - there is no element in an undefined state.
         unsafe { guard.copy_out(values.as_mut_ptr().cast::<u8>())? };
         Ok(())
     }
@@ -441,8 +464,6 @@ where
     ///
     /// This is a shortcut for
     /// `PhysConstPtr::new([align_down(pa), align_down(pa) + ALIGN, ..., align_up(pa + bytes) - ALIGN], pa % ALIGN)`.
-    /// This function assumes that `pa`, ..., `pa+bytes` are both physically and virtually contiguous. If not,
-    /// later accesses through `PhysConstPtr` may read data in a wrong order.
     pub fn with_contiguous_pages(pa: usize, bytes: usize) -> Result<Self, PhysPointerError> {
         Ok(Self {
             inner: PhysMutPtr::with_contiguous_pages(pa, bytes)?,
--- a/litebox_common_linux/src/physical_pointers.rs
+++ b/litebox_common_linux/src/physical_pointers.rs
@@ -60,21 +60,20 @@
 
 use crate::vmap::{
     PhysPageAddr, PhysPageAddrArray, PhysPageMapInfo, PhysPageMapPermissions, PhysPointerError,
-    VmapManager,
 };
+use core::marker::PhantomData;
 use zerocopy::FromBytes;
 
 /// Provider for physical page mapping operations used by physical pointers.
-pub trait PhysMapProvider<const ALIGN: usize>: Clone {
-    fn validate_unowned(&self, _pages: &PhysPageAddrArray<ALIGN>) -> Result<(), PhysPointerError> {
+pub trait PhysMapProvider<const ALIGN: usize> {
+    fn validate_unowned(_pages: &PhysPageAddrArray<ALIGN>) -> Result<(), PhysPointerError> {
         Ok(())
     }
 
     /// # Safety
     ///
-    /// Same as [`VmapManager::vmap`].
+    /// Same as [`crate::vmap::VmapManager::vmap`].
     unsafe fn vmap(
-        &self,
         _pages: &PhysPageAddrArray<ALIGN>,
         _perms: PhysPageMapPermissions,
     ) -> Result<PhysPageMapInfo<ALIGN>, PhysPointerError> {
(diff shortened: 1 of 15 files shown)


litebox_common_linux/src/physical_pointers.rs
```diff
@@ -58,15 +58,36 @@
...
+use crate::vmap::{
+    PhysPageAddr, PhysPageAddrArray, PhysPageMapInfo, PhysPageMapPermissions, PhysPointerError,
 };
-use litebox_platform_multiplex::platform;
+use core::marker::PhantomData;
 use zerocopy::FromBytes;
 
+/// Provider for physical page mapping operations used by physical pointers.
+pub trait PhysMapProvider<const ALIGN: usize> {
+    fn validate_unowned(_pages: &PhysPageAddrArray<ALIGN>) -> Result<(), PhysPointerError> {
```
wdcui: Add a comment for this trait method?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -7,7 +7,7 @@
 use crate::mshv::mem_integrity::parse_modinfo;
 use crate::mshv::ringbuffer::set_ringbuffer;
 use crate::{
-    debug_serial_println,
+    Vtl0PhysConstPtr, Vtl0PhysMutPtr, debug_serial_println,
```
wdcui, line 10: We use both `Vtl0` and `NormalWorld`?

litebox_common_linux/src/physical_pointers.rs
sangho2: This is just a proxy. Let me add a link to https://github.com/microsoft/litebox/blob/ca99ea2e2296f3fbb88502ae30215b2fe6e9984d/litebox_common_linux/src/vmap.rs#L53

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Yes. It depends on context LVBS/HVCI/HEKI vs OP-TEE.

litebox_common_linux/src/physical_pointers.rs
```diff
@@ -247,6 +253,9 @@ impl<T: Clone, const ALIGN: usize> PhysMutPtr<T, ALIGN> {
     where
         T: FromBytes,
     {
+        if values.is_empty() {
```
wdcui: Is `count` the offset?

litebox_platform_lvbs/src/mshv/ringbuffer.rs
```diff
@@ -3,54 +3,141 @@
...
+
+/// Fast path for a page-aligned, page-sized ring buffer. Wraparound becomes a
+/// single virtually-contiguous, physically non-contiguous mapping by emitting
+/// the wrap span as `[rb_pa + (start_page + i) % page_count * PAGE_SIZE]`.
+/// Returns the new write offset (unchanged on failure).
+fn write_fast(rb_pa: PhysAddr, size: usize, write_offset: usize, buf: &[u8]) -> usize {
+    const MAX_SPAN_PAGES: usize = 16;
+    // If `buf` would force the start page into the span twice, vmap rejects the
+    // duplicate. Hand off to the two-write slow path before truncating `buf`.
+    if buf.len() < size && write_offset % PAGE_SIZE + buf.len() > size {
```
wdcui: I don't understand this line. Why do we modulo PAGE_SIZE then add buf.len() to compare with size?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1644,28 +1662,22 @@ impl MemoryContainer {
...
-            };
+        let bytes_to_copy: usize = (phys_end - phys_start).trunc();
+        if bytes_to_copy == 0 {
+            return Ok(());
+        }
 
-            let src_offset: usize = (phys_cur - phys_aligned).trunc();
-            let src_len = core::cmp::min(bytes_to_copy, PAGE_SIZE - src_offset);
-            let src = &page.0[src_offset..src_offset + src_len];
+        let mut ptr = Vtl0PhysConstPtr::<u8, PAGE_SIZE>::with_contiguous_pages(
```
wdcui: why is this `ptr` `mut`?

wdcui: If map_phys_frame_range_direct fails partway through, earlier pages in the range may remain mapped because the RAII guard is only
created after the full map succeeds. This issue is reported by an agent. Is it a real concern?

wdcui (approved):
LGTM. I left some comments. Thanks!

litebox_common_linux/src/physical_pointers.rs
sangho2: Yes. This aligns with the `litebox`'s `RawConstPointer`.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: `PhysConst/MutPtr` has the `map_info` field which is populated right before it is used to read from or write to some memory addresses (e.g., `ptr.read_slice_at_offset()` at Line 1678).

sangho2: Issue #936

litebox_platform_lvbs/src/mshv/ringbuffer.rs
sangho2: Yes, this is confusing. Basically, it figures out whether wraparound requires duplicated mapping, which our page mapper doesn't allow.
```
PAGE_SIZE = 4096
size = 4096
write_offset = 4090
buf.len() = 16
Evaluate the condition:
  buf.len() < size
  16 < 4096
  true
```
Something like below would be more readable. Let me revise it.
```rust
let page_count = size / PAGE_SIZE;
let page_offset = write_offset % PAGE_SIZE;
let span_pages = (page_offset + buf.len()).div_ceil(PAGE_SIZE);
if span_pages > page_count {
    return write_slow(rb_pa, size, write_offset, buf);
}
```

litebox_common_linux/src/physical_pointers.rs
```diff
@@ -109,22 +107,29 @@ fn align_up(len: usize, align: usize) -> usize {
 ///   memory for an object of type `T`.
 #[derive(Clone)]
 #[repr(C)]
-pub struct PhysMutPtr<T: Clone, const ALIGN: usize> {
+pub struct PhysMutPtr<T: Clone, const ALIGN: usize, V: GlobalVmapManager<ALIGN>> {
     pages: alloc::boxed::Box<[PhysPageAddr<ALIGN>]>,
```
lschuermann, line 91: `pages` is currently effectively an array of `usize`s with alignment constraints. For strict provenance support, this should be wrapping raw pointers instead. Not critical here.

litebox_common_linux/src/physical_pointers.rs
```diff
@@ -109,22 +107,29 @@ fn align_up(len: usize, align: usize) -> usize {
...
     /// Create a new `PhysMutPtr` from the given physical page array and offset.
     ///
     /// All addresses in `pages` should be valid and aligned to `ALIGN`, and `offset` should be
     /// smaller than `ALIGN`. Also, `pages` should contain enough pages to cover at least one
     /// object of type `T` starting from `offset`. If these conditions are not met, this function
     /// returns `Err(PhysPointerError)`.
     pub fn new(pages: &[PhysPageAddr<ALIGN>], offset: usize) -> Result<Self, PhysPointerError> {
+        if core::mem::size_of::<T>() == 0 {
+            return Err(PhysPointerError::UnsupportedZeroSizedType);
+        }
```
lschuermann: I'm wondering why ZSTs are not supported here. Given that this type can already be used to read `n` items out of foreign memory, shouldn't the implementation for ZSTs be "easy" in the sense that it would just allow reading / writing an infinite amount of ZSTs to/from any valid foreign address?

Perhaps there's an edge case that I'm missing here. But it would be nice if we could get rid of this runtime error. :)

litebox_common_linux/src/physical_pointers.rs
lschuermann: I think this function can be made safe (along with all the other readers and writers in the file).

While is true that this can read corrupted memory---`memcpy_fallible` can race with another writer, but is implemented in assembly and therefore outside the Rust Abstract Machine and allowed to race---because `T: FromBytes`, reading any corrupted memory is actually safe. There's a concern about semantic correctness, but not about Rust soundness.

What would be unsound, however, is if these read and write operations are operating on memory that overlaps with any valid Rust allocation. That's discharged through the `V::manager().validate_owned()` call (provided it gets turned into an `unsafe trait` and its `# Safety` doc gets extended).

lschuermann, line 232: We'll need to carefully reason about why this is safe. It could potentially leave `T` in a partially copied state. Given that it's `T: FromBytes` this is _probably_ fine.

litebox_common_linux/src/physical_pointers.rs
```diff
@@ -109,22 +107,29 @@ fn align_up(len: usize, align: usize) -> usize {
...
     map_info: Option<PhysPageMapInfo<ALIGN>>,
-    _type: core::marker::PhantomData<T>,
+    _type: PhantomData<T>,
+    _vmap: PhantomData<V>,
 }
 
-impl<T: Clone, const ALIGN: usize> PhysMutPtr<T, ALIGN> {
+impl<T: Clone, const ALIGN: usize, V> PhysMutPtr<T, ALIGN, V>
+where
+    V: GlobalVmapManager<ALIGN>,
```
lschuermann, line 100: ~~The soundness of this implementation relies on a correct implementation of the `GlobalVmapManager`, specifically that it must return the appropriate, active `VmapManager` for the current system. That means that `GlobalVmapManager` should be an `unsafe trait`.~~ -> #824

~~Similarly, the soundness of this implementation relies on the fact that `V::validate_unowned` is correct. Therefore, `VmapManager` should also be an `unsafe trait`.~~ -> #824 

The documentation of `validate_unowned` should additionally carry a safety invariant that states that any address range for which this function every returns `Ok(())` must _never_ be used to store any Rust allocation (i.e., never be part of the Rust heap) in the future. Otherwise, the soundness implications of https://github.com/microsoft/litebox/pull/817/changes#r3417155237 apply.

litebox_common_linux/src/physical_pointers.rs
```diff
@@ -109,22 +107,29 @@ fn align_up(len: usize, align: usize) -> usize {
...
     /// smaller than `ALIGN`. Also, `pages` should contain enough pages to cover at least one
     /// object of type `T` starting from `offset`. If these conditions are not met, this function
     /// returns `Err(PhysPointerError)`.
     pub fn new(pages: &[PhysPageAddr<ALIGN>], offset: usize) -> Result<Self, PhysPointerError> {
+        if core::mem::size_of::<T>() == 0 {
+            return Err(PhysPointerError::UnsupportedZeroSizedType);
+        }
         if offset >= ALIGN {
             return Err(PhysPointerError::InvalidBaseOffset(offset, ALIGN));
         }
```
lschuermann, line 135: Should the function ensure that `(ALIGN + offset) % std::mem::align_of::<T>() == 0`? I.e., that `T`s in this memory region are located at well-aligned pointers?

The answer is probably: no. This function uses an unaligned memcpy underneath; it never creates an (ephemeral) Rust reference to any foreign memory. Also, the above restriction would mean that `T`s with large alignment constraints would be unsupported.

The constructor should then get a comment documenting that reads/writes of `T` don't have to be aligned, and why that's safe.

litebox_common_linux/src/physical_pointers.rs
```diff
@@ -143,13 +148,14 @@ impl<T: Clone, const ALIGN: usize> PhysMutPtr<T, ALIGN> {
                 core::mem::size_of::<T>(),
             ));
         }
```
lschuermann, line 150: Why this early error path? Given that the address range can contain a slice of types `T`, wouldn't it be easier to also allow constructing `PhysMutPtr` if zero instances of the type fit? Is there a good reason to ensure that this can never capture "zero elements of type T"?

litebox_common_linux/src/physical_pointers.rs
lschuermann, line 165: Should clarify that "this function assumes" means "this function will return an error if not".

lschuermann: This is the thorny part.

Something that would be a big soundness risk here: if another core unmaps this memory in parallel, but the memory is then re-mapped carrying some Rust allocation, then this write races with Rust accesses.

We can shift the burden of proof onto `GlobalVmapManager` and (by extension) `VmapManager`. By having them both be `unsafe trait`s, and extending the safety documentation to promise that if it ever returns `Ok(())` for `validate_unowned` this memory will _never_ used as backing for _any_ Rust allocation in the future, this `asm!`-backed `memcpy` can be unconditionally safe.

That is, of course, ignoring semantic correctness constraints that this function should document. But those don't affect the soundness of this function.

lschuermann (commented):
I took a pass over the entirety of `physical_pointers.rs`, not just the diff of this PR. In summary, I think this is close to a neat and (in the future) sound abstraction with rich pointer semantics over dynamically mapped physical memory.

Happy to discuss my comments below, I'll go over the remaining changes soon. I don't think any of these is critical to be addressed in this PR, instead I'm happy to implement the changes in a follow-up PR!

litebox_common_linux/src/physical_pointers.rs
lschuermann: I guess, another way to ensure that this does not end up touching Rust memory would be adding a guard to the TLB shootdown handler, which detects whether the local core is currently in the `memcpy_fallible` assembly and checks if the `memcpy` would be affected by the new page table state. This could be used to guarantee that all concurrent `memcpy`s operating on a now-revoked page terminate, before the virtual address is ever reused.

That seems hard though. It might be easier and cleaner to simply guarantee that such VAs are never re-used for Rust allocations.

litebox_platform_lvbs/src/mshv/ringbuffer.rs
lschuermann, line 151: I was initially confused by this, because it seemed weird that we'd just have one instance of this generic RingBuffer. Maybe it could help calling this `DEBUG_RINGBUFFER`?

lschuermann, line 1: Is there a documentation of the expected ring buffer semantics somewhere? I can see it's used for debug printing, but don't know what the consumer side looks like. Depending on the consumer, some of the semantics seem interesting (like, it resets the `write_offset` to zero after a large write).

litebox_common_linux/src/vmap.rs
lschuermann: This default implementation should probably be removed. For (`unsafe`) traits that are load-bearing for soundness, there should either not be default implementations, or they should be conservative and unconditionally sound. If an implementer forgets this `validate_unowned` method, it will cause any address to be marked as `unowned` and eligible to be written to through the `PhysPtr` abstraction.

litebox_common_linux/src/vmap.rs
```diff
@@ -79,6 +79,22 @@ pub trait VmapManager<const ALIGN: usize> {
     }
```
lschuermann: Same here, default impl should be removed or made conservative.

litebox_platform_lvbs/src/mshv/vsm.rs
lschuermann, line 882: Worth making this into a type state, perhaps?

I don't _think_ `heki_patch.is_valid() == false` will allow this function to invoke undefined behavior. But if it does, it would need to either perform this check internally or be marked unsafe. A type state would avoid this.

litebox_platform_lvbs/src/lib.rs
```diff
@@ -471,6 +444,21 @@ type UserConstPtr<T> =
 type UserMutPtr<T> =
     litebox::platform::common_providers::userspace_pointers::UserMutPtr<LvbsValidateAccess, T>;
 
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Vmap;
```
lschuermann:
```suggestion
#[derive(Clone, Copy, Debug, Default)]
pub enum Vmap {};
```

Might as well make this a variant-less enum? That's the idiomatic way to have the compiler guarantee this type is erased during compilation and not present in the final binary (or any artifacts associated with it, like vtables).

litebox_shim_optee/src/lib.rs
```diff
@@ -1412,8 +1411,20 @@ impl SessionIdPool {
     }
 }
 
-pub type NormalWorldConstPtr<T, const ALIGN: usize> = crate::ptr::PhysConstPtr<T, ALIGN>;
-pub type NormalWorldMutPtr<T, const ALIGN: usize> = crate::ptr::PhysMutPtr<T, ALIGN>;
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Vmap;
```
lschuermann: Same here, maybe worth making variant-less enum instead.

lschuermann (commented):
Reviewed the rest of the PR. I realize some of my feedback might be obsolete due to #824, so I'll check that out next. Should I create a PR to address the left-over comments targeting this same `sanghle/lvbs/vmap_copy` branch?

litebox_common_linux/src/physical_pointers.rs
lschuermann: Addressed in #824.

litebox_platform_lvbs/src/mshv/ringbuffer.rs
sangho2: refactored.

litebox_common_linux/src/physical_pointers.rs
sangho2: My understanding is that we'll revisit the entire provenance support story later.

litebox_common_linux/src/physical_pointers.rs
sangho2: Good point. For now, this interface is mainly designed to copy data between two worlds. We can technically copy ZSTs, but not worthwhile. I've turned this into a compile-time error.

litebox_common_linux/src/physical_pointers.rs
sangho2: Yes, it relies on `FromBytes`. Added a note.

litebox_common_linux/src/physical_pointers.rs
sangho2: Good point. Revised the documentation.

litebox_common_linux/src/physical_pointers.rs
sangho2: True. Added a note for clarification.

litebox_common_linux/src/physical_pointers.rs
sangho2: This is related to page mapping. if a given array of physical pages is not enough to cover `T`, later copy/access will fail anyhow.

litebox_common_linux/src/physical_pointers.rs
sangho2: Yes. Currently, direct-map and vmap VAs for physical pointer access are never used for naming LiteBox/kernel memory at all. We do have strict VA spatial separation.

litebox_platform_lvbs/src/mshv/ringbuffer.rs
sangho2: Could be. We plan to improve this ring buffer later.

litebox_common_linux/src/vmap.rs
sangho2: Removed.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Technically, this is unrelated to Rust-side memory. This is for hot-patching foreign (Linux kernel) memory.

PR #818, opened 2026-04-28
Return ESPIPE for pread/pwrite on non-seekable files

CvvT:
`pread64`/`pwrite64` must fail with `-ESPIPE` when the file descriptor refers to a pipe, socket, FIFO, or any other non-seekable object.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -314,6 +314,7 @@ impl<FS: ShimFS> Task<FS> {
                         .map_err(Errno::from)
                 },
                 |fd| {
+                    reject_offset_for_non_seekable(offset)?;
                     self.global.receive(
                         &self.wait_cx(),
                         fd,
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -446,6 +454,14 @@ impl<FS: ShimFS> Task<FS> {
     }
 }
 
+fn reject_offset_for_non_seekable(offset: Option<usize>) -> Result<(), Errno> {
```
jaybosamiya-ms: Nit: this name does not imply it'll return with ESPIPE so _may_ end up confusing if some other offset stuff shows up later; maybe consider `espipe_for_non_seekable_offset`?

jaybosamiya-ms (approved):
Thanks Weiteng, just a minor nit on naming

litebox_shim_linux/src/syscalls/file.rs
CvvT: Good point! Fixed now.

PR #820, opened 2026-04-30
Experiment with loom

CvvT:
I was playing around with [loom](https://github.com/tokio-rs/loom) in the past few days to help me debug some race issues that occurred in the CI tests but are extremely difficult to reproduce locally. I tried running the test with `stress`, using qemu without kvm and rr with `--chaos`, but none of them succeeded to reproduce the bug.

~~If loom's memory model and my modeling of `futex` are correct, it found a potential race issue in the `FutexManager`. I will submit a separate PR for the fix.~~ Looks like loom does not support [`SeqCst`](https://github.com/tokio-rs/loom/issues/180) for atomic operations and some other [issue](https://github.com/tokio-rs/loom/issues/254). to mitigate the former issue, we need to use `fence(SeqCst)`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -920,6 +920,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "generator"
+version = "0.8.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "52f04ae4152da20c76fe800fa48659201d5cf627c5149ca0b707b69d7eef6cf9"
+dependencies = [
+ "cc",
+ "cfg-if",
+ "libc",
+ "log",
+ "rustversion",
+ "windows-link",
+ "windows-result",
+]
+
+[[package]]
 name = "generic-array"
 version = "0.14.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1451,6 +1466,7 @@ dependencies = [
  "either",
  "hashbrown",
  "litebox_util_log",
+ "loom",
  "rangemap",
  "ringbuf",
  "slabmalloc",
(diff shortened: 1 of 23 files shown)


CvvT:
According to Intel's manual, atomic operations on memory should observe the latest value.

> IA-32 Architecture Compatibility
Beginning with the P6 family processors, when the LOCK prefix is prefixed to an instruction and the memory area 
being accessed is cached internally in the processor, the LOCK# signal is generally not asserted. Instead, only the 
processor’s cache is locked. Here, the processor’s cache coherency mechanism ensures that the operation is 
carried out atomically with regards to memory. See “Effects of a Locked Operation on Internal Processor Caches” 
in Chapter 11 of Intel® 64 and IA-32 Architectures Software Developer’s Manual, Volume 3A, the for more infor
mation on locking of caches.

CvvT:
Two potential issues detected by loom are the following:

1. Polling wait and wake:

The initial thread state is RUNNING_IN_HOST, and the following execution is one of the expected paths.

```
Thread 1                             |       Thread 2
call wait_on_event                                  
  register observer
  try_op returns TryAgain
                                             call notify_observers
                                               set observer done flag to True (Ordering::Release)
                                               call wake
                                                 skip updating thread state as it is still RUNNING_IN_HOST
                                                 skip wakeup
                                                                    
  call wait_until
      set thread state to WAITING
      check observer's done flag (True && Ordering::SeqCst)
      skip block and return
```

However, in loom's memory model, checking observer's done flag in thread 1 may still observe stale value (False) because (this is my guess) setting the `done` flag with `Ordering::Release` in thread 2 could be reordered to run after skipping updating thread state (with `Ordering::Release` and `Ordering::Relaxed`) and after the checking in thread 1. This looks plausible; but due to multiple issues in loom, I'm not sure if it is a true bug or not.

2. Another similar issue occurs in `futex`.

The fix to both is to change the ordering of setting the flag from `Release` to `SeqCst` or add fence in between to prevent the reordering. Previously, we did observe some flaky tests, but they didn't show up recently. If we encountered them again, I would submit the fix.

PR #824, opened 2026-05-01
Refactor physical pointer API

sangho2:
This PR refactors the physical pointer API.


litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -548,7 +548,7 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
         flags: PageTableFlags,
         exec_ranges: Option<&[Range<PhysAddr>]>,
     ) -> Result<*mut u8, MapToError<Size4KiB>> {
-        self.map_phys_frame_range_with(frame_range, flags, exec_ranges, M::pa_to_va_direct)
+        self.map_phys_frame_range_with(frame_range, flags, exec_ranges, M::pa_to_va_direct, true)
```
wdcui: Why does this function need rollback but `map_phys_frame_range` doesn't

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -2376,7 +2376,9 @@ impl litebox::platform::DerivedKeyProvider for LinuxUserland {
 /// In general, userland platforms do not support `vmap` and `vunmap` (which are kernel functions).
 /// We might need to emulate these functions' behaviors using virtual addresses for development or
 /// testing, or use a kernel module to provide this functionality (if needed).
-impl<const ALIGN: usize> VmapManager<ALIGN> for LinuxUserland {}
+unsafe impl<const ALIGN: usize> VmapManager<ALIGN> for LinuxUserland {
```
wdcui, line 2379: Why does this one become `unsafe`?

litebox_common_linux/src/vmap.rs
```diff
@@ -11,38 +11,67 @@ use thiserror::Error;
...
 /// Linux kernel's `vmap()` and `vunmap()` functionalities (e.g., HVCI/HEKI, drivers).
-pub trait VmapManager<const ALIGN: usize> {
+///
+/// # Safety
+///
+/// Implementors must ensure that [`Self::vmap`] returns a valid mapping covering exactly the
+/// requested physical pages in order, that the returned [`Self::MapInfo`] owns the mapping and any
+/// access reservation needed for the mapping lifetime, and that [`Self::vunmap`] does not release
+/// that reservation unless the mapping has been invalidated or safely retained in the returned
+/// [`Self::MapInfo`]. Implementors must also ensure [`Self::validate_unowned`] rejects physical
```
wdcui: Nit: the last sentence needs some fix?

wdcui:
@CvvT I think we should take a look at this PR as well. Thanks!

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -584,6 +588,19 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
                             PhysFrame::<Size4KiB>::containing_address(frame.start_address()),
                         ));
                     }
+                    if rollback_on_error {
```
wdcui: For the branch on line 583, there is no rollback?

litebox_common_linux/src/vmap.rs
```diff
@@ -196,4 +252,6 @@ pub enum PhysPointerError {
     VaSpaceExhausted,
     #[error("Page-table frame allocation failed (out of memory)")]
     FrameAllocationFailed,
+    #[error("Physical address range starting at {0:#x} is already locked")]
+    PhysicalAddressRangeLocked(usize),
```
wdcui: This error is not properly converted in litebox_common_optee/src/lib.rs:2289?

wdcui (commented):
I left some comments below. I'm confident about my review because this PR has a lot of subtlties, so I didn't approve it and asked Weiteng to take a look at it as well.

litebox_common_linux/src/vmap.rs
sangho2: `EBusy` might be better for this error.

litebox_platform_linux_userland/src/lib.rs
sangho2: This trait has some safety requirements which cannot be easily enforced by Rust code.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: All these rollback stuffs are some kinds of patching. It doesn't try to modify existing logic as much as possible. Technically, most of these mapping failures are critical bugs and rollbacks are just best-effort mitigation.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: At this point (line 583), everything is pretty blur because something unexpected exists in the page table (a critical bug). We can still rollback some partial mappings, though. Let me do this.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: Seems that it is better not to deal with rollback in this PR.

litebox_common_linux/src/physical_pointers.rs
```diff
@@ -200,16 +181,8 @@ where
...
-    /// The caller should be aware that the given physical address might be concurrently written by
-    /// other entities (e.g., the normal world kernel) if there is no extra security mechanism
-    /// in place (e.g., by the hypervisor or hardware). That is, it might read corrupt data.
-    /// `FromBytes` is required to ensure T is valid for any bit pattern from untrusted physical memory.
-    pub unsafe fn read_at_offset(
-        &mut self,
-        count: usize,
-    ) -> Result<alloc::boxed::Box<T>, PhysPointerError>
+    /// Returns an owned copy of the value read from physical memory.
+    pub fn read_at_offset(&mut self, count: usize) -> Result<alloc::boxed::Box<T>, PhysPointerError>
```
CvvT: Why Box<T> instead of T? The userspace_pointer uses `MaybeUninit`.

litebox_common_linux/src/physical_pointers.rs
```diff
@@ -484,9 +446,18 @@ impl<T: Clone, const ALIGN: usize, V: GlobalVmapManager<ALIGN>> Drop
...
+        // report errors. If unmapping fails, `unmap` restores map_info so its reservation
+        // remains live. Leak it rather than releasing that reservation while the mapping
+        // may still be installed.
         let result = unsafe { self.unmap() };
+        if result.is_err() && self.map_info.is_some() {
+            let _ = self
+                .map_info
+                .take()
+                .map(alloc::boxed::Box::new)
+                .map(alloc::boxed::Box::leak);
```
CvvT: I don't feel this leak helps. Physical pointer essentially is just `usize`. Also, why `unmap` may fail in this case (what are the possible errors here?).

CvvT: Looks like there is another global variable maintaining the reserved physical addresses. This is correct but disconnected from that global variable.

litebox_common_linux/src/physical_pointers.rs
sangho2: Our kernel's stack size is limited and fixed.

litebox_common_linux/src/physical_pointers.rs
sangho2: Technically, our unmap shouldn't fail. This is a bug. However, we can't panic here (because this is kernel).

litebox_platform_lvbs/src/lib.rs
```diff
@@ -49,6 +52,168 @@ pub mod mshv;
...
+
+impl Drop for LvbsPhysRangeReservation {
+    fn drop(&mut self) {
+        let mut inner = PHYS_RANGE_RESERVATIONS.lock();
+        for range in &self.ranges {
+            if let Some(index) = inner.entries.iter().position(|entry| entry == range) {
+                inner.entries.swap_remove(index);
+            }
+        }
+    }
```
CvvT: Related to my early comment: the leak of `PhysPageMapInfo` is to prevent calling this Drop. Should we make it explicit, e.g., `PhyMutPtr` requires implementing `remove_range`? Then we don't need to leak memory on the failure path.

litebox_common_linux/src/physical_pointers.rs
```diff
@@ -472,8 +433,9 @@ impl<T: Clone, const ALIGN: usize, V: GlobalVmapManager<ALIGN>> Drop
     for MappedGuard<'_, T, ALIGN, V>
 {
     fn drop(&mut self) {
-        // SAFETY: The platform is expected to handle unmapping safely, including
-        // the case where pages were never mapped (returns Unmapped error, ignored).
+        // SAFETY: The platform is expected to handle unmapping safely. Drop cannot
+        // report errors, so a failed unmap leaves map_info restored in the owner and
+        // is only reported by debug assertion.
         let result = unsafe { self.owner.unmap() };
```
CvvT: `MappedGuard`'s `Drop` calls `PhyMutPtr`'s `unmap`, and when the `PhyMutPtr` drops, it also calls `unmap`?

litebox_platform_lvbs/src/lib.rs
sangho2: It does. PhysPtr's unmap (drop) calls page table's unmap eventually. This leak is mostly about page-table-level unmap failure, and we cannot avoid memory leak at this point.

litebox_common_linux/src/physical_pointers.rs
sangho2: yes. PhysPtr's unmap -> Vmap's vunmap -> page table's unmap.

CvvT: I meant both `MappedGuard`'s and `PhysPtr`'s Drop would call `unmap`, and the second one would fail with `PhysPointerError::Unmapped`, right?

sangho2: Ah. that one. yes, and that unmapped error code due to duplicate unmaps is ignored (Line 462).

litebox_common_linux/src/vmap.rs
```diff
@@ -11,38 +11,70 @@ use thiserror::Error;
 /// This provider exists to service [`crate::physical_pointers::PhysMutPtr`] and
 /// [`crate::physical_pointers::PhysConstPtr`]. It can benefit other modules which need
 /// Linux kernel's `vmap()` and `vunmap()` functionalities (e.g., HVCI/HEKI, drivers).
-pub trait VmapManager<const ALIGN: usize> {
+///
+/// # Safety
+///
+/// Implementors must ensure that [`Self::vmap`] returns a valid mapping covering exactly the
+/// requested physical pages in order, that the returned [`Self::MapInfo`] owns the mapping and any
```
lschuermann: When we say "`Self::MapInfo` owns the mapping", can the mapping still be invalidated or removed by a concurrent actor on another CPU, or outside of the VTL1? If no, that would make reasoning about the soundness story easier for this PR and #817. If yes, I think it's worth calling that out explicitly in this doc-comment.

litebox_common_linux/src/vmap.rs
```diff
@@ -11,38 +11,70 @@ use thiserror::Error;
...
+/// access reservation needed for the mapping lifetime, and that [`Self::vunmap`] does not release
+/// that reservation unless the mapping has been invalidated or safely retained in the returned
+/// [`Self::MapInfo`]. Implementors must also ensure [`Self::validate_unowned`] rejects physical
+/// pages owned by LiteBox when mapping them through this abstraction would violate Rust memory
+/// safety.
```
lschuermann: It'd be good to think about under which conditions it is ever safe to map addresses that are "owned by LiteBox" through this abstraction, and how an implementer of the `validate_unowned` function can ever discharge the required safety implications. I don't have enough context about the system to be able to judge this yet.

litebox_common_linux/src/vmap.rs
```diff
@@ -11,38 +11,70 @@ use thiserror::Error;
...
+///
+/// An access reservation is the platform's cooperative physical-range exclusion mechanism for
+/// LiteBox mappings. It does not imply Rust ownership of the foreign physical memory and does not
```
lschuermann: What does "cooperative physical-range exclusion mechanism" mean? What guarantees does it give? This is similar to my comment above, as a user of this infrastructure, I'd want to know which guarantees I'm given.

litebox_common_linux/src/vmap.rs
```diff
@@ -11,38 +11,70 @@ use thiserror::Error;
...
     ///
-    /// The caller should ensure that `pages` are not in active use by other entities
-    /// (especially, there should be no read/write or write/write conflicts).
-    /// Unfortunately, LiteBox itself cannot fully guarantee this and it needs some helps
-    /// from the caller, hypervisor, or hardware.
-    /// Multiple LiteBox threads might concurrently call this function with overlapping
-    /// physical pages, so the implementation should safely handle such cases.
+    /// This function exposes raw mapped physical memory. The caller must not create Rust
+    /// references from the returned pointer unless it can uphold Rust's aliasing and validity
+    /// rules for the full lifetime of those references.
```
lschuermann: Is there ever a case where a caller can ensure this and, if so, how?

litebox_platform_linux_userland/src/lib.rs
lschuermann: I think, more precisely, the soundness of the overall system depends on this trait implementation being semantically correct and adhering to its contract. Hence, this must be an `unsafe trait` to discharge those safety obligations.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
lschuermann: I don't know what `LiteBox`'s internal policy for this is, but is this maybe a good point to insert a deliberate kernel panic? To me, it seems like there is value in distinguishing between
- a map failure that is benign and expected, or one that can be caused by some external actor,
- a map failure as a result of internal inconsistencies in OS-managed data structures.

In the latter case, it is very hard to gain confidence in the OS still maintaining basic correctness/security properties, esp. if a critical subsystem such as the pagetable is inconsistent. This makes me think that a `panic` policy is the right move here?

lschuermann (commented):
I want to take a slightly more in-depth pass over `lib.rs`, but otherwise this looks pretty good to me!

It's hard to judge what this + the other changes in #817 look like. Is it work merging this one into the `sanghle/lvbs/vmap_copy` branch first, and then doing another thorough pass over the merged result to make sure it's consistent?

Given the comparisons between this infrastructure and DMA, maybe it's worth looking at https://github.com/tock/tock/pull/4702? That PR to Tock addresses two separate but related concerns:
- It prevents the Rust compiler from making incorrect assumptions about the state of memory it can dereference (Rust slices). This is not directly applicable here, as memory is only ever accessed through `asm!()`-backed memcpy helpers, so there is no buffer with "shared custody" between Rust and the foreign entity.
- It emits the appropriate synchronization instructions (i.e., fences) for when Rust-issued writes need to be made visible to the foreign entity, or when foreign-issued writes need to be made visible to Rust. This may still be missing here.

litebox_common_linux/src/vmap.rs
sangho2: Technically, if this is correctly implemented and developers don't use any other uncontrolled interface, it cannot be invalidated or removed by others. Non-VTL1 entities cannot unmap this because they don't share virtual address space.

litebox_common_linux/src/vmap.rs
sangho2: VTL1 currently doesn't need to access its own memory through physical addresses because it doesn't interact with devices (directly). However, the LiteBox kernel platform in general doesn't have such restriction (LVBS/VTL1 is just a special case).

litebox_common_linux/src/vmap.rs
sangho2: As discussed yesterday, this is all about Rust-side protocol. If we have robust page table implementation which blocks or well-coordinate shared mappings, we don't need this.

litebox_common_linux/src/vmap.rs
sangho2: technically, no way.  better to drop this sentence.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: removed and opened an issue. If we have better page table impl, we don't need to worry about this issue.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: Agree with you. Fail-safe is not our goal but fail-fast is also not desired (we can't just crash VMs in the cloud). This should be discussed and handled separately.

litebox_common_linux/src/vmap.rs
sangho2: I dropped "reservation" for now.

PR #825, opened 2026-05-02
Add secure random number generator to LVBS platform

sangho2:
This PR adds Cryptographically Secure Pseudo-Random Number Generator (CSPRNG) to the LVBS platform. Userland applications (e.g., TAs) can use this CSPRNG using `fill_bytes_crng` though the shim. For now, it uses ChaCha20, so this is not for applications/scenarios demanding FIPS certification.

### DIFF
--- a/litebox_platform_lvbs/src/host/lvbs_impl.rs
+++ b/litebox_platform_lvbs/src/host/lvbs_impl.rs
@@ -232,7 +232,7 @@ fn rdrand_seed() -> Option<CrngSeed> {
         if !ok {
             return None;
         }
-        chunk.copy_from_slice(&word.to_ne_bytes()[..chunk.len()]);
+        chunk.copy_from_slice(&word.to_le_bytes()[..chunk.len()]);
     }
     Some(seed)
 }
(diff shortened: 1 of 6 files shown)


litebox_platform_lvbs/src/host/lvbs_impl.rs
```diff
@@ -136,6 +190,49 @@ pub(crate) fn set_platform_root_key(key: &[u8]) {
     });
 }
 
+fn rdrand_seed() -> <rand_chacha::ChaCha20Rng as SeedableRng>::Seed {
+    let mut seed = <rand_chacha::ChaCha20Rng as SeedableRng>::Seed::default();
+    for chunk in seed.chunks_mut(8) {
+        let mut word = 0;
+        loop {
+            // Safety: `RDRAND` is available on the LVBS target CPUs. A false
+            // carry flag means random data is temporarily unavailable.
```
wdcui, line 225: What happens when it's unavailable? Would we get a bad seed in this case?

sangho2: AFAIK, `RDRAND` returns zero if there is not enough HW entropy.

wdcui: If it returns zero, then we don't have a godo random seed. Would the current code return an error or silently continue?

sangho2: This function is for very first RNG initialization. If RDRAND is malfunctioning at that phase, there is no hope, so it panics.

wdcui: Panic is fine. Just wanted to make sure it will not silently ignore it.

wdcui:
GPT-5.5 reported this issue. Is it a valid concern?

   - High — litebox_platform_lvbs/src/host/lvbs_impl.rs:147: fill_bytes() serves ChaCha output before mixing fresh non-rollback entropy; RDRAND is    
  sampled only at CRNG initialization and after the 1 MiB reseed budget. A VM snapshot/clone between reseeds can replay identical random syscall bytes
  and TA stack canaries (litebox_shim_optee/src/loader/ta_stack.rs:262-265) until the next reseed. Mix fresh RDRAND/RDSEED or a VM-generation source  
  before serving externally visible/security-sensitive output, or detect resume/clone and reseed before output.

sangho2:
> GPT-5.5 reported this issue. Is it a valid concern?
> 
> * High — litebox_platform_lvbs/src/host/lvbs_impl.rs:147: fill_bytes() serves ChaCha output before mixing fresh non-rollback entropy; RDRAND is
>   sampled only at CRNG initialization and after the 1 MiB reseed budget. A VM snapshot/clone between reseeds can replay identical random syscall bytes
>   and TA stack canaries (litebox_shim_optee/src/loader/ta_stack.rs:262-265) until the next reseed. Mix fresh RDRAND/RDSEED or a VM-generation source
>   before serving externally visible/security-sensitive output, or detect resume/clone and reseed before output.

Yes, but this is out of our control and its threat model is unclear (someone should fork a VM).

wdcui (approved):
LGTM. Thanks.

PR #828, opened 2026-05-04
Add supports for generating identity signing key (IDK_S)

sangho2:
This PR adds supports for generating identity signing key (IDK_S) to the OP-TEE shim. A new function, `generate_identity_signing_key`, generates an IDK_S key pair based on the platform's CSPRNG and writes the public portion of it to the VTL0-side buffer. Currently, it generates and returns an uncompressed SEC1 P-384 public key.


sangho2:
~~VTL Call ABI for generating or getting IDK_S is straightforward. The main question is which layer should own IDK_S (LVBS platform/runner vs OP-TEE shim). If IDK_S is used only for OP-TEE scenarios, better to manage it in the OP-TEE Shim/PTA. Otherwise, if IDK_S is for a more general attestation service, better to maintain it in the platform/runner. This PR currently maintains IDK_S stuffs in the platform for simplicity (no cross-layer call is needed).~~

Now it does have it in the OP-TEE Shim (plan to include it in the IDK_S PTA).

litebox_platform_lvbs/src/mshv/mod.rs
```diff
@@ -131,6 +131,9 @@ pub const VSM_VTL_CALL_FUNC_ID_ALLOCATE_RINGBUFFER_MEMORY: u32 = 0x1_ffec;
 // This VSM function ID for setting the platform root key is subject to change
 pub const VSM_VTL_CALL_FUNC_ID_SET_PLATFORM_ROOT_KEY: u32 = 0x1_ffed;
 
+// This VSM function ID for generating the identity signing key is subject to change
+pub const VSM_VTL_CALL_FUNC_ID_GENERATE_IDENTITY_SIGNING_KEY: u32 = 0x1_ffee;
```
sangho2: - `0x1_ffee` if this is an LVBS command
- `0x1_fff1` if this is an OP-TEE/PTA command

sangho2:
Currently, this PR just returns a P-384 public key in an uncompressed SEC1 format. Instead, we can return it in a DER format. Also, in addition to the pub key, we can return a self-signed signature for validation.

litebox_shim_optee/src/idk.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+fn generate_identity_signing_private_key() -> Zeroizing<[u8; IDENTITY_SIGNING_PRIVATE_KEY_LEN]> {
+    let mut private_key_bytes = Zeroizing::new([0u8; IDENTITY_SIGNING_PRIVATE_KEY_LEN]);
+
+    loop {
+        litebox_platform_multiplex::platform().fill_bytes_crng(&mut *private_key_bytes);
+        if is_valid_identity_signing_private_key(&private_key_bytes) {
+            return private_key_bytes;
+        }
```
tgopinath-microsoft, line 122: do you think we should bound the loop here ?

tgopinath-microsoft (commented):
Do you think we should bound the loop here

tgopinath-microsoft (approved):

litebox_shim_optee/src/idk.rs
sangho2: loop bounding is a bit optional here, but adding it doesn't hurt. Let me do that.

PR #829, opened 2026-05-04
Support multiple built-in TAs

athvu:
This PR adds support for having multiple built-in TAs.

### DIFF
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -40,6 +40,7 @@ use litebox_platform_lvbs::{
     serial_println,
 };
 use litebox_platform_multiplex::Platform;
+use litebox_shim_optee::loader::ElfLoaderError;
 use litebox_shim_optee::msg_handler::{
     decode_ta_request, handle_optee_msg_args, handle_optee_smc_args, update_optee_msg_args,
 };
@@ -725,6 +726,8 @@ fn open_session_new_instance(
     client_identity: Option<litebox_common_optee::TeeIdentity>,
     ta_req_info: &litebox_shim_optee::msg_handler::TaRequestInfo<PAGE_SIZE>,
 ) -> Result<(), OpteeSmcReturnCode> {
+    let ta_bin = find_ta_binary(ta_uuid).map_err(|_| OpteeSmcReturnCode::ENotAvail)?;
+
     // Create and switch to new page table
     let task_pt_id = create_task_page_table()?;
 
@@ -756,7 +759,7 @@ fn open_session_new_instance(
         shim.load_ldelf(
             LDELF_BINARY,
             ta_uuid,
-            find_ta_binary(ta_uuid),
+            Some(ta_bin),
             client_identity,
             runner_session_id,
         )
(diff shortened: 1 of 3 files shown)


litebox_shim_optee/src/lib.rs
```diff
@@ -227,6 +227,9 @@ impl OpteeShim {
         client: Option<TeeIdentity>,
         session_id: u32,
     ) -> Result<LoadedProgram, loader::elf::ElfLoaderError> {
+        if ta_bin.is_none() {
+            return Err(loader::elf::ElfLoaderError::InvalidUuid);
+        }
```
sangho2: `ta_bin == None` will trigger dynamic TA loading (once it is ready). So, `None` is an expected value.

sangho2 (changes requested):
Hi Angelina, Thank you for your PR!

I've requested a minor change to maintain the core behavior of `load_ldelf`.

litebox_shim_optee/src/lib.rs
athvu: Thank you for the review! 

There is an issue with loading the OP-TEE module if I remove this.
As part of optee_probe(), there is a call to optee_enumerate_devices() which opens a session with device enumeration PTA. On the LiteBox side, this is handled as a normal TA.
Currently, Some(TA_BINARY) is passed to load_ldelf() as the ta_bin: Option<&[u8]> argument and the function returns Err(loader::elf::ElfLoaderError::InvalidUuid) at this point:
```
        if let Some(ta_bin) = ta_bin
            && !entrypoints.task.global.store_ta_bin(&ta_uuid, ta_bin)
        {
            return Err(loader::elf::ElfLoaderError::InvalidUuid);
        }
```
Then, returning to VTL0 and skipping the rest of the device enumeration code.

With my changes, None is passed to load_ldelf(), instead of Some(TA_BINARY), so the rest of load_ldelf() runs and returns Ok() and eventually there is a panic when ldelf tries to load the TA:

> ldelf/TA_CreateEntryPoint failed: return_code=ItemNotFound
> panicked at litebox_runner_lvbs/src/lib.rs:1209:5:

How would you suggest handling this?

sangho2: I think we shouldn't call `load_ldelf` in that case. `load_ldelf` is for loading a TA ELF binary through `ldelf`. PTA is not an ELF binary. I think it might be better to call `find_ta_binary` before `load_ldelf` and return `InvalidUuid` if it fails to find a corresponding TA.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -1314,6 +1317,21 @@ fn write_rpc_args_to_normal_world(
 // use include_bytes! to include ldelf and (KMPP) TA binaries
 const LDELF_BINARY: &[u8] = &[0u8; 0];
 const TA_BINARY: &[u8] = &[0u8; 0];
+const TA_BINARIES: &[&[u8]] = &[TA_BINARY];
+
+// Look up TA binary by UUID.
```
sangho2: nits: `//` -> `///` for rust doc.

Also, we could have TODO here to note we might need to handle some PTA UUIDs.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -1314,6 +1317,21 @@ fn write_rpc_args_to_normal_world(
...
+fn find_ta_binary(ta_uuid: litebox_common_optee::TeeUuid) -> Result<&'static [u8], ElfLoaderError> {
```
sangho2: It seems that return `Err(ElfLoader::InvalidUuid)` is less appropriate here because this error is not from the ELF loader and the TA file themselves. We could just return `Option<&'static [u8]>`.

sangho2 (approved):
Thanks!

PR #830, opened 2026-05-05
Add basic AArch64 support

sangho2:
This PR adds basic AArch64 support to the LiteBox core.  AArch64 and x86_64 have differences including assembly code, exception info, address space size, `c_char` (`i8` on x86_64 whereas `u8` on AArch64), and bitflags. Subsequent PRs (for Linux common, shim, platform, runner, and rewriter) are needed to run AArch64 Linux guests.


sangho2:
Will bring this up later.

PR #833, opened 2026-05-06
Harden OP-TEE-driven memory allocation and transfer

sangho2:
This PR hardens the memory allocation and transfer driven by OP-TEE to avoid potential out-of-memory issues. Note that OP-TEE OS does not specify this cap because it lets the secure-world kernel directly access untrusted memory (which is subject to TOCTTOU attacks). In contrast, the OP-TEE shim always copies data from the normal world or secure-world user space to the secure-world kernel space, such that it should specify the cap and check actual transfer size.


wdcui (approved):
LGTM. Thanks.

PR #834, opened 2026-05-06
Check the number of OP-TEE message parameters

sangho2:
This PR adds a check for the number of OP-TEE message parameters to deal with potentially malicious normal-world clients/kernel.


wdcui (approved):
LGTM. Thanks.

PR #835, opened 2026-05-07
Fix 9p client

CvvT:
This PR fixes multiple issues in the 9p client reported by Hyenas.

1. Previously we assumed the 9p server should be trusted, but it may not hold in the broker setup. Add input validation at various places.
2. Fix tag number wraparound issue.
3. Fix fid leak: in `clunk`, fid is removed from the server side even if server returns an error.
4. Refcounted Fid handles: a potential race issue may happen if one thread is trying to close a fid and create a new fid while another thread is using that fid. Although unlikely, we may reuse the closed fid and cause the second thread to read the wrong file.
5. Fix race issue on file offset update


litebox/src/fs/nine_p/client.rs
```diff
@@ -15,36 +16,73 @@ use super::Error;
 use super::fcall::{self, Fcall, FcallStr, GetattrMask, TaggedFcall};
 use super::transport::{self, Read, Write};
 
-/// Fid generator with thread-safe access
-struct FidGenerator<Platform: RawSyncPrimitivesProvider> {
+/// Pool of 9P Fid values
```
wdcui, line 19: Does other fs in LiteBox also manage their own Fid pool?

wdcui (approved):
LGTM. Thanks.

litebox/src/fs/nine_p/client.rs
CvvT: What do you mean? `Fid` is specific to 9p protocol.

PR #836, opened 2026-05-07
Revise OP-TEE `ShmInfo` to ensure byte-length bound check

sangho2:
This PR revises OP-TEE shim's `ShmInfo` to ensure byte-length bound check. Previously, `ShmInfo` only maintained an array of physical addresses and a start offset, such that its bound check was coarse-grained. A fine-grained check is needed to avoid confused-deputy attacks (i.e., preventing VTL1 kernel from corrupting some VTL0 bytes due to a page/byte gap).


wdcui (approved):
LGTM. Thanks.

PR #837, opened 2026-05-07
Harden physical-page linked-list traversal (LVBS/OP-TEE)

sangho2:
This PR hardens physical-page linked-list traversal in both LVBS (HEKI) and OP-TEE shim. Each implementation uses a page linked list to share data across non-contiguous physical pages. This PR adds additional validation to ensure these linked lists are well formed.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -1364,18 +1364,16 @@ impl<'a> ModuleMemoryMetadataIters<'a> {
 /// This function copies `HekiPage` structures from VTL0 and returns a vector of them.
 /// `pa` and `nranges` specify the physical address range containing one or more than one `HekiPage` structures.
 fn copy_heki_pages_from_vtl0(pa: u64, nranges: u64) -> Option<Vec<HekiPage>> {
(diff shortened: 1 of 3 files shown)


litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1348,16 +1348,26 @@ impl<'a> ModuleMemoryMetadataIters<'a> {
 fn copy_heki_pages_from_vtl0(pa: u64, nranges: u64) -> Option<Vec<HekiPage>> {
     let mut next_pa = PhysAddr::try_new(pa).ok()?;
     let mut heki_pages = Vec::with_capacity(nranges.truncate());
+    let mut visited_pages = HashSet::new();
     let mut range: u64 = 0;
 
     while range < nranges {
+        let cur_pa = next_pa;
```
wdcui: Why do we need to create `cur_pa`?

wdcui (approved):
LGTM. Thanks.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Thanks for a good catch. This is a legacy. Let me fix it.

Issue #839, opened 2026-05-08
LVBS/OP-TEE test failure


sangho2:
LVBS/OP-TEE test failed likely due to #810

sangho2:
PR #841

PR #841, opened 2026-05-08
Fix OP-TEE ELF loading without trampoline bug

sangho2:
The return values of `parse_trampoline` have changed (i.e., it returns `UnpatchedBinary` for an ELF binary without trampoline), but litebox_shim_optee has not been updated accordingly.

Issue #839

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -1343,7 +1343,9 @@ impl<Host: HostInterface> StdioProvider for LinuxKernel<Host> {
 
 impl<Host: HostInterface> litebox::platform::SystemInfoProvider for LinuxKernel<Host> {
     fn get_syscall_entry_point(&self) -> usize {
-        syscall_callback as *const () as usize
+        // Currently this is only used in ELF loader to fix trampoline code.
+        // When running in kernel mode, we don't need a syscall trampoline.
+        0
     }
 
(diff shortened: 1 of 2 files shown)


wdcui (approved):
Why would OPTEE shim see unpatched binary?

sangho2:
> Why would OPTEE shim see unpatched binary?

This is for LVBS. It doesn't use trampoline and doesn't patch binaries.

wdcui:
> > Why would OPTEE shim see unpatched binary?
> 
> This is for LVBS. It doesn't use trampoline and doesn't patch binaries.

Then would `get_syscall_entry_point()` return 0?

sangho2:
> > > Why would OPTEE shim see unpatched binary?
> > 
> > 
> > This is for LVBS. It doesn't use trampoline and doesn't patch binaries.
> 
> Then would `get_syscall_entry_point()` return 0?

Currently, it returns the callback's address. Better to make it return zero.

wdcui:
> > > > Why would OPTEE shim see unpatched binary?
> > > 
> > > 
> > > This is for LVBS. It doesn't use trampoline and doesn't patch binaries.
> > 
> > 
> > Then would `get_syscall_entry_point()` return 0?
> 
> Currently, it returns the callback's address. Better to make it return zero.

Yeah, that's the contract: we expect zero for kernel platforms.

PR #842, opened 2026-05-08
Fix RFLAGS handling of kernel platforms

sangho2:
This PR fixes RFLAGS handling of the LVBS and SNP platforms to prevent critical ring-3 RFLAGS (i.e., AC, DF, TF, NT, and IOPL) from affecting ring-0 behaviors. Note that the SNP platform doesn't use SFMASK and doesn't enable SMAP.

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -1765,7 +1765,7 @@ macro_rules! XRSTOR_VTL1_ASM {
 ///
 /// Prerequisite:
 /// - Store user `rsp` in `r11` before calling this macro.
-/// - Store user `rflags` in `gs:[scratch]` before calling this macro.
+/// - Store user `rflags` in `gs:[user_rflags]` before calling this macro.
 /// - Store the userspace return address in `rcx` (`syscall` does this automatically).
 #[cfg(target_arch = "x86_64")]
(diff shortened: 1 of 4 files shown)


litebox_platform_lvbs/src/lib.rs
```diff
@@ -1765,14 +1765,15 @@ macro_rules! XRSTOR_VTL1_ASM {
...
+/// - Store user `rflags` in `gs:[scratch]` before calling this macro.
 /// - Store the userspace return address in `rcx` (`syscall` does this automatically).
 #[cfg(target_arch = "x86_64")]
 macro_rules! SAVE_SYSCALL_USER_CONTEXT_ASM {
     () => {
         "
         push 0x2b       // pt_regs->ss = __USER_DS
         push r11        // pt_regs->rsp
-        pushfq          // pt_regs->eflags
+        push qword ptr gs:[{scratch_off}] // pt_regs->eflags
```
wdcui: Do we need to make similar changes in userland platforms?

wdcui (approved):
LGTM. Thanks.

litebox_platform_lvbs/src/lib.rs
sangho2: I think so (for defense-in-depth) because a guest might invoke `syscall` while `DF=1`, which can result in unexpected behaviors at LiteBox (in particular, its `naked_asm` blocks).

PR #844, opened 2026-05-14
Harden `switch_to_user` (LVBS)

sangho2:
This PR hardens LVBS's `switch_to_user` by checking whether a given user `rsp` is valid before switching back to the userspace. A user `rsp` should be a canonical userspace address. If not, `iretq` could end up with a kernel exception.


wdcui (approved):
LGTM. Thanks.

PR #845, opened 2026-05-14
Scrub OP-TEE TA stack

sangho2:
This PR scrubs the OP-TEE TA stack at the initialization. This prevents potential information leakage because the TA stack is recycled across different sessions of the same single-instance TA.

### DIFF
--- a/litebox_shim_optee/src/loader/ta_stack.rs
+++ b/litebox_shim_optee/src/loader/ta_stack.rs
@@ -125,11 +125,12 @@ impl TaStack {
     }
 
     /// Zero the unused stack region before a new session writes its parameters,
-    /// to avoid leaking leftover data from a prior session whose pages were recycled.
+    /// to avoid leaking leftover data from a prior session whose stack region was recycled.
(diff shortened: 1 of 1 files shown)


litebox_shim_optee/src/loader/ta_stack.rs
```diff
@@ -124,6 +124,21 @@ impl TaStack {
         Some(())
     }
 
+    /// Zero the unused stack region before a new session writes its parameters,
```
wdcui, line 127: Do we need to zero all memory pages (not just stack) that are allocated to the guest?

wdcui (approved):
LGTM. Thanks.

litebox_shim_optee/src/loader/ta_stack.rs
sangho2: Don't have to because we do use LiteBox's mmap interface for normal guest memory allocation. TA stack (for a single-instance TA) is special because it does reuse the allocated stack by default.

PR #847, opened 2026-05-15
Fix kernel module validation bugs

sangho2:
This PR fixes some kernel module validation bugs which reject some empty sections and overlook some symbol membership.


wdcui (approved):
LGTM. Thanks.

PR #848, opened 2026-05-15
Fix potential use-after-unmap due to stale TLB

sangho2:
This PR fixes potential use-after-unmap due to stale TLB entries. `unmap_pages` now deallocates frames after flushing TLB entries across all cores.


wdcui (approved):
LGTM. Thanks.

PR #849, opened 2026-05-15
Fix breakage of kdata load

tgopinath-microsoft:
kdata load return prematurely because KexecTrampoline metadata is only 2048 and hence not page aligned. For now let kdata load continue and ignore KexecTrampoline metadata if the size does not match. This is a basic fix to prevent heki itself from breaking. A final fix requires discussion on whether the metadata should be atleast a page and code changes possibly in VTL0 and VTL1.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -347,6 +347,7 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, VsmError> {
 
     let mut system_certs_mem = MemoryContainer::new();
     let mut kexec_trampoline_metadata = KexecMemoryMetadata::new();
+    let mut kexec_trampoline_insert_failed = false;
     let mut patch_info_mem = MemoryContainer::new();
     let mut kinfo_mem = MemoryContainer::new();
     let mut kdata_mem = MemoryContainer::new();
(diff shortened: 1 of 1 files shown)


femiatl (approved):
Looks good. Nit on title spelling "prem-maturely" -> prematurely

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -412,11 +418,13 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, VsmError> {
     vtl0_info.set_system_certificates(certs.clone());
     debug_serial_println!("VSM: Loaded {} system certificate(s)", certs.len());
 
-    for kexec_trampoline_range in &kexec_trampoline_metadata {
-        protect_physical_memory_range(
-            kexec_trampoline_range.phys_frame_range,
-            MemAttr::MEM_ATTR_READ,
-        )?;
+    if !kexec_trampoline_insert_failed {
```
sangho2, line 425: Please inline TODO here to note we'll protect this metadata in the future (to avoid false alarm).

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -361,7 +362,12 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, VsmError> {
                     .extend_range(heki_range)
                     .map_err(|_| VsmError::InvalidInputAddress)?,
                 HekiKdataType::KexecTrampoline => {
-                    kexec_trampoline_metadata.insert_heki_range(heki_range)?;
+                    if let Err(e) = kexec_trampoline_metadata.insert_heki_range(heki_range) {
```
sangho2, line 365: it seems that `mshv_vsm_validate_kexec` does have the same issue. Can ignore this for now because we don't validate kexec but worthwhile to consider.

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -361,7 +362,12 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, VsmError> {
...
+                        debug_serial_println!(
+                            "VSM: KexecTrampoline insert_heki_range failed ({e:?}); skipping kexec trampoline protection"
+                        );
+                        kexec_trampoline_insert_failed = true;
```
sangho2, line 369: nits: this might be fine because it is a temporal solution for now. But we don't need `insert_heki_range` anymore if failed is true (or can just clear the entire data structure).

sangho2 (approved):

litebox_platform_lvbs/src/mshv/vsm.rs
tgopinath-microsoft: I don't understand.

PR #851, opened 2026-05-15
Fix heki memory protect breakage

tgopinath-microsoft:
Currently if a memory range for which a permission change is requested, overlaps with VTL1 working memory, the entire request is ignored and a error is returned. This leads to initial heki protect memory call to fail because for VTL0, the VTL1 working memory is still part of the kernel memory(reserved). Fix protect_physical_memory_range so that if the incoming range has an overlap with VTL1 memory, split the range out, ignore the VTL1 memory range and proceed with setting the permissions for rest of the memory in the requested range.

### DIFF
--- a/litebox_platform_lvbs/src/mshv/vsm.rs
+++ b/litebox_platform_lvbs/src/mshv/vsm.rs
@@ -1384,23 +1384,57 @@ fn copy_heki_pages_from_vtl0(pa: u64, nranges: u64) -> Option<Vec<HekiPage>> {
     Some(heki_pages)
 }
 
-/// This function protects a VTL0 physical memory range from potentially compromised VTL0 by
-/// restricting its access permissions using VTL protection mask (e.g., kernel code integrity).
-/// It is a safe wrapper for `hv_modify_vtl_protection_mask`. `phys_frame_range` specifies the
-/// physical frame range to protect (must belong to VTL0) `mem_attr` specifies the memory
-/// attributes (VTL0's allowed access) to be applied to the range
-#[inline]
+/// Protects a VTL0 physical memory range from potentially compromised VTL0 by restricting its
+/// access permissions using VTL protection mask (e.g., kernel code integrity).
+///
+/// If the requested range overlaps with VTL1 working memory, the VTL1 portion is silently
+/// skipped and only the remaining VTL0 portions are protected. If the range falls entirely
+/// within VTL1, this function returns `Ok(())` without issuing a hypercall.
+///
+/// `phys_frame_range` specifies the physical frame range to protect (must belong to VTL0).
+/// `mem_attr` specifies the memory attributes (VTL0's allowed access) to be applied.
 pub(crate) fn protect_physical_memory_range(
     phys_frame_range: PhysFrameRange<Size4KiB>,
     mem_attr: MemAttr,
 ) -> Result<(), VsmError> {
     let vtl1_range = crate::platform_low().vtl1_phys_frame_range();
-    if phys_frame_range.start < vtl1_range.end && vtl1_range.start < phys_frame_range.end {
-        return Err(VsmError::Vtl1MemoryOverlap);
+
+    // Fast path: no overlap with VTL1 — protect the entire range directly.
+    let overlaps_vtl1 =
+        phys_frame_range.start < vtl1_range.end && vtl1_range.start < phys_frame_range.end;
+
+    if !overlaps_vtl1 {
+        let pa = phys_frame_range.start.start_address().as_u64();
+        let num_pages = phys_frame_range.count() as u64;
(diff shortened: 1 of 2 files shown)


litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Please update comment to explain now it (silently) skips VTL1 range. Also, please remove `#[inline]` (this function is now lengthy).

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1383,14 +1383,41 @@ pub(crate) fn protect_physical_memory_range(
...
-        hv_modify_vtl_protection_mask(pa, num_pages, mem_attr_to_hv_page_prot_flags(mem_attr))
-            .map_err(VsmError::HypercallFailed)?;
+
+    // Determine whether `phys_frame_range` overlaps the VTL1 reserved range.
+    // If it does, split the request into the portions before and after the
+    // VTL1 range and skip the overlapping pages entirely. Otherwise, protect
+    // the whole range as-is.
+    let overlaps_vtl1 =
+        phys_frame_range.start < vtl1_range.end && vtl1_range.start < phys_frame_range.end;
+
```
sangho2, line 1405: If it doesn't overlap, this function can quickly call `hv_modify_vtl_protection_mask` without dealing with two sub ranges. Also, if a given range fully belongs to VTL1, this function can simply return (with `Vtl1MemoryOverlap` if helpful, or remove `Vtl1MemoryOverlap` which is now dead.)

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1383,14 +1383,41 @@ pub(crate) fn protect_physical_memory_range(
...
+        [
+            phys_frame_range,
+            PhysFrame::range(phys_frame_range.end, phys_frame_range.end),
+        ]
+    };
+
+    for sub_range in sub_ranges {
+        if sub_range.start >= sub_range.end {
+            continue;
+        }
```
sangho2, line 1435: this check and the below `count() > 0` are redundant.

sangho2 (approved):

sangho2:
> Currently if a memory range for which a permission change is requested, overlaps with VTL1 working memory, the entire request is ignored and a error is returned. This leads to initial heki protect memory call to fail because for VTL0, the VTL1 working memory is still part of the kernel memory(reserved). Fix protect_physical_memory_range so that if the incoming range has an overlap with VTL1 memory, split the range out, ignore the VTL1 memory range and proceed with setting the permissions for rest of the memory in the requested range.

This sounds like VTL0-side HEKI code is unaware of the VTL1 memory range. Why does it need to touch VTL1's memory? If this is due to byte/page protection granularity gaps, this overlap skip might not fully work because the page that VTL0 and VTL1 share is still protected.

femiatl (commented):
This could be a temporary fix, but long term, VTL0 should be responsible for presenting the right memory ranges to LiteBox. If LiteBox silently accounts for the reserved VTL memory range, VTL0 now has an incomplete view of memory range permissions. When VTL0 makes future requests on this memory range, believing it can (un)set the memory range's permissions, LiteBox will possibly have to perform the same skip logic, leaving VTL0 to think it succeeded. I think it would be better to return error and let VTL0 deal with it.

tgopinath-microsoft:
> > Currently if a memory range for which a permission change is requested, overlaps with VTL1 working memory, the entire request is ignored and a error is returned. This leads to initial heki protect memory call to fail because for VTL0, the VTL1 working memory is still part of the kernel memory(reserved). Fix protect_physical_memory_range so that if the incoming range has an overlap with VTL1 memory, split the range out, ignore the VTL1 memory range and proceed with setting the permissions for rest of the memory in the requested range.
> 
> This sounds like VTL0-side HEKI code is unaware of the VTL1 memory range. Why does it need to touch VTL1's memory? If this is due to byte/page protection granularity gaps, this overlap skip might not fully work because the page that VTL0 and VTL1 share is still protected.

So not really. It is how Linux  looks at memory. Linux reserves this memory and never uses for its operations. But as far as Linux is concerned this is still kernel memory. And we send the entire kernel memory for protection to VTL1.

tgopinath-microsoft:
> This could be a temporary fix, but long term, VTL0 should be responsible for presenting the right memory ranges to LiteBox. If LiteBox silently accounts for the reserved VTL memory range, VTL0 now has an incomplete view of memory range permissions. When VTL0 makes future requests on this memory range, believing it can (un)set the memory range's permissions, LiteBox will possibly have to perform the same skip logic, leaving VTL0 to think it succeeded. I think it would be better to return error and let VTL0 deal with it.

So the vtl1_range is just the initial memory presented to VTL1. Nothing else. So a) Linux will not touch that memory because it is reserved. b) If an attacker touches it and sends request to modify memory permissions after boot, we should disallow it in VTL1. I think we do that with the end_of_boot_flag

PR #858, opened 2026-05-20
Add setitimer/getitimer

CvvT:
Implement syscalls `setitimer` and `getitime`, but `setitimer` does not support periodic timers yet (i.e., it is equivalent to `alarm` for now).

Also add a new feature `alarm_fallback` to enable/disable the fallback alarm check.

### DIFF
--- a/litebox_shim_linux/src/syscalls/process.rs
+++ b/litebox_shim_linux/src/syscalls/process.rs
@@ -25,7 +25,8 @@ use litebox::platform::{RawMutPointer as _, TimerHandle, TimerProvider};
 use litebox::sync::Mutex;
 use litebox::utils::TruncateExt as _;
 use litebox_common_linux::{
-    ArchPrctlArg, CloneFlags, FutexArgs, PrctlArg, TimeParam, errno::Errno,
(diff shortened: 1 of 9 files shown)


litebox_shim_linux/src/syscalls/process.rs
```diff
@@ -1129,25 +1144,24 @@ impl<FS: ShimFS> Task<FS> {
...
+        if prev.is_zero() {
+            Ok(0)
+        } else {
+            let extra = u64::from(prev.subsec_nanos() > 0);
+            Ok(u32::try_from(prev.as_secs() + extra).unwrap_or(u32::MAX))
+        }
+    }
+
+    /// Arm or disarm the per-process `ITIMER_REAL` timer (the state `alarm()`
+    /// and `setitimer(ITIMER_REAL, ...)` both manipulate). Returns the raw
```
wdcui: Nit: "both manipulate" to something else?

wdcui (approved):
LGTM. Thanks.

litebox_shim_linux/src/syscalls/process.rs
CvvT: Removed this comment.

PR #861, opened 2026-05-21
Add syscall recvmmsg

CvvT:


litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -1719,6 +1727,104 @@ impl<FS: ShimFS> Task<FS> {
...
+
+        // Linux's `do_recvmmsg` validates the timespec before looking up the fd,
+        // so a bad timeout takes precedence over EBADF.
+        let timeout_duration = timeout.read()?;
+
+        let Ok(sockfd) = u32::try_from(fd) else {
+            return Err(Errno::EBADF);
+        };
+
+        let vlen = (vlen as usize).min(UIO_MAXIOV);
```
wdcui: We don't return an error when `vlen` is greater than `UIO_MAXIOV`?

wdcui (approved):
LGTM. Thanks.

litebox_shim_linux/src/syscalls/net.rs
CvvT: Good catch. Looks like Linux does not have any explicit check on `vlen`. When `copy_to_user` fails, it returns `EFAULT`.

PR #862, opened 2026-05-22
Suppress LVBS/OP-TEE `todo!` macros

sangho2:
This PR suppresses some `todo!` macros which can be triggered by userspace processes and might result in kernel panic.

### DIFF
--- a/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
+++ b/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
@@ -893,7 +893,7 @@ impl<M: MemoryProvider, const ALIGN: usize> PageTableImpl<ALIGN> for X64PageTabl
                 #[cfg(not(debug_assertions))]
                 {
                     crate::serial_println!("Invalid frame address: {:#x}", pa);
-                    return Err(PageFaultError::AllocationFailed);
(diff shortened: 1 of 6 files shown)


litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -826,7 +888,13 @@ impl<M: MemoryProvider, const ALIGN: usize> PageTableImpl<ALIGN> for X64PageTabl
                 }
             }
             TranslateResult::InvalidFrameAddress(pa) => {
+                #[cfg(debug_assertions)]
                 todo!("Invalid frame address: {:#x}", pa);
+                #[cfg(not(debug_assertions))]
+                {
+                    crate::serial_println!("Invalid frame address: {:#x}", pa);
+                    return Err(PageFaultError::AllocationFailed);
```
wdcui: Double check: Is the error code `AllocationFailed` correct here?

wdcui (approved):
LGTM. Thanks.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: Thanks. Replaced it with `AccessError`.

PR #864, opened 2026-05-22
Add preadv/pwritev

CvvT:
This PR supports syscalls preadv/pwritev. 

The implementation is functionally correct but semantically slightly different from Linux. Linux's kernel does fdget() once at syscall entry and fdput() at exit, so a concurrent close() mid-vectored-op is not observed by later iovs. Our current per-iov lookup would observe it (returning EBADF mid-stream with partial progress preserved).


wdcui (approved):
LGTM. Thanks.

PR #865, opened 2026-05-22
Refactor OP-TEE TA lifecycle management

sangho2:
This PR refactors OP-TEE TA lifecycle management to ensure the safety and compatibility with the original OP-TEE TA lifecycle management. Specifically, it serializes the execution of the same single-instance TA with its UUID and the execution of the same multi-instance TA with its session ID.

### DIFF
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -44,11 +44,11 @@ use litebox_shim_optee::msg_handler::{
     decode_ta_request, handle_optee_msg_args, handle_optee_smc_args, update_optee_msg_args,
 };
 use litebox_shim_optee::session::{
-    CreationReservation, SessionIdGuard, SessionManager, TaInstance, allocate_session_id,
+    CreationReservation, SessionIdGuard, SessionManager, SessionTarget, TaInstance,
+    allocate_session_id,
 };
 use litebox_shim_optee::{NormalWorldConstPtr, NormalWorldMutPtr, UserConstPtr};
 use once_cell::race::OnceBox;
-use spin::mutex::SpinMutex;
 
 /// Seed the initial heap regions so the global allocator has enough memory
 /// for slab-backed allocations (the slab needs >= 2 MB backing pages).
@@ -358,6 +358,35 @@ unsafe fn delete_task_page_table(task_pt_id: usize) -> Result<(), OpteeSmcReturn
     }
 }
 
+/// Guard that restores the base page table when leaving a TA page table scope.
+struct TaskPageTableGuard {
+    active: bool,
+}
+
+impl TaskPageTableGuard {
+    fn enter(task_pt_id: usize) -> Result<Self, OpteeSmcReturnCode> {
+        unsafe { switch_to_task_page_table(task_pt_id)? };
+        Ok(Self { active: true })
+    }
+
+    fn leave(mut self) {
+        unsafe { switch_to_base_page_table() };
+        self.active = false;
+    }
+
+    fn deactivate(&mut self) {
+        self.active = false;
+    }
+}
+
+impl Drop for TaskPageTableGuard {
+    fn drop(&mut self) {
+        if self.active {
+            unsafe { switch_to_base_page_table() };
+        }
+    }
+}
+
 /// Tears down a TA's memory mappings and page table.
 ///
 /// This performs the following steps in order:
@@ -378,6 +407,15 @@ unsafe fn teardown_ta_page_table(shim: &litebox_shim_optee::OpteeShim, task_pt_i
         // Now delete the TA's page table without memory leak.
         let _ = delete_task_page_table(task_pt_id);
     }
+}
+
+unsafe fn teardown_active_ta_page_table(
+    guard: &mut TaskPageTableGuard,
+    shim: &litebox_shim_optee::OpteeShim,
+    task_pt_id: usize,
+) {
+    guard.deactivate();
+    unsafe { teardown_ta_page_table(shim, task_pt_id) };
 }
 
 /// Handler for OP-TEE SMC calls.
@@ -504,13 +542,17 @@ fn handle_open_session(
         .get_known_flags(&ta_uuid)
         .is_none_or(|f| f.is_single_instance());
 
-    // Resolve or create the TA instance.
-    // For single-instance TAs, `with_creation_slot` re-checks the cache
-    // under its lock and serializes instance creation per UUID.
-    // If a cache hit returns a zombie (an instance torn down by a
-    // concurrent close/panic), evict the dead entry and ask the Linux driver
-    // to retry so it can create a fresh TA instance.
-    match session_manager().with_creation_slot(&ta_uuid, is_single_instance, || {
+    let single_instance_lock =
+        is_single_instance.then(|| session_manager().single_instance_lock(ta_uuid));
+    let _single_instance_guard = if let Some(lock) = single_instance_lock.as_ref() {
+        Some(lock.try_lock().ok_or(OpteeSmcReturnCode::EThreadLimit)?)
+    } else {
+        None
+    };
+
+    // Resolve or create the TA instance. For single-instance TAs, the UUID
+    // lock above serializes creation, reuse, and teardown for this TA.
+    let result = match session_manager().with_creation_slot(&ta_uuid, is_single_instance, || {
         open_session_new_instance(
             msg_args,
             msg_args_phys_addr,
@@ -519,68 +561,48 @@ fn handle_open_session(
             client_identity,
             &ta_req_info,
         )
-    })? {
-        CreationReservation::ExistingSingleInstance(existing) => {
-            match open_session_single_instance(
-                msg_args,
-                msg_args_phys_addr,
-                existing.clone(),
-                params,
-                ta_uuid,
-                &ta_req_info,
-            )? {
-                OpenSessionOutcome::Handled => Ok(()),
-                OpenSessionOutcome::InstanceDestroyed => {
-                    // Evict the zombie. Analog of OP-TEE's `maybe_release_ta_ctx`
-                    // removing the dead ctx from `tee_ctxes`.
-                    let _ = session_manager().remove_single_instance_if_same(&ta_uuid, &existing);
-                    Err(OpteeSmcReturnCode::EThreadLimit)
-                }
-            }
-        }
-        CreationReservation::SlotReserved => Ok(()),
-    }
-}
+    }) {
+        Ok(CreationReservation::ExistingSingleInstance(existing)) => open_session_single_instance(
+            msg_args,
+            msg_args_phys_addr,
+            existing.clone(),
+            params,
+            ta_uuid,
+            &ta_req_info,
+        ),
+        Ok(CreationReservation::SlotReserved) => Ok(()),
+        Err(e) => Err(e),
+    };
 
-/// Outcome of [`open_session_single_instance`].
-enum OpenSessionOutcome {
-    /// Session was successfully opened, TA returned a non-fatal error, or TA panicked
-    /// and the instance was destroyed inline. No extra cleanup effort is needed.
-    Handled,
-    /// The cached `TaInstance` is `closed` and must not be entered.
-    InstanceDestroyed,
+    // If we conservatively held the per-UUID lock for an unknown TA that
+    // turned out to be multi-instance, evict the lock entry we created.
+    // Subsequent OpenSessions for this UUID skip the lock entirely (their
+    // `get_known_flags` will now return the multi-instance flags), so the
+    // entry would otherwise leak.
+    if single_instance_lock.is_some()
+        && let Some(actual_flags) = session_manager().get_known_flags(&ta_uuid)
+        && !actual_flags.is_single_instance()
+    {
+        session_manager().evict_single_instance_lock_if_unused(&ta_uuid);
+    }
+
+    result
 }
 
 /// Open a new session on an existing single-instance TA.
 ///
-/// Returns `Err(OpteeSmcReturnCode::EThreadLimit)` if the TA instance is currently in use.
-/// The Linux driver will wait and retry automatically.
-/// Returns [`OpenSessionOutcome::InstanceDestroyed`] if the cached TA is closed.
-///
 /// If the TA's OpenSession entry point returns an error, the session is not registered.
-/// On TARGET_DEAD the cached instance is destroyed unconditionally; any sibling sessions
-/// become orphans that fail-fast on next access via the `instance.closed` check.
+/// On TARGET_DEAD the cached instance is destroyed unconditionally.
 /// For cleanup semantics, see OP-TEE OS `tee_ta_open_session()` in `tee_ta_manager.c`.
-#[allow(clippy::type_complexity)]
 fn open_session_single_instance(
     msg_args: &mut OpteeMsgArgs,
     msg_args_phys_addr: u64,
-    instance_arc: Arc<SpinMutex<TaInstance>>,
+    instance_arc: Arc<TaInstance>,
     params: &[litebox_common_optee::UteeParamOwned],
     ta_uuid: litebox_common_optee::TeeUuid,
     ta_req_info: &litebox_shim_optee::msg_handler::TaRequestInfo<PAGE_SIZE>,
-) -> Result<OpenSessionOutcome, OpteeSmcReturnCode> {
-    // Use try_lock to avoid spinning - return EThreadLimit if TA is in use
-    // The Linux driver will handle this by waiting and retrying
-    let mut instance = instance_arc
-        .try_lock()
-        .ok_or(OpteeSmcReturnCode::EThreadLimit)?;
-
-    // `closed == true` means the instance is terminal and must not be entered.
-    if instance.closed {
-        return Ok(OpenSessionOutcome::InstanceDestroyed);
-    }
-    let task_pt_id = instance.task_page_table_id;
+) -> Result<(), OpteeSmcReturnCode> {
+    let task_pt_id = instance_arc.task_page_table_id;
 
     // Allocate session ID BEFORE calling load_ta_context so TA gets correct ID.
     // Use SessionIdGuard to ensure the ID is recycled on any error path
@@ -597,13 +619,12 @@ fn open_session_single_instance(
         runner_session_id
     );
 
-    let ta_flags = instance.loaded_program.ta_flags;
+    let ta_flags = instance_arc.loaded_program.ta_flags;
 
-    // Switch to the existing TA's page table
-    unsafe { switch_to_task_page_table(task_pt_id)? };
+    let mut task_pt_guard = TaskPageTableGuard::enter(task_pt_id)?;
 
     // Load TA context with parameters for OpenSession - pass actual session_id
-    instance
+    instance_arc
         .loaded_program
         .entrypoints
         .as_ref()
@@ -620,13 +641,13 @@ fn open_session_single_instance(
     let mut ctx = litebox_common_linux::PtRegs::default();
     unsafe {
         litebox_platform_lvbs::reenter_thread_ref(
-            instance.loaded_program.entrypoints.as_ref().unwrap(),
+            instance_arc.loaded_program.entrypoints.as_ref().unwrap(),
             &mut ctx,
         );
     }
 
     // Read TA output parameters from the stack buffer
-    let params_address = instance
+    let params_address = instance_arc
         .loaded_program
         .params_address
         .ok_or(OpteeSmcReturnCode::EBadAddr)?;
@@ -647,8 +668,9 @@ fn open_session_single_instance(
         );
 
         // Write error response BEFORE switching page tables (accesses user memory).
-        // Keep the instance lock held until this completes so another core cannot
-        // tear down the active page table while this core is copying TA outputs.
+        // The per-UUID lock held by the caller of `handle_open_session` keeps
+        // another core from tearing down the active page table while this core
+        // is copying TA outputs.
         let write_result = write_msg_args_to_normal_world(
             msg_args,
             msg_args_phys_addr,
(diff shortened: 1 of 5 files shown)


litebox_runner_lvbs/src/lib.rs
```diff
@@ -1203,36 +1242,32 @@ fn handle_close_session(
...
-                drop(instance);
                 debug_serial_println!(
                     "CloseSession complete: session_id={}, TA kept alive (INSTANCE_KEEP_ALIVE flag)",
                     session_id
                 );
                 return write_result;
             }
 
-            // Clear single-instance cache if this was a single-instance TA
+            // Clear single-instance cache (and evict its per-UUID lock) if this
```
wdcui: The single-instance cache may contain TAs that are not single-instance?

litebox_shim_optee/src/session.rs
```diff
@@ -115,26 +131,31 @@ pub struct TaInstance {
...
-    /// Set when the TA is committed to teardown (panic or last session closed). Any lock
-    /// holders should check `closed` before touching `task_page_table_id` and bail if true.
-    ///
-    /// The per-instance lock must be held when setting `closed = true` and across
-    /// the subsequent `teardown_ta_page_table`.
-    pub closed: bool,
 }
 
-// SAFETY: TaInstance is protected by SpinMutex and try_lock (`SessionEntry`)
+// SAFETY: TaInstance is shared as `Arc<TaInstance>`, but only one core is ever
```
wdcui: I'm a little confused by this comment. We will have a separate instance for each session for a multiple-instance TA but we will have a single instance for a single-instance TA?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -358,6 +358,27 @@ unsafe fn delete_task_page_table(task_pt_id: usize) -> Result<(), OpteeSmcReturn
     }
 }
 
+/// Guard that restores the base page table when leaving a TA page table scope.
```
wdcui: Why is this a runner concept instead of OPTEE?

litebox_shim_optee/src/session.rs
```diff
@@ -369,6 +411,11 @@ pub struct SessionManager {
     creation_state: SpinMutex<CreationState>,
     /// Cached TA flags by UUID, populated on first successful session registration.
     known_flags: SpinMutex<HashMap<TeeUuid, TaFlags>>,
+    /// Per-UUID serialization locks for single-instance TA handling.
+    single_instance_locks: SpinMutex<HashMap<TeeUuid, Arc<SpinMutex<()>>>>,
+    /// Session ids currently being handled (Invoke/Close). Guards a session
+    /// against concurrent SMC entry by another core that targets the same id.
+    active_sessions: SpinMutex<HashSet<u32>>,
```
wdcui, line 465: Session id is a u32?

litebox_shim_optee/src/session.rs
```diff
@@ -418,11 +464,34 @@ impl SessionManager {
...
+            .lock()
+            .entry(uuid)
+            .or_insert_with(|| Arc::new(SpinMutex::new(())))
+            .clone()
+    }
+
+    /// Try to mark `session_id` as actively handled. Returns a guard that
+    /// removes the marker on drop, or `None` if another core is already
+    /// inside an Invoke/Close handler for this session id.
+    pub fn try_activate_session(&self, session_id: u32) -> Option<ActiveSessionGuard<'_>> {
```
wdcui: What's the relation between `try_activate_session` and `register_session`?

litebox_runner_lvbs/src/lib.rs
sangho2: Our OP-TEE shim doesn't have a capability to manage address spaces or page tables (yet). Also, this is independent of shims because it is more about how to manage page tables in general.

litebox_runner_lvbs/src/lib.rs
sangho2: Let me revise the comment. It's confusing. We clear an entry of single-instance cache when the last session of a single-instance TA is closed.

litebox_shim_optee/src/session.rs
sangho2: Yes. Each session of a multi-instance TA has a separate instance. Each session of a single-instance TA share the same instance. Let me clarify this.

litebox_shim_optee/src/session.rs
sangho2: Yes. The OP-TEE Linux driver assumes it.

litebox_shim_optee/src/session.rs
sangho2: This function aims to detect whether the normal world concurrently uses the same session ID. Technically, it is just defense-in-depth. The OP-TEE protocol has no way to prevent session ID forgery.

litebox_shim_optee/src/session.rs
```diff
@@ -222,33 +274,35 @@ impl Default for SessionMap {
...
-    pub fn insert(&self, uuid: TeeUuid, instance: Arc<SpinMutex<TaInstance>>) {
+    fn insert(&self, uuid: TeeUuid, instance: Arc<TaInstance>) {
         self.inner.lock().insert(uuid, instance);
     }
 
-    /// Remove a cached single-instance TA only if it is the expected instance.
-    fn remove_if_same(&self, uuid: &TeeUuid, expected: &Arc<SpinMutex<TaInstance>>) -> bool {
+    /// Evict only if the cached instance matches `task_page_table_id`.
+    /// Distinguishes the live instance from a freshly-created one with the
+    /// same UUID when the caller wants to remove a specific one.
```
wdcui, line 302: How does it distinguish them?

litebox_shim_optee/src/session.rs
```diff
@@ -222,33 +274,35 @@ impl Default for SessionMap {
...
+    fn remove_if_pt(&self, uuid: &TeeUuid, task_page_table_id: usize) -> bool {
```
wdcui: In `remove_if_pt`, what does `pt` stand for?

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +326,498 @@ impl Default for SingleInstanceCache {
...
 ///
 /// Delegates to `SessionIdPool::recycle`.
-pub fn recycle_session_id(session_id: u32) {
+fn recycle_session_id(session_id: u32) {
     SessionIdPool::recycle(session_id);
 }
 
-/// RAII guard that recycles a session ID on drop unless disarmed.
+/// RAII token bundling the serialization primitives required to safely
+/// execute an OP-TEE TA operation.
```
wdcui: This sentence needs fix.

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +326,498 @@ impl Default for SingleInstanceCache {
...
-    /// Coordination state for concurrent instance creation.
-    creation_state: SpinMutex<CreationState>,
+    /// Number of instances currently being created (not yet registered).
+    /// Added to [`SessionManager::instance_count`] for the capacity check
+    /// in [`SessionManager::with_ta`] so two concurrent loads cannot both
+    /// pass the limit before either registers.
+    pending_count: SpinMutex<usize>,
     /// Cached TA flags by UUID, populated on first successful session registration.
+    ///
+    /// TODO: a TA's flags (in particular single- vs multi-instance) can
```
wdcui, line 442: Log an issue for this?

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +326,498 @@ impl Default for SingleInstanceCache {
...
+    /// insert), we return `EThreadLimit` so the Linux driver retries.
+    fn try_acquire_for_session(
+        &self,
+        session_id: u32,
+    ) -> Result<(SessionToken<'_>, SessionEntry), OpteeSmcReturnCode> {
+        let entry = self
+            .sessions
+            .get_entry(session_id)
+            .ok_or(OpteeSmcReturnCode::EBadCmd)?;
+        let snapshot_uuid = entry.ta_uuid();
```
wdcui: Why is this called `snapshot`?

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +326,498 @@ impl Default for SingleInstanceCache {
...
+        let snapshot_single = entry.ta_flags().is_single_instance();
+
+        if !self.active_sessions.lock().insert(session_id) {
```
wdcui, line 654: Where do we remove this session from active sessions?

litebox_shim_optee/src/session.rs
```diff
@@ -467,78 +829,96 @@ impl SessionManager {
...
-                // hit means another core finished creating the instance for
-                // this UUID; reuse it instead of starting a new load.
-                if let Some(existing) = self.single_instance_cache.get(uuid) {
-                    return Ok(CreationReservation::ExistingSingleInstance(existing));
-                }
+        let mut token = self.try_acquire_for_open(*uuid)?;
+        // Whether we started on the unknown-load path is determined by the
+        // identity of the lock the token carries. Captured before `f` runs
+        // so we know which branch to take in the post-`f` adoption step.
+        let on_unknown_path = token
```
wdcui: Why is it called `unknown_path`?

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +326,498 @@ impl Default for SingleInstanceCache {
...
+    ///
+    /// # Errors
+    /// - `EBusy` if the id pool is exhausted.
+    pub fn try_acquire_open_session_token(&self) -> Result<SessionToken<'_>, OpteeSmcReturnCode> {
+        let session_id = allocate_session_id().ok_or(OpteeSmcReturnCode::EBusy)?;
+        // The id pool's hint+wrap allocator defers reuse of recycled ids,
+        // so a freshly-allocated id can never collide with a marker slot
+        // that's still held by a previous owner.
+        let inserted = self.active_sessions.lock().insert(session_id);
+        debug_assert!(
```
wdcui, line 506: This check is not in release build. Is this expected?

wdcui:
@CvvT, can you also take a look at this PR? It has some subtle usage of locks.

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +326,498 @@ impl Default for SingleInstanceCache {
...
+    /// the session was registered and what flags it had (the latter for
+    /// callers that need to dispatch on `is_single_instance` /
+    /// `is_keep_alive` after removal).
+    pub fn unregister_session(&self, session_id: u32) -> Option<TaFlags> {
         let entry = self.sessions.remove(session_id);
         if entry.is_some() {
             recycle_session_id(session_id);
         }
-        entry
+        entry.map(|e| e.ta_flags())
```
wdcui, line 802: GPT-5.5 reports this issue:

  At session.rs:793, the ID is recycled while the with_session token can still be alive. SessionIdPool::recycle only clears the bitmap bit; it does not  
  know about active_sessions. The underlying IdPool usually avoids immediate reuse because allocation scans from the last hint, but if the recyclable    
  pool is full except for that just-recycled ID, the next allocation can return it while the old active marker still exists.                             
                                                                                                                                                         
  So the real issue is not necessarily that line 793 recycles, but that try_acquire_open_session_token at session.rs:494-498 treats marker collision as  
  debug-only impossible. If the recycled ID is reallocated before the old token drops, inserted == false must be handled in release builds, e.g. recycle 
  the just-allocated ID and return EThreadLimit.

wdcui (commented):
I left some comments and asked Weiteng to take a look at this PR as well.

litebox_shim_optee/src/session.rs
sangho2: it uses page table IDs (i.e., CR3 values).

litebox_runner_lvbs/src/lib.rs
```diff
@@ -751,30 +714,19 @@ fn open_session_new_instance(
 ) -> Result<(), OpteeSmcReturnCode> {
     let ta_bin = find_ta_binary(ta_uuid).ok_or(OpteeSmcReturnCode::ENotAvail)?;
 
-    // Create and switch to new page table
-    let task_pt_id = create_task_page_table()?;
+    // Token is declared before `task_pt_guard` so it drops AFTER it.
+    // Marker only releases once CR3 is back to base. See
+    // `try_acquire_open_session_token` for why.
+    let mut session_token = session_manager().try_acquire_open_session_token()?;
```
wdcui, line 725: GPT-5.5: Missing TA lookup is surfaced as an outer SMC transport error instead of writing an OpenSession command result into msg_args.ret/ret_origin.

litebox_shim_optee/src/session.rs
sangho2: It stands for page table ID. Let me revise it.

wdcui:
GPT-5.5: litebox_runner_lvbs/src/lib.rs:97, VTL1 minimum-size validation is smaller than later fixed-layout assumptions, so intermediate sizes can still panic later.

litebox_shim_optee/src/session.rs
sangho2: Yes. This is a bug that shouldn't happen but worthwhile to keep eye on. Release build shouldn't panic (maybe, better to log if this happens).

litebox_shim_optee/src/session.rs
sangho2: There is no particular reason. Let me change it.

litebox_shim_optee/src/session.rs
sangho2: `SessionToken` has `Drop`.

litebox_shim_optee/src/session.rs
sangho2: Known issue, but we don't immediately recycle session IDs. This could be an issue if we fully utilize the 32-bit ID space.

litebox_shim_optee/src/session.rs
sangho2: This is for handling an unknown TA/UUID which we didn't get its flag (single- vs multi-session) yet. `on_uknown_ta_path` or `on_uknown_uuid_path` might be better.

litebox_runner_lvbs/src/lib.rs
sangho2: Perhaps this is about we should return an (OP-TEE) item not found error instead of an SMC error? Let me re-check OP-TEE code and fix it accordingly.

sangho2:
> GPT-5.5: litebox_runner_lvbs/src/lib.rs:97, VTL1 minimum-size validation is smaller than later fixed-layout assumptions, so intermediate sizes can still panic later.

Fair. However, these VTL1 base address and size are trusted input, so this check itself is somewhat redundant one.

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +326,502 @@ impl Default for SingleInstanceCache {
...
+    /// change across a version update of the same UUID. Key this map by
+    /// `(uuid, version)` — or invalidate on version mismatch — once TA
+    /// versioning is wired through, so a re-loaded TA isn't serialized
+    /// under the old flags.
     known_flags: SpinMutex<HashMap<TeeUuid, TaFlags>>,
+    /// Per-UUID serialization locks for single-instance TA handling.
+    /// Entries are created lazily only for UUIDs that have been observed
+    /// to be single-instance — never for unknown UUIDs whose load might
+    /// fail or turn out to be multi-instance.
+    single_instance_locks: SpinMutex<HashMap<TeeUuid, Arc<SpinMutex<()>>>>,
```
CvvT: Why single instance's lock has to be separate from the instance? It used to be:
```
pub struct SingleInstanceCache {
    inner: SpinMutex<HashMap<TeeUuid, Arc<SpinMutex<TaInstance>>>>,
}
```

litebox_shim_optee/src/session.rs
```diff
@@ -467,78 +833,97 @@ impl SessionManager {
...
+                };
+            return f(target);
+        }
 
+        {
+            let mut pending = self.pending_count.lock();
             // Capacity check including in-flight creations.
-            let total = self.instance_count() + state.pending_count;
-            if total >= MAX_TA_INSTANCES {
+            if self.instance_count() + *pending >= MAX_TA_INSTANCES {
```
CvvT, line 896: Instead of using `SpinMutex` for pending_count, could we just use `Atomic`?

Also, instead of counting the total number by enumerating through the SessionMap, why not just have an Atomic `total_count`?

litebox_shim_optee/src/session.rs
sangho2: Only a single-instance TA is locked by a single instance lock. A multi-instance TA is locked by a session ID lock. All these are now managed via `SessionToken`.

litebox_shim_optee/src/session.rs
sangho2: Technically doable, but that means we need to maintain an extra atomic which is redundant with `len()`. I'm ssing mutexes here for consistency.

CvvT: I was thinking having one atomic `total_len` to replace `SpinMutex<pending>` and then we don't need `instance_count`, which should be simpler and slightly performant, but this is a minor improvement. You don't have to do it.

litebox_shim_optee/src/session.rs
CvvT: For sessions, you had 

```
pub struct SessionEntry {
    /// The TA instance (may be shared with other sessions for single-instance TAs).
    pub instance: Arc<SpinMutex<TaInstance>>,
```

I don't understand why lock needs to be separate from `TaInstance`. It feels more intuitive and might be simpler. We could discuss offline.

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +326,502 @@ impl Default for SingleInstanceCache {
...
+            active_session_id: Some(session_id),
+            owns_id_recycling: true,
+        })
     }
 
-    /// Get a session by ID.
-    pub fn get_session(&self, session_id: u32) -> Option<Arc<SpinMutex<TaInstance>>> {
-        self.sessions.get(session_id)
+    /// Mark every session currently pointing at `instance` as `Dead`. Must
+    /// be paired with [`SessionManager::evict_cached_instance`]
```
CvvT: If it must be paired with `evict_cached_instance` why not call it inside this function?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -696,48 +668,39 @@ fn open_session_single_instance(
...
         return Err(e);
     }
 
-    // Success: register session and disarm the guard (ownership transfers to session map)
-    session_manager().register_session(runner_session_id, instance_arc.clone(), ta_uuid, ta_flags);
-    session_id_guard.disarm();
-
-    drop(instance);
+    // Success: register a sibling session pointing at the existing instance.
+    session_manager().register_sibling_session(runner_session_id, instance)?;
```
CvvT, line 690: This needs to go over the session map again to get `Arc<TaInstance>`, why not just pass `Arc<TaInstance>` directly instead of `&TaInstance` to this function. `OpenSessionTarget::Sibling` can be `Sibling(Arc<TaInstance>)` because `with_ta` already get `Arc<TaInstance>`.

litebox_shim_optee/src/session.rs
```diff
@@ -467,78 +833,97 @@ impl SessionManager {
...
+        // so concurrent `with_ta(other_uuid)` cannot adopt our lock.
+        if result.is_ok() && on_unknown_uuid_path && self.single_instance_cache.get(uuid).is_some()
         {
-            let mut state = self.creation_state.lock();
-            if is_single_instance {
-                state.pending_uuids.remove(uuid);
+            let per_uuid = self.uuid_lock_arc(*uuid);
+            if let Some(old) = token.uuid_lock.replace(per_uuid) {
+                // SAFETY: see `SessionToken::drop` — same invariant.
+                unsafe { old.force_unlock() };
```
CvvT: This looks suspicious because the new `per_uuid` does not call `try_lock`, then when the token drops, it would call `force_unlock`.

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +326,502 @@ impl Default for SingleInstanceCache {
...
+    /// success returns the `Arc` whose forgotten guard is owned by the
+    /// caller — release via `force_unlock` on the returned `Arc`.
+    fn try_acquire_uuid_lock(&self, uuid: TeeUuid) -> Option<Arc<SpinMutex<()>>> {
+        let lock = self.uuid_lock_arc(uuid);
+        let guard = lock.try_lock()?;
+        // The lock now belongs to the SessionToken about to wrap us. Forget
+        // the guard so its `Drop` does not unlock; the token's `Drop` calls
+        // `force_unlock` via the retained `Arc`.
+        core::mem::forget(guard);
+        Some(lock)
```
CvvT: How about `single_instance_locks: SpinMutex<HashMap<TeeUuid, Option<SpinMutex<()>>>>`, and here we take the lock from `single_instance_locks` and return `guard`; in this way we don't need to use `forget` and `force_unlock`.

CvvT: To distinguish it from unknown uuid lock, we can define it as `single_instance_locks: SpinMutex<HashMap<TeeUuid, Option<SpinMutex<(boolean: is_unknown)>>>>`.

litebox_shim_optee/src/session.rs
sangho2: Issue #938

litebox_shim_optee/src/session.rs
sangho2: Let me revise the doc. in short, this pairing is conditional.

litebox_runner_lvbs/src/lib.rs
sangho2: using `Arc<TaInstance>` here can suffer from a lifetime issue because it can outlive `with_ta`.

litebox_shim_optee/src/session.rs
sangho2: Revised based on our offline discussion.

litebox_shim_optee/src/session.rs
sangho2: Revised based on our offline discussion.

litebox_shim_optee/src/session.rs
```diff
@@ -277,185 +327,501 @@ impl Default for SingleInstanceCache {
...
+        } else {
+            *held = true;
+            Some(HeldUuidLock::SingleInstance(uuid))
+        }
+    }
+
+    fn release_uuid_lock(&self, lock: HeldUuidLock) {
+        match lock {
+            HeldUuidLock::SingleInstance(uuid) => {
+                if let Some(held) = self.single_instance_locks.lock().get_mut(&uuid) {
```
CvvT, line 560: One minor issue: per-uuid lock is never removed from the map.

CvvT (approved):
LGTM, only one minor comment left. Thanks!

litebox_shim_optee/src/session.rs
sangho2: Yes. I'm still debating with myself: performance vs memory usage. Technically, this is a cache and normally we only work with a few TAs.

PR #867, opened 2026-05-26
Add syscall sendfile

CvvT:
Support syscall [sendfile](https://man7.org/linux/man-pages/man2/sendfile.2.html)


wdcui (approved):
LGTM. Thanks.

PR #869, opened 2026-05-26
Support syscall faccessat/faccessat2

CvvT:
Support syscalls `faccessat` and `faccessat2`, Similar to other `*at` syscalls, it does not support fd relative path yet.

To match the fixed UID/GID in in-mem fs, update runner to use the same UID/GID.


wdcui (approved):
LGTM. Thanks.

PR #873, opened 2026-05-28
Refactor pipe code in the Linux shim

wdcui:
This PR refactors the pipe related code in the Linux shim by moving related code into the newly-created pipe.rs file.


CvvT (approved):
LGTM, thanks!

PR #880, opened 2026-05-29
Initial broker design and implementation

wdcui:
This PR implements the broker based on a new design and adds the end-to-end implementation for the broker to provide the support for non-blocking evenfd.


litebox_broker_protocol/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+pub use error::ErrorCode;
+pub use event::{
+    AddEventRequest, AddEventResponse, ConsumeEventRequest, ConsumeEventResponse,
+    CreateEventRequest, CreateEventResponse, EventConsumeMode, EventConsumption, ReadinessState,
+    WaitEventRequest, WaitEventResponse, WaitOutcome,
+};
+pub use message::{
+    BrokerRequest, BrokerResponse, CoreRequest, CoreResponse, EventRequest, EventResponse,
+};
+pub use object::{ObjectHandle, ObjectReferenceGeneration, ObjectReferenceId};
```
jaybosamiya-ms, line 35: A large number of `pub use` make it difficult to follow things along, basically a bunch of modularity gets lost if everything is re-exposed publicly

litebox_broker_protocol/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// Returns whether this requested version is supported by `supported`.
+    ///
+    /// Minor revisions are backward-compatible within a major version, so a
+    /// broker can serve a peer requesting the same major version and a minor
+    /// version no newer than the broker supports.
+    pub const fn is_supported_by(self, supported: Self) -> bool {
+        self.major == supported.major && self.minor <= supported.minor
+    }
```
jaybosamiya-ms, line 59: Rather than `supported: Self`, prefer `broker: Self`, that way it is at least documenting which side.  

More broadly though, I am also not entirely sure what scenario we might wish to support a minor version bump on the broker side, but _not_ on the guest side.

litebox_broker_protocol/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/// Initial broker protocol version implemented by the split-broker POC.
+pub const INITIAL_PROTOCOL_VERSION: ProtocolVersion = ProtocolVersion::new(0, 1);
```
jaybosamiya-ms, line 63: Not sure why this constant exists; is it meant to be "current protocol version"?

litebox_broker_protocol/src/message.rs
```diff
@@ -0,0 +1,101 @@
...
+pub enum EventRequest {
+    /// Create a broker-owned event object.
+    Create(CreateEventRequest),
+    /// Check whether an event wait would complete now.
+    Wait(WaitEventRequest),
+    /// Add readiness credits to an event.
+    Add(AddEventRequest),
+    /// Consume readiness credits from an event.
+    Consume(ConsumeEventRequest),
+}
```
jaybosamiya-ms, line 47: I'm not immediately seeing the benefits of this 3 layer indirection

litebox_broker_protocol/src/message.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+use crate::ProtocolVersion;
+use crate::{
+    AddEventRequest, AddEventResponse, ConsumeEventRequest, ConsumeEventResponse,
+    CreateEventRequest, CreateEventResponse, ErrorCode, WaitEventRequest, WaitEventResponse,
+};
```
jaybosamiya-ms, line 8: this is the sort of stuff that becomes unreadable / unfollowable when the crate top becomes `pub use everything`

agents _really_ prefer making everything `pub` and everything just into a big top level thing, we need to steer them away from that :)

litebox_broker_protocol/src/object.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Creates an object reference identifier from its raw protocol value.
+    pub const fn new(raw: u64) -> Self {
+        Self(raw)
+    }
+
+    /// Returns the raw protocol value.
+    pub const fn get(self) -> u64 {
+        self.0
+    }
+}
```
jaybosamiya-ms, line 45: If it is just clone/copy/new/get, we can just use `pub struct ObjectReferenceId(pub u64);` and skip the unnecessary methods going in and out of it; similarly _everywhere_ code like this can be cleaned up wherever the methods are not actually gaining anything.  agents really like following the "encapsulation" approach from object oriented programming but to a ridiculous amount, we need to tell them not to make unnecessary getters+setters where a normal field would do.  There _are_ good reasons at times to make getters/setters, this is not it.

litebox_broker_protocol/src/wire.rs
jaybosamiya-ms, line 1: might wish to add a quick todo here stating that we should eventually move to Vest

litebox_broker_protocol/src/wire.rs
```diff
@@ -0,0 +1,314 @@
...
+        )
+    }
+
+    const fn event_request(request: EventRequest) -> BrokerRequest {
+        BrokerRequest::Core(CoreRequest::Event(request))
+    }
+
+    const fn event_response(response: EventResponse) -> BrokerResponse {
+        BrokerResponse::Core(CoreResponse::Event(response))
+    }
```
jaybosamiya-ms, line 313: this is the sort of funny unnecessary indirection hurts readability; the "correct" move would probably be to set up `From` impls _if_ we want to keep the 3-layer indirection, but as I mentioned in my other comment, I am not convinced that the 3 layer indirection is helping readability

litebox_broker_protocol/src/channel.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Broker authority response received from a control channel.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum ReceivedBrokerResponse {
+    /// A response understood by the current protocol crate.
+    Response(BrokerResponse),
+    /// A response emitted by a newer broker and not understood by this process.
+    Unknown,
+}
```
jaybosamiya-ms, line 47: Some of this complexity exists due to the allowance of the minor version.  What we really want to do is to have version checks early that decide ok or not, but after that, everything else should assume that versions are aligned enough.  If something could ever reach a scenario where the peers do not even understand each other's messages and decide to ignore them, that leads to eventual problematic scenarios, since it is _effectively_ equivalent to a designed-in parser differential.

litebox_broker_protocol/src/error.rs
```diff
@@ -0,0 +1,107 @@
...
+    InvalidRights,
+    /// Broker-side resource exhaustion.
+    ResourceExhausted,
+    /// The operation would block in the current event state.
+    WouldBlock,
+    /// Error code emitted by a newer broker and not understood by this local peer.
+    ///
+    /// This variant is reserved for raw codes not assigned by this protocol
+    /// version.
+    Unknown(u16),
```
jaybosamiya-ms, line 38: `Unknown` + `non_exhaustive` is very surprising.  `non_exhaustive` is reasonable; the unknown should not exist (see my other comment about preventing parser differentials)

litebox_broker_protocol/src/error.rs
```diff
@@ -0,0 +1,107 @@
...
+            11 => Self::UnsupportedOperation,
+            12 => Self::Internal,
+            4 => Self::PolicyDenied,
+            5 => Self::UnknownObject,
+            6 => Self::StaleHandle,
+            7 => Self::WrongObjectType,
+            8 => Self::InvalidRights,
+            9 => Self::ResourceExhausted,
+            13 => Self::WouldBlock,
+            raw => Self::Unknown(raw),
```
jaybosamiya-ms, line 63: can we pick a more readable ordering here? this feels like an agent tried to maintain compatibility with some older version even though there is no older version :)

litebox_broker_protocol/src/event.rs
```diff
@@ -0,0 +1,167 @@
...
+impl ReadinessState {
+    /// Creates a readiness state.
+    pub const fn new(read_ready: bool, write_ready: bool, generation: u64) -> Self {
+        Self {
+            read_ready,
+            write_ready,
+            generation,
+        }
+    }
+}
```
jaybosamiya-ms, line 26: this `new` is gonna lead to hard-to-read code because it takes two booleans in, so now every time we see it, we need to remember whether it is read-write or write-read.  The `ReadinessState` struct already has `pub` fields, not entirely sure this `new` is carrying its weight.

litebox_broker_protocol/src/event.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// Result of checking whether a broker event read wait would complete now.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum WaitOutcome {
+    /// The object is read-ready now.
+    Ready(ReadinessState),
+    /// The object is not read-ready; deployment-specific wait plumbing may block.
+    WouldBlock(ReadinessState),
+}
```
jaybosamiya-ms, line 36: Is there a reason this is non-exhaustive?  Is there any possibility between "read ready" and "not read ready"?  The agent seems to have over-used `non_exhaustive` all over the place, would be good to make sure it does not do that; things are exhaustive or not for a reason; overusing `non_exhaustive` causes code to become fragile everywhere the enum is used.

litebox_broker_protocol/src/event.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// Broker-authoritative readiness state for one object.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct ReadinessState {
+    /// Whether an event read/consume operation can complete without blocking.
+    pub read_ready: bool,
+    /// Whether an event write/add operation can complete without blocking.
+    pub write_ready: bool,
+    /// Monotonic readiness generation used to invalidate user-side readiness caches.
+    pub generation: u64,
```
jaybosamiya-ms, line 14: Do we have readiness caches?  I didn't quite see where this is being handled.  Not entirely sure we need to add more complexity when it is not used yet

docs/broker-design.md
jaybosamiya-ms, line 1: discussed IRL: we want to remove these docs before merging; the intention with including docs for any significant design changes is to keep track of the human inputs that went into the change, so for future PRs we should probably have the temporary docs only list the human inputs that went in, so that we have a better understanding of decisions made

docs/impl-plan.md
jaybosamiya-ms, line 1: reminder to drop this file

litebox_broker_core/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+
+/// Resource limits for broker-owned authority state.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct BrokerCoreLimits {
+    /// Maximum live broker objects.
+    pub max_objects: usize,
+    /// Maximum live object references.
+    pub max_references: usize,
+}
```
jaybosamiya-ms, line 47: per-local process, or global numbers?

litebox_broker_core/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+
+fn allocate_id(next_id: &mut u64) -> Result<u64> {
+    if *next_id == EXHAUSTED_ID {
+        return Err(BrokerError::ResourceExhausted);
+    }
+
+    let id = *next_id;
+    *next_id = id.checked_add(1).unwrap_or(EXHAUSTED_ID);
+    Ok(id)
+}
```
jaybosamiya-ms, line 122: I'm surprised by the use of `0` as a sentinel here, unnecessary complexity

litebox_broker_core/src/object.rs
```diff
@@ -0,0 +1,355 @@
...
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub struct ObjectRights(u32);
+
+impl ObjectRights {
+    /// Empty rights set.
+    pub const NONE: Self = Self(0);
+    /// Right to wait for readiness.
+    pub const WAIT: Self = Self(1 << 0);
+    /// Right to mutate object state, such as adding event readiness credits.
+    pub const WRITE: Self = Self(1 << 1);
```
jaybosamiya-ms, line 30: Use `bitflags` rather than spinning up custom equivalent

litebox_broker_core/src/object.rs
```diff
@@ -0,0 +1,355 @@
...
+pub(crate) struct ObjectId(u64);
+
+impl ObjectId {
+    /// Creates an object identifier from its raw value.
+    const fn new(raw: u64) -> Self {
+        Self(raw)
+    }
+}
+
+const FIRST_REFERENCE_GENERATION: ObjectReferenceGeneration = ObjectReferenceGeneration::new(1);
```
jaybosamiya-ms, line 69: I'm not quite sure why this is a constant; almost equivalent to saying `one = 1`, no?

litebox_broker_core/src/object.rs
```diff
@@ -0,0 +1,355 @@
...
+    }
+}
+
+impl BrokerCore {
+    /// Inserts a broker object and mints its first owned reference.
+    ///
+    /// The current POC never reuses reference slots, so the reference
+    /// generation starts at the authority-owned first generation. Any future
+    /// reference-slot reuse path must bump the generation before reissuing a
+    /// slot so stale handles cannot validate against a recycled reference.
```
jaybosamiya-ms, line 104: Rather than doing our own generational allocation, we should just use `slotmap` or similar

litebox_broker_core/src/object.rs
```diff
@@ -0,0 +1,355 @@
...
+
+impl ObjectKind {
+    pub(crate) const fn object_type(self) -> ObjectType {
+        match self {
+            Self::Event(_) => ObjectType::Event,
+        }
+    }
+}
+
+impl BrokerCore {
```
jaybosamiya-ms, line 98: I'm kinda surprised that these methods are separated from the actual definition of `BrokerCore`; there are some good reasons to do that sometimes, but it kinda diffuses responsibility and makes it harder to tell what the interface of a particular struct really is

litebox_broker_core/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+/// lifetime. Constructors return [`BrokerError::BrokerCoreAlreadyExists`] if a
+/// core has already been constructed.
+pub struct BrokerCore {
+    policy: PolicyEngine,
+    limits: BrokerCoreLimits,
+    next_process_id: u64,
+    next_object_id: u64,
+    next_reference_id: u64,
+    objects: BTreeMap<ObjectId, ObjectEntry>,
+    references: BTreeMap<ObjectReferenceId, ObjectReference>,
```
jaybosamiya-ms, line 83: Any reason for the `BTreeMap` rather than the (usually) faster `HashMap`?

litebox_broker_core/src/policy.rs
jaybosamiya-ms, line 1: We probably should explicitly document that the policy engine itself is a placeholder for now; a proper design of it is out of scope for this PR, but we'll need to come up with a good design for it at some point

litebox_broker_core/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+            Self::ResourceExhausted => f.write_str("broker resource exhausted"),
+            Self::BrokerCoreAlreadyExists => f.write_str("broker core already exists"),
+            Self::WouldBlock => f.write_str("broker operation would block"),
+            Self::UnsupportedOperation => f.write_str("unsupported broker operation"),
+            Self::InvalidPolicyDecision => f.write_str("invalid broker policy decision"),
+        }
+    }
+}
+
+impl core::error::Error for BrokerError {}
```
jaybosamiya-ms, line 49: Prefer using `thiserror` rather than manually written impls over error enums

litebox_broker_core/src/identity.rs
```diff
@@ -0,0 +1,71 @@
...
+///
+/// The first userland proof of concept does not authenticate Unix-socket peers,
+/// but BrokerCore still accepts an explicit credential value so authenticated
+/// servers or hosts can plumb identity through the same association-creation seam.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[non_exhaustive]
+pub enum CallerCredential {
+    /// Explicit deployment mode for the initial unauthenticated userland POC.
+    Unauthenticated,
+}
```
jaybosamiya-ms, line 16: Rather than unauthenticated, we should probably say "host-guaranteed", because in some sense, the unix socket is guaranteed by the underlying host to be a good enough auth, right?

litebox_broker_local/src/error.rs
```diff
@@ -0,0 +1,127 @@
...
+/// Errors returned by the broker-local control adapter.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum BrokerLocalError<E> {
+    /// The control channel failed.
+    Channel(E),
+    /// An operation requiring an active broker session was called before negotiation.
+    NotNegotiated,
+    /// Negotiation was requested after the local adapter was already active.
+    AlreadyNegotiated,
```
jaybosamiya-ms, line 17: Errors like this should be disallowed entirely by design, rather than having to be handled everywhere.  If for example, something never attempts the negotiation, it is totally unrecoverable, and a panic is more useful than trying to handle it.

litebox_broker_local/src/error.rs
```diff
@@ -0,0 +1,127 @@
...
+        /// Effective protocol version negotiated for this connection.
+        negotiated_protocol_version: ProtocolVersion,
+    },
+    /// The broker does not support the requested protocol version.
+    UnsupportedVersion {
+        /// Protocol version requested by this local adapter.
+        requested: ProtocolVersion,
+        /// Protocol version advertised by the broker.
+        broker_protocol_version: ProtocolVersion,
+    },
```
jaybosamiya-ms, line 49: all of these should probably fold away, and are unnecessary complexity

litebox_broker_local/src/event.rs
```diff
@@ -0,0 +1,101 @@
...
+
+use crate::{BrokerLocal, BrokerLocalError, Result};
+
+const EVENT_PROTOCOL_VERSION: ProtocolVersion = INITIAL_PROTOCOL_VERSION;
+
+impl<T: LocalControlChannel> BrokerLocal<T> {
+    /// Creates a broker-owned event object.
+    pub fn create_event(&mut self) -> Result<ObjectHandle, T::Error> {
+        self.create_event_with_count(0)
+    }
```
jaybosamiya-ms, line 19: not sure why this helper is here; the agent seems to add quite a lot of unnecessary indirections in various places that do something _very_ minor, but now require extra reasoning about the API surface

litebox_broker_local/src/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+        Self {
+            channel,
+            state: ConnectionState::AwaitingNegotiation,
+        }
+    }
+
+    /// Returns the underlying control channel for deployment-specific configuration.
+    pub fn control_channel_mut(&mut self) -> &mut T {
+        &mut self.channel
+    }
```
jaybosamiya-ms, line 54: Can just make a `pub channel` at the struct instead of this

litebox_broker_host/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+        BrokerError::UnknownObject => ErrorCode::UnknownObject,
+        BrokerError::StaleHandle => ErrorCode::StaleHandle,
+        BrokerError::WrongObjectType => ErrorCode::WrongObjectType,
+        BrokerError::InvalidRights => ErrorCode::InvalidRights,
+        BrokerError::ResourceExhausted => ErrorCode::ResourceExhausted,
+        BrokerError::WouldBlock => ErrorCode::WouldBlock,
+        BrokerError::UnsupportedOperation => ErrorCode::UnsupportedOperation,
+        _ => ErrorCode::Internal,
+    }
+}
```
jaybosamiya-ms, line 237: Agents really prefer to make such functions for errors rather than use `From<...>` machinery

litebox_broker_host/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+
+fn handle_core_result<T>(
+    result: litebox_broker_core::Result<T>,
+    into_response: impl FnOnce(T) -> BrokerResponse,
+) -> BrokerResponse {
+    match result {
+        Ok(value) => into_response(value),
+        Err(error) => BrokerResponse::Error(to_protocol_error(error)),
+    }
+}
```
jaybosamiya-ms, line 219: This is a surprisingly heavy-handed helper; not sure it helps readability at its callers

litebox/src/litebox.rs
```diff
@@ -30,6 +34,35 @@ impl<Platform: RawSyncPrimitivesProvider> LiteBox<Platform> {
...
+        broker_local: BrokerLocal<T>,
+    ) -> Self
+    where
+        T: LocalControlChannel + Send + 'static,
+    {
+        Self::new_inner(
+            platform,
+            Some(broker::control_from_local::<Platform, T>(broker_local)),
+        )
+    }
```
jaybosamiya-ms, line 60: do we need all 3?  what is the use case for each?

litebox/src/broker/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+/// LiteBox owns broker-backed local objects and constructs broker protocol
+/// requests. Deployment code owns endpoint selection and supplies the connected
+/// transport behind this protocol-level boundary.
+pub trait BrokerControl: Send + Sync {
+    /// Sends one active BrokerCore request and returns its response.
+    fn request(
+        &self,
+        request: CoreRequest,
+    ) -> core::result::Result<CoreResponse, BrokerControlError>;
+}
```
jaybosamiya-ms, line 25: This interfaces forces a blocking design, should document that we want to move away from this, at minimum

litebox/src/broker/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+
+struct BrokerLocalControl<Platform: RawSyncPrimitivesProvider, T> {
+    local: Mutex<Platform, BrokerLocal<T>>,
+}
```
jaybosamiya-ms, line 29: Prefer descriptive names like `Channel` rather than generic `T`, whenever it is going to be used via some sort of trait restriction.

Additionally, we generally prefer putting the trait restriction up-front at the struct itself when basically all its usability is gated behind it, it forces acknowledgement at the type itself, rather than diffusing obligations.

litebox/src/broker/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+pub(crate) fn control_from_local<Platform, T>(local: BrokerLocal<T>) -> Arc<dyn BrokerControl>
+where
+    Platform: RawSyncPrimitivesProvider,
+    T: LocalControlChannel + Send + 'static,
+{
+    Arc::new(BrokerLocalControl::<Platform, T>::new(local))
+}
```
jaybosamiya-ms, line 72: Is this helper function carrying its weight?  Why not inline it?

litebox/src/event/counter.rs
```diff
@@ -0,0 +1,191 @@
...
+            handle: response.handle,
+            pollee: Pollee::new(),
+            blocking_operations_supported: true,
+        })
+    }
+
+    /// Returns whether blocking reads and writes are supported.
+    pub fn supports_blocking_operations(&self) -> bool {
+        self.blocking_operations_supported
+    }
```
jaybosamiya-ms, line 83: I have not fully read the code for whether blocking/non-blocking is handled right, but from an interface design, it is kinda surprising that `new` _always_ sets it to true; if that's the case, why would anything above even care to check whether blocking is supported or not?

litebox/src/broker/error.rs
jaybosamiya-ms, line 1: `thiserror` at all error things

litebox/src/broker/error.rs
```diff
@@ -0,0 +1,85 @@
...
+/// Error returned by the deployment-provided broker control path.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum BrokerControlError {
+    /// The broker control transport failed.
+    Transport,
+    /// The broker returned an operation error.
+    Broker(ErrorCode),
+    /// The broker returned a response shape that does not match the request.
+    UnexpectedResponse,
```
jaybosamiya-ms, line 17: Is there any recovery mechanism if this happens?  I can't think of any sensible recovery mechanism on this side, so the only option is to panic, no?

litebox/src/broker/error.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// The broker rejected the cached object handle, type, or rights.
+    InvalidObject,
+    /// The object operation would block in its current broker-side state.
+    WouldBlock,
+    /// The object or broker-side state cannot grow further.
+    ResourceExhausted,
+    /// The broker returned a response shape that does not match the request.
+    UnexpectedResponse,
+    /// The broker reported a non-recoverable or unsupported object error.
+    Internal,
```
jaybosamiya-ms, line 37: similar

litebox_broker_userland/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+use litebox_broker_transport::unix_socket::UnixStreamHostControlChannel;
+
+const SESSION_TIMEOUT: Duration = Duration::from_secs(5);
+
+#[derive(Parser, Debug)]
+struct CliArgs {
+    /// Broker Unix socket path to bind.
+    #[arg(long, value_name = "PATH", value_hint = clap::ValueHint::FilePath)]
+    socket: PathBuf,
+}
```
jaybosamiya-ms, line 21: I don't think the broker should bind to a particular path, instead it should choose a path and make the local runner side bind to it, no?  In some sense, the broker should be in charge of spinning up the local runner, right?

litebox_broker_userland/tests/userland_broker.rs
```diff
@@ -0,0 +1,139 @@
...
+fn unique_socket_path() -> PathBuf {
+    let now = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_nanos();
+    env::temp_dir().join(format!(
+        "litebox-broker-userland-{}-{now}.sock",
+        std::process::id()
+    ))
+}
```
jaybosamiya-ms, line 139: this is used only for the test, it should sit within the text stuff

litebox_broker_userland/tests/userland_broker.rs
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+use std::env;
+use std::fs;
+use std::io;
```
jaybosamiya-ms, line 6: Prefer never directly importing modules; imports should be types, traits, consts, and macros.  It makes readability quite harder.  There are sometimes good reasons to import modules, but `std::env` is never a good reason.

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -481,6 +481,75 @@ impl LinuxUserland {
                     .unwrap(),
                 ],
             ),
+            // Broker control-channel I/O runs through a host Unix socket in the
+            // current POC. The transport refreshes read/write timeouts around
+            // each request.
```
jaybosamiya-ms, line 486: This seems like an unnecessary weakening of our seccomp filters.  I'm not quite sure why read/write timeouts are needed if the broker is healthy

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -894,6 +895,78 @@ impl<FS: ShimFS> Task<FS> {
         self.check_raw_fd_exists(fd)?;
         check_iovcnt(iovcnt)?;
         let iovs: &[IoReadVec<MutPtr<u8>>] = &iovec.to_owned_slice(iovcnt).ok_or(Errno::EFAULT)?;
+        let raw_fd = usize::try_from(fd).map_err(|_| Errno::EBADF)?;
+        {
```
jaybosamiya-ms: the sys_readv changes do not actually seem relevant to the other changes in this PR?  similarly, writev

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -935,16 +1008,98 @@ fn check_iovcnt(iovcnt: usize) -> Result<(), Errno> {
...
         if total > SSIZE_MAX {
             return Err(Errno::EINVAL);
         }
     }
+    Ok(total)
+}
+
+fn eventfd_iovec_total_len(iov_lens: impl IntoIterator<Item = usize>) -> Result<usize, Errno> {
+    iovec_total_len(iov_lens)
+}
```
jaybosamiya-ms: I'm not sure what this sort of indirection is actually gaining us

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -935,16 +1008,98 @@ fn check_iovcnt(iovcnt: usize) -> Result<(), Errno> {
...
+
+fn validate_eventfd_iovec_len(total_len: usize) -> Result<(), Errno> {
+    if total_len < size_of::<u64>() {
+        return Err(Errno::EINVAL);
+    }
     Ok(())
 }
```
jaybosamiya-ms: does this need to be its own function?  can't it be just where it is needed?  seems unnecessary indirection

litebox_shim_linux/src/syscalls/file.rs
jaybosamiya-ms, line 1: Basically a ton of readv/writev complexity that is going and touching systems that I'd consider not relevant to the changes being made in this PR.  I tried looking at why the changes are needed, but seems mostly just that _incidentally_ the test C file had readv/writev in it, and that suddenly forced the agent to go and start making other bits complicated?  Can we back out the readv/writev changes since they are not relevant to the core initial broker design (as I understand it).

jaybosamiya-ms (commented):
Bunch of comments all throughout the place about lower-level details.  As discussed IRL, the plan is to merge this almost as-is (modulo removing the `docs/` dir); a lot of the other comments are mostly to be read right now, but not really change the PR; future PRs will be doing the necessary cleanup.  As a 1st order approximation, this current PR is directionally correct, but the agents have left a bunch of code smells that we do need to clean up to keep the project healthy; that's what most of the suggestions are focusing on.

On major bit (other than the `docs/`) that we should try to remove from this PR before merging, if possible, is the readv/writev changes that are mostly irrelevant to this PR at least as I understand it.

If one of my comments mentions just adding a tiny TODO or such _trivially_ tweaked thing, might be good to just resolve that up front rn; but anything more significant than that probably should go into its own PR.

I'm making a tracking issue for the changes needed btw (https://github.com/microsoft/litebox/issues/911)

litebox/src/event/counter.rs
wdcui: `blocking_operations_supported` is removed.

litebox/src/litebox.rs
wdcui: I removed one.

litebox_broker_core/src/lib.rs
wdcui: Changed to HashMap.

litebox_broker_core/src/lib.rs
wdcui: Removed.

litebox_broker_core/src/object.rs
wdcui: Changed to bitflags.

litebox_broker_core/src/object.rs
wdcui: Removed.

litebox_broker_core/src/object.rs
wdcui: Changed to slotmap.

litebox_shim_linux/src/syscalls/file.rs
wdcui: Removed.

litebox/src/broker/error.rs
wdcui: fixed.

litebox_broker_core/src/error.rs
wdcui: fixed.

litebox_broker_core/src/lib.rs
wdcui: Global, not per-process.

litebox_broker_protocol/src/lib.rs
wdcui: Resolved in PR #944.

litebox_broker_protocol/src/object.rs
wdcui: Resolved.

litebox_broker_protocol/src/wire.rs
wdcui: Track it in Issue #946.

litebox_broker_protocol/src/event.rs
wdcui: Removed.

litebox_broker_core/src/object.rs
wdcui: Resolved in PR #944.

litebox_broker_core/src/identity.rs
wdcui: This file is removed in PR #944.

litebox_broker_local/src/error.rs
wdcui: They are removed in #944.

litebox_broker_protocol/src/lib.rs
wdcui: Removed in PR #944.

litebox_broker_protocol/src/channel.rs
wdcui: Resolved in PR #947.

litebox_broker_protocol/src/error.rs
wdcui: Resolved in PR #949.

litebox_broker_protocol/src/event.rs
wdcui: Resolved in PR #950.

litebox_broker_protocol/src/error.rs
wdcui: Resolved in PR #951.

litebox_broker_protocol/src/event.rs
wdcui: Resolved in PR #951.

litebox_broker_local/src/error.rs
wdcui: Resolved in PR #952.

litebox_broker_host/src/lib.rs
wdcui: Resolved in PR #953.

litebox_broker_host/src/lib.rs
wdcui: Resolved in PR #954.

litebox/src/broker/mod.rs
wdcui: Resolved in PR #955.

litebox/src/broker/error.rs
wdcui: Resolved in PR #956.

litebox/src/broker/error.rs
wdcui: Resolved in PR #956.

litebox_broker_userland/src/main.rs
wdcui: Resolved in PR #957.

litebox_broker_userland/tests/userland_broker.rs
wdcui: Removed in PR #957.

litebox_broker_userland/tests/userland_broker.rs
wdcui: Resolved in PR #959.

litebox_platform_linux_userland/src/lib.rs
wdcui: Resolved in PR #960.

litebox_broker_protocol/src/lib.rs
wdcui: Resolved in PR #962.

litebox_broker_protocol/src/message.rs
wdcui: Resolved in PR #962.

litebox_broker_protocol/src/message.rs
wdcui: Resolved in PR #965.

litebox_broker_protocol/src/wire.rs
wdcui: Resolved in PR #965.

litebox_broker_local/src/event.rs
wdcui: Removed in PR #965.

litebox_broker_local/src/lib.rs
wdcui: This function is removed in PR #969. I will keep `channel` private for now.

litebox_broker_core/src/policy.rs
wdcui: Resolved in PR #970.

litebox/src/broker/mod.rs
wdcui: Resolved in PR #970.

PR #885, opened 2026-05-29
Add per-session OP-TEE client identity handling

sangho2:
This PR adds support for handling OP-TEE client identity per session not per instance. Default client identity is now `Public` not `User`. It also does some additional sanity checks.

### DIFF
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -684,6 +684,9 @@ fn open_session_single_instance(
                 teardown_ta_page_table(instance.shim(), task_pt_id);
             };
         } else {
+            // The session id is forgotten (never recycled), so the token's drop
+            // won't clear the recorded identity. Remove the client identity here.
+            session_manager().clear_session_client_identity(runner_session_id);
             session_token.disarm();
         }
         return Err(e);
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -42,9 +42,8 @@ use litebox_shim_optee::msg_handler::{
     decode_ta_request, handle_optee_msg_args, handle_optee_smc_args, packed_msg_args_lock,
     update_optee_msg_args,
 };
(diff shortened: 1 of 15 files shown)


sangho2:
We should merge #865 first because this PR should have one more fix: per-session `ClientIdenty` rather than per-instance `ClientIdentity`. #865 heavily refactored related data structures and functions.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -681,6 +679,9 @@ fn open_session_single_instance(
                 teardown_ta_page_table(instance.shim(), task_pt_id);
             };
         } else {
+            // The session id is forgotten (never recycled), so the token's drop
+            // won't clear the recorded identity. Remove the client identity here.
+            session_manager().clear_session_client_identity(runner_session_id);
```
wdcui, line 689: Do we need to call clear_session_client_identity in line 901 as well?

litebox_shim_optee/src/session.rs
```diff
@@ -790,12 +814,37 @@ impl SessionManager {
...
+        self.session_client_identities.lock().remove(&session_id);
+    }
+
     /// Unregister a session and recycle its session ID. Returns whether
     /// the session was registered and what flags it had (the latter for
     /// callers that need to dispatch on `is_single_instance` /
     /// `is_keep_alive` after removal).
     pub fn unregister_session(&self, session_id: u32) -> Option<TaFlags> {
         let entry = self.sessions.remove(session_id);
+        self.session_client_identities.lock().remove(&session_id);
```
wdcui: Call `clear_session_client_identity`?

wdcui (approved):
LGTM. Thanks.

I left some comments for your reference.

litebox_runner_lvbs/src/lib.rs
sangho2: `SessionToken`'s `Drop` handles it in a legitimate case. This path is an exception (a normal world memory error).

litebox_shim_optee/src/session.rs
sangho2: I misunderstood the comment. Let me do that.

PR #891, opened 2026-06-02
fix(loader): page-align head/tail munmap in MapMemory::reserve

willportnoy:
**fix(loader): page-align head/tail munmap in `MapMemory::reserve`**

`MapMemory::reserve` allocates an over-sized `PROT_NONE` mapping and trims the unaligned head + tail with `munmap` so the caller gets a block aligned to `align`. The tail `munmap` was passed `(aligned_ptr + len, mapping_end - end)` directly — but `len` (an ELF's `max_vaddr - min_vaddr` span) is in general **not** page-aligned, and `munmap` rejects non-page-aligned start addresses with `EINVAL`. This surfaced as `execve` → `ENOEXEC` (`exit_code: 127`) on every guest `fork+exec` of binaries whose PT_LOAD span ends mid-page — e.g. prebuilt linux-x64 node.js (`0x6403D68`).

**Fix:** factor the trim arithmetic into a pure helper `compute_reserve_regions` in `litebox_common_linux::loader` that rounds the tail start up to `PAGE_SIZE` and the mapping end up to the kernel's actual page-rounded extent. Both `litebox_shim_linux` and `litebox_shim_optee` now call the helper (the bug was duplicated in both shims).

Covered by new unit tests in `reserve_regions_tests` (regression case pins the EINVAL pattern).

### DIFF
--- a/litebox_common_linux/src/loader.rs
+++ b/litebox_common_linux/src/loader.rs
@@ -622,12 +622,12 @@ pub trait MapMemory {
 /// The result of computing the head/tail trim regions for an over-sized
 /// anonymous reservation made by [`MapMemory::reserve`].
 ///
-/// See [`compute_reserve_regions`] for details.
+/// See [`compute_reserved_regions`] for details.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct ReservedRegions {
     /// Base address of the requested `len` bytes inside the over-sized
     /// reservation, aligned up to the `align` argument passed to
-    /// [`compute_reserve_regions`].
+    /// [`compute_reserved_regions`].
     pub aligned_ptr: usize,
     /// `(start, len)` of the page-aligned head slice that should be
     /// released with `munmap`, or `None` if no head trim is needed.
@@ -662,7 +662,7 @@ pub struct ReservedRegions {
 /// PT_LOAD span of `0x6403D68`), the kernel rejected the `munmap` with
 /// `EINVAL`, surfacing as `execve` → `ENOEXEC` for any guest fork+exec
 /// of node.
-pub fn compute_reserve_regions(
+pub fn compute_reserved_regions(
     mapping_ptr: usize,
     mapping_len: usize,
     len: usize,
@@ -767,7 +767,7 @@ impl Protection {
 #[cfg(test)]
 mod reserve_regions_tests {
     extern crate std;
-    use super::{PAGE_SIZE, ReservedRegions, compute_reserve_regions};
+    use super::{PAGE_SIZE, ReservedRegions, compute_reserved_regions};
 
     /// The exact non-page-aligned PT_LOAD span observed for the prebuilt
     /// linux-x64 node.js binary in the `litebox-test` Docker image, which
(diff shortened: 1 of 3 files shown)


litebox_common_linux/src/loader.rs
```diff
@@ -619,6 +619,82 @@ pub trait MapMemory {
...
 }
 
+/// The result of computing the head/tail trim regions for an over-sized
+/// anonymous reservation made by [`MapMemory::reserve`].
+///
+/// See [`compute_reserve_regions`] for details.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct ReserveRegions {
+    /// Page- and `align`-aligned base address of the requested `len` bytes
+    /// inside the over-sized reservation.
```
wdcui: Where is `align` provided?

litebox_common_linux/src/loader.rs
```diff
@@ -619,6 +619,82 @@ pub trait MapMemory {
...
+    pub aligned_ptr: usize,
+    /// `(start, len)` of the page-aligned head slice that should be returned
```
wdcui: Why should these regions be returned to the kernel? Should we use `kernel` or a more general term?

litebox_common_linux/src/loader.rs
```diff
@@ -619,6 +619,82 @@ pub trait MapMemory {
     -> Result<(), Self::Error>;
 }
 
+/// The result of computing the head/tail trim regions for an over-sized
+/// anonymous reservation made by [`MapMemory::reserve`].
+///
+/// See [`compute_reserve_regions`] for details.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct ReserveRegions {
```
wdcui: `ReservedRegions`?

wdcui (approved):
This PR looks good to me. I left some comments. Please update the PR description to shorten it because it will be included in the commit history.

wdcui:
@CvvT can you also take a look at this PR? Thanks!

litebox_common_linux/src/loader.rs
```diff
@@ -686,3 +762,144 @@ impl Protection {
         flags
     }
 }
+
+#[cfg(test)]
```
wdcui, line 767: We probably just need the first two tests?

CvvT (approved):
LGTM, thanks for the fix!

litebox_common_linux/src/loader.rs
```diff
@@ -619,6 +619,83 @@ pub trait MapMemory {
...
+///   slice is computed in page units: release everything from the first
+///   page strictly after `aligned_ptr + len` to that page-aligned end.
+///
+/// Prior to this helper, callers used `(aligned_ptr + len, mapping_end -
+/// (aligned_ptr + len))` directly as the tail `munmap` args. Whenever
+/// `len` ended mid-page (e.g. node.js's prebuilt linux-x64 binary has a
+/// PT_LOAD span of `0x6403D68`), the kernel rejected the `munmap` with
+/// `EINVAL`, surfacing as `execve` → `ENOEXEC` for any guest fork+exec
+/// of node.
+pub fn compute_reserve_regions(
```
wdcui: `compute_reserved_regisions`?

PR #896, opened 2026-06-03
Harden ldelf/TA ELF loading

sangho2:
This PR hardens ldelf/TA ELF loading by checking whether `.ta_head` is correct and whether a TA entry point reported by ldelf is within TA's executable ranges.

### DIFF
--- a/litebox_shim_optee/src/syscalls/ldelf.rs
+++ b/litebox_shim_optee/src/syscalls/ldelf.rs
@@ -250,17 +250,21 @@ impl Task {
         } else if flags.contains(LdelfMapFlags::LDELF_MAP_FLAG_EXECUTABLE) {
             prot |= ProtFlags::PROT_EXEC;
         }
+        let region_start = align_down(padded_start, PAGE_SIZE);
+        let region_len = (num_bytes + padded_start - region_start).next_multiple_of(PAGE_SIZE);
+        let region_is_executable = prot.contains(ProtFlags::PROT_EXEC);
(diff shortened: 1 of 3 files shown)


sangho2:
merge #814 first.

litebox_shim_optee/src/syscalls/ldelf.rs
```diff
@@ -250,19 +250,23 @@ impl Task {
         } else if flags.contains(LdelfMapFlags::LDELF_MAP_FLAG_EXECUTABLE) {
             prot |= ProtFlags::PROT_EXEC;
         }
+        let region_start = align_down(padded_start, PAGE_SIZE);
+        let region_len = (num_bytes + padded_start - region_start).next_multiple_of(PAGE_SIZE);
```
wdcui, line 254: Do we need to use checked_add here?

sangho2: We don't have to. We already use checked arithmetic to calculate `total_size` in Lines 175-178, which is greater than or equal to `region_len`.

sangho2:
I mentioned this to Weidong IRL. I checked Linux kernel source code, `fs/binfmt_elf.c:load_elf_binary()`, and confirmed that it just uses the `BAD_ADDR` macro against `elf_entry` --- it only checks whether that value is greater than or equal to `TASK_SIZE`. The LVBS kernel platform already does it, so we don't need this PR.

PR #898, opened 2026-06-03
Add syscall mkdirat

CvvT:
Add support for syscall `mkdirat` but ignore `dirfd` (i.e., fd relative path).


wdcui (approved):
LGTM. Thanks!

PR #900, opened 2026-06-03
Add minimal AArch64 support (scaffolding)

sangho2:
This PR introduces minimal AArch64 support to the LiteBox core and Linux common crates. It includes essential architecture-specific data structures and data fields as well as fallible memory copy assembly blocks. It also updates CI to build and test the two crates with an AArch64 host.

### DIFF
--- a/litebox/src/platform/arch.rs
+++ b/litebox/src/platform/arch.rs
@@ -3,13 +3,12 @@
 
 //! Architecture-specific platform interfaces.
 //!
-//! As it currently stands, the interfaces here are only considered for x86-64, in the future other
+//! As it currently stands, the interfaces here are mostly considered for x86-64, in the future other
 //! architectures might be supported.
 
 use thiserror::Error;
 
 /// A provider of architecture-specific functionality.
-#[cfg(target_arch = "x86_64")]
 pub trait ArchSpecificProvider {
     /// Get the architecture-specific `reg`, for the current guest context.
     ///
@@ -44,6 +43,11 @@ pub enum ArchSpecificRegister {
     FsBase,
     GsBase,
 }
+
+/// Architecture-specific registers for AArch64.
+#[cfg(target_arch = "aarch64")]
+#[non_exhaustive]
+pub enum ArchSpecificRegister {}
 
 /// Errors that can be produced by a [`ArchSpecificProvider`] operation.
 #[non_exhaustive]
--- a/litebox/src/platform/arch.rs
+++ b/litebox/src/platform/arch.rs
@@ -3,8 +3,8 @@
 
 //! Architecture-specific platform interfaces.
 //!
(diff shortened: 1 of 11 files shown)


sangho2:
merging #806 before this might be better. indirectly related (i.e., `ArchSpecificProvider`).

wdcui:
GPT-5.5 reported the following issues:

   1. litebox/src/mm/exception_table.rs:101 — High: AArch64 inline asm uses label 10: / 10b, which can trigger Rust’s denied binary-asm-labels lint. Use a
  non-binary-looking label like 22: / 22b.
   2. litebox_common_linux/src/lib.rs:907 — Medium: AArch64 suseconds_t is aliased to u64, but Linux defines it as signed long/i64. Negative tv_usec values
   can be misinterpreted. Use i64 and validate range explicitly.
   3. litebox_common_linux/src/lib.rs:2584 — Medium: AArch64 disables legacy mkdir but does not decode mkdirat, so directory creation on AArch64 likely
  falls through to ENOSYS. Add a mkdirat decode path with the intended dirfd semantics.

wdcui (approved):
LGTM. Thanks!

sangho2:
> GPT-5.5 reported the following issues:
> 
> 1. litebox/src/mm/exception_table.rs:101 — High: AArch64 inline asm uses label 10: / 10b, which can trigger Rust’s denied binary-asm-labels lint. Use a
>    non-binary-looking label like 22: / 22b.
> 2. litebox_common_linux/src/lib.rs:907 — Medium: AArch64 suseconds_t is aliased to u64, but Linux defines it as signed long/i64. Negative tv_usec values
>    can be misinterpreted. Use i64 and validate range explicitly.
> 3. litebox_common_linux/src/lib.rs:2584 — Medium: AArch64 disables legacy mkdir but does not decode mkdirat, so directory creation on AArch64 likely
>    falls through to ENOSYS. Add a mkdirat decode path with the intended dirfd semantics.

1. Interestingly, clippy for arm64 doesn't complain this. Anyway, let me change this.
2. Looks like both x86_64 and aarch64 use i64, so this might be a stale bug or known one (`tv_usec` shouldn't be negative even though it is a signed integer).
3. Let me change this once #898 is merged.

litebox/src/platform/arch.rs
```diff
@@ -3,13 +3,13 @@
 
 //! Architecture-specific platform interfaces.
 //!
-//! As it currently stands, the interfaces here are only considered for x86-64, in the future other
+//! As it currently stands, the interfaces here are mostly considered for x86-64, in the future other
 //! architectures might be supported.
```
jaybosamiya-ms: Minor nit: change this to "only x86-64 and aarch64", rather than "mostly"?

PR #903, opened 2026-06-04
Fix deadlock in layered FS for migrated files

scaspin:
Fix deadlock in layered FS where the `node_info_lookup` lock is read-acquired and then write-acquired


scaspin:

jaybosamiya-ms (approved):
Thanks Shai!  Good catch, let's chat more about this IRL soon.  I'll mildly tweak the PR title/description, since that's what the merge queue uses as the commit message, and then will hit merge shortly.

wdcui:
@jaybosamiya-ms should we cherry pick this one to ulitebox?

jaybosamiya-ms:
Answered IRL, but briefly: due to upcoming changes to layered FS, we don't need to cherry pick this one over rn

PR #906, opened 2026-06-09
Add support for multiple PTAs

sangho2:
This PR adds support for multiple PTAs to manage them through common interfaces. It lets a TA open and interact with PTAs using their unique (PTA) session IDs.

### DIFF
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -165,8 +165,13 @@ struct GlobalState {
     _litebox: litebox::LiteBox<Platform>,
     /// The TA UUID to binary map for TA loading.
     ta_uuid_map: TaUuidMap,
-    /// Maintain whether non-concurrent PTAs (i.e., PTAs w/o `TaFlags::CONCURRENT`)
-    /// are busy or not to serialize their command invocation
+    /// Tracks which non-concurrent PTAs (i.e., PTAs w/o `TaFlags::CONCURRENT`)
+    /// are currently busy. A busy PTA is *rejected* with `TeeResult::Busy`
+    /// rather than queued.
+    ///
+    /// TODO: OP-TEE serializes concurrent access to a non-concurrent PTA by
+    /// blocking/queuing the caller until the PTA is free. We currently reject
+    /// instead of serialize; revisit if a PTA needs true serialization.
     pta_busy: spin::mutex::SpinMutex<HashSet<PseudoTa>>,
 }
 
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -14,7 +14,7 @@ use aes::{Aes128, Aes192, Aes256};
 use alloc::{sync::Arc, vec};
 use core::cell::Cell;
 use ctr::Ctr128BE;
-use hashbrown::HashMap;
+use hashbrown::{HashMap, HashSet};
 use litebox::{
     LiteBox,
     mm::{PageManager, linux::PAGE_SIZE},
@@ -149,6 +149,7 @@ impl OpteeShimBuilder {
             pm: PageManager::new(&self.litebox),
             _litebox: self.litebox,
             ta_uuid_map: TaUuidMap::new(),
+            pta_busy: spin::mutex::SpinMutex::new(HashSet::new()),
         });
         OpteeShim(global)
     }
@@ -164,6 +165,8 @@ struct GlobalState {
     _litebox: litebox::LiteBox<Platform>,
     /// The TA UUID to binary map for TA loading.
     ta_uuid_map: TaUuidMap,
+    /// Non-concurrent PTAs currently entered in this shim instance.
+    pta_busy: spin::mutex::SpinMutex<HashSet<PseudoTa>>,
 }
 
 impl GlobalState {
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -9,6 +9,7 @@
 extern crate alloc;
 
 use crate::loader::elf::ElfLoaderError;
+use crate::syscalls::pta::PseudoTa;
 use aes::{Aes128, Aes192, Aes256};
 use alloc::{sync::Arc, vec};
 use core::cell::Cell;
@@ -244,6 +245,7 @@ impl OpteeShim {
                 tee_cryp_state_map: TeeCrypStateMap::new(),
                 tee_obj_map: TeeObjMap::new(),
                 ta_handle_map: TaHandleMap::new(),
+                pta_sessions: spin::mutex::SpinMutex::new(HashMap::new()),
                 ta_entry_point: Cell::new(0),
                 ta_stack_base_addr: Cell::new(0),
                 ta_prepared: Cell::new(false),
@@ -439,7 +441,7 @@ impl Task {
                 if let Some(ta_uuid) = ta_uuid.read_at_offset(0)
                     && let Some(usr_params) = usr_params.read_at_offset(0)
                 {
-                    Task::sys_open_ta_session(
+                    self.sys_open_ta_session(
                         ta_uuid,
                         cancel_req_to,
                         usr_params,
@@ -450,7 +452,7 @@ impl Task {
                     Err(TeeResult::BadParameters)
                 }
             }
-            SyscallRequest::CloseTaSession { ta_sess_id } => Task::sys_close_ta_session(ta_sess_id),
+            SyscallRequest::CloseTaSession { ta_sess_id } => self.sys_close_ta_session(ta_sess_id),
             SyscallRequest::InvokeTaCommand {
                 ta_sess_id,
                 cancel_req_to,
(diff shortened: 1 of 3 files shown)


litebox_shim_optee/src/lib.rs
```diff
@@ -163,6 +165,8 @@ struct GlobalState {
     _litebox: litebox::LiteBox<Platform>,
     /// The TA UUID to binary map for TA loading.
     ta_uuid_map: TaUuidMap,
+    /// Non-concurrent PTAs currently entered in this shim instance.
```
wdcui: Nit: this comment is a little confusing. Is it that a TA tries to call multiple PTAs?

litebox_shim_optee/src/lib.rs
```diff
@@ -1357,6 +1369,8 @@ pub(crate) struct SessionIdPool {
     pool: litebox::utils::id_pool::IdPool,
     /// Next one-time ID when the recyclable pool is exhausted.
     fallback_next: u32,
+    /// Whether all fallback IDs have been issued.
```
wdcui, line 1374: We should have a unified implementation for various IDs in LiteBox. Should we log an issue?

litebox_shim_optee/src/syscalls/pta.rs
```diff
@@ -13,18 +13,74 @@ use litebox::platform::{
...
-    time_hi_and_version: 0x11e8,
-    clock_seq_and_node: [0x9c, 0x2d, 0xfa, 0x7a, 0xe0, 0x1b, 0xbe, 0xbc],
-};
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub(crate) struct SystemPta;
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub(crate) enum PseudoTa {
+    System(SystemPta),
+}
```
wdcui: Why do we need both `SystemPta` and `PseudoTa`?

litebox_shim_optee/src/syscalls/pta.rs
```diff
@@ -13,18 +13,74 @@ use litebox::platform::{
...
+
+impl PseudoTa {
+    pub(crate) fn from_uuid(uuid: &TeeUuid) -> Option<Self> {
+        match *uuid {
+            SystemPta::UUID => Some(Self::System(SystemPta)),
+            _ => None,
+        }
+    }
+
+    /// Open a session to this PTA, returning the allocated session ID.
```
wdcui, line 40: Each TA runs in its own session. When it calls a PTA, we open another PTA session? How do these two sessions co-exist?

litebox_shim_optee/src/syscalls/pta.rs
```diff
@@ -68,36 +124,101 @@ pub enum PtaSystemCommandId {
...
+    }
+
+    pub(crate) fn close_pta_session(&self, ta_session_id: u32) -> Option<PseudoTa> {
+        let pta = self.pta_sessions.lock().remove(&ta_session_id)?;
+        pta.close_session(self, ta_session_id);
+        crate::SessionIdPool::recycle(ta_session_id);
+        Some(pta)
+    }
+
+    pub(crate) fn pta_for_inter_ta_session(&self, ta_sess_id: u32) -> Option<PseudoTa> {
```
wdcui: What is this for?

litebox_shim_optee/src/syscalls/pta.rs
```diff
@@ -68,36 +124,101 @@ pub enum PtaSystemCommandId {
...
+        self.pta_sessions.lock().get(&ta_sess_id).copied()
+    }
+
+    pub(crate) fn close_all_pta_sessions(&self) {
```
wdcui, line 199: What happens to the TA session that invoked a PTA session?

litebox_shim_optee/src/syscalls/pta.rs
```diff
@@ -68,36 +124,101 @@ pub enum PtaSystemCommandId {
...
+    const FLAGS: TaFlags = PTA_DEFAULT_FLAGS.union(TaFlags::CONCURRENT);
+
+    const UUID: TeeUuid = TeeUuid {
+        time_low: 0x3a2f_8978,
+        time_mid: 0x5dc0,
+        time_hi_and_version: 0x11e8,
+        clock_seq_and_node: [0x9c, 0x2d, 0xfa, 0x7a, 0xe0, 0x1b, 0xbe, 0xbc],
+    };
+
+    fn open_session(params: &UteeParams) -> Result<u32, TeeResult> {
```
wdcui, line 220: We assume the PTA is already loaded into memory?

litebox_shim_optee/src/syscalls/pta.rs
```diff
@@ -68,36 +124,101 @@ pub enum PtaSystemCommandId {
...
+            TeeParamType::None,
+            TeeParamType::None,
+        ]) {
+            return Err(TeeResult::BadParameters);
+        }
+
+        crate::SessionIdPool::allocate().ok_or(TeeResult::Busy)
+    }
+
+    fn close_session(_task: &Task, _session_id: u32) {}
```
wdcui: Why is `close_session` a no-op?

wdcui (approved):
Overall, LGTM. I left some comments below.

wdcui:
GPT-5.5 found the following issue. I understand the second issue is by design. But should we just have a better handling of session id allocatioin? In that case, we don't need to introduce this new restriction.

   1. litebox_shim_optee/src/syscalls/tee.rs:179 — sys_open_ta_session opens and records the PTA session before writing session_id back to ta_sess_id. If
  that write fails, the session remains in pta_sessions and the ID is not recycled, leaving the TA unable to close it and likely causing later opens of the
   same PTA to return Busy.

   2. litebox_shim_optee/src/syscalls/pta.rs:151 — the new per-task “one session per PTA” cap rejects a second open of the System PTA even though
  SystemPta::FLAGS includes MULTI_SESSION at line
    193. That breaks OP-TEE-style multi-session PTA semantics for TAs that legitimately hold multiple handles to the same PTA.

litebox_shim_optee/src/syscalls/pta.rs
sangho2: The system PTA is one of PTAs. `PseudoTa` is a common interface to handle various PTAs (e.g., system, IDK, ...).

litebox_shim_optee/src/syscalls/pta.rs
sangho2: A PTA is something like a set of extended (blocking) system calls. A PTA session is just an identifier. There is no additional process/thread, page tables, ... This is different from real inter-TA interaction which requires substantial amount of change/code.

litebox_shim_optee/src/syscalls/pta.rs
sangho2: A TA invokes PTA commands though a session ID (opens a session for PTA and then invokes commands). We need to figure out the corresponding PTA to route the commands to it.

litebox_shim_optee/src/syscalls/pta.rs
sangho2: The TA session blocks (like a blocking syscall).

litebox_shim_optee/src/syscalls/pta.rs
sangho2: Each PTA is just a set of in-kernel functions. It is not a separate binary.

litebox_shim_optee/src/syscalls/pta.rs
sangho2: The System PTA is stateless, so it doesn't require extra cleanup. Other PTAs might be stateful and this common interface can handle both cases.

litebox_shim_optee/src/lib.rs
sangho2: There are concurrent and non-concurrent PTAs. Concurrent PTAs can be concurrently invoked by multiple TA sessions whereas the invocation of non-concurrent PTAs must be serialized. `pta_busy` represents whether a non-concurrent PTA is busy or not.

litebox_shim_optee/src/lib.rs
sangho2: `litebox::utils::id_pool::IdPool` is a unified implementation.

sangho2:
> GPT-5.5 found the following issue. I understand the second issue is by design. But should we just have a better handling of session id allocatioin? In that case, we don't need to introduce this new restriction.
> 
> 1. litebox_shim_optee/src/syscalls/tee.rs:179 — sys_open_ta_session opens and records the PTA session before writing session_id back to ta_sess_id. If
>    that write fails, the session remains in pta_sessions and the ID is not recycled, leaving the TA unable to close it and likely causing later opens of the
>    same PTA to return Busy.
> 2. litebox_shim_optee/src/syscalls/pta.rs:151 — the new per-task “one session per PTA” cap rejects a second open of the System PTA even though
>    SystemPta::FLAGS includes MULTI_SESSION at line
>    193. That breaks OP-TEE-style multi-session PTA semantics for TAs that legitimately hold multiple handles to the same PTA.

1. Let me fix it.
2. This is intentional and I don't see real usage of holding multiple sessions for the same PTA (at least for now). Our PTAs still support multiple sessions and can be concurrently invoked by "different" TA sessions. One way to deal with this would be introducing something like `ulimit` and `cgroups`. For now, allowing up to 100 PTA sessions per TA instance.

PR #907, opened 2026-06-10
Add preemption timer to bound TA execution

sangho2:
This PR adds a timer-based mechanism for preempting TAs on the LVBS platform. In OP-TEE, TA execution is unbounded because there is no secure-world scheduler: if a TA spins in an infinite loop, it can hold the (V)CPU indefinitely. On Arm TrustZone, the normal world can eventually regain control by interrupting the secure world. However, in Hyper-V, VTL0 is not allowed to interrupt VTL1. 
To overcome this problem, this PR implements a VTL1-local one-shot preemption timer based on the Hyper-V synthetic timer to limit the execution time of each TA command invocation (which could span multiple syscalls). If it fires, the VTL1 kernel kills the TA and returns back to VTL0.

### DIFF
--- a/litebox_platform_lvbs/Cargo.toml
+++ b/litebox_platform_lvbs/Cargo.toml
@@ -48,6 +48,9 @@ default = ["optee_syscall"]
 optee_syscall = []
 linux_syscall = []
 devbox = []
+# Tighten the preemption timer quantum so runaway-TA preemption can be tested
+# without waiting out the production budget. Test builds only.
+preemption_test_quantum = []
 
 [lints]
 workspace = true
--- a/litebox_platform_lvbs/src/arch/x86/timer.rs
+++ b/litebox_platform_lvbs/src/arch/x86/timer.rs
@@ -1,0 +1,225 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//! Hyper-V synthetic-timer (STIMER) preemption timer: VTL1 preempts a runaway TA.
+//!
+//! VTL0 cannot interrupt VTL1 and OP-TEE has no scheduler, so a TA that
+//! spins without returning would hold the VP forever and freeze VTL0 too.
+//! VTL1 arms a VTL1-local Hyper-V synthetic timer (STIMER0 in direct mode)
+//! that the TA cannot tamper with; on expiry it is delivered as
+//! `STIMER_VECTOR` and the shim kills the TA with
+//! `TEE_ERROR_TARGET_DEAD`.
+//!
+//! `scoped` brackets a whole TA entry: arm before `run_thread_arch`,
+//! disarm once it fully returns to the VTL1 kernel: i.e., after the TA has
+//! left ring 3, and before the VP is handed back to VTL0. The timer stays
+//! armed across the TA's syscalls and faults (VTL1's own kernel work is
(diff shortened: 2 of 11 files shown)


sangho2 (commented):
.

litebox_platform_lvbs/Cargo.toml
```diff
@@ -46,6 +46,9 @@ default = ["optee_syscall"]
 optee_syscall = []
 linux_syscall = []
 devbox = []
+# Tighten the preemption timer quantum so runaway-TA preemption can be tested
+# without waiting out the production budget. Test builds only.
```
wdcui, line 52: What's the production budget?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -1632,7 +1632,15 @@ fn run_thread_inner(
     // `thread_ctx` will be passed to `syscall_handler` later.
     // `ctx_ptr` is to let `run_thread_arch` easily access `ctx` (i.e., not to deal with
     // member variable offset calculation in assembly code).
-    unsafe { run_thread_arch(&mut thread_ctx, ctx_ptr, u8::from(reenter)) };
+    //
+    // Arm the preemption timer for this TA entry; disarmed at the VTL0-return
```
wdcui: Nit: "bracket the entry" is a little confusing.

litebox_platform_lvbs/src/arch/x86/timer.rs
```diff
@@ -0,0 +1,243 @@
...
+
+//! Hyper-V synthetic-timer (STIMER) preemption timer: forcefully terminates
+//! runaway user-mode code in VTL1.
+//!
+//! VTL1 has no preemptive scheduler and VTL0 cannot interrupt VTL1, so
+//! user-mode code that spins without returning holds the VP forever and freezes
+//! VTL0 too. In lieu of a scheduler, VTL1 arms a VTL1-local Hyper-V synthetic
+//! timer (STIMER0 in direct mode) that user-mode code cannot tamper with; on
+//! expiry it fires `STIMER_VECTOR` and the shim terminates the offending
+//! thread.
```
wdcui, line 12: Where does the shim terminate the thread?

sangho2:
Potential conflict with PR #865. Merge it first and then this one.

wdcui (approved):
LGTM. Thanks!

litebox_platform_lvbs/src/arch/x86/timer.rs
sangho2: A user-mode STIMER interrupt will be delivered to the OP-TEE shim, and it'll terminate the thread. No extra code in this PR for this because the OP-TEE has no notion of interrupt such that its shim currently just terminates the thread if there is an interrupt.

litebox_platform_lvbs/Cargo.toml
sangho2: This PR for now uses 8 seconds, which is slightly below Linux kernel's hard CPU lockup timeout (10 seconds).

sangho2: update: intentionally trigger this issue, and confirm the Linux kernel suffers from RCU preempt stalls in this case. Azure Linux's `CONFIG_RCU_CPU_STALL_TIMEOUT` is 60, so we are using 50 (we can make this configurable if needed).

PR #908, opened 2026-06-11
fix(loader): page-align head/tail munmap in MapMemory::reserve (cherry-pick of #891)

willportnoy:
Cherry-pick of #891 (commit `5d4a5b7f`) onto `ulitebox`.

---

**fix(loader): page-align head/tail munmap in `MapMemory::reserve`**

`MapMemory::reserve` allocates an over-sized `PROT_NONE` mapping and trims the unaligned head + tail with `munmap` so the caller gets a block aligned to `align`. The tail `munmap` was passed `(aligned_ptr + len, mapping_end - end)` directly — but `len` (an ELF's `max_vaddr - min_vaddr` span) is in general **not** page-aligned, and `munmap` rejects non-page-aligned start addresses with `EINVAL`. This surfaced as `execve` → `ENOEXEC` (`exit_code: 127`) on every guest `fork+exec` of binaries whose PT_LOAD span ends mid-page — e.g. prebuilt linux-x64 node.js (`0x6403D68`).

**Fix:** factor the trim arithmetic into a pure helper `compute_reserved_regions` in `litebox_common_linux::loader` that rounds the tail start up to `PAGE_SIZE` and the mapping end up to the kernel's actual page-rounded extent. Both `litebox_shim_linux` and `litebox_shim_optee` now call the helper (the bug was duplicated in both shims).

Covered by new unit tests in `reserve_regions_tests` (regression case pins the EINVAL pattern).

---

Verified on this branch:
- `cargo test -p litebox_common_linux` → 5/5 reserve_regions tests pass.
- `cargo test -p litebox_shim_optee` → 2/2 pre-existing tests pass.
- `cargo clippy --all-targets -- -D warnings` clean on `litebox_common_linux`, `litebox_shim_linux` (`--features platform_linux_userland`), and `litebox_shim_optee`.

### DIFF
--- a/litebox_common_linux/src/loader.rs
+++ b/litebox_common_linux/src/loader.rs
@@ -619,6 +619,83 @@ pub trait MapMemory {
     -> Result<(), Self::Error>;
 }
 
+/// The result of computing the head/tail trim regions for an over-sized
+/// anonymous reservation made by [`MapMemory::reserve`].
+///
+/// See [`compute_reserved_regions`] for details.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct ReservedRegions {
+    /// Base address of the requested `len` bytes inside the over-sized
+    /// reservation, aligned up to the `align` argument passed to
+    /// [`compute_reserved_regions`].
+    pub aligned_ptr: usize,
+    /// `(start, len)` of the page-aligned head slice that should be
+    /// released with `munmap`, or `None` if no head trim is needed.
+    pub head_unmap: Option<(usize, usize)>,
+    /// `(start, len)` of the page-aligned tail slice that should be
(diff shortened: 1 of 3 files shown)


jaybosamiya-ms:
Just as a future note @willportnoy , it would be good to not include the last bits in your description as part of it, since that gets added into the history (the queue takes the commit message directly from the PR title + description).

If clippy or tests were not clean, the CI would have blocked it, so that text is not adding anything useful.  It's fine for this one, but generally speaking, we're trying to keep the signal-to-noise ratio good in the git history, so mentioning this for future PRs.  Thanks!

wdcui:
@jaybosamiya-ms that bloated description may be my fault. I saw Will shortened the description, but I guess the agent may have updated it when the last commit was pushed. I didn't check the description again when I hit the merge button.

PR #913, opened 2026-06-13
Strengthen x86_64 user return context sanitization

jaybosamiya-ms:


sangho2 (approved):
Looks good to me. Now we have `USER_ADDR_MIN`/`USER_ADDR_END` in `litebox_common_linux`. We could let the platforms use them to initialize `TASK_ADDR_MIN`/`TASK_ADDR_MAX` (optional as this PR is not for this).

jaybosamiya-ms:
Yep, that's why I made them public, but I didn't want to make those tweaks in this same PR, we can do that separately

PR #921, opened 2026-06-15
Fix a rare iced-x86 panic

jaybosamiya-ms:
https://github.com/microsoft/litebox/pull/697 fixed a rare panic due to a boundary issue.  Unfortunately, it was not fully sufficient, and there was still a (rarer) panic that could occur (e.g., https://github.com/microsoft/litebox/actions/runs/27575376051/job/81522248856).  Basically, #697 handled "don't advance a chunk across 4GiB," but missed "don’t pass the overlap bytes across 4GiB either." This PR fixes it.

### DIFF
--- a/litebox_syscall_rewriter/src/lib.rs
+++ b/litebox_syscall_rewriter/src/lib.rs
@@ -819,32 +819,94 @@ fn decode_section_instructions(
         let chunk_start_ip = section_base_addr + offset as u64;
         let chunk_end_ip = chunk_start_ip + chunk_advance_len as u64;
 
-        let mut decoder = iced_x86::Decoder::new(
+        append_decoded_instructions(
             bitness,
             &remaining[..decode_window_len],
-            iced_x86::DecoderOptions::NONE,
-        );
-        decoder.set_ip(chunk_start_ip);
-
-        for inst in &mut decoder {
-            if inst.len() == 0 {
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
As I'd mentioned on #697: 
> We plan to potentially remove iced-x86 from our dependencies, esp since we might want future ARM support, but at least this PR should fix some of the rare CI failures we see from time to time.

wdcui:
GPT-5.5 found the following issue:

   Medium —  litebox_syscall_rewriter/src/lib.rs:843                                                                                                                                                                               
   The 4GiB-boundary guard uses  >= window.len() , so a decode window ending exactly at the next 4GiB boundary is still passed directly to iced-x86. iced-x86 1.21 still does 32-bit pointer-delta arithmetic after decoding, so an
   instruction consuming bytes up to that exact boundary can underflow/panic. The scratch path has the same  >= /assert issue. Use strict  >  and shift/copy when  dist <= window.len() .

jaybosamiya-ms:
Thanks Weidong, that seems valid (I did not dive deeper into iced-x86 to confirm, since the change is a conservative fix).  Pushed the fix.  Also opened #927 to track the fact that we want to migrate away from iced-x86.

wdcui (approved):
LGTM. Thanks!

PR #922, opened 2026-06-16
.gitignore: ignore direnv files

lschuermann:
Ignores direnv-specific files. Direnv is useful to set up dynamic development environments, but there is not a single canonical configuration that should be upstreamed. Adding it to `.gitignore` prevents contributors from accidentally including their particular configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -193,3 +193,7 @@ venv/
 dev_bench/unixbench/prepared/
 dev_bench/unixbench/byte-unixbench-*/
 dev_bench/unixbench/*.zip
+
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms (approved):
LGTM, thanks!  Merging this, but also for future reference, `.git/info/exclude` (for project specific) or `$XDG_CONFIG_HOME/git/ignore` / `$HOME/.config/git/ignore` (for a global ignore config) is helpful for local setups, without needing to have a project-specific `.gitignore` need to change.

PR #925, opened 2026-06-17
Add key-value field visitor to HostLogger

sangho2:
This PR implements a simple key-value field visitor for host logging.


wdcui (approved):
LGTM. Thanks!

PR #937, opened 2026-06-18
Fix OP-TEE TA reentry stack alignment

sangho2:
This PR fixes OP-TEE TA's stack alignment for reentry (`reenter_thread`). In particular, `reenter_thread` uses `jmp` to enter the TA entry point function (`__ta_entry`) which is a normal function and thus suffers a function stack alignment issue.


CvvT (approved):
LGTM, thanks!

PR #961, opened 2026-06-24
Fix TLB flush hypercall (LVBS platform)

sangho2:
This PR fixes the TLB flush hypercall of the LVBS platform. Currently, it does use incorrect data structure encoding, which can result a `HV_STATUS_INVALID_PARAMETER` error if multiple VPs enter VTL1 concurrently.


wdcui (approved):
LGTM. Thanks.

PR #963, opened 2026-06-25
Make the fs `Backend` dyn-compatible

jaybosamiya-ms:
#887 introduced the `Backend` abstraction to the file system, but as I was working on more functionality, I discovered that the better design is to guarantee that `Backend` is [dyn-compatible](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility), so that the composition of multiple backends can be done without a huge amount of type-level machinery + macros; it also keeps the interfaces more ergonomic for consumers of LiteBox core.  To make things dyn-safe, we need to move out the `Backend`-specific handles; however, rather than the naive handling of this, there's a convenient abstraction such that backend implementations still end up with a largely convenient way to use their own typed handles.

### DIFF
--- a/litebox/src/fs/backend.rs
+++ b/litebox/src/fs/backend.rs
@@ -3,7 +3,12 @@
 
 //! [`Backend`] for filesystems supported by [`super::resolver`]
 
+use alloc::boxed::Box;
 use alloc::vec::Vec;
+use core::any::{Any, TypeId};
+use core::marker::PhantomData;
+
+use crate::utilities::anymap::AnyCloneSendSync;
 
 use super::errors::{
     ChmodError, ChownError, FileStatusError, MkdirError, OpenError, ReadDirError, ReadError,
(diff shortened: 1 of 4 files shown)


wdcui (approved):
LGTM. Thanks!

GPT-5.5 reported the following two findings. Please double check them.

1. litebox/src/fs/resolver.rs: O_PATH is accepted, but normal operations on those fds hit unimplemented!(), so a caller can panic LiteBox instead of getting an errno-style error such as EBADF.

2. litebox/src/fs/devices.rs: The devices backend gives / and /dev the same (dev, ino): dev_dir_inode starts at inode 0, and root status also reports inode 0. This breaks inode identity for distinct directories.

jaybosamiya-ms:
Thanks, the `unimplemented!` panic GPT likes to complain unnecessarily imho, if you don't mind, can you tell your future runs of it something like:
> `unimplemented!()` is an explicit signal of "we will implement this later", the panic there is intentionally indicating "we need to solve this in a future PR", do not consider this to be an issue

For the inodes for the dirs, yep, known and will be resolved in future PRs, see https://github.com/microsoft/litebox/pull/887#pullrequestreview-4500926069

PR #972, opened 2026-06-26
Fix cross-page-table sharing of mutable mappings (LVBS platform)

sangho2:
This PR fixes cross-page-table sharing of mutable mappings. It now only clones the fixed kernel PML4 slots (`>= KERNEL_OFFSET`) when it creates user (task) page tables to avoid duplicating transient mappings and intermediate tables for direct-map and vmap. It also sets stable `PRESENT | WRITABLE` to the parent table flags.

### DIFF
--- a/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
+++ b/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
@@ -38,10 +38,11 @@ const TLB_SINGLE_PAGE_FLUSH_CEILING: usize = 33;
 /// PML4 index of the first VTL1-kernel slot (`PA + KERNEL_OFFSET`).
 ///
 /// Only slots `>= KERNEL_PML4_START` are safe to share between page tables:
-/// the kernel mapping structure is fixed after boot. Lower slots (user,
-/// direct-map, vmap) have mappings that change at runtime and are not synced
-/// across page tables, so sharing a stale snapshot would make the base and a
-/// task observe different leaves for the same page (cross-PT write fault).
+/// their intermediate tables (P3/P2/P1) are fixed after boot, so sharing them
+/// is read-only. Lower slots (user, direct-map, vmap) get intermediate tables
+/// allocated and freed at runtime on whichever page table is active. Sharing
+/// those would let a task mutate the base's intermediate tables (and make
+/// frame ownership ambiguous at teardown), so each page table must own them.
 ///
 /// `KERNEL_OFFSET` is 512 GiB (PML4-slot) aligned, so this is an exact cutoff.
 pub(crate) const KERNEL_PML4_START: usize = ((crate::KERNEL_OFFSET >> 39) & 0x1FF) as usize;
--- a/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
+++ b/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
@@ -35,6 +35,17 @@ use crate::mm::{
 #[cfg(not(test))]
 const TLB_SINGLE_PAGE_FLUSH_CEILING: usize = 33;
 
+/// PML4 index of the first VTL1-kernel slot (`PA + KERNEL_OFFSET`).
+///
+/// Only slots `>= KERNEL_PML4_START` are safe to share between page tables:
+/// the kernel mapping structure is fixed after boot. Lower slots (user,
+/// direct-map, vmap) have mappings that change at runtime and are not synced
+/// across page tables, so sharing a stale snapshot would make the base and a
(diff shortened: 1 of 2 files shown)


wdcui:
@ziqiaozhou can you please help review this PR? Thanks!

litebox_platform_lvbs/src/lib.rs
```diff
@@ -375,7 +376,8 @@ impl PageTableManager {
             pt.clear_shared_pml4_entries(&self.base_page_table);
 
             // Safety: We're about to delete this page table, so it's safe to
-            // free the remaining (user-space) intermediate page table frames.
+            // free the remaining task-owned intermediate page table frames
+            // (user, direct-map, and vmap slots).
             unsafe {
                 pt.cleanup_page_table_frames();
             }
```
CvvT: `cleanup_page_table_frames` frees page table frames from 0 to `0xffff_ffff_ffff_f000`. If we change the range to [0, KERNEL_OFFSET], then we don't need to call `clear_shared_pml4_entries`?

CvvT: What are the ranges for direct-map and vmap? I assume they are below `KERNEL_OFFSET`.

CvvT (approved):
LGTM! I left two questions.

litebox_platform_lvbs/src/lib.rs
sangho2: 1. Good point. Let me see whether I can simplify this (integrating those functions perhaps?)
2. yes. they are below `KERNEL_OFFSET`: https://github.com/microsoft/litebox/blob/58c99c7e1a241bed4f0c8b8c877250d259f3a644/litebox_platform_lvbs/src/lib.rs#L85-L113

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -612,7 +624,8 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
             } else {
                 flags
             };
-            let table_flags = page_flags - PageTableFlags::NO_EXECUTE; // parent entries should not have NO_EXECUTE
+            // Parent entries use a stable permissive constant, not leaf-derived flags.
+            let table_flags = PageTableFlags::PRESENT | PageTableFlags::WRITABLE;
```
ziqiaozhou, line 645: NIT: You can safely use`PageTableFlags::PRESENT | PageTableFlags::WRITABLE | PageTableFlags::ACCESSED | PageTableFlags::DIRTY`, which is a small optimization made in Linux kernel code to avoid unnecessary hardware changes: less cache-coherency traffic and slightly faster page table walk.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -755,42 +768,42 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
...
-    /// Entries already present in `self` are left unchanged.
+    /// Only entries present in `source` and absent in `self` are copied.
     pub(crate) fn copy_pml4_entries_from(&self, source: &Self) {
         let mut dst = self.inner.lock();
         let src = source.inner.lock();
         for (dst_entry, src_entry) in dst
             .level_4_table_mut()
             .iter_mut()
             .zip(src.level_4_table().iter())
+            .skip(KERNEL_PML4_START)
```
ziqiaozhou, line 801: I am a little confused on this. Are you supposed to only copy PML4 index >=KERNEL_PML4_START? Why this code looks like you are copying all except for the KERNEL_PML4_START?

sangho2: it discards up to `KERNEL_PML4_START` and does something from that point.

ziqiaozhou: I was mixed it with SVSM's page table which only use a single index for kernel mapping :>.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2: This flags is for parent entries, so whether we have A/D or not seems not important (D is ignored in this case anyhow). There are other code paths to set up flags for parent and leaf entries, so we can consider adding this optimization both places together (as a separate PR).

ziqiaozhou: Sure!

sangho2: Issue #980

ziqiaozhou (approved):

PR #979, opened 2026-06-29
Fix: Serialize packed-`OpteeMsgArgs` access

sangho2:
This PR serializes packed-`OpteeMsgArgs` access on the base page table. The OP-TEE driver packs multiple requests into sub-page slots of a physical page frame, resulting in concurrent mapping attempts by multiple cores on the kernel/base page table. These attempts are rejected since the LiteBox/LVBS kernel platform does not support shared mapping. This PR fixes this problem by introducing a new lock to serialize mappings for accessing `OpteeMsgArgs`.

### DIFF
--- a/dev_tests/src/ratchet.rs
+++ b/dev_tests/src/ratchet.rs
@@ -43,7 +43,7 @@ fn ratchet_globals() -> Result<()> {
             ("litebox_runner_lvbs/", 6),
             ("litebox_runner_snp/", 2),
             ("litebox_shim_linux/", 1),
-            ("litebox_shim_optee/", 3),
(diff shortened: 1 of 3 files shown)


sangho2:
PR #972 ensures that kernel/base and user/task page tables no longer share direct-map and vmap. Thus, this lock is not needed for cores/code running on task page tables (e.g., `write_msg_args_to_normal_world`).

wdcui (approved):
LGTM. Thanks.

Per our offline discussion, please update the doc comment.

sangho2:
> LGTM. Thanks.
> 
> Per our offline discussion, please update the doc comment.

Thanks! Added TODO to the comment.

PR #984, opened 2026-06-30
ci: speed up semver-checks by subsetting crates

jaybosamiya-ms:
Essentially, skip semver-checks on any crates that _provably_ cannot touch semver, by looking at reverse-deps of changed files.

### DIFF
--- a/.github/tools/determine_semver_packages.py
+++ b/.github/tools/determine_semver_packages.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python3
+"""Determine which workspace packages need cargo-semver-checks.
+
+A package needs checking if semver-relevant files in that package changed, or if
+it depends (directly or transitively) on such a package. This lets the CI avoid
+checking unrelated crates while still catching public API changes caused by
+workspace dependency changes/re-exports.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import subprocess
+import sys
+from collections import defaultdict, deque
+from pathlib import Path
+
+
+def cargo_metadata() -> dict:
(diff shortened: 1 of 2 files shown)


jaybosamiya-ms:
Note: the Python script was generated by GPT-5.5; I've briefly sanity checked it, it is not the most elegant but should do the job reasonably (it is just figuring out reverse-deps for the crates based on what was touched).

~~I'll be doing a bit more testing before I mark this as ready to review, esp since CI stuff cannot be tested locally.~~ Ready for review.

wdcui:
GPT-5.5 found the following issue. Is it valid?

High — semver package selection can include unsupported runner crates ( .github/tools/determine_semver_packages.py:139 , used by .github/workflows/semver-checks.yml:84 ). The script selects from all  workspace_members , so a core change can emit  litebox_runner_lvbs  and litebox_runner_optee_on_linux_userland . CI explicitly treats these specially: LVBS needs a custom/nightly target, and the OP-TEE runner is built separately due feature conflicts. Running them in the single semver-checks invocation can make unrelated PRs fail. Filter selected output to workspace_default_members  or an explicit semver-check allowlist.

jaybosamiya-ms:
Invocations on these crates have been working fine on older versions of semver-checks (see below).  Also, _by design_ the semver-checks CI right now will always return a "success" and does not act as a blocker (in the future we may make it a blocker once we make a real release, so that it is really also making sure we are tracking version numbers correctly).  For now, the worst that can happen is that the message that gets posted misses some info.

image

image

wdcui (approved):
Thanks for the optimization.

PR #985, opened 2026-06-30
Load ELF interpreter high so a low-loaded ET_EXEC's brk heap isn't capped

willportnoy:
Dynamically-linked ET_EXEC (non-PIE) binaries load at their low canonical vaddrs and grow their glibc `brk` heap upward from there, but the ELF interpreter was reserved at a fixed low offset directly above the main image — capping the heap, so a heap-heavy process (e.g. a bundled `node`) hit `ENOMEM` the moment it crossed that offset. This loads the interpreter at the top of the address space instead, mirroring how the kernel places `ld.so` for an ET_EXEC main and leaving the whole gap above the main image free for the heap; it is a no-op for PIE mains, whose interpreter already loads high.

### DIFF
--- a/litebox_shim_linux/src/loader/elf.rs
+++ b/litebox_shim_linux/src/loader/elf.rs
@@ -498,15 +498,20 @@ mod tests {
             .load_mapped(task.global.platform)
             .expect("interpreter should load");
 
-        let expected_top_down_base =
-            <Platform as PageManagementProvider<{ PAGE_SIZE }>>::TASK_ADDR_MAX - PAGE_SIZE;
-        assert_eq!(
-            interp.base_addr,
-            expected_top_down_base,
-            "ET_EXEC interpreter loaded at {:#x}, leaving heap capped near {:#x}, not top-down at {:#x}",
+        // The interpreter must land high — via the top-down search — so the
+        // low ET_EXEC brk heap below it is not capped. The exact address is
+        // not asserted: `get_unmmaped_area` returns the highest free gap, and
+        // host mappings seeded into the userland VMA tree can sit near the top
+        // and push that gap below the very top slot (see `mm/linux.rs`). Assert
+        // the invariant that matters — placement in the high half of the
+        // address space, far above the low-heap region — not one exact slot.
+        let addr_max = <Platform as PageManagementProvider<{ PAGE_SIZE }>>::TASK_ADDR_MAX;
+        assert!(
+            interp.base_addr >= addr_max / 2,
+            "ET_EXEC interpreter loaded at {:#x}, near the low-heap region {:#x} rather than top-down high (>= {:#x})",
             interp.base_addr,
             crate::loader::DEFAULT_LOW_ADDR,
-            expected_top_down_base
+            addr_max / 2,
         );
     }
 }
--- a/litebox_shim_linux/src/loader/elf.rs
+++ b/litebox_shim_linux/src/loader/elf.rs
@@ -26,6 +26,11 @@ struct ElfFile<'a, FS: ShimFS> {
     task: &'a Task<FS>,
     fd: i32,
     load_high: bool,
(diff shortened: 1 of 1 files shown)


litebox_shim_linux/src/loader/elf.rs
```diff
@@ -317,3 +333,191 @@ impl From<ElfLoaderError> for litebox_common_linux::errno::Errno {
...
+            expected_top_down_base,
+            "ET_EXEC interpreter loaded at {:#x}, leaving heap capped near {:#x}, not top-down at {:#x}",
+            interp.base_addr,
+            crate::loader::DEFAULT_LOW_ADDR,
+            expected_top_down_base
+        );
+    }
+
+    #[test]
+    fn out_of_range_mmap_hint_fails_instead_of_falling_back_top_down() {
```
wdcui: This test isn't very useful. Should we remove it?

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -75,10 +80,17 @@ impl<FS: ShimFS> litebox_common_linux::loader::MapMemory for ElfFile<'_, FS> {
         // Allocate a mapping large enough that even if it's maximally misaligned we can
         // still fit `len` bytes.
         let mapping_len = len + (align.max(PAGE_SIZE) - PAGE_SIZE);
+        let hint = if self.load_high {
```
wdcui, line 83: This assumes load_high can only be done once? If so, is this assumption checked somewhere?

litebox_shim_linux/src/loader/elf.rs
willportnoy: Yes, removed.

litebox_shim_linux/src/loader/elf.rs
willportnoy: It’s added as a debug assertion (tell me if you prefer another method).

wdcui:
Please shorten the PR summary to 2-3 sentences.

litebox_shim_linux/src/loader/elf.rs
```diff
@@ -75,10 +87,32 @@ impl<FS: ShimFS> litebox_common_linux::loader::MapMemory for ElfFile<'_, FS> {
...
+            {
+                debug_assert!(
+                    !self.high_reserved,
+                    "load_high reserve() called more than once for one ELF image",
+                );
+                self.high_reserved = true;
+            }
+            <litebox_platform_multiplex::Platform as PageManagementProvider<{ PAGE_SIZE }>>::TASK_ADDR_MAX
+                .checked_sub(mapping_len)
+                .ok_or(Errno::ENOMEM)?
```
CvvT: I just realized `hint` has effect on Linux even if the specified address is not available:

```
If addr is not NULL, then the
kernel takes it as a hint about where to place the mapping; on
Linux, the kernel will pick a nearby page boundary
```

LiteBox does not follow it, but `hint` is directly forwarded to Linux `mmap` on the Linux platform. This does not work on Windows platform because we ignore the `hint` if the address is not available. It might be better to just use zero as hint as LiteBox would pick a proper hint in a top-down manner (see https://github.com/microsoft/litebox/blob/b60389fc38fe888755f469a61e394040660d90d4/litebox/src/mm/linux.rs#L913). Then we don't need `debug_assertions`.

willportnoy: That makes sense - updated.

(this fix was ported from a branch based long ago on the agent-sandbox-fork branch that had a top-down search implemented for invalid addresses - but your suggestion to use the platform’s support is better)

CvvT (approved):
LGTM, thanks!

wdcui:
@willportnoy, please feel free to merge your PR, and cherry pick it over to ulitebox. I hope you were not waiting for my approval. Thanks!

PR #987, opened 2026-06-30
Add OP-TEE `GetSystemTime` and revise LVBS platform `Instant`

sangho2:
This PR adds OP-TEE `GetSystemTime` and revises the LVBS platform's `Instant`. The OP-TEE system time is per-instance monotonic time. Also, it now uses Hyper-V's partition reference-counter tick that is a monotonic, frequency-invariant counter.

### DIFF
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -18,9 +18,9 @@ use hashbrown::{HashMap, HashSet};
 use litebox::{
     LiteBox,
     mm::{PageManager, linux::PAGE_SIZE},
-    platform::{RawConstPointer as _, RawMutPointer as _},
+    platform::{Instant as _, RawConstPointer as _, RawMutPointer as _, TimeProvider},
     shim::ContinueOperation,
-    utils::{ReinterpretUnsignedExt, TruncateExt},
+    utils::TruncateExt,
 };
 use litebox_common_linux::{MapFlags, ProtFlags, errno::Errno};
 use litebox_common_optee::{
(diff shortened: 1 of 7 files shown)


litebox_shim_optee/src/syscalls/tee.rs
```diff
@@ -290,6 +290,35 @@ impl Task {
...
+            TeeTimeCategory::System => {
+                let elapsed = self.global.system_time();
+                TeeTime {
+                    // `seconds` is a u32 in the GP `TEE_Time`; saturate rather than wrap.
+                    seconds: u32::try_from(elapsed.as_secs()).unwrap_or(u32::MAX),
+                    millis: elapsed.subsec_millis(),
+                }
+            }
+            // Persistent time was never set by this TA; report it per the spec.
+            TeeTimeCategory::TaPersistent => return Err(TeeResult::TimeNotSet),
```
wdcui, line 343: What syscall is for a TA to set its persistent time?

litebox_shim_optee/src/lib.rs
```diff
@@ -372,8 +386,7 @@ impl Task {
         let request = match SyscallRequest::<Platform>::try_from_raw(ctx.orig_rax, ctx) {
             Ok(request) => request,
             Err(err) => {
-                // TODO: this seems like the wrong kind of error for OPTEE.
-                ctx.rax = (err.as_neg() as isize).reinterpret_as_unsigned();
+                ctx.rax = TeeResult::from(err) as usize;
```
wdcui: The fix in this line is not specific to this PR?

wdcui (approved):
LGTM. I left some comments for your reference. Thanks.

litebox_shim_optee/src/syscalls/tee.rs
sangho2: There is `set_ta_time` that our OP-TEE shim doesn't support. Basically, this function stores an `offset` value in persistent storage. `TeeTimeCategory::TaPersistent` is expected to use this stored value.

litebox_shim_optee/src/lib.rs
sangho2: True. we can separate this if necessary.

sangho2: reverted. I'll create a separate PR.

PR #989, opened 2026-06-30
Fix alarm spin test overflow

CvvT:
Fix the userspace alarm spin test so its loop does not escape within one second (switching from signed `int` to unsigned `long long`).


wdcui (approved):
LGTM. Thanks.

PR #995, opened 2026-07-01
Support mounting file systems & migrate devices to use mounts

jaybosamiya-ms:
This PR introduces the `Composer`, which acts as a mount point for the new core file system design (see #887).  It maintains a runtime list of dyn-safe `Backend` objects for each mount point, using as much walking as feasible under the mount points.

To exercise the `Composer` and demonstrate how it works, I've also switched `Devices` to stop being hardcoded to `/dev` _inside_ the backend but instead be a flat list of devices that is _mounted_ at `/dev`.

While this PR does not expose/use it this way yet, with this PR, finally LiteBox core now has some of the fundamental machinery needed to eventually support `mount(2)` family of syscalls in the Linux shim.

Future PRs will continue migrating more backends over, eventually allowing for the old layered file system to disappear, and replaced with this mount composition + (upcoming) overlay file system to support union mounting.

### DIFF
--- a/litebox/src/fs/devices.rs
+++ b/litebox/src/fs/devices.rs
@@ -135,19 +135,13 @@ where
 {
     /// Construct a new `Devices` backend.
     #[must_use]
-    pub(crate) fn new(litebox: &LiteBox<Platform>, allocator: InodeAllocator) -> Self {
+    pub fn new(litebox: &LiteBox<Platform>, allocator: InodeAllocator) -> Self {
         let dev_dir_inode = allocator.next();
         Self {
             litebox: litebox.clone(),
             dev_dir_inode,
             _alloc: allocator,
         }
-    }
-
-    /// Migration helper.  This function will disappear soon.
-    #[must_use]
-    pub fn migration_helper_standalone_new(litebox: &LiteBox<Platform>) -> Self {
-        Self::new(litebox, InodeAllocator::standalone())
     }
 }
 
--- a/litebox/src/fs/devices.rs
+++ b/litebox/src/fs/devices.rs
@@ -3,12 +3,10 @@
 
 //! Unix-y devices [`super::backend::Backend`].
 //!
-//! Provides `/dev/{stdin,stdout,null,urandom,...}`.
-
-// XXX(jayb): soon this will switch to just being {stdin,stdout,...}, so that it is _mounted_ at
-// `/dev/` rather than associated at `/`, but that will be later.
+//! Provides `{stdin,stdout,null,urandom,...}` entries, intended to be mounted at `/dev`.
 
 use alloc::string::String;
+use alloc::vec;
 use alloc::vec::Vec;
 
 use crate::LiteBox;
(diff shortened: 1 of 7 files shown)


wdcui:
GPT-5.5 reported two potential issues. I will try to review this PR tomorrow.

   1.  litebox/src/fs/composer.rs:257  — Empty composers can panic on root stat.  Composer::builder().build()  succeeds with no mounts, but  root_handle() 
       returns a virtual root and  virtual_dir_status([])  then  expect s a precomputed virtual dir that was never inserted.
   2.  litebox/src/fs/composer.rs:435  — Ancestor paths of nested mounts shadow real directories from an enclosing mounted backend. With mounts like  / 
      and  /mnt/foo , walking  /mnt  switches to  Virtual  before asking the root backend, so real  /mnt  contents are hidden and mutations under  /mnt 
      fail as read-only instead of operating on the backing backend while merging mount children.

litebox_runner_snp/src/main.rs
```diff
@@ -239,7 +239,12 @@ pub extern "C" fn sandbox_process_init(
     };
     let dev_stdio = litebox::fs::resolver::Resolver::new(
         litebox,
-        litebox::fs::devices::Devices::migration_helper_standalone_new(litebox),
+        litebox::fs::composer::Composer::builder()
```
wdcui: Why do we have similar code both here and `default_fs()` in `litebox_shim_linux/src/lib.rs`?

litebox/src/fs/devices.rs
```diff
@@ -257,10 +226,7 @@ where
         name: &str,
         flags: OFlags,
     ) -> Result<Permissioned<FileHandle>, OpenError> {
-        let dir = dir.into_typed::<Self>();
-        if dir.location != Location::Dev {
-            return Err(OpenError::PathError(PathError::NoSuchFileOrDirectory));
-        }
+        let _dir = dir.into_typed::<Self>();
```
wdcui, line 229: It looks like `into_typed` may cause panic by `expect`. It's a little hidden,and it's obvious that `into_typed` would have this kind of behavior.

wdcui (approved):
Overall the code looks good to me. Some minor comments below. Please also take a look at the potential issues reported by GPT-5.5. Thanks!

litebox_runner_snp/src/main.rs
jaybosamiya-ms: It's similar but a little different because the `default_fs` in the Linux shim does not deal with 9p.  We can probably abstract some of it out, but that's outside the scope of this particular PR.

litebox/src/fs/devices.rs
jaybosamiya-ms: Yes, it can panic, but only if you actively try to use a handle from one file system onto another.  The expectation is that any "normal" use of the file system will not actually panic, since users are not supposed to be invoking backends _directly_.  More concretely, if we have initialized the file systems correctly (in shim or runner), and no backends are actively trying to mess with each other, then no set of guest syscalls can cause a panic at these calls.

Essentially, I explicitly chose not to make these `Option` or `Result`, but instead intentionally short there.

Nonetheless, I added doc comments to each of these as to what their _intended_ use case is.

jaybosamiya-ms:
Thanks Weidong!  The two reported issues are reasonable, not critical in practice I think, but fixed since they were real issues.  I also added a little more documentation about the internal usage of `into_typed` and such, based on you noticing their potential for panicking if you misuse them.

PR #999, opened 2026-07-02
Fix error code return for OP-TEE syscall dispatch

sangho2:
This PR fixes the error code return of OP-TEE syscall dispatching. Instead of returning `Errno` (Linux), it now returns `TeeResult` (OP-TEE) to the TA/ldelf if its syscall request is invalid (e.g., nonexistent/unsupported syscall, invalid arguments, ...).


wdcui (approved):
LGMT. Thanks.

PR #1003, opened 2026-07-02
Enable `x86_64` Linux stack guard for OP-TEE TAs

sangho2:
This PR enables stack guard for OP-TEE TAs. Currently, we use `x86_64` Linux toolchains to build OP-TEE TA binaries with `-fstack-protector` (enabled by default), which expects glibc/musl stores a stack canary value in `%fs:0x28`. However, OP-TEE TAs don't use glibc/musl such that no code prepares memory for this. This PR lets the OP-TEE shim allocate, initialize (with a random number), and protect a stack guard page.

### DIFF
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -269,6 +269,7 @@ impl OpteeShim {
                 ta_entry_point: Cell::new(0),
                 ta_stack_base_addr: Cell::new(0),
                 ta_prepared: Cell::new(false),
+                #[cfg(target_arch = "x86_64")]
                 stack_guard_page_addr: Cell::new(0),
             },
         };
@@ -836,6 +837,11 @@ impl Task {
     }
 
     /// Allocate and initialize the page backing the x86-64 stack-guard slot.
+    ///
+    /// The x86-64 toolchain emits stack-protector accesses to `%fs:0x28`.
+    /// Normally glibc or musl initializes that ABI slot before application code
+    /// runs. OP-TEE TAs use neither runtime, so the shim must provide and
+    /// initialize the slot before entering a protected TA.
     fn allocate_stack_guard_page(&self) -> Result<(), ElfLoaderError> {
         use litebox::platform::CrngProvider as _;
 
@@ -853,6 +859,9 @@ impl Task {
             })?;
         let mut guard = [0u8; core::mem::size_of::<usize>()];
         self.global.platform.fill_bytes_crng(&mut guard);
+        // Terminator-canary convention (matches glibc `_dl_setup_stack_chk_guard`):
+        // zero the lowest-addressed byte of the guard to stop the overflow by C string func.
+        guard[0] = 0;
         page.copy_from_slice(loader::ta_stack::ABI_STACK_GUARD_FS_OFFSET, &guard)
             .ok_or(ElfLoaderError::InvalidStackAddr)?;
         self.sys_mprotect(page, PAGE_SIZE, ProtFlags::PROT_READ)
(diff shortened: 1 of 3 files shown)


wdcui:
@CvvT , can you also take a look at this PR? The current approach feels a little hacky to me. It would be good to see if you have any suggestions.

CvvT (commented):
Looks like the fix is about setting the canary and page permission properly; I don't see why the rest of the change (i.e., renaming stuffs) is necessary. I feel `allocate_guest_tls` is more accurate than `allocate_stack_guard_page` (the later one sounds like we allocate a guard page for stack, but the page is for tls; we just initialize the canary in it).

CvvT:
> @CvvT , can you also take a look at this PR? The current approach feels a little hacky to me. It would be good to see if you have any suggestions.

I don't have a better idea. The hardcoded offset looks fine to me (it's a fixed value per target ABI). If we cannot control how to compile the TAs, I don't think there is other way around it.

sangho2:
> Looks like the fix is about setting the canary and page permission properly; I don't see why the rest of the change (i.e., renaming stuffs) is necessary. I feel `allocate_guest_tls` is more accurate than `allocate_stack_guard_page` (the later one sounds like we allocate a guard page for stack, but the page is for tls; we just initialize the canary in it).

The reference OP-TEE version (which is outdated btw) doesn't use thread pointers at all. That's why I thought the guest TLS is misleading. I've checked the lasted OP-TEE (for Arm) and confirmed that it now has guest TLS just for compiler compatibility (like this PR). In that sense, `allocate_guest_tls` does make sense. Let me revert this name change.

sangho2:
> > @CvvT , can you also take a look at this PR? The current approach feels a little hacky to me. It would be good to see if you have any suggestions.
> 
> I don't have a better idea. The hardcoded offset looks fine to me (it's a fixed value per target ABI). If we cannot control how to compile the TAs, I don't think there is other way around it.

The latest OP-TEE (for Arm) now does this TLS allocation in user-mode `libutee` for compiler compatibility (like this PR). Our reference OP-TEE (for x86) is, unfortunately, pretty outdated and thus doesn't have this code. We could get rid of this hack once our toolchain gets updated (no ETA AFAIK).

litebox_shim_optee/src/lib.rs
```diff
@@ -839,51 +836,47 @@ impl Task {
...
-            MapFlags::MAP_PRIVATE | MapFlags::MAP_ANONYMOUS,
-            -1,
-            0,
-        )?;
-        // Store TLS address for later restoration
-        self.tls_base_addr.set(addr.as_usize());
-        self.restore_guest_tls();
+    /// The x86-64 toolchain emits stack-protector accesses to `%fs:0x28`.
+    /// Normally glibc or musl initializes that ABI slot before application code
+    /// runs. OP-TEE TAs use neither runtime, so the shim must provide and
```
praveen-pk, line 843: > OP-TEE TAs use neither runtime

Is this specific to TAs running in LiteBox?

litebox_shim_optee/src/lib.rs
```diff
@@ -839,51 +836,47 @@ impl Task {
...
     }
 
-    /// Restore the guest TLS (FS base) before entering the TA.
-    ///
-    /// FS base is cleared across VTL switches, so we must restore it before
-    /// every TA entry.
-    #[cfg(target_arch = "x86_64")]
-    fn restore_guest_tls(&self) {
+    /// Restore the guest FS base so `%fs:0x28` reads the stack guard page.
+    fn restore_stack_guard_fs_base(&self) {
```
praveen-pk, line 874: nit: As you are setting FSBase register value here, `restore` in the function name is a bit confusing.

litebox_shim_optee/src/lib.rs
sangho2: *this is a stale/draft PR.*

The main issue is that libutee we are using is unaware of x86-gcc/libc's convention, such that it does not prepare FSBase as gcc expects. Revising the devkit (or its build configuration) would be a fundamental solution.

litebox_shim_optee/src/lib.rs
sangho2: Using **restore** here because user-mode LiteBox and guests share FSBase.

PR #1011, opened 2026-07-07
Harden LVBS/HEKI module/kexec validation against TOCTOU and confused deputy

sangho2:
This PR hardens LVBS/HEKI's module and kexec validation against TOCTOU and confused-deputy attacks. Currently, there is a time window which allows VTL0 (on another core) to tamper with validated module/kexec bytes before they turn into executable bytes, including direct modification and indirect modification through VTL1 (confused deputy). This PR eliminates this race condition by rejecting overlapped or already-protected page frames and freezing frames before copy/validation.

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -453,61 +453,10 @@ type UserMutPtr<T> =
 /// Type-level marker for the VTL0 physical-pointer provider.
 pub enum Vmap {}
 
-/// Type-level marker for MSHV operations authorized to modify protected VTL0 frames.
-pub(crate) struct PrivilegedVmap;
-
 impl<const ALIGN: usize> GlobalVmapManager<ALIGN> for Vmap {
     type Manager = crate::host::LvbsLinuxKernel;
     fn manager() -> &'static Self::Manager {
         crate::platform_low()
-    }
-}
-
-impl<const ALIGN: usize> GlobalVmapManager<ALIGN> for PrivilegedVmap {
-    type Manager = PrivilegedVmap;
-    fn manager() -> &'static Self::Manager {
-        &PrivilegedVmap
-    }
-}
-
-unsafe impl<const ALIGN: usize> VmapManager<ALIGN> for PrivilegedVmap {
-    type MapInfo = LvbsPhysPageMapInfo;
-
-    unsafe fn vmap(
-        &self,
-        pages: &PhysPageAddrArray<ALIGN>,
-        perms: PhysPageMapPermissions,
-    ) -> Result<Self::MapInfo, PhysPointerError> {
-        // SAFETY: callers uphold the raw mapping contract. This provider is used only
-        // for independently authorized HEKI patch and ring-buffer writes.
-        unsafe { crate::platform_low().vmap_privileged(pages, perms) }
-    }
-
-    unsafe fn vunmap(
-        &self,
-        map_info: Self::MapInfo,
-    ) -> Result<(), (PhysPointerError, Self::MapInfo)> {
-        // SAFETY: `map_info` came from the same LVBS platform mapper and has no outstanding uses
-        // beyond the physical-pointer guard that is dropping it.
-        unsafe {
-            <crate::host::LvbsLinuxKernel as VmapManager<ALIGN>>::vunmap(
-                crate::platform_low(),
-                map_info,
-            )
-        }
-    }
-
-    fn validate_unowned(&self, pages: &PhysPageAddrArray<ALIGN>) -> Result<(), PhysPointerError> {
-        crate::platform_low().validate_unowned(pages)
-    }
-
-    unsafe fn protect(
-        &self,
-        pages: &PhysPageAddrArray<ALIGN>,
-        perms: PhysPageMapPermissions,
-    ) -> Result<(), PhysPointerError> {
-        // SAFETY: callers uphold `VmapManager::protect`; this forwards unchanged to LVBS.
-        unsafe { crate::platform_low().protect(pages, perms) }
     }
 }
 
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -1344,7 +1344,7 @@ unsafe impl<Host: HostInterface, const ALIGN: usize> VmapManager<ALIGN> for Linu
                 PhysFrame::<Size4KiB>::containing_address(x86_64::PhysAddr::new(range.start)),
                 PhysFrame::<Size4KiB>::containing_address(x86_64::PhysAddr::new(range.end)),
             );
-            crate::mshv::vsm::protect_physical_memory_range(frame_range, mem_attr)
+            crate::mshv::vsm::set_vtl0_memory_protection(frame_range, mem_attr)
                 .map_err(|_| PhysPointerError::UnsupportedPermissions(perms.bits()))?;
         }
 
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -46,12 +46,12 @@ pub mod mshv;
 
 pub mod syscall_entry;
 
-/// Mapping info returned by LVBS vmap paths. Ordinary writable mappings retain their
-/// `protected_frames` guard here until unmap.
+/// Mapping metadata. Ordinary writable mappings retain an opaque protected-frame access guard for
+/// the mapping's lifetime.
 pub struct LvbsPhysPageMapInfo {
     base: *mut u8,
     size: usize,
-    protected_frames_guard: Option<spin::rwlock::RwLockReadGuard<'static, rangemap::RangeSet<u64>>>,
+    protected_frame_access: Option<crate::mshv::vsm::ProtectedFrameAccessGuard<'static>>,
 }
 
 impl LvbsPhysPageMapInfo {
@@ -59,7 +59,7 @@ impl LvbsPhysPageMapInfo {
         Self {
             base,
             size,
-            protected_frames_guard: None,
+            protected_frame_access: None,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


litebox_platform_lvbs/src/mshv/ringbuffer.rs
```diff
@@ -108,9 +108,12 @@ fn write_fast(rb_pa: PhysAddr, size: usize, write_offset: usize, buf: &[u8]) ->
 /// after attempting the write.
 fn write_slow(rb_pa: PhysAddr, size: usize, write_offset: usize, buf: &[u8]) -> usize {
     let write_slice = |pa: PhysAddr, slice: &[u8]| -> bool {
-        Vtl0PhysMutPtr::<u8, PAGE_SIZE>::with_contiguous_pages(pa.as_u64().trunc(), slice.len())
-            .and_then(|ptr| ptr.write_slice_at_offset(0, slice))
-            .is_ok()
+        PrivilegedVtl0PhysMutPtr::<u8, PAGE_SIZE>::with_contiguous_pages(
```
wdcui, line 111: `write_slow` is a utility funciton. How can we make sure its callers meet the requirement for `PrivilegedVtl0PhysMutPtr`?

litebox_platform_lvbs/src/lib.rs
```diff
@@ -447,17 +453,63 @@ type UserMutPtr<T> =
...
+unsafe impl<const ALIGN: usize> VmapManager<ALIGN> for PrivilegedVmap {
+    type MapInfo = LvbsPhysPageMapInfo;
+
+    unsafe fn vmap(
+        &self,
+        pages: &PhysPageAddrArray<ALIGN>,
+        perms: PhysPageMapPermissions,
+    ) -> Result<Self::MapInfo, PhysPointerError> {
+        // SAFETY: callers uphold the raw mapping contract. This provider is used only
+        // for independently authorized HEKI patch and ring-buffer writes.
```
wdcui: How can we make sure it's only used for this purpose?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -465,6 +468,132 @@ pub fn mshv_vsm_load_kdata(pa: u64, nranges: u64) -> Result<i64, VsmError> {
...
+            // `protected` holds existing non-writable frames, this reservation's earlier claims, and
+            // any other concurrent reservation's in-flight claims, so this one check rejects
+            // cross-object downgrade, intra-overlap, and concurrent double-claim alike.
+            let range = start..end;
+            let status = if seen.overlaps(&range) || (start < vtl1_end && vtl1_start < end) {
+                Err(VsmError::ProtectedFrameOverlap)
+            } else {
+                classify_reservation(&owned_before, &protected, range.clone())
+            };
+            seen.insert(range.clone());
```
wdcui: This insertion can ignore the `status`?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -528,6 +657,19 @@ pub fn mshv_vsm_validate_guest_module(pa: u64, nranges: u64, _flags: u64) -> Res
         }
     }
 
+    // Reject overlap and reserve this module's frames. Any overlap is adversarial (legitimate
+    // module frames are never shared). The reservation attempts to release the frames on failure.
+    let mut frame_guard = FrameReservation::new();
+    let _ = frame_guard.reserve(module_memory_metadata.iter().map(|r| r.phys_frame_range))?;
+
+    // Write-protect the frames that will become non-writable (RX/RO) BEFORE copying, so VTL0 cannot
```
wdcui: This comment is a little confusing. How does a frame become non-writable later (in other words, do we need to write protect it again at that time)?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -600,32 +744,49 @@ pub fn mshv_vsm_free_guest_module_init(token: i64) -> Result<i64, VsmError> {
...
+    // later be applied to recycled frames.
+    if !freed_init_patch_targets.is_empty() {
+        crate::platform_low()
+            .vtl0_kernel_info
+            .precomputed_patches
+            .remove_patch_data(&freed_init_patch_targets);
+    }
+
+    result?;
     Ok(0)
```
wdcui: Nit: This coding pattern is a little confusing. Shall we change it to return based on `result`?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1333,6 +1501,42 @@ impl ModuleMemoryMetadataMap {
         map.remove(&key).is_some()
     }
 
+    /// Drop a module's freed init ranges from its metadata after [`mshv_vsm_free_guest_module_init`]
+    /// hands them back to VTL0, so a later free/unload does not re-release them. Return the patch
+    /// targets that fell within the freed init frames to drop their precomputed patches.
```
wdcui: Nit: this comment is a little confusing: "return patches ... to drop ... patches"?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1411,9 +1615,62 @@ fn copy_heki_pages_from_vtl0(pa: u64, nranges: u64) -> Option<Vec<HekiPage>> {
...
+                litebox_common_linux::vmap::PhysPointerError::InvalidPhysicalAddress(
+                    page.as_usize(),
+                ),
+            );
+        }
+    }
+    Ok(())
+}
+
+/// Update the locked registry after a protection change. Holding the exclusive guard across the
```
wdcui: Who is holding the exclusive guard?

litebox_platform_lvbs/src/mshv/vsm.rs
```diff
@@ -1424,22 +1681,47 @@ pub(crate) fn protect_physical_memory_range(
...
     mem_attr: MemAttr,
 ) -> Result<(), VsmError> {
+    apply_vtl_protection(phys_frame_range, mem_attr)
+}
+
+/// Hands a VTL0 physical memory range back to VTL0 (read-write), which also clears its protection.
+fn release_physical_memory_range(
+    phys_frame_range: PhysFrameRange<Size4KiB>,
+) -> Result<(), VsmError> {
+    apply_vtl_protection(
```
wdcui: Why making these pages read and write equals to release them to VTL0?

litebox_platform_lvbs/src/mshv/vsm.rs
wdcui, line 1: A high level question: should we move some of the physical memory manipulation code out of mshv?

wdcui (commented):
I left some comments below. I don't have a high confidence of my review. Should we have someone else to take a look at this PR as well?

litebox_platform_lvbs/src/mshv/ringbuffer.rs
sangho2: `write_slow` is a private method which is accessible only via `RingBuffer`. It only works with a pre-specified physical address range.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Let me clarify it. We write-protect a frame in advance before validation to avoid TOCTOU. If the frame is valid, it will remain as RO or turn into RX (two-staged transition). If it is invalid, it can revert back to RW only if it was RW before.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Let me replace it with `result.map(|()| 0)`.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Let me revise it. Basically, it returns an array of target addresses to drop them later. We cannot drop them here due to locks.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: Let me clarify it. This function assumes that the caller holds the write lock for `PROTECTED_FRAMES`.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: This function releases the protection. Let me rename it to avoid confusion.

litebox_platform_lvbs/src/lib.rs
sangho2: This is mostly a document-based safety claim in addition to the scope (crate-local).

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: What mshv/vsm do here is mostly changing VTL protection mask via hypercalls and tracking them. This is what mshv should do IMO. mshv/vsm doesn't directly touch page tables.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: ignoring `status` here is harmless as `seen` is just a local bookkeeping. Still, it should be better to move it after `match status` to avoid unnecessary insertion. Let me revise it.

sangho2: reordered.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: revised.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: revised.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: revised.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: code is refactored. now all these related functions become the methods of `ProtectedFrameRegistry`.

litebox_platform_lvbs/src/mshv/vsm.rs
sangho2: renamed: `unprotect_physical_memory_range`.

wdcui (approved):
LGTM. Thanks.

PR #1012, opened 2026-07-08
Cache stdio TTY state

jaybosamiya-ms:
This PR caches TTY state of stdio early for the Linux platform, so that we do not need to depend upon it for seccomp purposes later


CvvT (approved):
LGTM, thanks!

PR #1013, opened 2026-07-08
Improve stdio tty ioctl + polling behavior

jaybosamiya-ms:
This PR tracks stdio stream identity to better manage ioctl (scoped only to tty now), and to better handle polling of the streams

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -377,10 +377,16 @@ impl<FS: ShimFS> syscalls::file::FilesState<FS> {
             .unwrap();
         let mut dt = global.litebox.descriptor_table_mut();
         let mut rds = self.raw_descriptor_store.write();
-        for (raw_fd, fd) in [(0, stdin), (1, stdout), (2, stderr)] {
+        for (raw_fd, fd, stream) in [
+            (0, stdin, litebox::platform::StdioStream::Stdin),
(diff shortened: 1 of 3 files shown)


wdcui:
gpt-5.6 found the following two potential issues. Should they be addressed in this PR?

1.  file.rs:1952-1957  — Reopened  /dev/stdin ,  /dev/stdout , or  /dev/stderr  lose  StdioStream  metadata, causing incorrect  ENOTTY  and poll behavior.
2.  epoll.rs:134-144  — Stdin is always reported readable, bypassing timeouts and potentially causing busy loops or blocking reads.

jaybosamiya-ms:
The first concern somewhat related but also kinda out of scope for this PR. We don't (currently) have a "backend level" metadata in our metadata handling, we only have 2-level metadata (at fd level, and at entry level).  We _could_ hack it in via a normalized check for `/dev/{stdin, ...}` during `open`, but that is fragile, since it breaks as soon as we introduce symlinks.  The correct fix is to have a backend-level metadata layer, but yeah, mostly out of scope for this PR.  I'll add a quick comment to the code to document this though.

The second one: this is what the TODO is documenting already; this PR is not changing the (existing) behavior of using dummy events, it is simply just passing _correct_ dummy events instead of the incorrect ones from before.

wdcui (approved):
LGTM. Thanks.

PR #1014, opened 2026-07-08
Handle pselect signal masks

jaybosamiya-ms:
This PR fixes the type of `SigSetPack` to actually use the pointer (see the [`pselect6` part of `man 2 pselect`](https://linux.die.net/man/2/pselect#:~:text=The%20final%20argument,a%20system%20call.)), and adds support for its use as a temporary signal mask.


CvvT (approved):
LGTM, thanks!

PR #1017, opened 2026-07-09
Add Python REPL test + reenable previously-disabled test

jaybosamiya-ms:
The Python test has been disabled on the CI for many months (see #484).  We've resolved other CI flakiness since then, so hopefully re-enabling it works now.  If it gets flaky again, we will at least get a heads up, and can work on debugging it again.

Additionally, our Python test only used to check non-interactive usage of Python.  Essentially, we had no TTY tests.  This PR adds a TTY test so that going forward, that remains checked.


jaybosamiya-ms:
Currently blocked on #1013 for being merged, but otherwise ready for review

wdcui (approved):
LGTM. Thanks!

PR #1025, opened 2026-07-13
Add OP-TEE ShortBuffer required-size handling

sangho2:
This PR adds the required-size handling for OP-TEE ShortBuffer error paths. OP-TEE OS returns a required buffer size when it returns a ShortBuffer error to a TA, but the current OP-TEE shim doesn't return this information.


wdcui (approved):
LGTM. Thanks!

PR #1026, opened 2026-07-13
Add IDK_S PTA

sangho2:
This PR introduces the IDK_S PTA. It provides a method which endorses TA-provided data with a flat structure signed by an IDK_S key. A wire format is `TA_DATA || REPORT (MAGIC, VERSION, TA_DATA_LEN, TA_UUID, ..., TA_SIGNING_CERT_DER) || SIGNATURE_PARAMS || SIGNATURE`, where `TA_DATA_LEN` is an observed length of `TA_DATA`. All other values are from the LiteBox/PTA and `SIGNATURE` covers all except for itself.

### DIFF
--- a/litebox_shim_optee/src/idk.rs
+++ b/litebox_shim_optee/src/idk.rs
@@ -1,11 +1,23 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-use crate::NormalWorldMutPtr;
-use litebox::{mm::linux::PAGE_SIZE, platform::CrngProvider, utils::TruncateExt};
+use crate::{NormalWorldMutPtr, Task, UserConstPtr, UserMutPtr, syscalls::Cleanup};
+use alloc::vec::Vec;
+use litebox::{
+    mm::linux::PAGE_SIZE,
+    platform::{CrngProvider, RawConstPointer as _, RawMutPointer as _},
+    utils::TruncateExt,
+};
 use litebox_common_linux::errno::Errno;
+use litebox_common_optee::{
+    PTA_DEFAULT_FLAGS, TaFlags, TeeParamType, TeeResult, TeeUuid, UteeParams,
+};
 use num_enum::TryFromPrimitive;
-use p384::{NonZeroScalar, elliptic_curve::sec1::ToEncodedPoint};
+use p384::{
+    NonZeroScalar,
+    ecdsa::{Signature, SigningKey, signature::Signer},
+    elliptic_curve::sec1::ToEncodedPoint,
+};
 use spin::Once;
 use zeroize::Zeroizing;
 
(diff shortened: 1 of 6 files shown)


litebox_shim_optee/src/idk.rs
```diff
@@ -14,11 +28,155 @@ const IDENTITY_SIGNING_PUBLIC_KEY_LEN: usize = 97;
...
+    endorsement.extend_from_slice(IDKS_ENDORSEMENT_MAGIC);
+    endorsement.extend_from_slice(&IDKS_ENDORSEMENT_VERSION.to_le_bytes());
+    endorsement.extend_from_slice(ta_data);
+    endorsement.extend_from_slice(&ta_uuid.to_le_bytes());
+    endorsement.extend_from_slice(&ta_svn.to_le_bytes());
+    endorsement.extend_from_slice(ta_digest);
+    endorsement.push(IDKS_DEBUG_FLAG);
+    endorsement.extend_from_slice(ISOLATION_SOLUTION);
+    Some(endorsement)
+}
```
sangho2, line 164: This function would be the most important one within this PR. Need to decide the data to include and the claim format. Don't prefer to use any specific format (e.g., C-struct, JSON, CBOR/COSE, ...) here because it heavily depends on what the verifier expects.

sangho2:
~~Prototyped the KeyIso claim structure as well.~~

litebox_shim_optee/src/idk.rs
```diff
@@ -149,25 +611,182 @@
...
+
     #[test]
-    fn identity_signing_private_key_signs_and_verifies_message() {
-        use crate::syscalls::tests::init_platform;
-        use p384::ecdsa::{
-            Signature, SigningKey, VerifyingKey,
-            signature::{Signer, Verifier},
+    fn keyiso_claim_has_expected_layout() {
+        let ta_data = b"TA data";
+        let nonce = [0x5a; 32];
```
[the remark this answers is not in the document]
sangho2: This nonce is on a test path.

litebox_shim_optee/src/idk.rs
```diff
@@ -149,25 +611,182 @@
...
+        assert_eq!(offset, endorsement.len());
+    }
+
+    #[test]
+    fn keyiso_signature_covers_entire_claim_prefix() {
+        use p384::ecdsa::{Signature, VerifyingKey, signature::Verifier};
+
+        let mut private_key = [0u8; IDENTITY_SIGNING_PRIVATE_KEY_LEN];
+        private_key[IDENTITY_SIGNING_PRIVATE_KEY_LEN - 1] = 1;
+        let nonce = [0x5a; 32];
```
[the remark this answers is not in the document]
sangho2: This nonce is on a test path.

sangho2:
- [ ] Drop `TaDigest`
- [x] Add a placeholder for cert/pubkey
- [x] Embedded vs dynamically loaded
- [ ] Support CBOR/COSE? (https://crates.io/crates/coset)

sangho2:
`TA_SIGNING_CERT_LEN` and `TA_SIGNING_CERT_DER` are technically redundant if a TA is embedded because it isn't signed. However, for now, we do not make them optional.

litebox_shim_optee/src/idk.rs
```diff
@@ -15,11 +28,200 @@ const KEY_ALGORITHM_MASK: u64 = 0xff00;
...
+
+    fn endorse_data<Platform: crate::OpteeShimPlatform>(
+        task: &Task<Platform>,
+        params: &mut UteeParams,
+    ) -> Result<(), TeeResult> {
+        use TeeParamType::{MemrefInput, MemrefOutput, None};
+        // [in]  params[0].memref.buffer   Opaque TA data
+        // [in]  params[0].memref.size     TA data size
+        // [out] params[1].memref.buffer   Output buffer for signed endorsement
+        // [out] params[1].memref.size     Buffer size
```
sangho2, line 95: Since a signed endorsement is just an appendix now, we don't have to copy the received TA data back to the output buffer if the TA concatenates these two by itself (performance optimization).

litebox_shim_optee/src/idk.rs
```diff
@@ -15,11 +28,200 @@ const KEY_ALGORITHM_MASK: u64 = 0xff00;
...
+    endorsement.extend_from_slice(&ta_uuid.to_le_bytes());
+    endorsement.extend_from_slice(&ta_svn.to_le_bytes());
+    endorsement.extend_from_slice(ta_digest);
+    endorsement.push(u8::from(ta_dynamic));
+    endorsement.push(IDKS_DEBUG_FLAG);
+    endorsement.extend_from_slice(ISOLATION_SOLUTION);
+    endorsement.extend_from_slice(&cert_len.to_le_bytes());
+    endorsement.extend_from_slice(ta_signing_cert);
+    Some(endorsement)
+}
```
thara-gopinath, line 253: So.. I don't think we can have ta_data as the first element. It will be impossible to parse the claims structure. I propose MAGIC || VERSION || TA_DATA_LEN || TA_UUID || TA_SVN || TA_DIGEST ||
/// TA_DYNAMIC || DEBUG || ISOLATION_SOLUTION || TA_SIGNING_CERT_LEN ||
/// TA_SIGNING_CERT_DER || TA_DATA || SIGNATURE

Now I re-read the Attestation doc from EnS multiple times ! I think they are also looking for some kind of info on the signing cert (IDK_S) itself. Like the type of algorithm/ hash algorithm etc which makes sense so that we can switch to PQC later. The doc talks about following structure for this

""""
typedef struct _KEYISO_ATTESTATION_SIGNATURE_PARAMS
{
ULONG Magic; // KEYISO_ATTESTATION_SIGNATURE_PARAMS_MAGIC
ULONG Version; // KEYISO_ATTESTATION_SIGNATURE_PARAMS_CURRENT_VERSION
ULONG cbAlgID; // Length of algorithm identifier string
ULONG cbAlgPramas; // Length of algorithm parameters structure
ULONG cbHashAlg; // Length of hash algorithm name string
// UCHAR AlgID[cbAlgID] -- Algorithm name (e.g., "ECDSA_P384")
// UCHAR AlgParams[cbAlgPramas] -- Algorithm params (e.g., KEYISO_ATTESTATION_ECC_SIGNATURE_PARAMS)
// UCHAR HashAlg[cbHashAlg] -- Hash algorithm name (e.g., "SHA384") } KEYISO_ATTESTATION_SIGNATURE_PARAMS, * PKEYISO_ATTESTATION_SIGNATURE_PARAMS;

For LVBS, the IDKS key is an ECDSA P-384, so:
• AlgID: "ECDSA_P384"
• AlgParams: The ECC signature params structure KEYISO_ATTESTATION_ECC_SIGNATURE_PARAMS, contains only the Magic and Version - no additional parameters beyond the curve identified in AlgID.
• HashAlg: "SHA384"
"""""

But I am not sure if we want to follow a structure like this. Just defining the relevant fields might also suffice. Like char idk_alg_name[] and char idk_hash_name[] ? I don't know if these should be fixed length strings or we need to add fields to specify the size

litebox_shim_optee/src/idk.rs
```diff
@@ -15,11 +28,200 @@ const KEY_ALGORITHM_MASK: u64 = 0xff00;
...
+    let ta_data_len = u32::try_from(ta_data.len()).ok()?;
+    let cert_len = u32::try_from(ta_signing_cert.len()).ok()?;
+    let mut endorsement = Vec::with_capacity(capacity);
+    endorsement.extend_from_slice(ta_data);
+    endorsement.extend_from_slice(IDKS_ENDORSEMENT_MAGIC);
+    endorsement.extend_from_slice(&IDKS_ENDORSEMENT_VERSION.to_le_bytes());
+    endorsement.extend_from_slice(&ta_data_len.to_le_bytes());
+    endorsement.extend_from_slice(&ta_uuid.to_le_bytes());
+    endorsement.extend_from_slice(&ta_svn.to_le_bytes());
+    endorsement.extend_from_slice(ta_digest);
```
thara-gopinath, line 229: Do we need a field for ta_ digest size ?

litebox_shim_optee/src/idk.rs
thara-gopinath: Isn't TA_DATA part of the endorsement claim in params1 ?

litebox_shim_optee/src/idk.rs
sangho2: - This change is based on the discussion with Dikla. My understanding is that, `TA_DATA` will internally has a length field to recognize its end.
- That's what our old commit implemented :) If we want to have them, it might be better to just use EnS's KeyIso structure, instead of inventing our own thing.

litebox_shim_optee/src/idk.rs
sangho2: For now, we only use SHA-256, whose length is fixed. it depends on whether we want to support other hash algorithms like SHA-384 and SHA-512.

litebox_shim_optee/src/idk.rs
sangho2: If we don't have `MAGIC` and `VERSION` prefix, this claim is technically fully decoupled from `TA_DATA` itself. TA can concatenate them. Anyhow, this is just perf optimization.

litebox_shim_optee/src/idk.rs
thara-gopinath: I think we should add this  . It is very possible that we want to support SHA-384 or SHA-512 later. The attestation verifier will have to change then

litebox_shim_optee/src/idk.rs
thara-gopinath: Hey.. yes. From yesterday's discussion, they don't care. But I think another TA could. And if we change it then the remote verifier will also have to be changed. I think we should make it parsable

litebox_runner_lvbs/src/lib.rs
```diff
@@ -1413,6 +1417,7 @@ fn write_rpc_args_to_normal_world(
 const LDELF_BINARY: &[u8] = &[0u8; 0];
 const TA_BINARY: &[u8] = &[0u8; 0];
 const TA_BINARIES: &[&[u8]] = &[TA_BINARY];
+const TA_SIGNING_CERT_DER: &[u8] = &[0u8; 0];
```
thara-gopinath: So I think Angelina has a different way of sending the cert. But let us check in your PR and then Angelina can fix it up with her signing PR

litebox_shim_optee/src/idk.rs
sangho2: I decided to provide two commands which simply have difference placements of `TA_DATA`. In terms of parseability, I think the verifier can still parse the claim anyway because it knows golden UUID, SVN, TA_DIGEST, ... which will work as anchors.

sangho2: Question is more about whether we want to implement table-based one (as EnS expects) or use the current flat layout.

litebox_shim_optee/src/idk.rs
sangho2: Added fields for digest algorithm and lengths.

litebox_runner_lvbs/src/lib.rs
sangho2: Yes. This is a placeholder for now.

PR #1029, opened 2026-07-15
Add broker-backed pipes

wdcui:
This PR adds broker-backed pipes, with the broker owning pipe state and endpoint lifetime. The existing in-process pipe remains available when no broker is configured.


wdcui:
Superseded by the focused PR stack #1041, #1042, #1043, #1044, and #1045.

PR #1028, opened 2026-07-14
Enable concurrent mappings of shared normal-world/VTL0 physical pages

sangho2:
This PR enables concurrent mappings of shared normal-world (VTL0) physical pages. Currently, our physical pointer abstraction uses direct/identity maps for contiguous physical pages, rejecting concurrent mappings. PR #979 bypasses this problem through *over*-serialization. Instead, this PR solves it by making the physical pointer abstraction always use `vmap`, returning private virtual addresses for shared physical pages.

PR #1031, opened 2026-07-16
Use canonical address for cleanup

athvu:
This PR changes the cleanup_page_table_frames function to calculate the end address using KERNEL_OFFSET directly so that the address is canonical.

### DIFF
--- a/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
+++ b/litebox_platform_lvbs/src/arch/x86/mm/paging.rs
@@ -42,9 +42,6 @@ const PML4_SHIFT: u32 = 39;
 /// Mask for a 9-bit page-table index (512 entries per table).
 const PML4_INDEX_MASK: u64 = 0x1FF;
 
-/// Number of bytes of virtual address space covered by one PML4 slot (512 GiB).
-const PML4_SLOT_SIZE: u64 = 1 << PML4_SHIFT;
-
 /// PML4 index of the first VTL1-kernel slot (`PA + KERNEL_OFFSET`).
 ///
 /// Only slots `>= KERNEL_PML4_START` are safe to share between page tables:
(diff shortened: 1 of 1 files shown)


sangho2:
Oh. PR #1028 does include a fix for this issue (https://github.com/microsoft/litebox/pull/1028/changes#diff-88109e004129104c151bad9c778ad09fd0ef7a63bf21e60837b918f2970e4ffcR305-R320). A separate PR like this one should be better. One difference is that my PR explicitly handles two canonical ranges, while yours relies on the API itself?

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
sangho2, line 314: It'd better to add a comment that this function crosses the non-canonical hole.

litebox_platform_lvbs/src/arch/x86/mm/paging.rs
```diff
@@ -308,7 +308,7 @@ impl<M: MemoryProvider, const ALIGN: usize> X64PageTable<'_, M, ALIGN> {
         // `0 ..= KERNEL_PML4_START * PML4_SLOT_SIZE - 1`. The kernel region at
         // and above `KERNEL_PML4_START` is base-owned/shared.
         let start = Page::<Size4KiB>::from_start_address(VirtAddr::new(0)).unwrap();
-        let end = Page::<Size4KiB>::containing_address(VirtAddr::new(
+        let end = Page::<Size4KiB>::containing_address(VirtAddr::new_truncate(
```
sangho2: `VirtAddr::new(crate::KERNEL_OFFSET - 1)` would be better.

athvu:
Thanks for the review @sangho2. Should I change this PR to handle the low and high canonical halves separately or leave it as is crossing the non-canonical hole?

sangho2:
> Thanks for the review @sangho2. Should I change this PR to handle the low and high canonical halves separately or leave it as is crossing the non-canonical hole?

I think there is no technical difference, so up to you (both have pros and cons). I'll approve this PR. Thanks!

sangho2 (approved):

wdcui:
@sangho2 , should this commit be cherry picked to ulitebox?

sangho2:
> @sangho2 , should this commit be cherry picked to ulitebox?

Yes. that's my plan.

PR #1040, opened 2026-07-17
Migrate Tar-RO file system to new backend

jaybosamiya-ms:
This PR switches our Tar-RO filesystem to the new core file system design (see #887).  Concretely, it adds a `TarRo` backend, migrates all old usage of `tar_ro::FileSystem` to a resolver-backed one to use the new `TarRo` backend, and then removes the old `tar_ro::FileSystem`.  It also makes minor tweaks to the `Backend` interface to correctly account for read-only file systems.  Interestingly, this move to the backend-based handling also gives us a chance to clean up the hardcoded dir stuff inside the tar-ro to instead properly use inode numbering :)

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -15,6 +15,7 @@
 
 extern crate alloc;
 
+use alloc::borrow::Cow;
 use alloc::vec;
 use alloc::vec::Vec;
 
@@ -62,7 +63,7 @@ pub(crate) type LinuxFS<Platform> = litebox::fs::layered::FileSystem<
     litebox::fs::layered::FileSystem<
         Platform,
         litebox::fs::resolver::Resolver<Platform, litebox::fs::composer::Composer>,
-        litebox::fs::tar_ro::FileSystem<Platform>,
+        litebox::fs::resolver::Resolver<Platform, litebox::fs::composer::Composer>,
     >,
 >;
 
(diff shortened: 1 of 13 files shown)


jaybosamiya-ms:
This PR is easiest to review one commit in it at a time, rather than the whole thing at once.  Also, just noting that semver-checks should (expectedly) mention a warning about the `Backend` trait having been tweaked + also `tar_ro::FileSystem` will get marked as disappeared, both are expected.

wdcui:
GPT-5.6 found two potential issues. I will add my review soon.

1. Explicit directories can overwrite directory nodes —  litebox/src/fs/tar_ro.rs:397 
A  dir/  tar entry may replace the directory created for  dir/file , causing descendant lookups to fail with  ENOTDIR .
2. Tar directory  .  and  ..  entries now have inode 0 —  litebox_shim_linux/src/lib.rs:347 
 Resolver::read_dir  supplies no inode metadata for synthesized dot entries, changing prior behavior and potentially causing consumers to ignore them.

jaybosamiya-ms:
Thanks Weidong, for those two: 
1. `tar_no_std` does not give us dir entries (see https://docs.rs/tar-no-std/latest/tar_no_std/struct.ArchiveEntry.html).  This is not a concern.
2. this is expected, I already have this documented in the test. it is a resolver-level concern that needs to be fixed up there: https://github.com/microsoft/litebox/blob/10be5936c1b6d79b283d1cc6a4a2f2facc8f5545/litebox/src/fs/tests.rs#L1057-L1062

litebox_shim_linux/src/lib.rs
```diff
@@ -343,13 +344,22 @@ fn default_fs(
             .build()
             .unwrap(),
     );
+    let tar_ro = litebox::fs::resolver::Resolver::new(
+        litebox,
+        litebox::fs::composer::Composer::builder()
+            .mount("/", |allocator| {
```
wdcui, line 396: In the future, should we handle the case where it's not mounted at "/"?

wdcui (approved):
LGTM. Thanks!

litebox_shim_linux/src/lib.rs
jaybosamiya-ms: The resolver currently _already_ supports handling it at non-`/` mount, I am setting it to `/` just to maintain consistency with existing code :)

PR #1050, opened 2026-07-19
Enable shared broker pipes over Unix sockets

wdcui:
## Summary
- create one sealed memfd per broker pipe and transfer it with SCM_RIGHTS
- map separate read and write staging regions on both control-channel peers
- allow only the Linux-userland syscalls needed to receive and validate attachments
- cover two-way mapping visibility and the runner broker integration path

## Stack
1. #1049 — protocol foundation
2. #1048 — broker-host staging
3. #1051 — broker-local data path
4. **#1050** — Unix transport enablement

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1515,6 +1515,7 @@ dependencies = [
 name = "litebox_broker_transport"
 version = "0.1.0"
 dependencies = [
+ "libc",
  "litebox_broker_protocol",
 ]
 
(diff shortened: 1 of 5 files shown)


wdcui:
Superseded by #1052. The generic Unix shared-memory transport now lives in #1049, while #1052 owns the shared-memory-only broker-pipe protocol and data path.

PR #1051, opened 2026-07-19
Add broker-local shared pipe data path

wdcui:
## Summary
- use attached staging memory for bounded broker pipe reads and writes
- retain inline operations when the channel has no shared-memory attachment
- manage per-endpoint staging cursors, preallocation, and close lifecycle

## Stack
1. #1049 — protocol foundation
2. #1048 — broker-host staging
3. **#1051** — broker-local data path
4. #1050 — Unix transport enablement

### DIFF
--- a/litebox_broker_local/src/lib.rs
+++ b/litebox_broker_local/src/lib.rs
@@ -20,6 +20,9 @@ mod error;
 mod event;
 mod pipe;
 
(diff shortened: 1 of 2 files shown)


wdcui:
Superseded by #1052, which makes shared memory the only broker-pipe data path and consolidates the host, local, and protocol changes.

PR #1054, opened 2026-07-20
fix(net): accept advisory socket options (IP_TOS, SO_RCVBUF, SO_SNDBUF)

willportnoy:
setsockopt returned EOPNOTSUPP for IP_TOS, SO_RCVBUF and SO_SNDBUF, which Node/libuv treat as fatal — Socket.setTypeOfService and TLS buffer sizing raise uncaught exceptions that tear down the connection. These options are advisory hints, so accept them silently and keep the fixed internal buffer size that getsockopt already reports, for both INET and UNIX sockets. This matches how unprivileged sockets behave on native Linux.

### DIFF
--- a/litebox_shim_linux/src/syscalls/net.rs
+++ b/litebox_shim_linux/src/syscalls/net.rs
@@ -385,11 +385,15 @@ impl<FS: ShimFS> GlobalState<FS> {
 
         match optname {
             SocketOptionName::IP(ip) => match ip {
-                // IP_TOS is an advisory traffic-class hint. Accept it silently
-                // (we don't propagate the bit anywhere) instead of returning
-                // EOPNOTSUPP, which Node's Socket.setTypeOfService treats as
-                // fatal and cascades into tearing down the connection.
-                litebox_common_linux::IpOption::TOS => return Ok(()),
+                // IP_TOS is an advisory traffic-class hint. Accept it (we don't
+                // propagate the bit anywhere) instead of returning EOPNOTSUPP,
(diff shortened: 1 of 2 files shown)


litebox_shim_linux/src/syscalls/net.rs
```diff
@@ -385,7 +385,11 @@ impl<FS: ShimFS> GlobalState<FS> {
 
         match optname {
             SocketOptionName::IP(ip) => match ip {
-                litebox_common_linux::IpOption::TOS => return Err(Errno::EOPNOTSUPP),
+                // IP_TOS is an advisory traffic-class hint. Accept it silently
+                // (we don't propagate the bit anywhere) instead of returning
+                // EOPNOTSUPP, which Node's Socket.setTypeOfService treats as
+                // fatal and cascades into tearing down the connection.
+                litebox_common_linux::IpOption::TOS => return Ok(()),
```
CvvT: Maybe add `litebox_util_log::debug!` to all three places so that it's visible? A minor concern I have is that some applications may break if we just silently accept these.

CvvT (approved):
LGTM, thanks!

litebox_shim_linux/src/syscalls/net.rs
willportnoy: Good call — done in a2182c49. Added `litebox_util_log::debug!` at all three accept sites so an accepted-but-ignored option stays visible in the logs if an app misbehaves.

PR #1067, opened 2026-07-22
Pre-set ACCESSED and DIRTY flags in page table entries

dywongcloud:
Pre-set the ACCESSED and DIRTY flags when creating page table entries, matching Linux’s `_KERNPG_TABLE` behavior. This avoids atomic read-modify-write operations by the CPU page-table walker on first access.


CvvT (changes requested):
Hi @dywongcloud Thanks for breaking down your previous PR. This one look reasonable to me. One minor concern is the long PR message that is unnecessary and makes `git log` harder to read. I would appreciate it if you could shorten it. For this PR, only keeping the text from summary should be fine. We would like to keep a consistent style. Thanks!

dywongcloud:
> Hi @dywongcloud Thanks for breaking down your previous PR. This one look reasonable to me. One minor concern is the long PR message that is unnecessary and makes `git log` harder to read. I would appreciate it if you could shorten it. For this PR, only keeping the text from summary should be fine. We would like to keep a consistent style. Thanks!

Hi I shortened it! Please let me know!

dywongcloud:
sorry i accidentally closed it but I reopened it

CvvT (commented):
LGTM, thanks!

CvvT (approved):

Issue #514, opened 2025-11-21
gsbase-less kernel


sangho2:
> > Side note, Coconut avoids use of gsbase in kernel mode altogether and instead has a per-CPU region of the kernel-mode address space (and per-CPU kernel-mode page tables). Interesting design choice...
> 
> Interesting. Our LiteBox kernel platform is still WiP, so we can adopt coconut's idea if it makes sense. Using `gsbase` and `swapgs` is not a strict requirement (might have a minor performance impact, though).

I'm trying to get more information about why they made this design choice. I think with their current design it's less efficient than using `gsbase`, but I don't think that has to be the case...

_Originally posted by @jstarks in https://github.com/microsoft/litebox/issues/510#issuecomment-3559546285_

PR #1087, opened 2026-07-24
Make the VTL1 kernel gsbase-less (#514)

dywongcloud:
Closes issue #514.

Remove all kernel-mode GSBASE/swapgs usage from VTL1, replacing it with RSP-based per-CPU lookup and per-core LSTAR syscall stubs. Also removes obsolete GSBASE infrastructure and fixes a vtl_switch RSP bug. ABI unchanged..


sangho2:
Hi @dywongcloud, Thank you for the contribution. Since we have not yet decided whether the VTL1 kernel will be GSBASE-less, I'd like to defer changes related to this area for now. Thus, I'm going to close this PR. We can revisit it once the design direction is settled.

I took a quick look at the PR and saw that the page table flag changes overlap with the changes in your previous PR. That might be helpful. 

Thanks again.

PR #1090, opened 2026-07-25
Retire host-backed broker objects exactly once

wdcui:
Events and pipes keep all of their state inside `litebox_broker_core`, so dropping the last reference to one releases everything it owned; a socket will not, since its state is a host descriptor and a poll registration that only the broker host can release. This adds the seam a socket will need: an opaque `HostResourceId`, a per-session retirement queue, and a host-backed object entry whose drop records its identity for the host to release. Retirement is deferred because references drop under the process-wide reference lock, where releasing a host resource would hold a spin lock across a syscall, and the queue reserves capacity once so recording never allocates inside `Drop`. What bounds the queue is a charge taken when a resource is named and released when the host takes it back, since a close during an in-flight operation leaves a resource in neither the reference table nor the queue. Nothing names a host resource yet.

### DIFF
--- a/litebox_broker_core/src/event.rs
+++ b/litebox_broker_core/src/event.rs
@@ -25,7 +25,7 @@ pub fn add(session: &BrokerSession, handle: ObjectHandle, value: u64) -> Result<
     let required_rights = ObjectRights::WRITE;
     session.with_authorized_object_mut(handle, required_rights, |object| match object {
         ObjectEntry::Event(event) => event.add(value),
-        ObjectEntry::Pipe(_) => Err(BrokerError::InvalidRights),
+        ObjectEntry::Pipe(_) | ObjectEntry::HostBacked(_) => Err(BrokerError::InvalidRights),
     })
 }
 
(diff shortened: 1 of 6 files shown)


wdcui:
Closing unmerged. The network plan's PR 4 scope says to "introduce the concrete host-resource identity/lifecycle seam together with the first socket consumer rather than a general host-object framework", and splitting the seam into its own PR produced exactly the general framework that line warns against — a `host_resource` module abstracting over a single planned consumer, with no natural home for the object's `Drop` until `socket.rs` exists.

This work lands instead alongside the core socket object, with the object and its `Drop` in `socket.rs` (mirroring `PipeObject`/`PipeCapacityReservation` in `pipe.rs`) and the per-session retirement queue in `session.rs` next to `BrokerSession`. The socket protocol DTOs and wire codec go first as a self-contained protocol-only PR.

The four fixes review found here carry over: charging a resource at naming rather than counting reference-table membership (an in-flight operation holds an object after its table entry is removed, so the queue could outgrow its reserved capacity), returning `InvalidRights` rather than `UnsupportedOperation` from readiness (which maps to a `panic!` and would abort the guest), leasing the resource identity rather than returning it (a returned identity outlives the object protecting it), and reserving the queue before taking the session-id lock rather than under it.

PR #1091, opened 2026-07-25
Define the broker socket protocol

wdcui:
Adds the socket message family to the broker protocol crate: typed address family, socket type, IP protocol, address, shutdown, and message-flag values, the six request and response pairs for create, connect, send, receive, shutdown, and status, and their wire codec. Send and receive carry shared-buffer descriptors rather than inline bytes, matching pipes, and every value is a closed set instead of a passthrough integer, so a local endpoint cannot name an address family, protocol, or flag the broker has not agreed to support. Send and receive flags are separate types because their flag sets are disjoint, so a receive-only flag such as `PEEK` cannot be named on a send. Connection failures use a separate `SocketError` rather than widening `ErrorCode`, because `ErrorCode` maps to a guest panic for codes a local endpoint should never see, while network failures are ordinary outcomes. Nothing serves these operations yet: the broker host answers a socket request with `UnsupportedOperation` until the core gains a socket object, which follows in the next change together with host-resource retirement.

### DIFF
--- a/litebox/src/event/counter.rs
+++ b/litebox/src/event/counter.rs
@@ -485,7 +485,9 @@ mod tests {
                 BrokerOperation::CheckReadiness(_) => {
                     BrokerResult::Readiness(ReadinessFlags::WRITE)
                 }
-                request @ (BrokerOperation::Event(_) | BrokerOperation::Pipe(_)) => {
+                request @ (BrokerOperation::Event(_)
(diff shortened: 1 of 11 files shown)


wdcui:
Closing this while we establish a clean portable transport/platform boundary first. The branch is retained so the socket protocol work can be revised and reused afterward.

PR #1098, opened 2026-07-27
Bound broker resource budgets per session

dywongcloud:
BrokerCoreLimits is global to the broker core, but one broker process serves every client association.


dywongcloud:
@sangho2

CvvT:
The broker is still work in progress.

PR #1104, opened 2026-07-29
Add KeyStack PTA

sangho2:
This PR introduces a KeyStack PTA which derives a stack of keys based on a requesting TA's UUID, Secure Version Number (SVN), and extra data, enabling anti-rollback key derivation.

### DIFF
--- a/litebox_shim_optee/src/keystack.rs
+++ b/litebox_shim_optee/src/keystack.rs
@@ -1,0 +1,192 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//! Implementation of the SVN key stack pseudo TA.
+//!
+//! This PTA derives a stack of TA-unique keys, one per Secure Version Number (SVN),
+//! so that a TA running at SVN `n` can unseal data sealed by any of its older
+//! versions while remaining unable to derive the keys of any newer version.
+
+use crate::syscalls::Cleanup;
+use crate::syscalls::pta::{
+    HmacSha256, PTA_DEFAULT_FLAGS, TA_DERIVED_EXTRA_DATA_MAX_SIZE, TA_DERIVED_KEY_MAX_SIZE,
+    TA_DERIVED_KEY_MIN_SIZE, huk_subkey_derive, open_default_pta_session,
+};
+use crate::{Task, UserConstPtr, UserMutPtr};
+use alloc::{vec, vec::Vec};
+use hmac::Mac;
+use litebox::platform::{RawConstPointer as _, RawMutPointer as _};
+use litebox::utils::TruncateExt;
+use litebox_common_optee::{HukSubkeyUsage, TaFlags, TeeParamType, TeeResult, TeeUuid, UteeParams};
+use num_enum::TryFromPrimitive;
+use zeroize::Zeroizing;
+
+/// Maximum number of keys in SVN key stack.
+const SVN_KEY_STACK_MAX_SIZE: u32 = 4096;
+
+pub(crate) struct KeyStackPta;
+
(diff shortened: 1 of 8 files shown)


sangho2:
Replace #831 with this PR to maintain the compatibility with the mainline OP-TEE OS.

litebox_shim_optee/src/keystack.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub(crate) const FLAGS: TaFlags = PTA_DEFAULT_FLAGS.union(TaFlags::CONCURRENT);
+
+    // TODO: Replace this placeholder with the UUID agreed upon with the
+    // consuming TA before this PTA is treated as a stable interface.
+    pub(crate) const UUID: TeeUuid = TeeUuid {
+        time_low: 0x978a_f7a7,
+        time_mid: 0x074f,
+        time_hi_and_version: 0x4f59,
+        clock_seq_and_node: [0xb3, 0xae, 0x33, 0xa5, 0x93, 0xc1, 0xd4, 0x88],
+    };
```
sangho2, line 46: UUID

litebox_shim_optee/src/syscalls/tests.rs
```diff
@@ -66,3 +79,309 @@
...
+        "C reference produced {} lines for {} cases",
+        c_output.len(),
+        cases.len()
+    );
+
+    let mut derived = 0;
+    let mut rejected = 0;
+    for (case, expected) in cases.iter().zip(c_output.iter()) {
+        let actual = run_rust_implementation(case);
+        assert_eq!(
```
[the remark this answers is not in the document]
sangho2: This is public information.

litebox_shim_optee/src/syscalls/pta.rs
```diff
@@ -56,28 +60,46 @@ impl PseudoTa {
...
-const PTA_DEFAULT_FLAGS: TaFlags = TaFlags::SINGLE_INSTANCE
+pub(crate) const PTA_DEFAULT_FLAGS: TaFlags = TaFlags::SINGLE_INSTANCE
     .union(TaFlags::MULTI_SESSION)
     .union(TaFlags::INSTANCE_KEEP_ALIVE);
 
 const MAX_PTA_SESSIONS_PER_TASK: usize = 100;
 
+/// Open a session to a PTA that carries no per-session state and takes no
+/// parameters at session-open time.
+pub(crate) fn open_default_pta_session(params: &UteeParams) -> Result<u32, TeeResult> {
```
praveen-pk: `default_pta` is a bit confusing. This implies a Default PTA that would be invoked.

Looking at the definition, this seems like some setup/validation.

litebox_shim_optee/src/keystack.rs
```diff
@@ -0,0 +1,192 @@
...
+        let mut stage_key = Zeroizing::new(vec![0u8; key_size]);
+        huk_subkey_derive(
+            task,
+            HukSubkeyUsage::UniqueTa,
+            &[&uuid_bytes, &extra_data],
+            &mut stage_key,
+        )?;
+
+        // Derive keys from max SVN down to 0
+        for svn_idx in (0..svn_key_stack_size).rev() {
```
praveen-pk, line 162: If the input parameter has `2047`, this section will create all the keys, but only send keys with svns <= `ta_svn`.

Does it make sense to only generate `min(svm_key_stack_size, ta_svn)` keys instead?

sangho2: We can't do that because this key derivation is based on a sequential hash chain. It should compute a key with the maximum SVN first. `h(k,x-1) != h(h(k,x),x-1)`.

PR #1106, opened 2026-07-29
Improve handle management

jaybosamiya-ms:
This PR merges some operations to use a single `Handle` enum, rather than do them individually.  This simplifies the `Backend` interface and also allows for supporting `chmod` of `/` (for example), which is needed as more backends move to the new design.

### DIFF
--- a/litebox/src/fs/backend.rs
+++ b/litebox/src/fs/backend.rs
@@ -147,9 +147,6 @@ pub trait Backend: private::Sealed + Send + Sync + Any {
     // XXX(jayb): I don't like that unlink and rmdir exist separately, we should probably merge them.
     fn rmdir_at(&self, dir: DirHandle, name: &str) -> Result<(), RmdirError>;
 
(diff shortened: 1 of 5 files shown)


wdcui (approved):
LGTM. Thanks. GPT-5.6 sol reported the following potential issue:

Medium —  litebox/src/fs/resolver.rs:211-217, 653-676 

 chmod / chown  now require execute permission on the target directory itself. An owner cannot restore  /locked  from mode  0600  to  0700 , although only the parent should require search permission. Exclude the final target directory from walk-permission checks for metadata operations.

The concern was:

• Previously,  chmod("/a/locked")  walked only  /a .
• Now  path_handle()  walks  /a/locked .
• For a directory,  check_walk_permissions()  checks execute permission on  locked  too.
• POSIX requires search permission on parent directories, not the final directory being changed.

jaybosamiya-ms:
Hmm, this is a valid concern, will fix up in a separate PR

PR #1107, opened 2026-07-29
Migrate in-memory file system to new backend

jaybosamiya-ms:
This PR switches our in-memory filesystem to the new core file system design (see https://github.com/microsoft/litebox/pull/887). Concretely, it adds a `InMem` backend, migrates all old usage of `in_mem::FileSystem` to a resolver-backed one to use the new `InMem` backend, and then removes the old `in_mem::FileSystem`. It also revamps the in-mem backend initialization design, moving away from the ugly "temporarily change user and do operations" to an actual controlled initialization, which reduces the chances of footguns, and also makes future improvements easier :)

### DIFF
--- a/litebox/src/fs/in_mem.rs
+++ b/litebox/src/fs/in_mem.rs
@@ -9,7 +9,6 @@ use alloc::vec::Vec;
 use hashbrown::HashMap;
 
 use crate::LiteBox;
-use crate::path::Arg;
 use crate::sync;
 
 use super::errors::{
@@ -567,10 +566,6 @@ fn assert_supported_oflags(flags: super::OFlags) {
     }
 }
 
-/// Just a random constant that is distinct from other file systems. In this case, it is
-/// `b'IMem'.hex()`.
-const DEVICE_ID: usize = 0x494d656d;
-
 /// Block size for file system I/O operations
 // TODO(jayb): Determine appropriate block size
 const BLOCK_SIZE: usize = 0;
@@ -583,14 +578,7 @@ const BLOCK_SIZE: usize = 0;
 /// is dropped.
 pub struct FileSystem<Platform: sync::RawSyncPrimitivesProvider> {
     litebox: LiteBox<Platform>,
-    // TODO: Possibly support a single-threaded variant that doesn't have the cost of requiring a
-    // sync-primitives platform, as well as cost of mutexes and such?
-    root: sync::RwLock<Platform, RootDir<Platform>>,
-    current_user: UserInfo,
-    // cwd invariant: always ends with a `/`
-    current_working_dir: String,
-    // a source of freshness for providing unique IDs
-    unique_id_freshness: core::sync::atomic::AtomicUsize,
+    resolver: super::resolver::Resolver<Platform, InMem<Platform>>,
 }
 
 impl<Platform: sync::RawSyncPrimitivesProvider> FileSystem<Platform> {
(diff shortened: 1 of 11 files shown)


jaybosamiya-ms:
This PR is easiest to review one commit at a time.  It is stacked on top of #1106 due to an ordering requirement but is possible to review independently.

wdcui:
Reported by GPT-5.6 sol. Will do my own review soon.

Found two high-confidence issues:

1. Critical — Missing parent-directory write checks
 litebox/src/fs/resolver.rs:439-495,704-760 
Create, mkdir, unlink, and rmdir now verify execute permission but not write permission on the parent. Unprivileged guests can modify root-owned  0755  directories. Restore  can_write  checks before backend mutations.
2. High — Guest-triggerable panic opening directories writable
 litebox/src/fs/in_mem.rs:303-319 
Opening a directory with  O_WRONLY  or  O_RDWR  reaches  unimplemented!() , allowing a guest to crash the runtime. Return an  EISDIR -equivalent  OpenError  instead.

jaybosamiya-ms:
I don't think it is critical, but yes to the first one.  There are some subtleties with the `/` that are already on my radar for other TODOs as more stuff migrates over, but I have added a check to handle this on all non-`/` directories.

The second one, yes, I've mentioned this before too, models like to complain about panics due to `unimplemented!()` but that is _exactly_ what `unimplemented!()` is for: to mark something as "this is not implemented right now, need to implement it in a future PR".  The panic there is _intentional_

litebox/src/fs/in_mem.rs
```diff
@@ -94,802 +169,502 @@ impl<Platform: sync::RawSyncPrimitivesProvider> FileSystem<Platform> {
...
+        let perms = file.read().perms.clone();
+        let handle = super::backend::FileHandle::from_typed::<Self>(InMemFileHandle { file });
+        if flags.contains(super::OFlags::TRUNC) && !flags.contains(super::OFlags::PATH) {
+            // Linux truncates whenever the open succeeds, regardless of the access mode (an
+            // `O_RDONLY|O_TRUNC` open of a writable file does truncate it); `O_PATH` opens ignore
+            // `O_TRUNC` entirely.
+            //
+            // TODO(jayb): Linux's `may_open` also adds `MAY_WRITE` for `O_TRUNC`, and checks
+            // permissions _before_ truncating; the resolver does neither, so a denied
+            // `O_RDONLY|O_TRUNC` open still empties the file here.
```
wdcui, line 364: Should we worry about Linux's semantics here?

wdcui:
More findings by AI agents:

1. High —  O_TRUNC  modifies files before authorization
 litebox/src/fs/in_mem.rs:357-365 
Truncation occurs before write-permission and  O_CREAT|O_EXCL  checks. A denied open can still empty the file.
2. High — Concurrent  O_CREAT  can replace an entry and panic
 litebox/src/fs/in_mem.rs:501-508 
Existence checking and insertion are not atomic. Two concurrent creates can overwrite the first entry and trigger  assert!(old.is_none()) .
3. High — Root-directory mutations bypass write permissions
 litebox/src/fs/resolver.rs:211-223 
Non-root directories are now checked, but  /  has no permission metadata and is treated as writable. A non-root guest can modify root-owned mode- 0755   / .
4. Medium —  O_APPEND  writes are not atomic
 litebox/src/fs/resolver.rs:616-626 
EOF is read separately from the write operation. Concurrent appenders can use the same offset and overwrite each other.

wdcui (approved):
LGTM. Thanks.

litebox/src/fs/in_mem.rs
jaybosamiya-ms: The Linux semantics are weird around this, and I am not particularly happy that we are still using `OFlags`, but changing that out is a bigger ask than the scope of this current change, thus I just wanted to document the weirdness that shows up around Linux-isms, that's all.

jaybosamiya-ms:
Thanks Weidong, the 4 issues your agents mention:

* 1, 3, and 4: agent is reporting an existing TODO/comment for this in the code.
* 2: yep, missing an existence check after grabbing the lock, fixing it shortly.

PR #1110, opened 2026-07-31
Fix checking of directory search/read perms

jaybosamiya-ms:
Require search/read perms on target/parents depending on situation

### DIFF
--- a/litebox/src/fs/resolver.rs
+++ b/litebox/src/fs/resolver.rs
@@ -179,6 +179,19 @@ struct WalkedDir<'a> {
     permissions: Option<PermissionCheck>,
 }
 
+/// Which directories along a walk must grant search (execute) permission.
+#[derive(Clone, Copy)]
+enum SearchScope {
+    /// Every walked directory, including a final directory component, must be searchable.
+    AllComponents,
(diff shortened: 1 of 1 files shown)


wdcui (approved):
LGTM. Thanks. The following issue is reported by AI agents.

Claude Opus 5: One medium-severity race: plain  O_CREAT  may incorrectly return  AlreadyExists  if another thread creates the file between lookup and creation ( litebox/src/fs/in_mem.rs:496-498 , propagated at  resolver.rs:573 ). Without  O_EXCL , it should retry by opening the existing file.

GTP-5.6 sol agrees with it:  The lookup and creation are not atomic. If another thread creates the file after  walk_path  reports it missing,  create_file_at  returns  AlreadyExists , even without  O_EXCL ; plain  O_CREAT  should instead open the newly existing file.

wdcui (approved):
LGTM. Thanks.

jaybosamiya-ms:
Thanks Weidong, am already aware of this one but it is outside the scope of this PR, so I am merging this PR as is. The fix will come as part of other semantics fixes related to oflags cleanups.

PR #1111, opened 2026-07-31
Reduce locking contention over fd iterations

jaybosamiya-ms:
By storing a tiny bit of data, we don't need to bother grabbing a lock before we know which subsystem an FD belongs to when iterating, and can selectively only grab locks when pointing at the correct subsystem.

### DIFF
--- a/litebox/src/fd/mod.rs
+++ b/litebox/src/fd/mod.rs
@@ -58,7 +58,8 @@ impl<Platform: RawSyncPrimitivesProvider> Descriptors<Platform> {
                 self.entries.push(None);
                 self.entries.len() - 1
             });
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
quick note: this change is independent of other changes in the stack, but I'm placing it here since it makes maintaining the stack on my end easier rather than maintaining a mega-merge.  fingers crossed the new github stacked PRs support ([finally in public preview today!](https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/)) will soon support merge queues + squash merges well, and we'll be able to use them; would make things a significantly easier to manage in our project

wdcui (approved):
LGTM. Thanks.

PR #1112, opened 2026-07-31
Fix handling of OP-TEE `ldelf` segment padding

sangho2:
This PR fixes handling of OP-TEE `ldelf` segment padding. `ldelf` can place paddings (unmapped/free pages) before and after a segment. Previously, we assumed that the paddings of different segments are disjoint, but they are not. Also, since `ldelf` is unaware of the syscall trampoline, we should ignore a trampoline page when we deal with paddings.


wdcui (approved):
LGTM. Thanks.

PR #1113, opened 2026-07-31
Migrate 9p file system to new backend

jaybosamiya-ms:
This PR switches our 9p filesystem to the new core file system design (see https://github.com/microsoft/litebox/pull/887). Like prior migrations, it adds a new backend, migrates all old usages to a resolver-backed one to use the new backend, and then removes the old `FileSystem` object.

Additionally, as a drive-by-fix, the old `read_dir`'s check reports every entry as a regular file (because it looked at `e.typ` rather than `e.qid.typ`); the new backend fixes this.

### DIFF
--- a/litebox/src/fs/nine_p/client.rs
+++ b/litebox/src/fs/nine_p/client.rs
@@ -83,6 +83,24 @@ impl<Platform: RawSyncPrimitivesProvider> Drop for FidInner<Platform> {
     }
 }
 
+/// The outcome of a [`Client::walk`].
+pub(super) struct WalkResult<Platform: RawSyncPrimitivesProvider> {
+    /// The qids of the components that were walked, in order.
+    pub(super) wqids: Vec<fcall::Qid>,
+    /// The fid for the final location.
+    ///
+    /// `Some` iff `wqids.len() == wnames.len()`: per 9P2000.L, a short walk does not establish a
+    /// new fid, so a partial walk yields qids but nothing to address them with.
+    pub(super) fid: Option<Fid<Platform>>,
+}
+
+impl<Platform: RawSyncPrimitivesProvider> WalkResult<Platform> {
+    /// The fid for a walk that reached the requested path, treating a short walk as `ENOENT`.
+    pub(super) fn into_complete_fid(self) -> Result<Fid<Platform>, Error> {
+        self.fid.ok_or(Error::Remote(super::ENOENT))
+    }
+}
+
 /// 9P client state for writing to the connection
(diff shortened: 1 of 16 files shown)


jaybosamiya-ms:
Similar to other migrations, it is easiest to review one commit at a time.

wdcui:
GPT-5.6 Sol: Requested changes—found 5 issues:

• High:  O_CREAT | O_EXCL | O_TRUNC  may truncate an existing file before returning  AlreadyExists  ( nine_p/mod.rs:417-425 ,  resolver.rs:526-530 ).
• High: Write-opening a directory can panic via  unimplemented!()  ( nine_p/mod.rs:355-358 ).
• Medium: Concurrent appenders can overwrite each other ( nine_p/mod.rs:500-513 ,  resolver.rs:661-671 ).
• Medium: Seeking beyond EOF is rejected, breaking sparse-file creation ( nine_p/mod.rs:486-488 ,  resolver.rs:716-720 ).
• Medium: A malformed 9P walk response can trigger an assertion panic ( nine_p/mod.rs:324-332 ).

litebox/src/fs/nine_p/client.rs
```diff
@@ -336,22 +359,26 @@ impl<Platform: RawSyncPrimitivesProvider, T: Read + Write> Client<Platform, T> {
             }
             // It means that the walk failed at the nwqid-th element
             if new_len < chunk.len() {
+                // XXX: Per 9P2000.L the server does not establish `new_f` on a short walk, so not
```
wdcui, line 362: @CvvT do you remember why we have a clunk here?

wdcui (approved):
LGTM. Thanks. Please take a look at the issues reported by GPT-5.6 sol.

jaybosamiya-ms:
Thanks Weidong, quick notes on the issues your agent raised:

1. same as the one on a prior PR, I already have a TODO on this
2. `unimplemented!()` panics are expected as "fixed in future PR"
3. this is already tracked at the resolver
4. this is already tracked in the resolver
5. yes there is an assertion there, that is literally stating "the server is working as intended", surviving a malformed 9P server is not a thing we need to protect against, right? Either way would be outside the purview of this PR. 

Note to myself: need to rebase after #1111 is merged before I merge

litebox/src/fs/nine_p/client.rs
jaybosamiya-ms: Since we have the comment to track it in the codebase, I'm marking this as resolved so that we can merge the code in

PR #1125, opened 2026-08-05
Add brokered timerfd support to ulitebox

willportnoy:
Adds Linux timerfd support to the ulitebox broker. Timer expiry is delegated to the host through the broker — a dedicated epoll reactor owns real host timerfds and publishes readiness through the notification ring, mirroring the brokered TCP and eventfd model. The guest gets timerfd_create/timerfd_settime/timerfd_gettime plus read, close, and epoll through a new shim subsystem.


wdcui:
@willportnoy: I ran multiple agents and got the following findings.

All 8 reviewers completed within the five-minute budgets.

Highest-confidence findings:

1. High — Timer readiness is lost after the first drain or rearm
 litebox_broker_core/src/timer.rs:135-158 ,  litebox_broker_platform_linux_userland/src/timer.rs:501-542 
Independently confirmed by three reviewers. Empty readiness transitions are not published, allowing the next  READ  publication to be coalesced and blocking reads to hang indefinitely.
2. High — One session’s publication failure can destroy all sessions’ timers
 litebox_broker_platform_linux_userland/src/timer.rs  in  TimerReactor::run  /  fail_all_timers 
A per-timer readiness publication error terminates the shared reactor, clears every session’s timers, and permanently disconnects subsequent operations.
3. Medium — Resource exhaustion is incorrectly exposed as  EAGAIN 
 litebox_common_linux/src/errno/mod.rs:590-593 
Confirmed by both shim reviewers. Blocking reads,  timerfd_create , and control operations can return an errno Linux callers do not expect, potentially causing busy retries or hard failures.
4. Medium — Valid Linux timerfd clocks are rejected
 litebox_shim_linux/src/syscalls/timerfd.rs:82-87 
Both shim reviewers identified rejection of  CLOCK_BOOTTIME  and alarm clock IDs.
5. Medium — Platform readiness is queried while holding the object lock
 litebox_broker_core/src/session.rs:327 
This differs from the socket path and can block close/teardown behind the timer snapshot mutex.
6. Medium — Timer count can exceed readiness-sink capacity
 litebox_broker_core/src/timer.rs:99-120 
More than 4,096 registrations can trigger the shared-reactor failure above, producing a broker-wide timer denial of service.
7. Medium — Queue backpressure leaks into syscall behavior
 litebox_broker_platform_linux_userland/src/timer.rs ,  TimerReactorClient::request 
A transiently full 64-entry command queue makes  settime / gettime  fail rather than applying internal backpressure.

Lower-confidence or disputed findings include strict timer-response boolean validation, protocol-version compatibility, epoll behavior after closing one duplicated descriptor, and a scheduler-sensitive 20 ms native test. The two protocol-version reviewers and two epoll reviewers reached opposing conclusions on those items.

willportnoy:
> _AI-authored, human-reviewed._

Thanks @wdcui — the multi-agent pass was genuinely useful. Pushed fixes for the three timer-specific findings; the rest look shared with the socket reactor (details below).

**Fixed**

- **#1 (readiness lost → blocking-read hang):** the reactor now `republish`es the READ edge instead of `publish`, so a parked reader is woken on *every* expiration, not just the first — the coalesced unchanged-READ was the hang. Mirrors how the socket reactor forces a wakeup on a readable-count change. Added a `timer_reactor` integration test that drives the real reactor through a coalescing-model sink and reproduces the hang, plus consecutive blocking reads in `timerfd.c`. (`c4a4498`)
- **#4 (clocks):** accept `CLOCK_BOOTTIME` in the shim and the broker clock mapping, with unit + e2e coverage. I left the `*_ALARM` clocks rejected on purpose — they need `CAP_WAKE_ALARM` and can wake a suspended system, which seems outside the sandbox’s authority. Happy to add them if you’d prefer parity with native. (`c4a4498`)
- **#5 (readiness under the object lock):** timer now clones the resource under the lock and reads readiness after releasing it, matching the socket path. There’s no lock inversion today, so this is contention hygiene rather than a live deadlock. (`715a193`)

**Shared with the socket reactor — proposing to keep as-is or fix holistically**

- **#2 / #6:** a publish failure ends the whole reactor via `fail_all_*` — `socket.rs` does the same today. Per-object isolation would be better handled for both providers together.
- **#7:** queue-full → `ResourceExhausted` is the same `try_send` pattern the socket reactor uses.
- **#3:** `ResourceExhausted` → `EAGAIN` mirrors the eventfd mapping; I kept it consistent rather than diverging a single provider.

Glad to take any of the shared ones in this PR if you’d like them addressed here.

PR #1056, opened 2026-07-21
Make litebox_shim_linux generic over its platform

jstarks:
The Linux shim was hard-wired to the global `litebox_platform_multiplex::Platform` singleton (set via `set_platform`, fetched via `platform()`). This made the shim non-generic and forced platform selection through cargo features. This PR parameterizes the shim over a platform type `Platform` end-to-end.

PR #1127, opened 2026-08-05
Remove platform singletons from the LVBS platform and OP-TEE shim

sangho2:
This PR removes platform singletons from the LVBS platform and OP-TEE shim crates, allowing us to drop `litebox_platform_multiplex` completely. Refactor both crates at once because they relied on the same `GlobalVmapManager` which is now removed as well.


sangho2:
A sequel of #1056

praveen-pk:
This is rather a large PR. would be nice if this can be broken down to smaller commits.

Overall, the PR LGTM!

wdcui:
@jaybosamiya-ms , can you help review this PR? Thanks!

wdcui:
@sangho2 , the agent found the following issue:

Medium —  OpteeShim  unintentionally loses  Clone  —  litebox_shim_optee/src/lib.rs:292 .  #[derive(Clone)]  adds  Platform: Clone , but neither platform implements it. Replace the derive with a manual  Clone  impl that clones the inner  Arc .

litebox_platform_lvbs/src/mshv/mod.rs
```diff
@@ -37,7 +48,7 @@ unsafe impl<const ALIGN: usize> VmapManager<ALIGN> for PrivilegedVmap {
     ) -> Result<Self::MapInfo, PhysPointerError> {
         // SAFETY: callers uphold the raw mapping contract. This provider is used only for
         // writes whose destination the caller has independently authorized.
-        unsafe { crate::platform_low().vmap_privileged(pages, perms) }
+        unsafe { self.0.vmap_privileged(pages, perms) }
```
jaybosamiya-ms: Nit: do we want to name it as `self.platform_low` for consistency/readability? Or just `self.platform`?  It is possible `.0` good enough.  No response needed here, feel free to mark as resolved either way.

litebox_shim_optee/src/session.rs
```diff
@@ -481,14 +498,13 @@ pub struct SessionManager {
...
-pub fn session_manager() -> &'static SessionManager {
-    static SESSION_MANAGER: once_cell::race::OnceBox<SessionManager> =
-        once_cell::race::OnceBox::new();
-    SESSION_MANAGER.get_or_init(|| alloc::boxed::Box::new(SessionManager::new()))
-}
+// NOTE: the session manager singleton lives in the composition root (each
+// runner), not here. A `static` cannot name a generic parameter, and a shim
+// instance is built per session, so the shim has nowhere to put it. The runner
+// knows its concrete platform, so it can hold the `static` and hand out
+// `&'static SessionManager<ConcretePlatform>`.
```
jaybosamiya-ms: Nit: this feels like a comment left in by an agent removing code leaving breadcrumbs; not sure it needs to stay?

jaybosamiya-ms (approved):
Thanks Sangho for getting rid of the final bits of the multiplex! The PR generally looks good to me. Most of it is fairly mechanical, so I haven't reviewed every line, but have done a look through for any bits that stood out to me, and have pointed out a few very minor nits. If there are any parts that you think I should specifically take a closer look at or sanity check, let me know.

litebox_platform_lvbs/src/mshv/mod.rs
sangho2: yes, `self.0` sometimes looks bad. `self.platform` might work well. Let me figure out better option.

litebox_shim_optee/src/session.rs
sangho2: Good catch. Let me go over all comments again.

sangho2:
Let me merge this one later because it might conflict with ongoing dynamic TA loading efforts.

PR #1130, opened 2026-08-05
Combine PRK and IDK_S generation VTL calls

athvu:


sangho2 (changes requested):
Hi Angelina,

Thank you for your effort! I have one recommendation: rather than modifying the `fill_bytes_crng` interface in the `litebox` crate, could you install a trusted seed through`Vtl1Gate` (e.g., set a new global var) and use that seed in LVBS's `fill_bytes_crng` implementation? As you already observed, the `litebox` crate is a shared common interface/abstraction. any changes of it affect many other subsystems, such that we require strong justification.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -256,10 +256,24 @@ fn vtlcall_dispatch(params: &[u64; NUM_VTLCALL_PARAMS]) -> i64 {
             let smc_args_pfn = params[1];
             optee_smc_handler_entry(smc_args_pfn)
         }
-        VsmFunction::GenerateIdentitySigningKey => {
-            let public_key_pa = params[1];
-            let key_alg = params[2];
-            litebox_shim_optee::idk::generate_identity_signing_key(public_key_pa, key_alg)
+        VsmFunction::SetPlatformRootKeyAndGenerateIdentitySigningKey => {
+            let tpm_random_pa = params[1];
```
sangho2: nits: Is this the address of an RNG seed? Technically, LiteBox doesn't know or doesn't care whether this is from TPM or not (e.g., could be from Pluton, HSM, ...). A generic name might be better (e.g., trusted seed, ...)

litebox_runner_lvbs/src/lib.rs
```diff
@@ -256,10 +256,24 @@ fn vtlcall_dispatch(params: &[u64; NUM_VTLCALL_PARAMS]) -> i64 {
...
+            let return_code = vsm_dispatch(
+                VsmFunction::SetPlatformRootKeyAndGenerateIdentitySigningKey,
+                &params[1..2],
+            );
+            if return_code < 0 {
+                return return_code;
+            }
+
+            litebox_shim_optee::idk::generate_identity_signing_key(
+                tpm_random_pa + PRK_LEN as u64,
```
sangho2: looks like this is a concatenation of two data: PRK || seed ? then, something like prk_and_seed might be better.

litebox_shim_optee/src/idk.rs
```diff
@@ -61,16 +62,30 @@ pub fn generate_identity_signing_key(public_key_pa: u64, key_alg: u64) -> i64 {
 /// This function assumes that the caller prepares a buffer at the given physical
 /// address (in a single or contiguous physical memory page(s)) whose length is equal to
 /// or greater than `IDENTITY_SIGNING_PUBLIC_KEY_LEN`.
-fn generate_identity_signing_key_inner(public_key_pa: u64, key_alg: u64) -> Result<i64, Errno> {
+fn generate_identity_signing_key_inner(
+    tpm_random_pa: u64,
```
sangho2: I got confused here because it looks like `tpm_random_pa = tpm_random_pa + PRK_LEN`. Anyway, as noted above, I think it is better not to be `tpm_random`.

litebox_platform_lvbs/src/host/lvbs_impl.rs
```diff
@@ -231,10 +244,10 @@ fn rdrand_seed() -> Option<CrngSeed> {
     Some(seed)
 }
 
-fn crng_seed_from_prk_and_rdrand(prk: &[u8; PRK_LEN], rdrand_seed: CrngSeed) -> CrngSeed {
+fn crng_seed_from_tpm_and_rdrand(seed: &[u8], rdrand_seed: CrngSeed) -> CrngSeed {
```
sangho2: tpm -> rot/hrot?

It'd be better to use `&[u8; CRNG_SEED_LEN]`.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -315,10 +321,10 @@ fn vsm_dispatch(func_id: VsmFunction, params: &[u64]) -> i64 {
         VsmFunction::AllocateRingbufferMemory => {
             heki.allocate_ringbuffer_memory(params[0], params[1])
         }
-        VsmFunction::SetPlatformRootKey => vtl1.set_platform_root_key(params[0]).map(|()| 0),
-        VsmFunction::GenerateIdentitySigningKey => {
-            Err(VsmError::OperationNotSupported("Identity key generation"))
-        }
+        VsmFunction::ExchangeSecrets => vtl1
+            .set_platform_root_key(params[0])
+            .and_then(|()| vtl1.set_crng_seed(params[0] + PRK_LEN as u64))
```
sangho2: nits: let's use `checked_add` to catch integer overflow.

litebox_platform_lvbs/src/host/lvbs_impl.rs
```diff
@@ -134,10 +134,10 @@ struct LvbsCrng {
 }
 
 impl LvbsCrng {
-    fn new(prk: &[u8; PRK_LEN], rdrand_seed: CrngSeed) -> Self {
+    fn new(seed: &[u8], rdrand_seed: CrngSeed) -> Self {
```
sangho2: `seed: &[u8; CRNG_SEED_LEN]` would be better.

litebox_runner_lvbs/src/lib.rs
sangho2: This depends on the ABI, but, at least for now, separating `set_platform_root_key` and `set_crng_seed` is a bit bad for performance and atomicity (redundant page mapping/copying and potential partial failure). That is, we can copy 64 bytes at once, rather than copy two 32 bytes in order. Anyway, it depends on VTL0-side argument passing, so it's up to you.

sangho2 (approved):
Overall, looks good to me. I left some comments and suggestion.

PR #1131, opened 2026-08-05
Fix forced signal delivery

sangho2:
This PR fixes forced delivery of a signal. Two issues fixed:
1. `handle_exception_request` covers `SIGFPE`, `SIGTRAP`, and `SIGILL`, but `force_signal_with_info` asserts the signal is `SIGKILL` or `SIGSEGV`.
2. `siginfo_exception` is used to return `SigInfo` with a fixed code (`SI_KERNEL`), ignoring other codes like `FPE_INTDIV` and `ILL_ILLOPN`.

### DIFF
--- a/litebox_common_linux/src/signal/mod.rs
+++ b/litebox_common_linux/src/signal/mod.rs
@@ -319,6 +319,10 @@ pub const SI_TKILL: i32 = -6;
 pub const SI_DETHREAD: i32 = -7;
 pub const SI_ASYNCNL: i32 = -60;
 
+pub const ILL_ILLOPN: i32 = 2;
+
+pub const FPE_INTDIV: i32 = 1;
+
 #[cfg(target_arch = "x86_64")]
 #[repr(C)]
 #[derive(Clone, FromBytes, IntoBytes)]
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -170,7 +170,9 @@ impl<Platform: ShimPlatform, FS: ShimFS> litebox::shim::EnterShim
                 return ContinueOperation::Terminate;
             }
         }
-        self.enter_shim(false, ctx, |task, _ctx| task.handle_exception_request(info))
+        self.enter_shim(false, ctx, |task, ctx| {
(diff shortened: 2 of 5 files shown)


CvvT (approved):
LGTM, thanks!

wdcui (approved):
I didn't find any issue, but the agent reported the following issue:

This PR makes SIGFPE/SIGILL delivery reachable, but  siginfo_exception()  still emits  SI_KERNEL  with address  0 . Linux handlers expect  FPE_INTDIV  or  ILL_ILLOPN  and the faulting RIP. The fix must also bypass  RLIMIT_SIGPENDING  explicitly for forced signals; simply changing  si_code  would allow these signals to be dropped. SIGTRAP can retain  SI_KERNEL .

sangho2:
> I didn't find any issue, but the agent reported the following issue:
> 
> This PR makes SIGFPE/SIGILL delivery reachable, but  siginfo_exception()  still emits  SI_KERNEL  with address  0 . Linux handlers expect  FPE_INTDIV  or  ILL_ILLOPN  and the faulting RIP. The fix must also bypass  RLIMIT_SIGPENDING  explicitly for forced signals; simply changing  si_code  would allow these signals to be dropped. SIGTRAP can retain  SI_KERNEL .

The review makes sense, which would bump up this one-line change though. Let me figure out minimal change.

sangho2:
> > I didn't find any issue, but the agent reported the following issue:
> > This PR makes SIGFPE/SIGILL delivery reachable, but  siginfo_exception()  still emits  SI_KERNEL  with address  0 . Linux handlers expect  FPE_INTDIV  or  ILL_ILLOPN  and the faulting RIP. The fix must also bypass  RLIMIT_SIGPENDING  explicitly for forced signals; simply changing  si_code  would allow these signals to be dropped. SIGTRAP can retain  SI_KERNEL .
> 
> The review makes sense, which would bump up this one-line change though. Let me figure out minimal change.

Added a minimal fix to return `FPE_INTDIV` and `ILL_ILLOPN` based on `Signal`. Left TODO because this is not fine-grained (additional machinery is needed).

PR #1142, opened 2026-08-06
optee: share TA binaries through a global UUID map

praveen-pk:
Move the TA UUID-to-binary map out of individual shim instances so TA binaries can be registered once and reused across instances. This global map is required to support Dyanamically Loading TAs and using them across sessions.

Protect the map with a spin-based RwLock and expose shim methods for storing and retrieving TA binaries.

Register the LVBS runner's embedded TA during BSP initialization and resolve new TA instances through the shared map.

### DIFF
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -224,6 +224,11 @@ pub fn init(is_bsp: bool) -> Option<&'static Platform> {
     // Per-CPU; safe to call on BSP and APs.
     timer::init();
 
+    if is_bsp {
+        let shim = litebox_shim_optee::OpteeShimBuilder::new().build();
+        register_embedded_tas(&shim);
+    }
+
     ret
 }
 
@@ -778,7 +783,8 @@ fn open_session_new_instance(
     client_identity: Option<litebox_common_optee::TeeIdentity>,
     ta_req_info: &litebox_shim_optee::msg_handler::TaRequestInfo<PAGE_SIZE>,
 ) -> Result<(), OpteeSmcReturnCode> {
-    let Some(ta_bin) = find_ta_binary(ta_uuid) else {
+    let shim = litebox_shim_optee::OpteeShimBuilder::new().build();
+    let Some(ta_bin) = find_ta_binary(&shim, &ta_uuid) else {
         msg_args.session = 0;
         msg_args.ret = TeeResult::ItemNotFound;
         msg_args.ret_origin = TeeOrigin::Tee;
@@ -801,9 +807,8 @@ fn open_session_new_instance(
     })?;
 
     // Load ldelf and TA - Box immediately to keep at fixed heap address
-    let shim = litebox_shim_optee::OpteeShimBuilder::new().build();
     let loaded_program = Box::new(
-        shim.load_ldelf(LDELF_BINARY, ta_uuid, Some(ta_bin))
+        shim.load_ldelf(LDELF_BINARY, ta_uuid, Some(&ta_bin))
             .map_err(|_| {
                 // Safety: We are about to tear down this TA instance;
                 // no references to user-space memory will be held afterwards.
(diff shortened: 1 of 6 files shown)


litebox_runner_lvbs/src/lib.rs
```diff
@@ -224,6 +224,11 @@ pub fn init(is_bsp: bool) -> Option<&'static Platform> {
     // Per-CPU; safe to call on BSP and APs.
     timer::init();
 
+    if is_bsp {
+        let shim = litebox_shim_optee::OpteeShimBuilder::new().build();
+        register_embedded_tas(&shim);
+    }
```
sangho2, line 230: It is a bit unclear whether this is a good place to register TA binaries. a bit early. Probably fine because this is a temporary solution.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -1337,24 +1342,41 @@ fn write_rpc_args_to_normal_world(
...
+) -> bool {
+    let Some(ta_head) = litebox_common_optee::parse_ta_head(ta_binary) else {
+        return false;
+    };
+    shim.store_ta_bin(&ta_head.uuid, ta_binary)
+}
+
+/// Register all TA binaries embedded in the runner image.
+fn register_embedded_tas(shim: &litebox_shim_optee::OpteeShim) {
+    static REGISTERED: spin::Once<()> = spin::Once::new();
```
sangho2: Do we need this `static REGISTERED`? `ta_uuid_map` itself should be able to reject redundant registration.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -1337,24 +1342,41 @@ fn write_rpc_args_to_normal_world(
...
+    REGISTERED.call_once(|| {
+        assert!(register_embedded_ta(
```
sangho2: It registers only a single TA, regression. We could keep using `const TA_BINARIES`.

litebox_shim_optee/src/lib.rs
```diff
@@ -169,8 +168,6 @@ struct GlobalState {
     pm: litebox::mm::PageManager<Platform, { PAGE_SIZE }>,
     /// The LiteBox instance used throughout the shim.
     _litebox: litebox::LiteBox<Platform>,
-    /// The TA UUID to binary map for TA loading.
-    ta_uuid_map: TaUuidMap,
```
sangho2, line 173: Dropping `ta_uuid_map` is here is a bit concerning because we can no longer use `shim`'s `self`. Instead, we could store `&'static TaUuidMap` here. This is an outstanding PR's direction (#1127).

litebox_shim_optee/src/lib.rs
```diff
@@ -1355,20 +1364,28 @@ impl TaUuidMap {
...
     // Lazy removal of TA binaries when they are no longer needed.
     pub(crate) fn remove(&self, uuid: &TeeUuid) -> Option<alloc::boxed::Box<[u8]>> {
-        self.inner.lock().remove(uuid).map(|info| info.binary)
+        self.inner.write().remove(uuid).map(|info| info.binary)
     }
 }
 
+/// Get the global TA UUID map.
+fn ta_uuid_map() -> Arc<TaUuidMap> {
+    static TA_UUID_MAP: once_cell::race::OnceBox<Arc<TaUuidMap>> = once_cell::race::OnceBox::new();
```
sangho2: Using `Arc` in general a good idea, but using it against `TaUuidMap` itself is not a good call IMO. This map is already protected by a lock and reference counting it less meaningful because we would not drop this map anyhow. Instead, we can `Arc` each TA binary which might be helpful to deal with future TA binary update as well.

litebox_shim_optee/src/lib.rs
sangho2: As noted, we could do `binary: Arc<[u8]>` here.

litebox_shim_optee/src/lib.rs
```diff
@@ -187,12 +184,12 @@ impl GlobalState {
     /// Returns `true` if the binary was successfully stored, `false` if the binary's
     /// UUID (from `.ta_head` section) doesn't match the provided UUID or parsing failed.
     pub(crate) fn store_ta_bin(&self, ta_uuid: &TeeUuid, ta_bin: &[u8]) -> bool {
-        self.ta_uuid_map.insert(*ta_uuid, ta_bin.into())
+        ta_uuid_map().insert(*ta_uuid, ta_bin.into())
     }
 
     /// Get the TA binary associated with the given TA UUID.
     pub(crate) fn get_ta_bin(&self, ta_uuid: &TeeUuid) -> Option<alloc::boxed::Box<[u8]>> {
```
sangho2: once we do use `Arc<[u8]>`, functions like this one can return `Arc`, avoiding deep heap copy.

sangho2 (commented):
Thanks for your effort! Yes, we should differentiate shim-global from runner-global. I left some comments/suggestions.

litebox_runner_lvbs/src/lib.rs
praveen-pk: Agreed! This will be dropped once dynamic TA is properly enabled.

litebox_shim_optee/src/lib.rs
praveen-pk: make sense. Dropped it in latest version.

litebox_runner_lvbs/src/lib.rs
praveen-pk: agreed. Dropped it.

litebox_runner_lvbs/src/lib.rs
praveen-pk: As embedded TA support would be dropped soon, I didn't see much value in supporting multiple TA_BINARIES.


I restored this in the latest version. we can revisit this as a clean up step after Dyn Ta support.

litebox_shim_optee/src/lib.rs
praveen-pk: Move binary to `Arc`.

litebox_shim_optee/src/lib.rs
praveen-pk: addressed

litebox_shim_optee/src/lib.rs
praveen-pk: addressed in the latest commit.

praveen-pk:
I confirmed that static TAs continue to function with this change.
```
$ sudo /usr/share/optee_examples/optee_example_hello_world
Invoking TA to increment 42
TA incremented value to 43
```

litebox_runner_lvbs/src/lib.rs
```diff
@@ -801,9 +807,8 @@ fn open_session_new_instance(
     })?;
 
     // Load ldelf and TA - Box immediately to keep at fixed heap address
-    let shim = litebox_shim_optee::OpteeShimBuilder::new().build();
     let loaded_program = Box::new(
-        shim.load_ldelf(LDELF_BINARY, ta_uuid, Some(ta_bin))
+        shim.load_ldelf(LDELF_BINARY, ta_uuid, Some(&ta_bin))
```
sangho2: I noticed another performance issue. `shim.load_ldelf` does call `store_ta_bin` if `ta_bin` is not `None`. This is because, previously, we don't have `register_embedded_ta` and `load_ldelf` is the one which populates this UUID map. We can avoid this issue by passing `None` to `shim.load_ldelf` instead of `Some(&ta_bin)`.

sangho2 (approved):
Looks good to me. Please take a look at the new comment.

Feel free to drop "[Draft]" from the title if it is ready to merge and replace optee with OP-TEE or OP-TEE shim. Thanks!

litebox_runner_lvbs/src/lib.rs
praveen-pk: Nice Catch!

Actually `ta_bin` is used in load_ldelf to solely store the TA binary. I will go ahead and drop `ta_bin` from the arg list for `load_ldelf`. That is cleaner than passing a `None`.

sangho2: `None` might be simpler than dropping it because `litebox_runner_optee_on_linux_userland` uses this path. It is also fine to revise `litebox_runner_optee_on_linux_userland`, if you prefer a clean signature :)

praveen-pk: did that :)

PR #1143, opened 2026-08-07
Fix python test

CvvT:
Fix Python runner tests failing with: `ModuleNotFoundError: No module named 'encodings'`

`rust-cache` removes staged files while preserving empty directory structures under `target/tmp`. The staging logic previously treated an existing destination directory as complete and skipped copying its contents.

Add a cache-excluded completion marker after each Python directory is staged. Missing markers now trigger a fresh copy, covering empty, partial, and interrupted staging directories.


wdcui (approved):
Thank you for the fix.

PR #1153, opened 2026-08-10
Add bottom-up mmap to OP-TEE shim

sangho2:
This PR makes the OP-TEE shim's mmap allocate virtual addresses (VAs) from low to high addresses. OP-TEE allocates userspace VAs bottom-up, whereas LiteBox allocates them top-down. This mismatch can lead to incompatibilities (e.g., padding management).

### DIFF
--- a/litebox_shim_optee/src/syscalls/mm.rs
+++ b/litebox_shim_optee/src/syscalls/mm.rs
@@ -3,10 +3,14 @@
 
 //! Implementation of memory management related syscalls, eg., `mmap`, `munmap`, etc.
 
-use litebox::mm::linux::{MappingError, PAGE_SIZE};
+use litebox::mm::linux::{MappingError, PAGE_SIZE, VmFlags};
+use litebox::platform::page_mgmt::PageManagementProvider;
 use litebox_common_linux::{MapFlags, ProtFlags, errno::Errno, user_pointers::UserPtrMut};
 
 use crate::{Platform, Task, UserMutPtr};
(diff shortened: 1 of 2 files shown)


litebox_shim_optee/src/syscalls/mm.rs
```diff
@@ -3,18 +3,55 @@
...
     debug_assert!(align.is_power_of_two());
     addr.checked_next_multiple_of(align)
 }
 
 impl Task {
+    /// Finds the first address-space gap from low to high.
+    ///
+    /// OP-TEE chooses user VAs bottom-up. Matching that order matters because
+    /// `ldelf`'s sequential segment allocations with padding rely on it, while
+    /// LiteBox's default VA placement is top-down.
```
wdcui, line 26: Is this controlled by the shim or the litebox core or the platform?

wdcui (approved):

litebox_shim_optee/src/syscalls/mm.rs
sangho2: Core does top-down VA placement, as Linux does:

https://github.com/microsoft/litebox/blob/7af6242f0729c1f0224161c7cec0afc114994cf6/litebox/src/mm/linux.rs#L913-L947

sangho2: Now explicitly mention LiteBox's `get_unmmaped_area` in this comment block.

PR #1161, opened 2026-08-12
Share broker transport coordination

CvvT:
To reduce the duplication on broker transport between Linux and Window platforms, this PR extracts platform-neutral broker transport coordination from the existing Linux implementation, including:

- Share pending-call coordination across transports.
- Share setup frame encoding and decoding while preserving partial I/O progress.
- Make control-ring memory access policy explicit and align Linux shared-memory setup.

I also did some clean up on shared memory construction to distinguish control ring buffer from regular ones.

### DIFF
--- a/litebox_broker_transport/src/control_ring.rs
+++ b/litebox_broker_transport/src/control_ring.rs
@@ -61,6 +61,38 @@ const PRODUCER_EPOCH_OFFSET: usize = 0;
 const CONSUMER_EPOCH_OFFSET: usize = 4;
 const CONSUMER_HEAD_OFFSET: usize = 8;
 
+/// Access model enforced by a concrete shared-memory mapping.
+///
+/// This keeps byte-copy memory from exposing typed control-ring operations and
+/// constrains control-ring accesses to the disjoint regions defined by its ABI.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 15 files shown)


wdcui (approved):
LGTM. Thank you, Weiteng.

Opus 5 reported the following issue, but I don't think it's a big deal for removing some regression tests (you probably have your own reasons for doing that).

• Medium — deleted regression tests:  setup_frame.rs:116-157  and  memfd.rs:~940 . Commit  ee3f2ff4a  removes the only tests covering partial I/O, invalid I/O counts, clean-close handling, and explicit control-ring memory policy. Restore them using annotated closure error types rather than invalid turbofish syntax.

CvvT:
> LGTM. Thank you, Weiteng.
> 
> Opus 5 reported the following issue, but I don't think it's a big deal for removing some regression tests (you probably have your own reasons for doing that).
> 
> • Medium — deleted regression tests:  setup_frame.rs:116-157  and  memfd.rs:~940 . Commit  ee3f2ff4a  removes the only tests covering partial I/O, invalid I/O counts, clean-close handling, and explicit control-ring memory policy. Restore them using annotated closure error types rather than invalid turbofish syntax.

Thanks! Regarding the issue, this PR does not delete any existing tests. The last commit I did deleted some new tests added by agents, which I don't feel are that useful.

PR #1165, opened 2026-08-12
Join the spawned handles in tests

jaybosamiya-ms:
This PR fixes some rare deadlocks that occur on Windows.  Specifically, a still-running thread races with process teardown, which on Windows can in rare cases get stuck.  Doing a `join` fixes this.  It also helps surface panics _if_ they occur on the thread.


jaybosamiya-ms:
Relevant example of failure that this PR should fix: https://github.com/microsoft/litebox/actions/runs/31642623868/job/94268620821

image

wdcui (approved):
LGTM. Thanks.

PR #1168, opened 2026-08-14
Stop mapping eventfd resource exhaustion to EAGAIN

willportnoy:
EventCounterError::ResourceExhausted mapped to EAGAIN, which tells a guest to retry eventfd operations against a permanent object-reference limit or allocation failure. In practice this arm is only reachable on eventfd2() creation, so it now maps to ENOMEM (eventfd2's documented exhaustion errno) while WouldBlock stays on EAGAIN, matching how the socket path already surfaces ResourceExhausted as ENOBUFS. If events ever gain an externally-backed readiness-sink path, the transient sink-full case should be split out rather than re-widening this arm, so a transient backpressure condition does not surface as ENOMEM on read/write.

### DIFF
--- a/litebox_common_linux/src/errno/mod.rs
+++ b/litebox_common_linux/src/errno/mod.rs
@@ -576,8 +576,12 @@ impl From<litebox::event::counter::EventCounterError> for Errno {
     fn from(value: litebox::event::counter::EventCounterError) -> Self {
         match value {
             litebox::event::counter::EventCounterError::InvalidInput => Errno::EINVAL,
-            litebox::event::counter::EventCounterError::WouldBlock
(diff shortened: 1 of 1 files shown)


wdcui:
@willportnoy , should this be merged into main instead of ulitebox? Also, I don't think the added test is particularly useful.

@CvvT, should we merge this into main?

CvvT:
> Will Portnoy (@willportnoy) , should this be merged into main instead of ulitebox? Also, I don't think the added test is particularly useful.
> 
> Weiteng Chen (@CvvT), should we merge this into main?

The main branch does not have this code.

wdcui:
Ah okay. Thanks, @CvvT.

@willportnoy, please remove the test then merge the PR. Thanks!

PR #1179, opened 2026-08-17
Fix resource limit reads and updates

### LINKED ISSUE #1159: semantic bug from interleaving relaxed stores over fields in AtomicRLimit
The `AtomicRLimit` struct has two atomic fields for current and max resource limits.  Defined here:
https://github.com/microsoft/litebox/blob/e7984422ce1aab181305ac7b9085c3e84e7bb27c/litebox_shim_linux/src/syscalls/process.rs#L755-L758

Both `set_rlimit` and `get_rlimit` functions use Ordering::Relaxed on both fields:
https://github.com/microsoft/litebox/blob/e7984422ce1aab181305ac7b9085c3e84e7bb27c/litebox_shim_linux/src/syscalls/process.rs#L809-L817

Concurrent sets could lead to an incorrect get that violates the assumption that` r.cur =< r.max`.  Static analysis indicated  potential racing operations on each atomic individually, and an agent uncovered this logical bug by proposing the following ordering: 

thread B sets cur -> thread A sets cur -> thread A sets max -> thread B sets max

| order | thread | store | state |
|-------|--------|--------|-------|
| 1 | B - (10,10) | cur = 10 | (10,100) |
| 2 | A - (90,100) | cur = 90 | (90,100)|
| 3 | A - (90, 100) | max = 100 | (90,100)|
|4 | B - (10,10)| max = 10 | (90,10)|

I couldn't find any logical check in the code that cur < max, other than when setting a new limit [here](https://github.com/microsoft/litebox/blob/7af6242f0729c1f0224161c7cec0afc114994cf6/litebox_shim_linux/src/syscalls/process.rs#L838), but this check would happen before `set_rlimit` is called with the new limits.

Stronger ordering guarantees wouldn't fix this, the resource limits need to be protected by a lock to ensure atomicity over the entire get/set operations.

CvvT:
Fixes #1159: replace independently atomic resource-limit fields with an `RwLock`-protected array.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1801,7 +1801,6 @@ dependencies = [
  "object",
  "once_cell",
  "ringbuf",
- "seq-macro",
  "spin 0.9.8",
  "syscalls",
  "tempfile",
@@ -2700,12 +2699,6 @@ dependencies = [
  "core-foundation-sys",
  "libc",
 ]
-
-[[package]]
-name = "seq-macro"
-version = "0.3.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1bc711410fbe7399f390ca1c3b60ad0f53f80e95c5eb935e52268a0e2cd49acc"
 
 [[package]]
 name = "serde"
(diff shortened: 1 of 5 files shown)


wdcui (approved):
LGTM. Thank you for the bug fix.

PR #1181, opened 2026-08-17
Fix races in Linux file descriptor duplication

### LINKED ISSUE #1170: shim_linux fd-table dup lock is released between inserting the new fd and closing it
In the Linux shim `do_dup_inner` when matching on the target fd , a `raw_descriptor_store` write lock is acquired for every case in the match statement that does an insert into the table and dropped at the end of each case like so:
https://github.com/microsoft/litebox/blob/9732adf0add28c663bd51ef9881a0b8992e20be8/litebox_shim_linux/src/syscalls/file.rs#L2396-L2399

If the new_fd (the lowest available) is greater than the max_fd, then the `raw_description_store` lock is acquired again later through `do_close` -> `do_close_and_replace:` 
https://github.com/microsoft/litebox/blob/9732adf0add28c663bd51ef9881a0b8992e20be8/litebox_shim_linux/src/syscalls/file.rs#L2414-L2415
https://github.com/microsoft/litebox/blob/9732adf0add28c663bd51ef9881a0b8992e20be8/litebox_shim_linux/src/syscalls/file.rs#L762

So for an insert and teardown when the fd is above the limit, we don't hold a lock across both which can lead to:
- if a concurrent thread reads slot N, it can read a descriptor that shouldn't be there
- if concurrent threads run close on N and then rewrite to N, we can close a completely different file (that would be below the max if the limit was adjusted.

Another issue is that the `max_fd` number is read at the beginning of the function and can be changed concurrently, so by the time we get to the comparison, it may be different and the file is closed unnecessarily. 

Seems like a similar pattern is handled correctly in `insert_raw_fd` where the lock is held across all the related operations.

CvvT:
Fix #1170: keep FD allocation and exact replacement atomic under the raw descriptor table lock.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -91,7 +91,7 @@ pub(crate) struct FilesState<Platform: ShimPlatform> {
     pub(crate) fs: alloc::sync::Arc<LinuxFS<Platform>>,
     pub(crate) raw_descriptor_store:
         litebox::sync::RwLock<Platform, litebox::fd::RawDescriptorStorage>,
-    fd_limit: AtomicUsize,
+    max_fd: AtomicUsize,
 }
 
 impl<Platform: ShimPlatform> FilesState<Platform> {
@@ -101,12 +101,12 @@ impl<Platform: ShimPlatform> FilesState<Platform> {
             raw_descriptor_store: litebox::sync::RwLock::new(
                 litebox::fd::RawDescriptorStorage::new(),
             ),
-            fd_limit: AtomicUsize::new(usize::MAX),
+            max_fd: AtomicUsize::new(usize::MAX),
         }
     }
 
-    pub(crate) fn set_fd_limit(&self, fd_limit: usize) {
-        self.fd_limit.store(fd_limit, Ordering::Relaxed);
+    pub(crate) fn set_max_fd(&self, max_fd: usize) {
+        self.max_fd.store(max_fd, Ordering::Relaxed);
     }
 
     // Returns Ok(raw_fd) if it fits within the max limits already set up; otherwise returns the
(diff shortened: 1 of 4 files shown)


wdcui:
@jaybosamiya-ms maybe you should review this fd-related fix?

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -79,7 +79,7 @@ pub(crate) struct FilesState<Platform: ShimPlatform, FS: ShimFS> {
     pub(crate) fs: alloc::sync::Arc<FS>,
     pub(crate) raw_descriptor_store:
         litebox::sync::RwLock<Platform, litebox::fd::RawDescriptorStorage>,
-    max_fd: AtomicUsize,
+    fd_limit: AtomicUsize,
```
jaybosamiya-ms: I'm not sure why this changed, is there a reason for this?  This change seems unnecessary to the PR's goals

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -89,12 +89,12 @@ impl<Platform: ShimPlatform, FS: ShimFS> FilesState<Platform, FS> {
...
             ),
-            max_fd: AtomicUsize::new(usize::MAX),
+            fd_limit: AtomicUsize::new(usize::MAX),
         }
     }
 
-    pub(crate) fn set_max_fd(&self, max_fd: usize) {
-        self.max_fd.store(max_fd, Ordering::Relaxed);
+    pub(crate) fn set_fd_limit(&self, fd_limit: usize) {
+        self.fd_limit.store(fd_limit, Ordering::Relaxed);
```
jaybosamiya-ms: Same

litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -112,14 +112,43 @@ impl<Platform: ShimPlatform, FS: ShimFS> FilesState<Platform, FS> {
...
+            return Err(typed_fd);
         }
+
+        let mut raw_fd = min_fd;
+        for occupied_raw_fd in rds.iter_alive().skip_while(|&fd| fd < min_fd) {
+            if occupied_raw_fd != raw_fd {
+                break;
+            }
+            raw_fd += 1;
+        }
```
jaybosamiya-ms, line 160: This is not new code and is just moving from existing stuff here, so no need to change it as part of this PR.  Nonetheless, might help to add a quick `// XXX: Can clean+speed this up by exposing a new method at RawDescriptorStorage` so that a future PR can do so.

jaybosamiya-ms (commented):
Generally seems reasonable to me, just a couple of points; the biggest being that I am not sure why `max_fd` changed to `fd_limit` as part of this change; seems unrelated to the goal of this PR to fix the races, no?

litebox_shim_linux/src/syscalls/file.rs
CvvT: Makes sense. I added one comment.

litebox_shim_linux/src/syscalls/file.rs
CvvT: This change was because `max_fd` was set to `RLIMIT_NOFILE_CUR - 1`, which does not work if user set `RLIMIT_NOFILE_CUR` to zero. I changed `max_fd` to represent `RLIMIT_NOFILE_CUR` instead, but I guess changing its name is unnecessary. I added some doc to the field to indicate that this is an exclusive upper bound.

litebox_shim_linux/src/syscalls/file.rs
CvvT: reverted

jaybosamiya-ms (approved):
LGTM, thanks Weiteng!

PR #1182, opened 2026-08-17
Fix pipe2 fd-table race when the table is full

### LINKED ISSUE #1171: pipe2 bug when the fd table is full and the writer's slot is closed concurrently
In [pipe2](https://github.com/microsoft/litebox/blob/9732adf0add28c663bd51ef9881a0b8992e20be8/litebox_shim_linux/src/syscalls/file.rs#L1728):

The raw_fd_store lock is acquired three times across the two inserts (1732, 1736) and the error path on the second insert (1739):
https://github.com/microsoft/litebox/blob/9732adf0add28c663bd51ef9881a0b8992e20be8/litebox_shim_linux/src/syscalls/file.rs#L1732-L1741

If the second operation fails, say because the table is full, and a concurrent close is run on the same slot, then any other concurrent inserts will end up at that slot too. That means that if a close , or close -> insert run in between the first two lock operations, the consume on the first fd as a result of the second insert failing, can produce an error (if there is nothing at that slot) or consume a different file instead of the original written one inserted by this thread.

CvvT:
Fixes #1171: take the `raw_descriptor_store` write lock once in `sys_pipe2` and hold it across
both inserts and the rollback, so the writer's slot cannot be recycled.

### DIFF
--- a/litebox_shim_linux/src/syscalls/file.rs
+++ b/litebox_shim_linux/src/syscalls/file.rs
@@ -103,9 +103,17 @@ impl<Platform: ShimPlatform, FS: ShimFS> FilesState<Platform, FS> {
         &self,
         typed_fd: TypedFd<Subsystem>,
     ) -> Result<usize, TypedFd<Subsystem>> {
+        let mut rds = self.raw_descriptor_store.write();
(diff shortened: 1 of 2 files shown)


wdcui (approved):
LGTM. Thanks.

PR #1191, opened 2026-08-18
Restrict exception fixups to synchronous faults

sangho2:
This PR prevents the exception fixup path from misclassifying asynchronous faults as memory faults. Currently, an asynchronous signal delivered during exception fixup execution is handled as a memory fault, causing `memcpy_fallible` to abort and the signal to be lost.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -2420,7 +2420,7 @@ impl litebox::mm::linux::VmemPageFaultHandler for LinuxUserland {
 
 #[cfg(test)]
 mod tests {
-    use core::sync::atomic::AtomicU32;
+    use core::sync::atomic::{AtomicBool, AtomicU32, Ordering};
     use std::thread::sleep;
 
     use litebox::{fs::OFlags, platform::RawMutex};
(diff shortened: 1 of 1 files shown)


wdcui:
@sangho2 is this a valid concern?

Claude Opus 5: Found one high-severity issue at  litebox_platform_linux_userland/src/lib.rs:2117-2133 .

Opus reports that asynchronous  SIGSEGV  delivered while  in_guest == 1  bypasses the new  si_code > 0  guard in  next_signal_handler . The handler may therefore construct a guest exception from stale  REG_TRAPNO ,  REG_ERR , and  REG_CR2  values, injecting a bogus fault into the guest. The proposed fix is to test synchronicity in  exception_signal_handler  before calling  signal_handler_exit_guest , and route asynchronous signals through the pending/next-handler path. The same concern may apply to  SIGBUS ,  SIGFPE ,  SIGILL , and  SIGTRAP .

sangho2:
> Sangho Lee (@sangho2) is this a valid concern?
> 
> Claude Opus 5: Found one high-severity issue at  litebox_platform_linux_userland/src/lib.rs:2117-2133 .
> 
> Opus reports that asynchronous  SIGSEGV  delivered while  in_guest == 1  bypasses the new  si_code > 0  guard in  next_signal_handler . The handler may therefore construct a guest exception from stale  REG_TRAPNO ,  REG_ERR , and  REG_CR2  values, injecting a bogus fault into the guest. The proposed fix is to test synchronicity in  exception_signal_handler  before calling  signal_handler_exit_guest , and route asynchronous signals through the pending/next-handler path. The same concern may apply to  SIGBUS ,  SIGFPE ,  SIGILL , and  SIGTRAP .

It seems that what Opus found is an interesting bug, but orthogonal to the issue that this PR covers. This PR aims to ensure that `memcpy_fallible` is not cancelled due to async SIGSEGV. In this case, `in_guest` must be `0`.

Further analysis on async SIGSEGV while `in_guest == 1` reveals that LiteBox would deliver it to the guest as a kernel-triggered SIGSEGV at the stale CR2 which cannot be masked (DoS).

#1193

wdcui (approved):

PR #1192, opened 2026-08-18
Pin GitHub Actions to full-length commit SHAs

danfiedler-msft:
This PR pins GitHub Actions to full-length commit SHAs and adds a 7 day cooldown to Dependabot configuration for GitHub Actions. See more detail at https://aka.ms/action-pinning.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    groups:
+      github-actions:
+        patterns: ["*"]
+    schedule:
+      interval: "weekly"
+    cooldown:
+      default-days: 7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,14 +30,14 @@ jobs:
       RUSTDOCFLAGS: -Dwarnings
     steps:
       - name: Check out repo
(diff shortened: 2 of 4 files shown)


jaybosamiya-ms (approved):
Thanks Dan, the changes look good to me.  As a heads up, I'll rewrite the PR description to shorten it before I merge it in, since that becomes part of the git history with the merge queues we use.

Sidenote: for future reference, quick script to validate the commit hashes:
```sh
grep -rhoP 'uses:\s*\K[\w.-]+/[\w.-]+@[0-9a-f]+\s*#\s*v?[\w.-]+' .github/workflows/ \
  | sed 's/#//' | sort -u \
  | while read -r ref tag; do
      repo=${ref%@*}; sha=${ref#*@}
      actual=$(gh api "repos/$repo/git/ref/tags/$tag" --jq '.object.sha' 2>/dev/null)
      typ=$(gh api "repos/$repo/git/ref/tags/$tag" --jq '.object.type' 2>/dev/null)
      [ "$typ" = tag ] && actual=$(gh api "repos/$repo/git/tags/$actual" --jq '.object.sha')
      if [ "$actual" = "$sha" ]; then echo "OK   $repo $tag"; else echo "BAD  $repo $tag: comment=$tag sha=$sha actual=$actual"; fi
    done
```

PR #1194, opened 2026-08-18
fix(broker): keep the shared socket reactor alive when one association's readiness publication fails

willportnoy:
The broker's single epoll reactor thread serves every guest socket across all sessions. On the asynchronous event path a per-object readiness *publication* failure was fatal: it propagated out of `Reactor::run()` and triggered `fail_all_sockets()`, tearing down every session's sockets and exiting the reactor, so one guest whose sink could not accept a notification disconnected all others. This change makes asynchronous publication best-effort — matching the UDP endpoint handler — so one association's failure no longer fails the shared reactor; `update_snapshot` commits the cached snapshot before publishing, so it stays authoritative, while genuine host-syscall and internal-consistency errors still propagate and remain fatal. Synchronous command paths and the UDP enqueue rollback are unchanged. In production an admitted socket's publication cannot fail — admission bounds live readiness registrations to the sink's capacity — so this hardens against a fault admission already prevents and shrinks its blast radius. Regression tests inject publication failures on the reactor's readable and connect-completion paths and verify the reactor stays alive, the faulting socket and a second session's socket keep transacting, and cached readiness stays authoritative.

### DIFF
--- a/litebox_broker_platform_linux_userland/src/socket/tcp.rs
+++ b/litebox_broker_platform_linux_userland/src/socket/tcp.rs
@@ -1093,12 +1093,21 @@ pub(super) fn handle_socket_event(
     socket: &mut SocketEntry,
     events: epoll::EventFlags,
 ) -> BrokerResult<bool> {
+    // Readiness publication is best-effort on the shared reactor path: a single
+    // association's publication failure must not fail the reactor for every
+    // other session using it (this mirrors the UDP endpoint handler).
+    // `update_snapshot` commits the cached snapshot before publishing, so a
+    // dropped notification leaves the cached readiness authoritative. In
+    // production an admitted socket's publication does not fail: the broker
+    // bounds live readiness registrations to the sink's capacity at association
+    // setup, so this only absorbs a synthetic sink fault. Genuine host-syscall
+    // or internal-consistency errors still propagate and remain fatal.
     if socket.tcp_state().is_ok_and(|tcp| tcp.listening) {
-        update_snapshot(socket, None, readiness_from_epoll(socket, events))?;
+        let _ = update_snapshot(socket, None, readiness_from_epoll(socket, events));
         return Ok(false);
     }
     if socket.kind() == SocketKind::Udp {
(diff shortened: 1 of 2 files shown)


willportnoy:
_AI-authored, human-reviewed._

Before opening, this change was put through two independent static reviewers — GPT-5.6 Sol (xhigh) and Claude Opus 5 (xhigh). Both converged on a **single** substantive point, with different severity calls:

- **Opus 5 — SHIP.** Independently verified: (a) the only two `ReactorFailure::Broker` escalations are the two handler calls in `run()`, and `process_commands` / `drain_wake` / `expire_deadlined_state` cannot escalate a publication failure; (b) the async-best-effort vs. synchronous-transactional split is correct — the guest-to-guest UDP enqueue rollback is untouched; (c) the new test genuinely exercises the async `handle_socket_event` path and is not flaky (the armed predicate cannot be consumed early, and the wait is deadline-asserted).
- **Sol — one High note:** the shared concern below.

**The shared concern — lost wakeup on a dropped async publication.** Because publication is now best-effort and `update_snapshot` only re-publishes on change, a dropped notification is not retried; combined with edge-triggered host epoll, a guest blocked purely on that notification could stall on that one socket.

**Resolution (not production-reachable for sockets).** The production sink only rejects publication with `TooManyObjects`, i.e. when registering a *new* object beyond capacity. Broker admission bounds live readiness registrations to the sink capacity at association setup (`litebox_broker_host` readiness admission), so an already-admitted socket's publication does not fail. This PR therefore hardens the reactor against a fault admission control already prevents, and is strictly better than the prior behavior (which forced every session's sockets to `ERROR` and exited the reactor for the same input). The same tradeoff is already accepted in the merged UDP handler. The reasoning is captured in the `handle_socket_event` doc comment and the commit body. If sink failure were ever made genuinely reachable, the follow-up is to keep the object dirty on a failed publish (track last-published separately) rather than re-escalate — noted for the timerfd rebase, where readiness admission accounting for timer registrations is a separate item.

PR #1195, opened 2026-08-18
Switch layered to overlay filesystem

jaybosamiya-ms:
This PR removes the old layered file system, which had to deal with two different semantics with non-trivial complexity, instead to an overlay-based file system.  This is aided by work I'd done to set up the ability to mount things (via the composer, #995), as well as a backend trait more amenable to the necessary composition (`Backend`, #887).  With this PR, the last of the remaining layered file systems are removed.


jaybosamiya-ms:
As with many of my prior PRs, this PR is easiest to review one commit at a time :)

litebox/src/fs/overlay.rs
```diff
@@ -0,0 +1,1174 @@
...
+        Ok(current)
+    }
+
+    /// Copy the lower file `lower` up into the upper backend as `name` within `upper_dir`.
+    ///
+    /// The namespace lock is held for the whole copy, so a partially written (or failed and
+    /// unlinked) upper file is never observable.
+    // XXX(jayb): holding the namespace lock across a whole-file byte copy blocks every other
+    // namespace operation for as long as the copy takes. Ideally, we would avoid this by doing an
+    // atomic link/rename, would need to update `Backend` for that.
```
wdcui, line 260: This is important for performance. We may want to fix it in the near future.

wdcui:
GPT-5.6 sol and Opus 5 reported the following issues:

1. High — Writable opens can bypass permissions enforced by a lower backend
 litebox/src/fs/overlay.rs:843-880 
For a modifying open, the overlay changes the lower open flags to  O_RDONLY , copies the file up, and returns the lower backend’s original  PermissionCheck . A self-enforcing backend such as NineP returns  ByBackend , so the resolver assumes the requested write access was already authorized—even though NineP only authorized the substituted read-only open. A user who may read but not write a lower file can therefore obtain a writable upper copy.
The overlay needs to authorize the caller’s original access mode independently from the read handle used for copy-up.
2. High — The SNP runner now permits guest writes directly into the host 9p share
 litebox_runner_snp/src/main.rs:214-247 
The previous filesystem stack placed writable in-memory/layered behavior above NineP. This PR instead mounts NineP directly at  / , with InMem mounted only at  /tmp . Creating or modifying files outside  /tmp  can now mutate the host export rather than remaining ephemeral or being rejected.
The SNP runner should mount an in-memory-upper  Overlay  over NineP at  / , with  /dev  and  /tmp  retained as nested mounts, unless direct host writes are an intentional security-policy change.
3. Medium — Copy-up breaks lower-layer hard links
 litebox/src/fs/overlay.rs:843-880 
If two lower names refer to the same inode, writable opens copy each pathname into an independent upper file. The second copy can start from stale lower contents, after which the aliases no longer share data or metadata. Although node identities and copied-up handles are tracked, the writable-open path does not reuse an existing copy-up for the same lower node.
Copy-up needs alias-aware handling keyed by lower identity, likely requiring upper-backend hard-link support or explicit tracking of all names associated with a copied inode.
4. Medium — Open directory handles can silently rebind to a replacement directory
 litebox/src/fs/overlay.rs:793-809,904-913,965-971 
 OverlayDir  stores only its pathname. Directory reads and status calls re-resolve that path every time. If an opened directory is removed and another directory is created at the same path, operations through the old descriptor address the replacement directory rather than the originally opened object.
Owned directory handles should retain stable underlying handles or an identity/generation that detects replacement.
5. Medium — Copied-up file handles are retained for the overlay’s entire lifetime
 litebox/src/fs/overlay.rs:60-65,573-593 
Every copy-up stores an upper  FileHandle  in  State::copied_up , and entries are never removed. Closing and unlinking the file does not release that handle. This retains deleted InMem data and prevents NineP’s final-handle  Tclunk , allowing unbounded memory or backend-resource growth.
The overlay needs lifecycle or alias tracking so mappings can be removed once no lower-backed handles or namespace aliases require them.
6. Medium — Failure partway through marker cleanup can expose lower entries
 litebox/src/fs/overlay.rs:422-438,1135-1144 
 clear_markers  deletes markers one at a time but returns only an error if a later deletion fails. The list of markers already removed is lost, so  rmdir  cannot restore them. Whiteouts removed before the failure can expose lower entries even though  rmdir  returned an error.
Marker cleanup should either be transactional or return the successfully removed markers alongside any error so the caller can restore them.
7. Medium — An upper  status  failure publishes an untracked copy-up
 litebox/src/fs/overlay.rs:298-308 
After creating and publishing an upper file,  copy_up_file  silently ignores failure of  upper.status . Without the upper identity,  bind_copy_up  is skipped. Existing lower-backed handles continue using stale lower contents while new path opens see the upper file.
Failure to obtain the upper identity should fail and roll back the copy-up rather than returning success.
8. Medium — Concurrent unlink and  chmod / chown  can recreate a deleted lower file
 litebox/src/fs/overlay.rs:342-378,1053-1090,1150-1172 
The resolver obtains a path handle before entering  Overlay::chmod  or  chown . If another thread unlinks the lower file between those steps, the metadata operation still uses the stale handle’s recorded parent path and name.  ensure_upper  can then copy the object into that pathname, overriding the whiteout and making the deleted file visible again.
Before path-based copy-up, the overlay should re-resolve the name under the namespace lock and verify that it still refers to the same overlay identity.

wdcui (approved):

wdcui (approved):
LGTM. Thanks!

jaybosamiya-ms:
Thanks Weidong!

Quick notes on the agent findings:
- The two "high" ones seem valid to fix in this PR, will fix them up before merging.  For the 9p one it is possible I misunderstood what the old snp side was doing, so accidentally restricted it to `/tmp` for the in-mem; will tweak it back to maintain existing semantics
- The medium ones are known things planned as part of future PRs (we do not have hardlink support yet at all, so unsurprisingly it behaves weirdly under hardlinks; various other things get auto-solved as part of supporting better transactionality at backend level)

litebox/src/fs/overlay.rs
jaybosamiya-ms: Agreed.  In practice, writing to an existing file from below is a rare operation (e.g., the tar file contains read-only things like executables and .so files), so this lock being taken should be quite rare.

PR #1197, opened 2026-08-18
Skip empty embedded TA binaries during registration

praveen-pk:
Avoid panicking during initialization when an embedded TA binary is an empty placeholder.

Fixes: e33f6ffd: share TA binaries through a global UUID map

### DIFF
--- a/litebox_runner_lvbs/src/lib.rs
+++ b/litebox_runner_lvbs/src/lib.rs
@@ -1355,7 +1355,9 @@ fn register_embedded_ta(shim: &litebox_shim_optee::OpteeShim, ta_binary: &'stati
 /// Register all TA binaries embedded in the runner image.
 fn register_embedded_tas(shim: &litebox_shim_optee::OpteeShim) {
     for ta_binary in TA_BINARIES {
-        assert!(register_embedded_ta(shim, ta_binary));
+        if !ta_binary.is_empty() {
+            assert!(register_embedded_ta(shim, ta_binary));
+        }
     }
 }
 


praveen-pk:
Without this fix LiteBox panics at:

```
panicked at litebox_runner_lvbs/src/lib.rs:1358:9:
assertion failed: register_embedded_ta(shim, ta_binary)
```

litebox_runner_lvbs/src/lib.rs
```diff
@@ -1355,7 +1355,9 @@ fn register_embedded_ta(shim: &litebox_shim_optee::OpteeShim, ta_binary: &'stati
 /// Register all TA binaries embedded in the runner image.
 fn register_embedded_tas(shim: &litebox_shim_optee::OpteeShim) {
     for ta_binary in TA_BINARIES {
-        assert!(register_embedded_ta(shim, ta_binary));
+        if !ta_binary.is_empty() {
+            assert!(register_embedded_ta(shim, ta_binary));
```
sangho2, line 1359: If an empty placeholder is legitimate, we might not need `assert`, or `debug_assert` might be better.

sangho2 (approved):
LGTM. I left a comment.

litebox_runner_lvbs/src/lib.rs
tgopinath-microsoft: you can instead do the below change. This will just skip the for loop  if there are no TAs 
-const TA_BINARY: &[u8] = &[0u8; 0];
-const TA_BINARIES: &[&[u8]] = &[TA_BINARY];
+const TA_BINARIES: &[&[u8]] = &[];

praveen-pk: >you can instead do the below change. This will just skip the for loop if there are no TAs

I left the existing implementation as is, so that users can use it for reference

praveen-pk: >we might not need assert,

the current `assert` will panic if the embedded TA is corrupt. It would be best to report this earlier in the boot.

The current version allows empty placeholder and panic if a non-empty TA fails to parse.

sangho2: If that's the goal, `debug_assert` would be better option IMO. Panicking the entire VM is a bit worrisome. Whether an embedded TA is corrupted or not doesn't affect LiteBox's own operations.

PR #1208, opened 2026-08-19
Revert "Complete Windows guest startup and file I/O hardening"

CvvT:
Reverts microsoft/litebox#1207


CvvT:
Revert #1207. It supposes to be a local WIP branch but was merged by agents without me noticing it.

PR #1213, opened 2026-08-20
Dynamic TA Support

praveen-pk:
If TA is not found within the LiteBox's cache, load the TA from Normal world. This version only supports ELF format for TA binaries.


sangho2 (commented):
[AI review]

Review of the LOAD_TA RPC initiation path. Draft-stage, so I focused on wire-protocol correctness and the VTL0-facing edges rather than polish. Comments only — no blocking verdict.

---

## Correctness

### 1. UUID is byte-swapped on the wire (`litebox_runner_lvbs/src/lib.rs:595-603`)

`TeeUuid::to_le_bytes()` is **not** the inverse of `TeeUuid::from_bytes()` / `from_u64_array()`.

The incoming UUID is decoded in `litebox_shim_optee/src/msg_handler.rs:453` via `TeeUuid::from_u64_array([a, b])`, which lays the two LE u64s down as 16 octets and then reads `time_low`/`time_mid`/`time_hi_and_version` as **big-endian** (RFC 4122 octet order, `litebox_common_optee/src/lib.rs:655-658`). `to_le_bytes()` re-emits those same three fields **little-endian**, so each is byte-swapped relative to the octets we received.

Using the existing test vector at `litebox_common_optee/src/lib.rs:2549`:

```
driver sent:  params[0].u.value.a = 0xe311f8e7_e0b34f38
this code sends:                    0x11e3e7f8_384fb3e0
```

tee-supplicant does `uuid_from_octets(&uuid, (void *)&params[0].a)` (RFC 4122 octets), matching optee_os's `tee_uuid_to_octets()` in `rpc_load()`. So normal world will look up the wrong TA.

Suggest adding a `to_bytes()` / `to_u64_array()` on `TeeUuid` that mirrors `from_bytes` / `from_u64_array`, with a round-trip unit test, and using that here.

(Note `to_le_bytes()`'s only other caller, `syscalls/pta.rs:309`, is a HUK KDF input where only self-consistency matters — that's why this has gone unnoticed.)

### 2. Stale rmem fields leak into the RPC (`prepare_load_ta_rpc`, `lib.rs:583-616`)

`rpc_args` is parsed directly out of normal-world memory (`read_optee_msg_args_from_phys`, `msg_handler.rs:182-204`), so `params[1].data` can contain whatever the driver/previous RPC left there. The function sets `attr` and the `size` field but never clears `offs` (`data[0..8]`) or `shm_ref` (`data[16..24]`).

optee_os emits an **all-zero** rmem for a NULL memref (`get_rpc_arg()` in `core/kernel/thread.c`), and the Linux driver will try to resolve a non-zero `shm_ref` cookie. Please zero `params[..num_params]` (or at least param 1) before populating.

### 3. `set_param_memref_size` and `set_param_rmem` clobber each other (`lib.rs:609-614`)

`set_param_rmem` does `data.copy_from_slice(rmem.as_bytes())` over all 24 bytes, so when `memref` is `Some`, the `memref_size` written on the previous line is silently discarded. Either drop the `memref_size` parameter when a full rmem is supplied, or set `rmem.size = memref_size` before writing. As written, the stage-2 call path is already broken.

### 4. `unwrap()` on a normal-world-driven path (`lib.rs:554`)

```rust
let rpc_args_ref = rpc_args.as_ref().unwrap();
```

A panic here is a VTL1 kernel panic (`#[panic_handler]` → `raise_vtl0_gp_fault`). Today it's guaranteed `Some` because `handle_open_session` is the only `RpcCmd` producer and it bails out earlier if `rpc_args` is `None` — but that's an implicit cross-function invariant that the next `RpcCmd` producer will break. Prefer:

```rust
let Some(rpc_args_ref) = rpc_args.as_ref() else {
    smc_args.set_return_code(OpteeSmcReturnCode::EBadCmd);
    return *smc_args;
};
```

### 5. Behavior regression when `rpc_args` is `None` (`lib.rs:642`)

For a plain `OpteeSmcFunction::CallWithArg` (driver without `RPC_ARG`), `rpc_args` is `None` (`msg_handler.rs:224-228`), so a cache miss now returns `Err(EBadCmd)` — the driver fails the whole SMC. Previously (and still, in `open_session_new_instance` at `lib.rs:857-864`) a missing TA produced a clean `TeeResult::ItemNotFound` in `msg_args` with `Ok(())`. Suggest falling through to the existing `ItemNotFound` path instead of `EBadCmd` when RPC isn't available.

---

## Design / follow-up

### 6. Nothing can resume the RPC yet

`OpteeSmcFunction` (`litebox_common_optee/src/lib.rs:2276-2288`) has no `OPTEE_SMC_FUNCID_RETURN_FROM_RPC` variant, so `func_id()` (`lib.rs:2226`) returns `EBadCmd` when the driver re-enters after servicing LOAD_TA, and no pending open-session state is persisted anywhere. Net effect of this PR standalone: a cached-TA miss goes from "clean `ItemNotFound`" to "failed OpenSession". Worth stating explicitly in the PR description which stage adds resume, and confirming stage 1 won't be merged to `main` ahead of it (or is gated).

### 7. Duplicated cache-miss check

`handle_open_session` (`lib.rs:637`) and `open_session_new_instance` (`lib.rs:858`) now both do `get_ta_bin(..).is_none()`. Consider a single detection point. Also note the single-instance/sibling path never reaches the new check — presumably intentional (TA already resident), but worth a comment.

### 8. Duplicate binding (`lib.rs:635` and `lib.rs:649`)

```rust
let ta_uuid = ta_req_info.uuid.ok_or(OpteeSmcReturnCode::EBadCmd)?;
```

appears twice; the second is a redundant shadow. Please drop it.

### 9. Inconsistent error mapping (`lib.rs:597`, `lib.rs:608`)

`.map_err(|_| OpteeSmcReturnCode::EBadCmd)` on `set_param_attr_type`, which already returns `OpteeSmcReturnCode` — and it discards the more accurate `ENotAvail`. The neighboring `set_param_value` / `set_param_memref_size` calls just use `?`. Use `?` throughout.

### 10. `prepare_load_ta_rpc` visibility and shape (`lib.rs:583`)

It's `pub` in the runner crate with no external caller. Make it private, or move it into `litebox_common_optee` alongside the other RPC helpers so both runners can use it. Also, `memref_size`/`memref` are always `0`/`None` today — per the repo's "no speculative flexibility" guidance, consider trimming until stage 2 actually needs them (and see #3).

### 11. New setters in `litebox_common_optee` (`lib.rs:2120-2162`)

- No unit tests, despite the crate having a test module with existing param round-trip coverage (`test_optee_rpc_args_roundtrip`). These are public API on the VTL0 boundary — worth covering, especially bounds behavior and the rmem layout offsets.
- `set_param_memref_size`'s doc says "rmem parameter" but it writes `data[8..16]` unconditionally; it silently "succeeds" on a value param. Either document that it's layout-based and attr-agnostic (it's also valid for tmem), or validate the attr type.
- `set_param_attr_type` overwrites the whole `attr` word, dropping `META`/`NONCONTIG` bits. Fine for RPC args, but worth a doc note given `OpteeMsgAttr` carries those flags.

### 12. Deleted rationale comment (`litebox_common_optee/src/lib.rs`, removed lines after `set_param_tmem`)

The "RPC does not use rmem params" note is now obsolete — but rather than deleting it outright, consider replacing it with the actual rule: optee_os maps a NULL memref to `RMEM_*` with an all-zero body and a registered-shm memref to `RMEM_*`, tmem otherwise. That's non-obvious and directly informs #2.

### 13. Minor

- `shim.get_ta_bin(&ta_uuid).is_none()` clones an `Arc<[u8]>` just for a presence check; a `contains_ta_bin`/`has_ta_bin` would be cheaper and clearer.
- `OpteeShimBuilder::new().build()` per OpenSession (`lib.rs:634`) constructs a fresh `LiteBox` + `PageManager` — consistent with existing sites (`lib.rs:229`, `lib.rs:857`), but this PR adds a second one per open-session call.
- Dropping `rpc_get_ta_bin` is a clean no-op removal (it always returned `None`, and no caller depended on the fallback) — nice cleanup. 👍

litebox_runner_lvbs/src/lib.rs
```diff
@@ -558,6 +572,47 @@ fn optee_smc_handler(smc_args_addr: usize) -> OpteeSmcArgs {
...
+    memref_size: u64,
+    memref: Option<OpteeMsgParamRmem>,
+) -> Result<(), OpteeSmcReturnCode> {
+    rpc_msg_args.cmd = OpteeRpcCommand::LoadTa;
+    rpc_msg_args.num_params = 2;
+
+    rpc_msg_args
+        .set_param_attr_type(0, OpteeMsgAttrType::ValueInput)
+        .map_err(|_| OpteeSmcReturnCode::EBadCmd)?;
+    let uuid_bytes = ta_uuid.to_le_bytes();
```
sangho2: We need to double-check this endian. AFAIR, Linux kernel OP-TEE driver transmits big-endian TA UUID to LiteBox. I doubt that it would expect a litte-endian UUID from LiteBox.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -558,6 +572,47 @@ fn optee_smc_handler(smc_args_addr: usize) -> OpteeSmcArgs {
...
+        },
+    )?;
+
+    rpc_msg_args
+        .set_param_attr_type(1, OpteeMsgAttrType::RmemOutput)
+        .map_err(|_| OpteeSmcReturnCode::EBadCmd)?;
+    rpc_msg_args.set_param_memref_size(1, memref_size)?;
+    if let Some(rmem) = memref {
+        rpc_msg_args.set_param_rmem(1, rmem)?;
+    }
```
sangho2: `set_param_rmem` would overwrite the memref size that `set_pram_memref_size` wrote. is this expected? If it is, better not to call `set_param_memref_size` if `memref` is `Some` to avoid confusion.

Also, I wonder whether `RmemOutput` is correct here. According to the old study (the deleted comment), RPC does use Tmem for this purpose. Is this difference because we are not doing real RPC yet? In that case, better to document it.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -548,7 +549,20 @@ fn optee_smc_handler(smc_args_addr: usize) -> OpteeSmcArgs {
         unsafe { switch_to_base_page_table() };
 
         if let Err(e) = result {
-            smc_args.set_return_code(e);
+            if e == OpteeSmcReturnCode::RpcCmd {
+                debug_serial_println!("OP-TEE SMC returning RPC command to normal world");
+                let rpc_args_ref = rpc_args.as_ref().unwrap();
```
sangho2: nits: Better not to use `unwrap()` here to avoid kernel panics. We could return an SMC error.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -569,12 +624,28 @@ fn optee_smc_handler(smc_args_addr: usize) -> OpteeSmcArgs {
...
     }
+    let shim: litebox_shim_optee::OpteeShim = litebox_shim_optee::OpteeShimBuilder::new().build();
+
+    let ta_uuid = ta_req_info.uuid.ok_or(OpteeSmcReturnCode::EBadCmd)?;
+    if shim.get_ta_bin(&ta_uuid).is_none() {
+        debug_serial_println!(
+            "TA binary not found for uuid={:?}, requesting load from normal world",
+            ta_uuid
+        );
+        let rpc_mut = rpc_args.as_mut().ok_or(OpteeSmcReturnCode::EBadCmd)?;
```
sangho2: Why do we return `EBadCmd` here? To work around a lack of true RPC? An SMC command itself seems valid.

litebox_common_optee/src/lib.rs
```diff
@@ -2117,6 +2117,48 @@ impl OpteeRpcArgs {
...
+        if index >= self.num_params as usize {
+            Err(OpteeSmcReturnCode::ENotAvail)
+        } else {
+            self.params[index].data[8..16].copy_from_slice(&size.to_le_bytes());
+            Ok(())
+        }
+    }
+
+    /// Set an rmem parameter by index with bounds checking against `num_params`.
+    pub fn set_param_rmem(
```
sangho2, line 2154: Looks like `set_param_rmem` does what `set_param_memref_size` does + alpha. Better to define two other methods for `offs` and `shm_ref` as well, or just use this function only.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -558,6 +572,47 @@ fn optee_smc_handler(smc_args_addr: usize) -> OpteeSmcArgs {
...
+
+    rpc_msg_args
+        .set_param_attr_type(0, OpteeMsgAttrType::ValueInput)
+        .map_err(|_| OpteeSmcReturnCode::EBadCmd)?;
+    let uuid_bytes = ta_uuid.to_le_bytes();
+    rpc_msg_args.set_param_value(
+        0,
+        OpteeMsgParamValue {
+            a: u64::from_le_bytes(uuid_bytes[..8].try_into().unwrap()),
+            b: u64::from_le_bytes(uuid_bytes[8..].try_into().unwrap()),
```
sangho2: same. better to re-check what the OP-TEE driver actually expects.

sangho2 (commented):
Left some comments.

litebox_runner_lvbs/src/lib.rs
praveen-pk: I confirmed that Kernel dies capture the UUIDs in big-endian format. I will fix this in next update.

litebox_runner_lvbs/src/lib.rs
praveen-pk: https://optee.readthedocs.io/en/latest/architecture/trusted_applications.html#loading-ree-fs-ta is the protocol for LOAD_TA.
From what I have traced in `optee_os` and `optee_client`:

## First LOAD_TA
send `LOAD_TA` with `memref_size` set to 0.
If `memref_size` it set to 0, tee-supplicant will return the size of the TA in bytes.

## SHM_ALLOC
Get REE to allocate some memory to load the TA.

## Second LOAD_TA 
Register the TMEM into a shm and invoke a Second LOAD_TA. REE will actually load the TA into the provided shm in this call.


While tracing the first load_ta, the code branches of based on the input `memref_size` of `0` without checking the type. Further digging I noticed that the argument should be `Tmem`. I will fix in the next update.


>If it is, better not to call set_param_memref_size if memref is Some to avoid confusion.

As I mentioned above code handles the first and second LOAD_TA calls as described above. I do see that memref_size does get clobbers if memref is valid (during second LOAD_TA). I will fix this.

litebox_runner_lvbs/src/lib.rs
praveen-pk: Yes.
The following `prepare_load_ta_rpc` needs a valid RPC_args. I am handling the case when the client incorrectly sends a command without valid RPC_args.

litebox_runner_lvbs/src/lib.rs
praveen-pk: Confirmed this is Big-endian.

litebox_runner_lvbs/src/lib.rs
praveen-pk: I noticed the kernel always sends `rpc_args` alongside msg_args. So, I was returning an error, if some corruption caused rpc_args to not be present.


I don't see any downsides to returning ItemNotFound here. Updated, in latest version.

litebox_common_optee/src/lib.rs
praveen-pk: fixed

litebox_common_optee/src/lib.rs
```diff
@@ -2148,10 +2176,6 @@ impl OpteeRpcArgs {
             Ok(())
         }
     }
-
-    // Note: RPC does not use rmem params. Rmem requires pre-registered shared memory
-    // references from the normal-world driver, which is a main-messaging-path concept.
-    // RPC uses tmem for buffer references since OP-TEE provides physical addresses directly.
```
praveen-pk, line 2136: Normal first sends a TMEM for LOAD_TA request. In response to SHM_ALLOC, normal later responds with RMEM. 

So above comment is not valid, based on what I observed.

litebox_common_optee/src/lib.rs
```diff
@@ -2130,10 +2177,52 @@ impl OpteeRpcArgs {
...
+}
+
+/// Prepare a LOAD_TA RPC request to be sent to normal world.
+///
+/// When `memref` is `None`, the request asks normal world to return the TA size.
+/// When it is `Some`, the request provides registered memory for the TA binary.
+pub fn prepare_load_ta_rpc(
+    rpc_msg_args: &mut OpteeRpcArgs,
+    ta_uuid: TeeUuid,
+    memref_size: u64,
```
sangho2, line 2189: Looks like `memref_size` is not needed. It is zero for the first LOAD_TA RPC, and is overwritten by `rmem.size` for the second RPC.

litebox_common_optee/src/lib.rs
```diff
@@ -2130,10 +2177,52 @@ impl OpteeRpcArgs {
...
+        0,
+        OpteeMsgParamValue {
+            a: uuid_bytes[0],
+            b: uuid_bytes[1],
+            c: 0,
+        },
+    )?;
+
+    if memref.is_none() {
+        // First call of LOAD_TA protocol: normal world returns the TA size in memref_size.
```
sangho2, line 2207: this reads like normal world writes the TA size the `memref_size` variable, but, in fact, `tmem.size` is the one.

litebox_common_optee/src/lib.rs
```diff
@@ -2130,10 +2177,52 @@ impl OpteeRpcArgs {
...
+        rpc_msg_args.set_param_tmem(
+            1,
+            OpteeMsgParamTmem {
+                buf_ptr: 0,
+                size: memref_size,
+                shm_ref: 0,
+            },
+        )?;
+    } else {
+        // Second call of LOAD_TA protocol: secure world provides a memref for the TA binary.
```
sangho2, line 2218: a bit misleading. `rmem` itself is from VTL0 (it is VTL0/normal world memory), and LiteBox/VTL1 just passes its information the VTL0 again to let it populate data.

litebox_common_optee/src/lib.rs
sangho2: yes. sounds like it becomes RMEM as we use dynamic SHM now.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -569,18 +586,44 @@ fn optee_smc_handler(smc_args_addr: usize) -> OpteeSmcArgs {
...
+            msg_args.ret = TeeResult::ItemNotFound;
+            msg_args.ret_origin = TeeOrigin::Tee;
+            write_non_ta_msg_args_to_normal_world(msg_args, msg_args_phys_addr)?;
+            return Ok(());
+        };
+        // LOAD_TA is a two-call protocol. The first call uses a zero-sized
+        // memref (a NULL buffer) so normal world returns the TA size.
+        prepare_load_ta_rpc(rpc, ta_uuid, 0, None)?;
+        return Err(OpteeSmcReturnCode::RpcCmd);
+    }
```
sangho2, line 619: We should move this code block into the `OpenSessionTarget::NewInstance` arm or the `open_session_new_instance` function. This code block makes an RPC call if a given TA is not in the in-memory TA binary cache/storage (which can be deleted if there is memory pressure or TA update with `remove_ta_bin`). However, if a TA is a single-instance TA which is already loaded, we don't need these bogus RPC and binary loading paths since we can reuse/share the loaded TA page table.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -569,18 +586,44 @@ fn optee_smc_handler(smc_args_addr: usize) -> OpteeSmcArgs {
...
+
     let client_identity = ta_req_info.client_identity;
     let params = &ta_req_info.params;
 
     session_manager().with_ta(&ta_uuid, |target| match target {
+        // A sibling points to an already-loaded single-instance TA, so no
+        // binary cache lookup or LOAD_TA request is needed on this path.
```
sangho2, line 626: Align with the above comment. `handle_open_session` somehow unconditionally does binary cache lookup and RPC.

litebox_shim_optee/src/lib.rs
```diff
@@ -190,17 +190,14 @@ impl GlobalState {
...
-        if let Some(ta_bin) = self.ta_uuid_map.get(ta_uuid) {
-            Some(ta_bin)
-        } else {
-            let ta_bin = Self::rpc_get_ta_bin(ta_uuid)?;
-            if !self.store_ta_bin(ta_uuid, &ta_bin) {
-                return None;
-            }
-            Some(ta_bin)
-        }
+        self.ta_uuid_map.get(ta_uuid)
```
sangho2, line 195: Fine for now, but I think we at least need to maintain TODO for RPC or TA binary pinning. This works now because we never call `remove_ta_bin`. However, if we exercise it, loading TA binaries only at the `handle_open_session` function can suffer from TOCTOU issues. We need to either implement real RPC for TA loading, or pin `Arc<ta_bin>` until we load it into the memory. Of course, not for this PR series.

litebox_runner_lvbs/src/lib.rs
praveen-pk: I will look at moving to `OpenSessionTarget::NewInstance`.

Another note, further investigating TA store (maps to ta_uuid_map) I realized the TA binary should be dropped from uuid_map cache as soon as ldelf loads it. 

If a TA is configured as single-instance, and KEEP_ALIVE is off, OP-TEE should use RPC to re-load the TA Binary from REE for every new request.

litebox_shim_optee/src/lib.rs
praveen-pk: From what I checked, OP-TEE does not support TA binary pinning at all. So, we should not implement it either.

sangho2: Clarification: we need pinning because we use `Arc` here. if no one refcounts a TA binary (including uuid map itself due to remove_ta_bin), the binary will be removed from the memory and `ldelf` might fail to read it.

litebox_runner_lvbs/src/lib.rs
sangho2: single instance with no keep alive, I agree with you, but keep-alive single instance and muti-instance TAs can suffer from tricky concurrency issues.

litebox_common_optee/src/lib.rs
praveen-pk: fixed in next version.

litebox_common_optee/src/lib.rs
praveen-pk: updated the comment.

litebox_common_optee/src/lib.rs
praveen-pk: addressed

litebox_runner_lvbs/src/lib.rs
praveen-pk: >tricky concurrency issues.
Yes. I made note of these. I will fix it once I merge the basic Dyn Ta Support.

praveen-pk:
fyi, I am working on a fix for race conditions while invoking Dynamic TA in multi-instance case. I will push that as a follow up PR.

litebox_common_optee/src/lib.rs
```diff
@@ -2143,10 +2255,102 @@ impl OpteeRpcArgs {
             Ok(())
         }
     }
+}
+
+/// Prepare a shared-memory allocation RPC request to be sent to normal world.
+pub fn prepare_shm_alloc_rpc(
```
sangho2, line 2261: nits: any particular reasons to make these functions (`prepare_shm_alloc_rpc`, `prepare_shm_free_rpc`, ...) free functions? It would be better to make them `OpteeRpcArgs` methods. We can use `&mut self` and drop `_rpc`.

litebox_common_optee/src/lib.rs
```diff
@@ -2143,10 +2255,102 @@ impl OpteeRpcArgs {
...
+) -> Result<(), OpteeSmcReturnCode> {
+    rpc_msg_args.cmd = OpteeRpcCommand::ShmAlloc;
+    // Match OP-TEE's get_rpc_arg(): default to failure in case normal world
+    // returns without updating the RPC result.
+    rpc_msg_args.ret = TeeResult::GenericError;
+    rpc_msg_args.num_params = 1;
+
+    rpc_msg_args
+        .set_param_attr_type(0, OpteeMsgAttrType::ValueInput)
+        .map_err(|_| OpteeSmcReturnCode::EBadCmd)?;
```
sangho2, line 2275: do we need `.map_err()`? Other `prepare_*_rpc`  functions don't do this. inconsistency.

litebox_common_optee/src/lib.rs
```diff
@@ -2143,10 +2255,102 @@ impl OpteeRpcArgs {
...
+
+/// Prepare a shared-memory free RPC request to be sent to normal world.
+/// This will free the memory allocated by SHM_ALLOC request in normal world.
+pub fn prepare_shm_free_rpc(
+    rpc_msg_args: &mut OpteeRpcArgs,
+    shm_type: OpteeRpcShmType,
+    shm_ref: u64,
+) -> Result<(), OpteeSmcReturnCode> {
+    rpc_msg_args.cmd = OpteeRpcCommand::ShmFree;
+    rpc_msg_args.ret = TeeResult::GenericError;
```
sangho2, line 2297: nits: this doesn't have a comment unlike the above and below functions. Perhaps better to make only the above one have a comment, and let this one and the below one refer to it.

litebox_common_optee/src/lib.rs
```diff
@@ -2499,6 +2727,23 @@ pub const HUK_SUBKEY_MAX_LEN: usize = 32;
...
+        }
+    }
+
+    #[cfg(target_pointer_width = "64")]
+    #[test]
+    fn test_rpc_context_id_rejects_upper_bits() {
+        let mut args = OpteeSmcArgs::default();
+        args.args[3] = (u32::MAX as usize) + 1;
+        assert_eq!(args.get_rpc_context_id(), Err(OpteeSmcReturnCode::EBadCmd));
+    }
```
sangho2, line 2745: These tests are too trivial, not worthwhile to maintain.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -542,48 +549,641 @@ fn optee_smc_handler(platform: &'static Platform, smc_args_addr: usize) -> Optee
...
+                })
+            },
+        )
+    } else {
+        handle_optee_smc_args(platform, &mut smc_args)
+    };
+    let Ok(smc_result) = smc_result else {
+        if is_return_from_rpc && let Ok(context_id) = smc_args.get_rpc_context_id() {
+            discard_rpc_context(context_id);
+        }
```
sangho2, line 584: The RPC branch (`is_return_from_rpc == true`) already knows `context_id`. why not call `discard_rpc_context` at the end of that branch?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -542,48 +549,641 @@ fn optee_smc_handler(platform: &'static Platform, smc_args_addr: usize) -> Optee
...
+                    // request identity in trusted state before overwriting it.
+                    let (registered_shm_ref, regd_shm_offset) =
+                        match smc_args.optee_regd_shm_ref_and_offset() {
+                            Ok(location) => location,
+                            Err(error) => {
+                                smc_args.set_return_code(error);
+                                return *smc_args;
+                            }
+                        };
+                    let ta_uuid = match decode_ta_request(platform, &msg_args)
```
sangho2, line 647: please fix clippy issues.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -542,48 +549,641 @@ fn optee_smc_handler(platform: &'static Platform, smc_args_addr: usize) -> Optee
...
+                    if smc_args.func_id() != Ok(OpteeSmcFunction::CallWithRegdArg) {
+                        smc_args.set_return_code(OpteeSmcReturnCode::EBadCmd);
+                        return *smc_args;
+                    }
+                    let Some(rpc_args_ref) = rpc_args.as_ref() else {
+                        smc_args.set_return_code(OpteeSmcReturnCode::EBadCmd);
+                        return *smc_args;
+                    };
+                    // RPC continuation reuses args[3] for the context ID. Preserve the
+                    // request identity in trusted state before overwriting it.
```
sangho2, line 638: Looks like this is out of place. perhaps it belongs to `rpc_context_map().allocate()`?

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -1022,3 +1135,28 @@ fn get_shm_info_from_optee_msg_param_rmem(
         rmem.size.trunc(),
     )
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn rpc_args_range_can_cross_page_boundary() {
```
sangho2, line 1144: without actual mapping, this test doesn't do meaningful stuffs (and actual mapping is not straightforward here).

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+            Some(RpcStage::ShmAlloc)
+        );
+        assert_eq!(
+            contexts.transition(context_id, RpcStage::LoadTaSize, RpcStage::LoadTaBinary),
+            Err(RpcContextError::UnexpectedStage)
+        );
+    }
+
+    #[test]
+    fn taking_a_context_rejects_replay() {
```
sangho2, line 471: looks trivial. It just removes the same entry twice from a map, which just tests the map itself. Many tests in this file are with the same issue.

litebox_shim_optee/src/session.rs
```diff
@@ -1073,6 +1073,29 @@ mod tests {
...
+    #[test]
+    fn keep_alive_instance_is_reused_after_last_session_closes() {
+        let manager = SessionManager::new();
+        let uuid = make_uuid(0xA5);
+        let flags = single_instance_flags() | TaFlags::INSTANCE_KEEP_ALIVE;
+
+        register_for_test(&manager, 107, flags, 12, uuid);
+        let instance = manager.single_instance_cache.get(&uuid).unwrap();
+
+        assert_eq!(manager.unregister_session(107), Some(flags));
```
sangho2, line 1085: `unregister_session` is not enough to close/teardown a TA instance. So, this test would be less meaningful.

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -935,6 +1016,38 @@ impl<const ALIGN: usize> ShmRefMap<ALIGN> {
...
+        .get(shm_ref)
+        .ok_or(OpteeSmcReturnCode::EBadAddr)?;
+    let (rpc_args_offset, rpc_args_size) =
+        rpc_args_range(msg_args_offset, msg_args_num_params, rpc_args.num_params)?;
+    let mut blob = alloc::vec![0u8; rpc_args_size];
+    rpc_args.serialize(&mut blob)?;
+    shm_info.write_at(platform, rpc_args_offset, &blob)
+}
+
+fn rpc_args_range(
```
sangho2, line 1037: this helper looks trivial. inlining might be better.

litebox_common_optee/src/lib.rs
```diff
@@ -2116,6 +2152,54 @@ impl OpteeRpcArgs {
...
+        if param.attr.attr_type() != OpteeMsgAttrType::TmemOutput as u8
+            || param.attr.0 & !(u64::from(u8::MAX) | OPTEE_MSG_ATTR_NONCONTIG) != 0
+        {
+            return Err(OpteeSmcReturnCode::EBadCmd);
+        }
+        OpteeMsgParamTmem::read_from_bytes(&param.data).map_err(|_| OpteeSmcReturnCode::EBadCmd)
+    }
+
+    /// Return whether an exactly validated TMEM output parameter uses a page list.
+    pub fn is_param_tmem_output_noncontiguous(
```
sangho2, line 2177: do we need this helper?

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+        self.ta_uuid
+    }
+
+    /// Return the registered shared-memory reference containing the RPC arguments.
+    pub fn registered_shm_ref(&self) -> u64 {
+        self.registered_shm_ref
+    }
+
+    /// Return the offset of the message arguments in registered shared memory.
+    pub fn registered_shm_offset(&self) -> usize {
```
sangho2, line 153: no caller.

litebox_shim_optee/src/msg_handler.rs
```diff
@@ -986,7 +1099,7 @@ fn get_shm_info_from_optee_msg_param_tmem(
 ///
 /// `rmem.offs` must be an offset within the shared memory region registered with `rmem.shm_ref` before
 /// and `rmem.offs + rmem.size` must not exceed the size of the registered shared memory region.
-fn get_shm_info_from_optee_msg_param_rmem(
+pub fn get_shm_info_from_optee_msg_param_rmem(
```
sangho2, line 1102: why `pub`?

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+                return Ok(context_id);
+            }
+        }
+
+        Err(RpcContextError::Full)
+    }
+
+    /// Insert a context with a caller-provided ID.
+    #[cfg(test)]
+    fn insert(
```
sangho2, line 226: looks like this is for testing only. do we need this function?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -542,48 +549,641 @@ fn optee_smc_handler(platform: &'static Platform, smc_args_addr: usize) -> Optee
...
+            // Always switch back to base page table before returning to VTL0
+            // Safety: No user-space memory references are held after this point
+            unsafe { switch_to_base_page_table(platform) };
+
+            if let Err(e) = result {
+                if e == OpteeSmcReturnCode::RpcCmd {
+                    debug_serial_println!("OP-TEE SMC returning RPC command to normal world");
+
+                    // CallWithArg is not supported for Dynamic TA RPC continuation.
+                    // This applies to how shm_ref is interpreted and used during RPC calls.
```
sangho2, line 628: Not only `CallWithArg` but also `CallWithRpcArg` are not supported. The second sentence is unclear. Better to explain a requirement because SHM should be already registered?

litebox_runner_lvbs/src/lib.rs
```diff
@@ -818,18 +1420,34 @@ fn open_session_new_instance(
...
+                "RPC args not present in incoming request, cannot request LOAD_TA from normal world"
+            );
+            msg_args.session = 0;
+            msg_args.ret = TeeResult::ItemNotFound;
+            msg_args.ret_origin = TeeOrigin::Tee;
+            write_non_ta_msg_args_to_normal_world(platform, msg_args, msg_args_phys_addr)?;
+            return Ok(());
+        };
+        // LOAD_TA is a two-call protocol. By passing `None`, we indicate that
+        // this is the first call, and the normal world should return TA binary size.
```
sangho2, line 1448: I think we don't need to explain what `prepare_load_ta_rpc` is here (function documentation should).

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+            registered_shm_ref,
+            regd_shm_offset,
+            requested_size: None,
+            shm_ref: None,
+            local_shm_registered: false,
+            completion: RpcCompletion::OpenSession,
+        }
+    }
+
+    /// Return the current RPC stage.
```
sangho2: looks like that method names are already self-contained, so comments are not that needed.

litebox_common_optee/src/lib.rs
```diff
@@ -2143,10 +2255,102 @@ impl OpteeRpcArgs {
...
+                shm_ref: 0,
+            },
+        )?;
+    } else {
+        // Second LOAD_TA: LiteBox passes VTL0-owned memory back to VTL0
+        // so normal world can populate it with the TA binary.
+        rpc_msg_args.set_param_attr_type(1, OpteeMsgAttrType::RmemOutput)?;
+        if let Some(rmem) = memref {
+            rpc_msg_args.set_param_rmem(1, rmem)?;
+        }
```
sangho2, line 2350: why do we need to check whether `memref` is `Some` again here? we already enter `else`.

litebox_shim_optee/src/rpc_context.rs
sangho2, line 137: Looks like all method names here are self-explanatory. Do we need comments?

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+        self.completion
+    }
+}
+
+/// Maps RPC context IDs to trusted continuation state.
+pub struct RpcContextMap {
+    next_id: AtomicU32,
+    inner: SpinMutex<HashMap<u32, RpcContext>>,
+    max_contexts: usize,
+}
```
sangho2, line 178: LiteBox provides an ID pool utility: `litebox::utils::id_pool::IdPool`. Please take a look at it.

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+        regd_shm_offset: usize,
+    ) -> Result<u32, RpcContextError> {
+        let mut contexts = self.inner.lock();
+        if contexts.len() >= self.max_contexts {
+            return Err(RpcContextError::Full);
+        }
+
+        // With N active entries, at least one of N + 1 consecutive IDs is free.
+        for _ in 0..=contexts.len() {
+            let context_id = self.next_id.fetch_add(1, Ordering::Relaxed);
```
sangho2, line 209: double synchronization: `self.inner.lock()` and `self.next_id.fetch_add()`. `next_id` doesn't need to be atomic in this case.

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+        }
+        contexts.insert(
+            context_id,
+            RpcContext::new(stage, ta_uuid, registered_shm_ref, regd_shm_offset),
+        );
+        Ok(())
+    }
+
+    /// Get the current stage for `context_id`.
+    pub fn get_curr_stage(&self, context_id: u32) -> Option<RpcStage> {
```
sangho2, line 249: most of this `get_*` methods are just indirections. Why not return `RpcContext` using `context_id` and call its methods? they are already public.

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+        self.inner.lock().remove(&context_id)
+    }
+
+    /// Return the number of active RPC contexts.
+    pub fn len(&self) -> usize {
+        self.inner.lock().len()
+    }
+
+    /// Return whether there are no active RPC contexts.
+    pub fn is_empty(&self) -> bool {
```
sangho2, line 399: test-only function.

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+/// Action to take after an in-flight shared-memory free RPC returns.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum RpcCompletion {
+    OpenSession,
+    ReturnError(OpteeSmcReturnCode),
+}
+
+/// Trusted state for an RPC-backed Dynamic TA request.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct RpcContext {
```
sangho2, line 105: I think `RpcContext` should be `enum` because its field have stage-specific meanings. The current representation looks more like C not Rust.

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,596 @@
...
+        Ok(())
+    }
+
+    /// Atomically advance the stage of a context from `expected` to `next` if it matches the current stage.
+    /// TODO: Check if the new stage is a valid transition from the current stage.
+    pub fn transition(
+        &self,
+        context_id: u32,
+        expected: RpcStage,
+        next: RpcStage,
```
sangho2, line 375: Why does this function need to get `next`? It should be able to figure out the next stage by itself because the current state machine simple enough (e.g., `LoadTaSize` should be followed by `ShmAlloc` if there is no error)?

sangho2 (commented):
I think this PR is functionally fine (except for known bugs which are already noted), but need some code cleaning and refactoring.

litebox_common_optee/src/lib.rs
praveen-pk: As most of these methods to modify rpc_args, makes sense to make them `OpteeRpcArgs` methods. I will update this in next version.

litebox_common_optee/src/lib.rs
praveen-pk: make sense to drop `map_err` and make this consistent with rest of the methods

litebox_common_optee/src/lib.rs
praveen-pk: I will find a way to make this consistent across these methods

litebox_runner_lvbs/src/lib.rs
praveen-pk: This was an attempt to consolidate `discard_rpc_context` calls. If not, this, I have to call it in every error case. 

As I am already doing this in the rpc helper methods, I will pull `discard_rpc_context` into the if block and call it when necessary.

litebox_runner_lvbs/src/lib.rs
praveen-pk: thanks for the catch, I will fix this

litebox_common_optee/src/lib.rs
praveen-pk: This was a miss during refactoring. I will dorp the check

litebox_shim_optee/src/rpc_context.rs
praveen-pk: dropped this. the code now has a stricter state machine

litebox_shim_optee/src/rpc_context.rs
praveen-pk: refactored

litebox_shim_optee/src/rpc_context.rs
praveen-pk: droped

litebox_common_optee/src/lib.rs
praveen-pk: moved

litebox_shim_optee/src/rpc_context.rs
praveen-pk: this is addressed in rpc_contex refactor.

litebox_shim_optee/src/rpc_context.rs
praveen-pk: Dropped this, now using `IdPool`

litebox_runner_lvbs/src/lib.rs
praveen-pk: fixed

litebox_runner_lvbs/src/lib.rs
praveen-pk: fixed

litebox_runner_lvbs/src/lib.rs
praveen-pk: moved these comments to appropriate location

litebox_shim_optee/src/msg_handler.rs
praveen-pk: dropped.

litebox_shim_optee/src/msg_handler.rs
praveen-pk: fixed

litebox_runner_lvbs/src/lib.rs
praveen-pk: reworded the comment

litebox_shim_optee/src/rpc_context.rs
praveen-pk: refactored to use IdPool.

litebox_shim_optee/src/lib.rs
praveen-pk: We should not preserve the dynamically loaded TA binary (in our cache) across requests. According to OP-TEE protocol, the TA Binary from RPC will be dropped once `ldelf` loads the binary into runtime memory. 

`KEEP_ALIVE` controls if the TA's runtime memory will be preserved across sessions. For a TA with `KEEP_ALIVE = OFF`, the TA binary should be loaded from Normal world for every request.


If the TA supports MULTI_INSTANCE, the current global ta_uuid cache can introduce synchronization issues. For example, one instance may cache the TA binary while another instance removes the same entry from the shared cache. I'm considering moving the ta_uuid cache into rpc_context, allowing each instance to maintain its own TA cache and use context_id as the reference key.

tgopinath-microsoft:
> fyi, I am working on a fix for race conditions while invoking Dynamic TA in multi-instance case. I will push that as a follow up PR.



> fyi, I am working on a fix for race conditions while invoking Dynamic TA in multi-instance case. I will push that as a follow up PR.

How hard are these fixes ? Can we get them in together with this PR ? I ask because we are breaking an existing functionality

litebox_common_optee/src/lib.rs
```diff
@@ -2144,9 +2253,160 @@ impl OpteeRpcArgs {
...
+            || self.num_params != 2
+        {
+            return Err(OpteeSmcReturnCode::EBadCmd);
+        }
+
+        let rmem = self.get_param_rmem_output(1)?;
+        if rmem.shm_ref != expected_shm_ref || rmem.offs != 0 || rmem.size != expected_size {
+            return Err(OpteeSmcReturnCode::EBadCmd);
+        }
+        read_binary(rmem.shm_ref, rmem.size)
```
sangho2: Do we need this callback? Currently, this function has only one caller and thus the caller itself can do this `read_binary` stuff. Better to avoid callbacks if possible.

litebox_common_optee/src/lib.rs
```diff
@@ -2116,6 +2152,51 @@ impl OpteeRpcArgs {
...
+        if param.attr.attr_type() != OpteeMsgAttrType::TmemOutput as u8
+            || param.attr.0 & !(u64::from(u8::MAX) | OPTEE_MSG_ATTR_NONCONTIG) != 0
+        {
+            return Err(OpteeSmcReturnCode::EBadCmd);
+        }
+        OpteeMsgParamTmem::read_from_bytes(&param.data).map_err(|_| OpteeSmcReturnCode::EBadCmd)
+    }
+
+    /// Return whether an exactly validated TMEM output parameter uses a page list.
+    fn is_param_tmem_output_noncontiguous(&self, index: usize) -> Result<bool, OpteeSmcReturnCode> {
```
sangho2: Do we need this helper? the caller can just check `self.params[0].attr.noncontig()`.

litebox_shim_optee/src/rpc_context.rs
```diff
@@ -0,0 +1,402 @@
...
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct RpcCommon {
+    ta_uuid: TeeUuid,
+    // RPC continuation reuses args[3] for the context ID. Use these fields to
+    // preserve the original registered SHM reference and offset
+    // before overwriting it.
+    registered_shm_ref: u64,
+    regd_shm_offset: usize,
+}
```
sangho2: We can inline this in to `RpcContext`. `RpcCommon` is unnecessary abstraction.
```rust
LoadTaSize {
  ta_uuid: TeeUuid,
  registered_shm_ref: u64,
  regd_shm_offset: usize,
},
ShmAlloc {
  ta_uuid: TeeUuid,
  registered_shm_ref: u64,
  regd_shm_offset: usize,
  requested_size: u64,
},
...
```

litebox_common_optee/src/lib.rs
```diff
@@ -2130,6 +2211,34 @@ impl OpteeRpcArgs {
         }
     }
 
+    /// Set a parameter's attribute type by index with bounds checking against `num_params`.
+    pub fn set_param_attr_type(
```
sangho2: this and below `set_param_rmem` helpers don't need to be `pub`.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -542,48 +549,523 @@ fn optee_smc_handler(platform: &'static Platform, smc_args_addr: usize) -> Optee
...
+            tmem.shm_ref,
+            RpcCompletion::ReturnError(OpteeSmcReturnCode::EBadCmd),
+        );
+        return;
+    };
+    if rpc_args
+        .prepare_load_ta_rpc(context.ta_uuid(), Some(rmem))
+        .is_err()
+    {
+        let _ = unregister_rpc_shm(tmem.shm_ref);
```
sangho2: this call is redundant because `start_shm_free_rpc` does it internally.

litebox_runner_lvbs/src/lib.rs
```diff
@@ -542,48 +549,523 @@ fn optee_smc_handler(platform: &'static Platform, smc_args_addr: usize) -> Optee
...
+            msg_args,
+            rpc_args,
+            context_id,
+            tmem.shm_ref,
+            RpcCompletion::ReturnError(OpteeSmcReturnCode::EBadCmd),
+        );
+        return;
+    }
+    if !write_next_rpc(platform, smc_args, msg_args, rpc_args, context_id) {
+        let _ = unregister_rpc_shm(tmem.shm_ref);
```
sangho2: same. `start_shm_free_rpc` does it.

PR #1219, opened 2026-08-21
Fix the OOM handling of page-sized memory allocation

sangho2:
This PR fixes the out-of-memory (OOM) handling of page-sized memory allocation. Currently, LiteBox's allocator panics when its page-sized memory allocation fails. This prevents us from using fallible APIs like `try_reserve` which expect the allocator reports OOM by returning `null`.

### DIFF
--- a/litebox/src/mm/allocator.rs
+++ b/litebox/src/mm/allocator.rs
@@ -162,14 +162,18 @@ unsafe impl<const ORDER: usize, M: MemoryProvider> GlobalAlloc
             Self::BASE_PAGE_SIZE => {
                 // Best to use the underlying backend directly to allocate pages
                 // to avoid fragmentation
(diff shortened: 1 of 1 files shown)


wdcui (approved):
LGTM. Thanks.

wdcui:
@sangho2 does it make sense to resolve #1222 at some point?

sangho2:
> Sangho Lee (@sangho2) does it make sense to resolve #1222 at some point?

At some point, `SafeZoneAllocator` might need to return `null` when `rescue` fails. LVBS doesn't go through this path because it doesn't have `rescue` (no dynamic heap expansion exists). But other general kernel (or user) platforms should consider it.

praveen-pk (approved):

PR #1224, opened 2026-08-21
[WIP] Add Initial "True Tales" for Modelling External Interactions

lschuermann:
This adds a work-in-progress "True Tales" development tool / crate to model and reason about interactions between LiteBox and external entities such as userspace processes and device drivers. It also adds a small end-to-end testing harness that can run a LiteBox platform & shim under Miri, to test core APIs such as the layered file system. It integrates the True Tales model into both the core LiteBox crate and the end-to-end harness, to exercise the model for the modeled read & write system calls.

As this is still work in progress, it would probably be best merged into a development branch. From that point onward, we can work towards integrating it into LiteBox proper.


jaybosamiya-ms:
Merged into `dev/truetales` as we discussed

PR #1226, opened 2026-08-25
Retire legacy `FileSystem` trait

jaybosamiya-ms:
This PR mechanically removes the legacy `FileSystem` trait, so that everything is now based on the new `Backend`-based interface (#887).  The relevant north interface now sits as inherent methods of `Resolver`.


jaybosamiya-ms:
More tweaks to the filesystem in future PRs, this one is mechanical but touches a lot of files so I've spun it out as its own PR to simplify reviewing

wdcui (approved):
LGTM. Thanks.

PR #1230, opened 2026-08-27
Keep epoll interest alive while a duplicate of the registered fd survives

willportnoy:
## Problem

Linux `epoll(7)` removes a descriptor from an interest list only after *every* descriptor referring to the underlying open file description (OFD) has been closed. The shim instead anchored each interest to the per-descriptor `TypedFd`, so closing the registered descriptor dropped the interest even when a `dup` referring to the same OFD was still open — `EpollEntry::poll` bailed on a dead `Weak<TypedFd>` and readiness was never delivered.

## Fix

Anchor epoll interest to the open file description:

- Add `WeakEntryHandle` to `litebox::fd` — a durable, `dup`-surviving weak reference to a descriptor's shared entry, plus `EntryHandle::downgrade`/`as_ptr`/`with_shared_metadata`.
- Re-point epoll's `DescriptorRef` at a per-subsystem `WeakEntryHandle`, key interests by the OFD's stable address, and re-poll through the shared entry (eventfd/unix/pipe via the entry's `IOPollable`, socket/file via aliased metadata). Observer registration is unchanged; it already targets the shared pollable.
- Expose `with_iopollable` on the pipe entry so a pipe can be polled without a live per-descriptor `PipeFd`.

Covers all six epoll-able fd types on `main` (eventfd, unix, pipe, socket, file, and the pre-existing epoll-on-epoll `unimplemented!()`), via exhaustive matches.

## Test

`tests/epoll_dup.c`: register an eventfd, `dup` it, close the original, and verify the interest still delivers events and survives re-arming. Passes natively, fails without this change under Litebox, and passes with it.

### DIFF
--- a/litebox/src/fd/mod.rs
+++ b/litebox/src/fd/mod.rs
@@ -146,11 +146,33 @@ impl<Platform: RawSyncPrimitivesProvider> Descriptors<Platform> {
             // Unique, so we can just return it if allowed.
             if can_close_immediately(old.x.entry.read().as_subsystem::<Subsystem>()) {
                 fd.x.mark_as_closed();
-                let entry = Arc::into_inner(old.x)
-                    .map(|shared| RwLock::into_inner(shared.entry))
-                    .map(DescriptorEntry::into_subsystem_entry::<Subsystem>)
-                    .unwrap();
-                Some(CloseResult::Closed(entry))
+                match Arc::try_unwrap(old.x) {
+                    Ok(shared) => {
+                        let entry = DescriptorEntry::into_subsystem_entry::<Subsystem>(
+                            RwLock::into_inner(shared.entry),
+                        );
+                        Some(CloseResult::Closed(entry))
+                    }
+                    Err(x) => {
+                        // The strong count was 1 above, but a lock-free
+                        // `WeakEntryHandle::upgrade` on another thread (e.g. an
+                        // epoll re-poll of a still-registered interest) can
+                        // transiently re-share the entry before we take
+                        // ownership. Fall back to the shared path: duplicate the
+                        // descriptor so it is closed once that reference drops,
+                        // rather than dropping the entry without an orderly
+                        // close.
+                        let replaced = self.entries[idx].replace(IndividualEntry {
+                            x,
+                            metadata: old.metadata,
+                        });
+                        assert!(replaced.is_none());
+                        Some(CloseResult::Duplicated(TypedFd {
+                            _phantom: PhantomData,
+                            x: OwnedFd::new(idx),
+                        }))
+                    }
+                }
             } else {
                 // Put it back
                 let old = self.entries[idx].replace(old);
--- a/litebox/src/fd/mod.rs
+++ b/litebox/src/fd/mod.rs
@@ -9,6 +9,7 @@
 )]
 
 use alloc::sync::Arc;
+use alloc::sync::Weak;
 use alloc::vec;
 use alloc::vec::Vec;
 use core::marker::PhantomData;
@@ -555,6 +556,77 @@ impl<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>
 
     pub fn with_entry_mut<R>(&self, f: impl FnOnce(&mut Subsystem::Entry) -> R) -> R {
         f(self.0.entry.write().as_subsystem_mut::<Subsystem>())
+    }
+
+    /// Runs `f` with the aliased (open-file-description-level) metadata of type
+    /// `T`, if present.
+    ///
+    /// This reads the metadata stored via [`Descriptors::set_entry_metadata`],
+    /// which is shared by every descriptor referring to the same open file
+    /// description. Returns `None` if no such metadata exists.
+    pub fn with_shared_metadata<T, R>(&self, f: impl FnOnce(&T) -> R) -> Option<R>
+    where
+        T: core::any::Any + Clone + Send + Sync,
+    {
+        self.0.entry.read().metadata.get::<T>().map(f)
+    }
+
+    /// The address of the shared open file description this handle refers to.
+    ///
+    /// Duplicates of a descriptor share one open file description, so this
+    /// address is stable across `dup` and uniquely identifies the description
+    /// for as long as any duplicate keeps it alive.
+    #[must_use]
+    pub fn as_ptr(&self) -> *const () {
+        Arc::as_ptr(&self.0).cast()
+    }
+
+    /// Downgrades to a [`WeakEntryHandle`] that survives `dup`.
+    ///
+    /// The resulting handle upgrades for as long as *any* descriptor referring
+    /// to the same open file description remains open, even after the specific
+    /// descriptor this handle was obtained from has been closed.
+    #[must_use]
+    pub fn downgrade(&self) -> WeakEntryHandle<Platform, Subsystem> {
+        WeakEntryHandle(Arc::downgrade(&self.0), PhantomData)
+    }
+}
+
+/// A durable, `dup`-surviving weak reference to a descriptor's open file
+/// description.
+///
+/// Unlike a [`TypedFd`], which is tied to one descriptor slot, this upgrades as
+/// long as *any* descriptor referring to the same open file description is
+/// open. It is the correct anchor for interest that must outlive the closure of
+/// the specific descriptor it was registered against (for example, epoll
+/// interest, per Linux `epoll(7)` semantics).
+pub struct WeakEntryHandle<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>(
+    Weak<SharedEntry<Platform>>,
+    PhantomData<fn(Subsystem) -> Subsystem>,
+);
+
+impl<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>
+    WeakEntryHandle<Platform, Subsystem>
+{
+    /// Upgrades to a strong [`EntryHandle`] if the open file description is
+    /// still alive (i.e. at least one duplicate remains open).
+    #[must_use]
+    pub fn upgrade(&self) -> Option<EntryHandle<Platform, Subsystem>> {
+        self.0
+            .upgrade()
+            .map(|entry| EntryHandle(entry, PhantomData))
+    }
+
+    /// The address of the shared open file description, stable across `dup`.
+    ///
+    /// This is safe to use as a durable identity key. A [`WeakEntryHandle`]
+    /// keeps the underlying allocation reserved even after the open file
+    /// description is closed (every strong reference dropped), so this address
+    /// is never recycled for a different open file description while this handle
+    /// exists. The pointer is only ever compared, never dereferenced.
+    #[must_use]
+    pub fn as_ptr(&self) -> *const () {
+        self.0.as_ptr().cast()
     }
 }
 
--- a/litebox/src/fd/mod.rs
+++ b/litebox/src/fd/mod.rs
@@ -541,11 +541,20 @@ impl<Platform: RawSyncPrimitivesProvider> Descriptors<Platform> {
     }
 }
 
+/// An opaque, stable identity for a descriptor's entry (its open file
+/// description).
+///
+/// Equal keys denote the same entry. A key is stable across `dup` and for the
+/// entry's lifetime, so it can identify an entry (for example as a map key)
+/// without dereferencing anything.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct EntryStableKey(usize);
+
 /// A handle to a descriptor entry (via [`Descriptors::entry_handle`]) that can be used without
 /// maintaining access to the descriptor table itself.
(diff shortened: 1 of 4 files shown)


litebox/src/fd/mod.rs
```diff
@@ -556,6 +557,77 @@ impl<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>
...
+    WeakEntryHandle<Platform, Subsystem>
+{
+    /// Upgrades to a strong [`EntryHandle`] if the open file description is
+    /// still alive (i.e. at least one duplicate remains open).
+    #[must_use]
+    pub fn upgrade(&self) -> Option<EntryHandle<Platform, Subsystem>> {
+        self.0
+            .upgrade()
+            .map(|entry| EntryHandle(entry, PhantomData))
+    }
```
CvvT, line 631: One subtle issue with this lies on https://github.com/microsoft/litebox/blob/e18e65a82e3fe9ef9155fedf336b4d53aca1329d/litebox/src/fd/mod.rs#L133-L153

Now `Arc::into_inner(old.x)...unwrap()` may panic. Closing a socket cannot be done simply via `Drop`. It needs to take the ownership of the entry. One potential fix suggested by copilot is to replace it with `Arc::try_unwrap` and return `CloseResult::Duplicated` on the error path.

willportnoy: _AI-authored, human-reviewed._

Good catch — fixed in fb437945. You're right the `Arc::strong_count == 1` check followed by `Arc::into_inner(...).unwrap()` is racy now: a `WeakEntryHandle::upgrade()` (which this PR newly lets epoll interests do) runs lock-free, so it can transiently bump the strong count between the check and the unwrap and panic it.

Switched to `Arc::try_unwrap`; on the error path it puts the entry back and returns `CloseResult::Duplicated`, so the descriptor is closed once the transient reference drops — the same self-healing queued-closure path already taken when the entry is genuinely shared.

I also checked the sibling `remove()`: it uses `Arc::into_inner(...).map(...)` (no unwrap), and its socket callers discard the entry via `let _ =`, so there's no analogous panic or orderly-close regression there.

litebox/src/fd/mod.rs
```diff
@@ -556,6 +579,77 @@ impl<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>
...
+
+    /// The address of the shared open file description this handle refers to.
+    ///
+    /// Duplicates of a descriptor share one open file description, so this
+    /// address is stable across `dup` and uniquely identifies the description
+    /// for as long as any duplicate keeps it alive.
+    #[must_use]
+    pub fn as_ptr(&self) -> *const () {
+        Arc::as_ptr(&self.0).cast()
+    }
```
jaybosamiya-ms: I am not sure I agree with exposing a `*const ()` here, over-broadens our interface and guarantees (if we introduce this, now we are promising that we will always give out the address specifically).  The correct move is probably to expose an opaque `PartialEq+Eq+Hash+Clone+Copy` `EntryStableKey`.  That would allow more stable utilization.

litebox/src/fd/mod.rs
```diff
@@ -556,6 +579,77 @@ impl<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>
...
+    ///
+    /// This reads the metadata stored via [`Descriptors::set_entry_metadata`],
+    /// which is shared by every descriptor referring to the same open file
+    /// description. Returns `None` if no such metadata exists.
+    pub fn with_shared_metadata<T, R>(&self, f: impl FnOnce(&T) -> R) -> Option<R>
+    where
+        T: core::any::Any + Clone + Send + Sync,
+    {
+        self.0.entry.read().metadata.get::<T>().map(f)
+    }
```
jaybosamiya-ms: Nit: doc comment + function naming could be better for consistency:

```suggestion
    /// Apply `f` on metadata at the entry, if it exists.
    ///
    /// In contrast to [`Descriptors::with_metadata`], this obtains entry-level metadata.  
    /// For FD-specific metadata, one necessarily needs the specific FD.
    pub fn with_entry_metadata<T, R>(&self, f: impl FnOnce(&T) -> R) -> Option<R>
    where
        T: core::any::Any + Clone + Send + Sync,
    {
        self.0.entry.read().metadata.get::<T>().map(f)
    }
```

litebox/src/fd/mod.rs
```diff
@@ -556,6 +579,77 @@ impl<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>
...
+
+    /// Downgrades to a [`WeakEntryHandle`] that survives `dup`.
+    ///
+    /// The resulting handle upgrades for as long as *any* descriptor referring
+    /// to the same open file description remains open, even after the specific
+    /// descriptor this handle was obtained from has been closed.
+    #[must_use]
+    pub fn downgrade(&self) -> WeakEntryHandle<Platform, Subsystem> {
+        WeakEntryHandle(Arc::downgrade(&self.0), PhantomData)
+    }
```
jaybosamiya-ms: Nit: doc comment, it is not about surviving `dup`, it is about getting a non-owning `WeakEntryHandle`, that's all.  Also, we do not need to document upgrade here.

litebox/src/fd/mod.rs
```diff
@@ -556,6 +579,77 @@ impl<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>
...
+}
+
+/// A durable, `dup`-surviving weak reference to a descriptor's open file
+/// description.
+///
+/// Unlike a [`TypedFd`], which is tied to one descriptor slot, this upgrades as
+/// long as *any* descriptor referring to the same open file description is
+/// open. It is the correct anchor for interest that must outlive the closure of
+/// the specific descriptor it was registered against (for example, epoll
+/// interest, per Linux `epoll(7)` semantics).
```
jaybosamiya-ms: Similarly, this is not `dup`-surviving, it is simply a weak reference to a descriptor entry.  This is too much superfluous documentation.  The upgrade stuff also is literally the obvious thing for any weak pointer, there is no reason to state that it works as long as something strong still exists.  The whole doc comment here could just be:

```suggestion
/// A weak-reference to a descriptor entry.
```

litebox/src/fd/mod.rs
```diff
@@ -556,6 +579,77 @@ impl<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>
...
+pub struct WeakEntryHandle<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>(
+    Weak<SharedEntry<Platform>>,
+    PhantomData<fn(Subsystem) -> Subsystem>,
```
jaybosamiya-ms, line 619: Why is this inconsistent with `EntryHandle`'s PhantomData?

litebox/src/fd/mod.rs
```diff
@@ -556,6 +579,77 @@ impl<Platform: RawSyncPrimitivesProvider, Subsystem: FdEnabledSubsystem>
...
+    ///
+    /// This is safe to use as a durable identity key. A [`WeakEntryHandle`]
+    /// keeps the underlying allocation reserved even after the open file
+    /// description is closed (every strong reference dropped), so this address
+    /// is never recycled for a different open file description while this handle
+    /// exists. The pointer is only ever compared, never dereferenced.
+    #[must_use]
+    pub fn as_ptr(&self) -> *const () {
+        self.0.as_ptr().cast()
+    }
```
jaybosamiya-ms: Similarly, this should switch to `EntryStableKey`

litebox/src/pipes.rs
```diff
@@ -729,3 +729,18 @@ crate::fd::enable_fds_for_subsystem! {
...
     -> PipeFd<Platform>;
 }
+
+impl<Platform: RawSyncPrimitivesProvider + TimeProvider> DescriptorEntry<Platform> {
+    /// Runs `f` with the [`IOPollable`] backing this pipe end.
+    ///
+    /// This lets a holder of a durable entry handle poll the pipe without a
+    /// live per-descriptor [`PipeFd`], which is required so that an epoll
+    /// interest survives closing the registered descriptor while a duplicate
+    /// referring to the same open file description remains open.
```
jaybosamiya-ms: This explanation is unnecessary to keep here

litebox/src/fd/mod.rs
jaybosamiya-ms: Also another issue with it being a `*const ()` is that, it would prevent us from making the _stronger_ guarantee in the future that re-allocations even at the same address would be caught.  If we make it an opaque `EntryStableKey` we can strengthen things without breaking any API contract, and automatically things get better.

jaybosamiya-ms (changes requested):
Thanks Will, I reviewed the changes in the `litebox` core, since this is changing our API surface.  I think some changes are needed to make this more compatible with future evolution.  The base of the changes is mostly reasonable, just needs some tweaks.

The changes should be fairly straightforward, but do let me know if you'd prefer I open up a PR with the changes I have in mind to the core, instead of tweaking those things over this PR.

Oh also, please do make sure to keep the PR description human-written, not AI-generated-human-reviewed.

litebox/src/fd/mod.rs
willportnoy: _AI-authored, human-reviewed._

Done — replaced `as_ptr() -> *const ()` on both `EntryHandle` and `WeakEntryHandle` with `stable_key() -> EntryStableKey`, an opaque `#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]` newtype, and keyed `EpollEntryKey` on it. Agreed on the future-strengthening point — the opaque type lets us later add e.g. a generation counter to catch same-address reallocations without touching the API. (It also derives `Ord`/`PartialOrd` since `EpollEntryKey` is a `BTreeMap` key.)

litebox/src/fd/mod.rs
willportnoy: _AI-authored, human-reviewed._

Done — renamed to `with_entry_metadata` with your doc wording.

litebox/src/fd/mod.rs
willportnoy: _AI-authored, human-reviewed._

Done — trimmed to "Obtains a non-owning `WeakEntryHandle` to this entry." and dropped the upgrade note.

litebox/src/fd/mod.rs
willportnoy: _AI-authored, human-reviewed._

Done — the doc is now just "A weak reference to a descriptor entry."

litebox/src/fd/mod.rs
willportnoy: _AI-authored, human-reviewed._

Good question — unified them: `EntryHandle` now also uses `PhantomData<fn(Subsystem) -> Subsystem>`. That invariant form is what makes the handle unconditionally `Send`/`Sync` regardless of `Subsystem` (the subsystem markers embed `Platform`), which `WeakEntryHandle` needs because it is stored in a cross-thread `Observer`. Unifying on it removes the inconsistency.

litebox/src/fd/mod.rs
willportnoy: _AI-authored, human-reviewed._

Done — `WeakEntryHandle` now exposes `stable_key() -> EntryStableKey` too.

litebox/src/pipes.rs
willportnoy: _AI-authored, human-reviewed._

Done — removed the explanation; the doc is now a single line.

wdcui:
@jaybosamiya-ms can you please take a look at Will's fixes when you get a chance? Thanks!

jaybosamiya-ms (changes requested):
Thanks Will for making the updates.  There are still a few changes needed:

1. The introduction of a weak entry handle means that we need to revisit the assumptions of all existing functions that were previously assuming or only working with strong counts.  As far as I can tell, `remove` (and/or its callers) break upon a racing `upgrade`. Specifically, if an `upgrade` occurs at _just_ the right time, then it is possible for _no_ `remove` to ever return the entry (i.e., it silently gets dropped), which can break other code.  However, if we update `remove` to account for this race, not close the descriptor, and keep it in, then a bunch of other code that assumes a `remove` will actually give it the entry will break.  Some sort of re-designing is needed, and it is unclear what the nicer redesign is.  It would take more effort to figure out what the correct design for this is.  I'm not necessarily suggesting a change here, but am saying that at least we need some documentation around `remove` that says `// XXX: Due to #1230, a race of a weak handle upgrade might cause a fully missed entry` or something similar to that, so that we are at least documenting the potential issue.
2. Similar, revisiting assumptions of `drain_entries_full_covered_by` which also can fail due to a race with an `upgrade`; basically the strong-count snapshot it takes might be invalidated by a racing `upgrade`.  The right move is to update the code to retain the entry when it loses the race, rather than relying entirely on the snapshot.  This one is less problematic for callers, since they already have the built-in retry-like logic.  So this one we should definitely fix.
3. The PR description is still AI generated.  I'd mentioned this in my prior review too, but please do make sure to keep the PR description human-written, not AI-generated-human-reviewed.

PR #1231, opened 2026-08-27
Switch to explicit file system `Context`s

jaybosamiya-ms:
This PR switches the file system resolver to explicit `Context`s, so that the underlying file system(s) and the context that they are used in are separated.  Essentially, this means that nothing within the file system is itself aware of CWD (current working dir) or acting user now, and the `Context` object explicitly carries this.  This means that the Linux shim no longer needs to maintain its own `cwd: String` field and manipulation of it, allowing resolution + permission decisions to live in one place.

Along with this, I also updated the in-mem backend to use the resolver context rather than maintain its own user management, closing out yet another place of unnecessary duplication and potential inconsistency.

Finally, as a drive-by fix: `getcwd` no longer returns a trailing `/`, making it more consistent with Linux.

### DIFF
--- a/litebox/src/fs/backend.rs
+++ b/litebox/src/fs/backend.rs
@@ -134,11 +134,16 @@ pub trait Backend: private::Sealed + Send + Sync + Any {
         &self,
         dir: DirHandle,
         name: &str,
-        node: NewNode,
+        metadata: CreationMetadata,
     ) -> Result<FileHandle, OpenError>;
 
     /// Create a new directory at `parent` with the given `name` and metadata.
-    fn mkdir_at(&self, dir: DirHandle, name: &str, node: NewNode) -> Result<DirHandle, MkdirError>;
+    fn mkdir_at(
+        &self,
+        dir: DirHandle,
+        name: &str,
+        metadata: CreationMetadata,
+    ) -> Result<DirHandle, MkdirError>;
 
     /// Remove the file `name` at `parent`.
     fn unlink_at(&self, dir: DirHandle, name: &str) -> Result<(), UnlinkError>;
@@ -337,7 +342,7 @@ pub(super) enum WalkStopReason {
 /// The metadata a backend stamps onto a newly created file or directory.
 #[derive(Clone, Copy, Debug)]
 #[non_exhaustive]
-pub struct NewNode {
+pub struct CreationMetadata {
     /// Permission bits for the new node.
     pub mode: Mode,
     /// Owner of the new node.
--- a/litebox/src/fs/backend.rs
+++ b/litebox/src/fs/backend.rs
@@ -129,16 +129,16 @@ pub trait Backend: private::Sealed + Send + Sync + Any {
     /// Status of an open file or directory handle.
     fn status(&self, h: HandleRef<'_>) -> Result<FileStatus, FileStatusError>;
 
-    /// Create a new file at `parent` with the given `name` and `mode`.
+    /// Create a new file at `parent` with the given `name` and metadata.
     fn create_file_at(
         &self,
         dir: DirHandle,
         name: &str,
-        mode: Mode,
+        node: NewNode,
     ) -> Result<FileHandle, OpenError>;
 
-    /// Create a new directory at `parent` with the given `name` and `mode`.
-    fn mkdir_at(&self, dir: DirHandle, name: &str, mode: Mode) -> Result<DirHandle, MkdirError>;
+    /// Create a new directory at `parent` with the given `name` and metadata.
+    fn mkdir_at(&self, dir: DirHandle, name: &str, node: NewNode) -> Result<DirHandle, MkdirError>;
 
     /// Remove the file `name` at `parent`.
     fn unlink_at(&self, dir: DirHandle, name: &str) -> Result<(), UnlinkError>;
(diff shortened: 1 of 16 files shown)


jaybosamiya-ms:
Again, easiest to review one commit at a time :)

wdcui:
GPT/Opus reported the following issues:

1. Root and mount-root  chmod / chown  bypass ownership checks

Files:  litebox/src/fs/resolver.rs:277-287,824-865 ,  litebox/src/fs/in_mem.rs:574-607 
Severity: High

This PR moves ownership authorization for  chmod  and  chown  from  InMem  into the resolver. However, filesystem roots and Composer mount roots are represented as  PermissionCheck::ByBackend , and  may_change_metadata  treats that value as automatically authorized. At the same time, the ownership checks were removed from  InMem , leaving neither layer responsible for authorization.

For example, a guest running as UID 1000 can call  chmod("/")  or  chown("/")  on a root directory owned by UID 0, and the operation reaches  InMem  without any ownership check.

There is a related TODO in  resolver.rs  noting that write permission on root directories is currently unchecked, but that TODO concerns permission to add or remove directory entries. It does not recognize this  chmod / chown  bypass. In fact,  path_handle  says the backend is expected to enforce metadata permissions, which is no longer true for  InMem .

Suggested fix: Return resolver-checkable permission metadata for filesystem and mount roots, or preserve ownership enforcement in mutable backends for  ByBackend  handles.

2. Resolver-side metadata authorization introduces a TOCTOU race

Files:  litebox/src/fs/resolver.rs:824-865 ,  litebox/src/fs/in_mem.rs:574-607 
Severity: High

The resolver now authorizes  chmod  and  chown  using a copied  PermissionInfo , then separately calls the backend to mutate the node. The backend acquires its metadata write lock only during the mutation, so another thread can change ownership between the authorization check and the update.

For example:

1. UID 1000 starts  chmod  on a file currently owned by UID 1000 and passes the resolver check.
2. Another thread changes the owner to UID 2000.
3. The first thread resumes and changes the mode of a file it no longer owns.

Before this PR,  InMem  checked ownership while holding the same write lock used for the mutation, so this race did not exist there.

Suggested fix: Pass the acting user into the backend metadata operation and perform authorization while holding the same lock used to update the metadata. Re-reading status in the resolver would still leave a race window.

3. Overlay copy-up can silently change ownership with a 9P upper layer

Files:  litebox/src/fs/overlay.rs:271-283,474-495 ,  litebox/src/fs/nine_p/mod.rs:500-539 
Severity: Medium
Recognition: Partially known

Overlay previously created the upper node and then called  chown  to preserve the lower node’s ownership. This PR removes that  chown , relying on  NewNode.owner  being applied atomically by every upper backend.

The 9P backend explicitly documents that  Tlcreate  and  Tmkdir  cannot set the requested owning UID: the new node is owned by the user attached to the 9P connection. Therefore, when Overlay copies up a lower file or directory owned by another user, the operation can succeed while silently changing its ownership.

The 9P protocol limitation is clearly documented in the code, so that part is known. What does not appear to be recognized is that removing Overlay’s post-create  chown  turns that limitation into incorrect copy-up behavior.

Suggested fix: Preserve the post-create  chown  and rollback for backends that cannot honor  NewNode.owner , or make creation fail when the requested ownership cannot be applied rather than returning a differently owned node.

4.  chdir("")  now succeeds instead of returning  ENOENT 

 litebox_shim_linux/src/syscalls/file.rs:1745-1753 
Severity: Medium

The old implementation routed  chdir  through  resolve_path , which explicitly rejected an empty pathname with  ENOENT . The new implementation calls  Context::resolve  directly.

 Context::resolve("")  ignores the empty path component and returns the current working directory. Because that path exists and is a directory,  sys_chdir  returns success without changing anything. Linux requires  chdir("")  to fail with  ENOENT .

Suggested fix: Explicitly reject an empty pathname with  ENOENT  before calling  Context::resolve .

litebox/src/fs/backend.rs
```diff
@@ -129,16 +129,16 @@ pub trait Backend: private::Sealed + Send + Sync + Any {
...
 
-    /// Create a new file at `parent` with the given `name` and `mode`.
+    /// Create a new file at `parent` with the given `name` and metadata.
     fn create_file_at(
         &self,
         dir: DirHandle,
         name: &str,
-        mode: Mode,
+        node: NewNode,
     ) -> Result<FileHandle, OpenError>;
```
wdcui, line 138: The comment says metadata, but the new argument is `node`?

litebox/src/fs/backend.rs
```diff
@@ -334,6 +334,16 @@ pub(super) enum WalkStopReason {
     Continue,
 }
 
+/// The metadata a backend stamps onto a newly created file or directory.
+#[derive(Clone, Copy, Debug)]
+#[non_exhaustive]
+pub struct NewNode {
```
wdcui: Why is it called `NewNode`?

wdcui (approved):
LGTM. Thanks.

litebox/src/fs/backend.rs
jaybosamiya-ms: Naming is hard :)  Renamed to `CreationMetadata` which is more descriptive.

litebox/src/fs/backend.rs
jaybosamiya-ms: fixed with new name

jaybosamiya-ms:
For the agent findings:
1. `/` is a known + documented "permissions are weird here, need to fix up", nothing to be done in this PR
2. this is again the "atomicity of operations" thing I've mentioned quite a few times before; agent's suggestion is straight up bad and breaks all the layering
3. not a particularly important concern for any use cases I'm aware of
4. valid even though unimportant compatibility thing, will fix before merging

PR #1237, opened 2026-08-28
Add a standalone hostname-filtering egress proxy

wdcui:
This PR adds a standalone `litebox_egress_proxy` executable that authorizes exact hostname-and-port rules, resolves each authorized hostname only through an explicitly configured DNS server, and forwards plain HTTP or CONNECT traffic to the returned numeric IPv4 addresses, with loopback or inherited-listener support, bounded HTTP and stream handling, hermetic tests, and cross-platform CI coverage.


wdcui:
Superseded by the reviewable stack in #1238, #1239, and #1240.

PR #1246, opened 2026-08-31
LVBS: Use reference counters for page tables

sangho2:
This PR uses `Arc` reference counters for LVBS/OP-TEE page tables to safely share them across cores. In particular, it makes the page-table lifetime/ownership explicit rather than relies on session-level serialization.

### DIFF
--- a/litebox_platform_lvbs/src/lib.rs
+++ b/litebox_platform_lvbs/src/lib.rs
@@ -247,12 +247,6 @@ impl PageTableManager {
             return PageTableHandle::task(pt);
         }
 
-        // The anchor can lag CR3 briefly while switching page tables.
-        let task_pts = self.task_page_tables.read();
-        if let Some(pt) = task_pts.get(&cr3_id) {
-            return PageTableHandle::task(Arc::clone(pt));
-        }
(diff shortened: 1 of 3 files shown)


litebox_platform_lvbs/src/lib.rs
```diff
@@ -268,7 +288,14 @@ impl PageTableManager {
     ///   after the switch (including the code being executed and stack)
     /// - No references to user-space memory are held across the switch
     pub unsafe fn load_base(&self) {
-        self.base_page_table.load();
+        x86_64::instructions::interrupts::without_interrupts(|| {
+            // Keep the previous page table alive until after switching CR3.
```
wdcui: Is this comment about the code below it or about a pre-condition before that code?

wdcui (approved):

litebox_platform_lvbs/src/lib.rs
sangho2: This comment is about ordering. We should update CR3 first and then decrease/drop Arc (`set_active_page_table` internally does it). Let me clarify it.

PR #1249, opened 2026-09-01
Improve CI usability for semver-checks across forks

jaybosamiya-ms:
When running on forks, the auth tokens do not allow the CI to put the normal automated semver comment, which makes it hard to keep track of when something actually might tweak semver behavior for PRs from a fork.  This PR changes it so that when something is semver-changing, it requires a comment (which it gives precisely the comment to copy-paste into place) so actually succeed on a fork.

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -8,6 +8,11 @@ on:
       - synchronize
       - reopened
       - edited
+  issue_comment:
+    types:
+      - created
+      - edited
+      - deleted
   merge_group:
 
 # If a new commit is pushed to the branch before ongoing runs finish, cancel the ongoing runs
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
Testing this is gonna be a bit non-trivial (specifically, we _may_ need some tweaks after being merged since I can't _really_ test until this is on `main` _and_ some PR is opened from a fork).  However, this is a low-risk one to merge, because it should not tweak usability for our most common scenarios, it mostly is an additional set of checks when a PR comes from a fork _and_ has a breaking change to our APIs.

jaybosamiya-ms:
Well, technically as a test, I can put this: :robot: SemverChecks :robot:

Which should trigger the particular sub-action to start and then immediately mark itself as skipped.  Will edit this message in-place as to the results of that test.

EDIT: hmm, even that test requires this to be merged into `main` first.  Oh well, not a big issue that this can't be tested fully until it hits `main`; at least statically, as far as I can tell (and as far as my agent reviews can tell) this should be fine.

wdcui (approved):

wdcui:
@jaybosamiya-ms I approved this PR since it's low risk. You can decide when to merge it.

PR #1252, opened 2026-09-01
Fix socketpair fd-table race when the table is full

### LINKED ISSUE #1172: do_socketpair possible bug when fd table is full
Similarly to #1171, the `raw_descriptor_store` lock is grabbed three times in succession for two inserts. If the second operation fails (because the table is full, for example) the first fd is removed. If a concurrent thread interleaves between these two lock acquisitions and closes the same slot, the second insert's failure mode will either try to close a non-existent fd or an unrelated fd.

https://github.com/microsoft/litebox/blob/9732adf0add28c663bd51ef9881a0b8992e20be8/litebox_shim_linux/src/syscalls/net.rs#L1075-L1083

CvvT:
Fix #1172: hold the descriptor-table write lock across both socketpair fd insertions and rollback.


wdcui (approved):
LGTM. Thanks for the bug fix.

PR #1253, opened 2026-09-01
Fix Windows reserve-only page allocations

CvvT:
Some applications may reserve large chunk of memory (e.g., 4GB) upfront but only uses a small portion of it. When allocating memory with no permission, we should only reserve memory without committing any pages on Windows platform. We then commit reserved pages when accessible permissions are later requested.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1666,6 +1666,7 @@ dependencies = [
  "getrandom 0.3.4",
  "litebox",
  "litebox_common_linux",
+ "rangemap",
  "windows-sys 0.60.2",
  "zerocopy",
 ]
@@ -2441,9 +2442,9 @@ dependencies = [
 
 [[package]]
 name = "rangemap"
-version = "1.6.0"
+version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f93e7e49bb0bf967717f7bd674458b3d6b0c5f48ec7e3038166026a69fc22223"
+checksum = "f60fcc7d6849342eff22c4350c8b9a989ee8ceabc4b481253e8946b9fe83d684"
 
 [[package]]
 name = "raw-cpuid"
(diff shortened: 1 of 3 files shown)


wdcui:
gpt/opus found the following two issues:

1.  MAP_POPULATE | PROT_NONE  can still call  do_prefetch_on_range  after  reserve_and_maybe_commit  leaves the range uncommitted. This occurs in the  MEM_FREE  path around lines 1772–1787 and the anonymous fallback around lines 1797–1808.  PrefetchVirtualMemory  requires a range accessible to the target process, and failure currently triggers an assertion. Please skip prefetching when  initial_permissions.is_empty() .

2. The collision check around lines 1710–1726 only treats  MEM_COMMIT  regions as occupied. Now that a live  PROT_NONE  mapping can remain  MEM_RESERVE , a subsequent allocation may treat its address as available.  Vmem::insert_mapping  protects the normal  NoReplace  path through its VMA map, but  Hint  and direct provider behavior can still incorrectly reuse the range. The provider needs a way to distinguish live reserve-only mappings from reusable reserved padding/decommitted regions and include live mappings in collision handling.

wdcui:
@CvvT, do you plan to address the comment above?

CvvT:
> Weiteng Chen (@CvvT), do you plan to address the comment above?

Yes, I am still working on it.

CvvT:
> * MAP_POPULATE | PROT_NONE  can still call  do_prefetch_on_range  after  reserve_and_maybe_commit  leaves the range uncommitted. This occurs in the  MEM_FREE  path around lines 1772–1787 and the anonymous fallback around lines 1797–1808.  PrefetchVirtualMemory  requires a range accessible to the target process, and failure currently triggers an assertion. Please skip prefetching when  initial_permissions.is_empty() .
> *

Now we skip `do_prefetch_on_range` if `PROT_NONE` is requested.

CvvT:
> 2\. The collision check around lines 1710–1726 only treats  MEM_COMMIT  regions as occupied. Now that a live  PROT_NONE  mapping can remain  MEM_RESERVE , a subsequent allocation may treat its address as available.  Vmem::insert_mapping  protects the normal  NoReplace  path through its VMA map, but  Hint  and direct provider behavior can still incorrectly reuse the range. The provider needs a way to distinguish live reserve-only mappings from reusable reserved padding/decommitted regions and include live mappings in collision handling.

Add tracking for live reserve-only mappings for collision detection and update tracking after permission changes and deallocation.

PR #1256, opened 2026-09-02
Fix Windows exception callback argument passing

CvvT:
The old code assumed that `rsp` is restored to `host_sp` and then read thread context via `*rsp` to `rcx`, but `rsp` is actually assigned to `host_sp - EXCEPTION_RECORD`. This PR fixes it by directly assigning `*host_sp` to `rcx`.


wdcui (approved):
LGTM. Thanks.

Issue #1138, opened 2026-08-06
LVBS: Direct memory copy between different foreign memory domains (i.e., no in-kernel bounce buffer)


sangho2:
In LVBS/OP-TEE, the normal-world client/kernel and secure-world TAs exchange data. Currently, we use a bounce buffer in the secure-world kernel to support this. However, since the exchanged data is often transparent to the secure-world kernel, this bounce buffer usage could result in unnecessary memory consumption and time delay. Since we do not trust both foreign memory domains anyways, it is okay to directly copy "data" between them. Of course, any control messages should still be checked by the secure kernel (similar to control/data plane separation). Also, we still have an in-TA private buffer to copy in/out data (because we **never** map normal-world memory to TA address spaces).

PR #1257, opened 2026-09-02
Add support for direct memory copy between VTL0 and VTL1-userspace

sangho2:
This PR enables direct memory copy between VTL0 (normal world) and VTL1 (secure world) userspace. Currently, we should use in-VTL1-kernel (in-LiteBox) bounce buffers to copy data between these two foreign memory domains regardless of whether the kernel/LiteBox checks the content. This results in pure time delay and memory pressure. Note that this direct memory copy does not affect LiteBox's memory safety because both source and destination are not Rust (kernel) memory.

### DIFF
--- a/litebox/src/platform/common_providers/userspace_pointers.rs
+++ b/litebox/src/platform/common_providers/userspace_pointers.rs
@@ -110,6 +110,24 @@ impl<V: ValidateAccess, T> UserConstPtr<V, T> {
         }
     }
 
+    /// Copy data from userspace to a raw pointer.
+    ///
+    /// # Safety
+    ///
+    /// `dst` must be either non-Rust memory or Rust memory with exclusive access.
+    pub unsafe fn copy_to_raw(self, dst: *mut T, len: usize) -> Option<()> {
+        if len == 0 {
+            return Some(());
+        }
+        let byte_len = len.checked_mul(core::mem::size_of::<T>())?;
+        self.inner.checked_add(byte_len)?;
+        let src =
+            V::validate_slice(core::ptr::slice_from_raw_parts(self.as_ptr(), len).cast_mut())?
+                .cast_const();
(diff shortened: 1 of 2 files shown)


sangho2:
Issue #1138

wdcui (approved):
I got the following comment from the agent:

The new  copy_to_raw / copy_from_raw  APIs appear to have incomplete safety contracts. They are generic over  T , but  copy_to_raw  can write arbitrary userspace bytes into Rust memory without requiring  T: FromBytes , while  copy_from_raw  can copy uninitialized padding from Rust memory without requiring  T: IntoBytes . The contracts also do not state that the raw pointer must be valid for  len * size_of::<T>()  bytes. The existing callers use  T = u8  with foreign physical memory and are safe, but future callers could satisfy the documented contracts and still trigger undefined behavior. Could these methods either operate directly on  *mut u8 / *const u8  with byte lengths, or add the appropriate zerocopy bounds and explicit buffer-size requirements?

sangho2:
> I got the following comment from the agent:
> 
> The new  copy_to_raw / copy_from_raw  APIs appear to have incomplete safety contracts. They are generic over  T , but  copy_to_raw  can write arbitrary userspace bytes into Rust memory without requiring  T: FromBytes , while  copy_from_raw  can copy uninitialized padding from Rust memory without requiring  T: IntoBytes . The contracts also do not state that the raw pointer must be valid for  len * size_of::()  bytes. The existing callers use  T = u8  with foreign physical memory and are safe, but future callers could satisfy the documented contracts and still trigger undefined behavior. Could these methods either operate directly on  *mut u8 / *const u8  with byte lengths, or add the appropriate zerocopy bounds and explicit buffer-size requirements?

Thanks!
Having zerocopy here might be better and strict, but for now `copy_to_raw` and `copy_from_raw` never create Rust `T` values internally, and are not intended to write bytes into Rust memory. Anyway, let me use zerocopy. It shouldn't affect the actual usage (physical pointers always work with *u8).

PR #1276, opened 2026-09-03
Direct data copy between normal-world shared memory and TA memory

### LINKED ISSUE #1138: LVBS: Direct memory copy between different foreign memory domains (i.e., no in-kernel bounce buffer)

sangho2:
This PR lets LiteBox/LVBS secure kernel directly copy data between normal-world shared memory and TA memory without relying on kernel-owned bounce buffers. It does not harm our security guarantee because source and destination addresses as well as their data sizes are tightly checked. It still ensures that TAs never directly access normal-world memory.

### DIFF
--- a/litebox_common_optee/src/lib.rs
+++ b/litebox_common_optee/src/lib.rs
@@ -582,12 +582,25 @@ impl UteeParams {
 #[derive(Clone)]
 pub enum UteeParamOwned {
     None,
-    ValueInput { value_a: u64, value_b: u64 },
+    ValueInput {
+        value_a: u64,
+        value_b: u64,
+    },
     ValueOutput,
-    ValueInout { value_a: u64, value_b: u64 },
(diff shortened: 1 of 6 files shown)


sangho2:
A sequel of #1257. Issue #1138.

wdcui (approved):
LGTM. Thanks.

PR #1279, opened 2026-09-03
Stabilize broker TCP client test

jaybosamiya-ms:
Handle TCP disconnect errors in the broker client test, preventing a flaky server-thread panic


jaybosamiya-ms:
Example case where I hit such flakiness in the test, which this PR fixes: https://github.com/microsoft/litebox/actions/runs/33698218503/job/100471623438

PR #1292, opened 2026-09-04
Fix race in TA classification to ensure single-instance loading

sangho2:
This PR fixes a race in TA classification by rechecking the TA flags (i.e., single- or multi-instance). It eliminates a small window where a single-instance TA could be loaded more than once.


wdcui (approved):
LGTM. Thanks.

PR #1296, opened 2026-09-04
Fix stack alignment in LVBS boot trampolines

sangho2:
This PR fixes a few remaining stack alignment issues in LVBS boot trampolines.


wdcui (approved):
LGTM. Thanks.

Issue #1268, opened 2026-09-02
A stale ELF patch plan can rewrite a newer fixed-address mapping


CvvT:
A potential issue found by [Specula](https://github.com/specula-org/Specula):

See the full [report](https://github.com/specula-org/specula-case-studies/blob/data/litebox-case-study-20260831/systems/litebox/modules/core/runs/litebox-known-aware-rerun-20260830/confirmed-bugs.md#entry-6-a-stale-elf-patch-plan-can-rewrite-a-newer-fixed-address-mapping) for more details.

## Trigger scenario
1. Map a read-only file-backed ELF page, creating ElfPatchState.file_mappings.
2. Replace the same address with a different read-only file using mmap(MAP_FIXED).
3. Call mprotect(addr, len, PROT_READ|PROT_EXEC).
4. maybe_patch_on_mprotect_exec collects the stale old mapping and maybe_patch_exec_segment rewrites the bytes currently at that address, which now belong to the replacement mapping.

Though benign programs should not do this, we should probably still fix it.

PR #1302, opened 2026-09-05
Resolve file descriptors once per syscall

### LINKED ISSUE #1261: One chunked syscall can consume multiple open-file descriptions after fd reuse
One bug found by [specula](https://github.com/specula-org/Specula)

sys_readv validates the raw fd once, then read_from_iovec calls self.sys_read(fd, ...) again for each iovec chunk. If another task sharing FilesState closes and reopens the same raw fd between chunks, one readv can return bytes from two different open-file descriptions.

See the full [report](https://github.com/specula-org/specula-case-studies/blob/data/litebox-case-study-20260831/systems/litebox/modules/core/runs/litebox-known-aware-rerun-20260830/confirmed-bugs.md#entry-3-one-chunked-syscall-can-consume-multiple-open-file-descriptions-after-fd-reuse) for more detail.

CvvT:
Resolve raw file descriptors once and retain the resulting `AnyTypedFd` throughout each syscall. This prevents chunked operations from switching to a different open-file description when another task concurrently closes and reuses the raw descriptor.

This updates file, memory-mapping, and socket syscalls, and consolidates typed-fd dispatch and narrowing.

Fixes #1261.

### DIFF
--- a/litebox_shim_linux/src/lib.rs
+++ b/litebox_shim_linux/src/lib.rs
@@ -22,7 +22,6 @@ use alloc::sync::Arc;
 use core::cell::{Cell, RefCell};
 use litebox::{
     LiteBox,
-    fd::TypedFd,
     mm::{PageManager, linux::PAGE_SIZE},
     net::Network,
     pipes::Pipes,
@@ -51,8 +50,6 @@ pub(crate) mod stdio;
 pub mod syscalls;
 pub mod transport;
 mod wait;
-
-use crate::syscalls::file::get_file_descriptor_flags;
 
 pub type DefaultFS<Platform> = LinuxFS<Platform>;
 
@@ -436,8 +433,9 @@ impl<Platform: ShimPlatform> Task<Platform> {
         let files = self.files.borrow();
         let alive_fds: Vec<usize> = files.raw_descriptor_store.read().iter_alive().collect();
         for raw_fd in alive_fds {
-            if let Ok(flags) = get_file_descriptor_flags(raw_fd, &self.global, &files)
-                && flags.contains(litebox_common_linux::FileDescriptorFlags::FD_CLOEXEC)
+            if let Ok(fd) = files.typed_fd_from_raw(raw_fd)
+                && syscalls::file::get_file_descriptor_flags(&fd, &self.global)
+                    .contains(litebox_common_linux::FileDescriptorFlags::FD_CLOEXEC)
             {
                 let _ = self.do_close(raw_fd);
             }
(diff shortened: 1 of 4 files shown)


wdcui:
## Static review findings (5-min budget, dual-model review: GPT-5.6 Sol + Opus 5)

Reviewed the actual diff vs `origin/main` (`litebox_shim_linux/src/lib.rs`, `syscalls/file.rs`, `syscalls/mm.rs`, `syscalls/net.rs`).

### 🟡 Medium — `mm.rs` ELF-patch path re-resolves the raw fd instead of reusing the pre-resolved handle

**Location:** `litebox_shim_linux/src/syscalls/mm.rs:136` and `:147` (feeding into `init_elf_patch_state` at `:548`, `:581`)

`do_mmap_file` resolves the descriptor once (`let typed_fd = self.typed_fd(fd)…` at line 121) and correctly threads `&typed_fd` into `try_cow_mmap_file`/`do_mmap_file_memcpy`. However, the ELF-rewriting path is still handed the raw `i32` fd:

- `maybe_patch_exec_segment(result, len, fd, …)` (:136) → `init_elf_patch_state(fd, …)` (:768)
- `init_elf_patch_state(fd, …)` (:147, direct call)

`init_elf_patch_state` then re-resolves the raw fd twice more via `self.sys_read(fd, &mut ehdr_buf, Some(0))` (:548) and `self.sys_read(fd, &mut phdrs_buf, Some(e_phoff))` (:581), each doing a fresh `raw_descriptor_store` lookup.

Since the fd table is shared across `CLONE_FILES` threads, a sibling thread can close/`dup2` over `fd` between the initial resolution (:121) and these later reads. Two concrete failure modes within a single `mmap` call:
1. `sys_read` returns `EBADF` → `init_elf_patch_state` silently skips patch-state init → `maybe_patch_exec_segment` treats it as "no patch needed" and reports the mapping successful but unpatched. For a pre-patched binary that requires the trampoline rewrite, this leads to a SIGSEGV on the first rewritten syscall instead of the intended fail path.
2. The slot was replaced by another file → ELF header/phdrs are read from a *different* file than the one actually mapped, and `elf_patch_cache` (keyed by the raw `i32`) gets populated with segment offsets from the wrong file.

This doesn't appear to be a regression introduced by this PR, but the PR's stated invariant ("resolve each raw fd once per syscall") isn't fully established for `mmap`'s ELF-patch path even though a resolved `AnyTypedFd` is already in scope — it reads as an incomplete migration.

**Suggested fix:** Change `maybe_patch_exec_segment` and `init_elf_patch_state` to take `&AnyTypedFd<Platform>` (keep the raw `i32` only as the `elf_patch_cache` key) and replace the two `self.sys_read(fd, …)` calls with `self.do_read(typed_fd, …)`, matching what's already done for `do_mmap_file_memcpy`.

**Confidence:** High — independently identified by both reviewers (GPT-5.6 Sol and Opus 5) reviewing this file/chunk separately.

---

### Verified clean (no action needed)
The rest of the diff (`file.rs`, `net.rs`, and the resolver logic in `lib.rs`) came back clean from both independent reviewers, with matching verified invariants:
- Every converted syscall resolves its fd exactly once per entry point (no re-resolution across chunked read/write, iovec, or `*mmsg` loops).
- Dispatch ordering and error-code precedence (EBADF/ENOTSOCK before EINVAL/EFAULT) match the removed `run_on_raw_fd`/`with_socket` code paths.
- No new use-after-free/refcount hazards: a concurrent close on a resolved `Arc<TypedFd>` fails closed (`EBADF`) rather than causing UAF or fd-confusion.
- No new lock re-entrancy or deadlock risk from holding `self.files.borrow()` across nested resolutions.

wdcui (approved):
LGTM. Thanks.

CvvT:
> ## Static review findings (5-min budget, dual-model review: GPT-5.6 Sol + Opus 5)
> Reviewed the actual diff vs `origin/main` (`litebox_shim_linux/src/lib.rs`, `syscalls/file.rs`, `syscalls/mm.rs`, `syscalls/net.rs`).
> 
> ### 🟡 Medium — `mm.rs` ELF-patch path re-resolves the raw fd instead of reusing the pre-resolved handle
> **Location:** `litebox_shim_linux/src/syscalls/mm.rs:136` and `:147` (feeding into `init_elf_patch_state` at `:548`, `:581`)
> 
> `do_mmap_file` resolves the descriptor once (`let typed_fd = self.typed_fd(fd)…` at line 121) and correctly threads `&typed_fd` into `try_cow_mmap_file`/`do_mmap_file_memcpy`. However, the ELF-rewriting path is still handed the raw `i32` fd:
> 
> * `maybe_patch_exec_segment(result, len, fd, …)` (:136) → `init_elf_patch_state(fd, …)` (:768)
> * `init_elf_patch_state(fd, …)` (:147, direct call)
> 
> `init_elf_patch_state` then re-resolves the raw fd twice more via `self.sys_read(fd, &mut ehdr_buf, Some(0))` (:548) and `self.sys_read(fd, &mut phdrs_buf, Some(e_phoff))` (:581), each doing a fresh `raw_descriptor_store` lookup.
> 
> Since the fd table is shared across `CLONE_FILES` threads, a sibling thread can close/`dup2` over `fd` between the initial resolution (:121) and these later reads. Two concrete failure modes within a single `mmap` call:
> 
> 1. `sys_read` returns `EBADF` → `init_elf_patch_state` silently skips patch-state init → `maybe_patch_exec_segment` treats it as "no patch needed" and reports the mapping successful but unpatched. For a pre-patched binary that requires the trampoline rewrite, this leads to a SIGSEGV on the first rewritten syscall instead of the intended fail path.
> 2. The slot was replaced by another file → ELF header/phdrs are read from a _different_ file than the one actually mapped, and `elf_patch_cache` (keyed by the raw `i32`) gets populated with segment offsets from the wrong file.
> 
> This doesn't appear to be a regression introduced by this PR, but the PR's stated invariant ("resolve each raw fd once per syscall") isn't fully established for `mmap`'s ELF-patch path even though a resolved `AnyTypedFd` is already in scope — it reads as an incomplete migration.
> 
> **Suggested fix:** Change `maybe_patch_exec_segment` and `init_elf_patch_state` to take `&AnyTypedFd<Platform>` (keep the raw `i32` only as the `elf_patch_cache` key) and replace the two `self.sys_read(fd, …)` calls with `self.do_read(typed_fd, …)`, matching what's already done for `do_mmap_file_memcpy`.
> 
> **Confidence:** High — independently identified by both reviewers (GPT-5.6 Sol and Opus 5) reviewing this file/chunk separately.
> 
> ### Verified clean (no action needed)
> The rest of the diff (`file.rs`, `net.rs`, and the resolver logic in `lib.rs`) came back clean from both independent reviewers, with matching verified invariants:
> 
> * Every converted syscall resolves its fd exactly once per entry point (no re-resolution across chunked read/write, iovec, or `*mmsg` loops).
> * Dispatch ordering and error-code precedence (EBADF/ENOTSOCK before EINVAL/EFAULT) match the removed `run_on_raw_fd`/`with_socket` code paths.
> * No new use-after-free/refcount hazards: a concurrent close on a resolved `Arc<TypedFd>` fails closed (`EBADF`) rather than causing UAF or fd-confusion.
> * No new lock re-entrancy or deadlock risk from holding `self.files.borrow()` across nested resolutions.

I intentionally left elf patch cache untouched because it requires larger changes than simply replacing raw fd with typed fd. I created an issue #1316 for it. It also has some other issue like #1268. It needs more thoughts than what this PR does.

PR #1304, opened 2026-09-05
Move filesystem authority into the broker

wdcui:
This PR moves filesystem authority, resolver and backend implementation, and open-file-description state into `litebox_broker_core` behind synchronous broker RPCs while keeping guest filesystem API types, path handling, and descriptor metadata in `litebox`, isolates Windows registry storage in a separate broker filesystem namespace, updates userland runners and broker startup to construct and require broker-owned filesystems, preserves device I/O, directory pagination, shared-offset, and mmap-copy behavior, and makes unsupported SNP filesystem startup fail explicitly.


wdcui:
Closing this large migration PR in favor of a smaller stack rebuilt from current ulitebox. The new stack starts with the file RPC ABI, then adds ObjectEntry::File and broker dispatch, moves the resolver/backend engine into broker core, wires transport and providers, and finally switches consumers atomically.

PR #1311, opened 2026-09-06
Add same-process userland broker connections

wdcui:
This PR adds `connect_in_process(&BrokerCore)` to `litebox_broker_local_userland`, composing the existing Linux Unix-socket/memfd and Windows named-pipe/shared-section transports with the reusable broker association runtime, authenticating the same-process peer, joining the broker host thread during association failure teardown, and covering concurrent requests and shutdown through the resulting production `BrokerConnection`.


wdcui:
Closing in favor of composing the existing production listener/connect APIs directly in the runner, without a separate connect_in_process abstraction.

PR #1314, opened 2026-09-07
Separate filesystem open-file-description state

wdcui:
This PR separates filesystem descriptor entries from shared open-file-description state by moving backend handles, access and status flags, and file offsets behind an Arc-backed platform mutex, allowing descriptor-entry locks to be released before backend I/O while preserving duplicated-descriptor sharing and independent-open semantics, with focused coverage for shared offsets, descriptor lifetime, and independent offsets.

### DIFF
--- a/litebox/src/fs/resolver.rs
+++ b/litebox/src/fs/resolver.rs
@@ -34,6 +34,24 @@ pub struct Resolver<
     backend: Backend,
 }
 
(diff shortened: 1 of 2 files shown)


wdcui:
Closing this preparation PR because the final design keeps only an immutable broker handle locally and introduces synchronized ResolverEntry state directly in broker core. The replacement stack starts with the durable file RPC ABI.

PR #1329, opened 2026-09-08
Read Linux process maps to EOF

### LINKED ISSUE #1328: Make Linux read_maps handle large /proc/self/maps output
`LinuxUserland::read_maps` reads `/proc/self/maps` into a fixed 8 KiB buffer and panics with `buffer too small` when the file reaches that limit. This caused the `litebox_runner_linux_userland::run::brokered_getrandom` test to fail in PR #1327 ([failed job](https://github.com/microsoft/litebox/actions/runs/34246791132/job/102130959539)); the failed job passed on rerun without code changes.

The failure is timing-dependent in in-process runner mode. The runner establishes the broker association before `Platform::new`, while the broker concurrently starts eight request workers, its readiness notifier, and transport threads. Their stack VMAs are added while `read_maps` snapshots the process mappings. A traced successful run read about 6.4 KiB while additional workers were still being created; CI scheduled enough thread creation before or during the read to fill the 8 KiB buffer. Repeated targeted local runs passed, confirming the threshold race.

Replace the fixed-size read with an EOF-driven implementation that handles allocation failure explicitly. The fix should also account for the existing stale-snapshot TODO: allocations and mappings created during or after the read may be omitted, so the implementation and tests should avoid implying a stronger reservation guarantee than the platform actually provides.

wdcui:
This PR replaces the Linux userland platform's fixed 8 KiB process-maps read with fallible EOF-driven accumulation and adds coverage for reads spanning multiple buffers, preventing in-process runner startup from failing as broker thread mappings grow. Fixes #1328.

### DIFF
--- a/litebox_platform_linux_userland/src/lib.rs
+++ b/litebox_platform_linux_userland/src/lib.rs
@@ -268,28 +268,23 @@ impl LinuxUserland {
         let Ok(fd) = fd else {
             return alloc::vec::Vec::new();
         };
-        let mut buf = [0u8; 8192];
-        let mut total_read = 0;
-        while total_read < buf.len() {
-            let n = unsafe {
+        let maps = Self::read_to_end(|buffer| {
+            // SAFETY: `fd` is open for reading and `buffer` is valid for writes
+            // of up to its full length.
+            unsafe {
                 syscalls::syscall3(
                     syscalls::Sysno::read,
                     fd,
-                    buf.as_mut_ptr() as usize + total_read,
-                    buf.len() - total_read,
+                    buffer.as_mut_ptr() as usize,
+                    buffer.len(),
                 )
             }
-            .expect("read failed");
-            if n == 0 {
-                break;
-            }
-            total_read += n;
-        }
-        assert!(total_read < buf.len(), "buffer too small");
+            .expect("read failed")
+        });
         unsafe { syscalls::syscall1(syscalls::Sysno::close, fd) }.expect("close failed");
 
         let mut reserved_pages = alloc::vec::Vec::new();
-        let s = core::str::from_utf8(&buf[..total_read]).expect("invalid UTF-8");
+        let s = core::str::from_utf8(&maps).expect("invalid UTF-8");
         for line in s.lines() {
             let parts: Vec<&str> = line.split_whitespace().collect();
             if parts.len() < 5 {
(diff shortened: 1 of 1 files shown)


litebox_platform_linux_userland/src/lib.rs
```diff
@@ -246,50 +246,74 @@ impl LinuxUserland {
...
+            // `openat` contract when no creation flag is present.
+            let fd = unsafe {
+                syscalls::syscall4(
+                    syscalls::Sysno::openat,
+                    AT_FDCWD,
+                    path.as_ptr() as usize,
+                    OFlags::RDONLY.bits() as usize,
+                    0,
+                )
+            };
```
CvvT, line 279: We don't need to open the file multiple times inside the loop?

litebox_platform_linux_userland/src/lib.rs
```diff
@@ -246,50 +246,74 @@ impl LinuxUserland {
...
         }
-        assert!(total_read < buf.len(), "buffer too small");
-        unsafe { syscalls::syscall1(syscalls::Sysno::close, fd) }.expect("close failed");
 
         let mut reserved_pages = alloc::vec::Vec::new();
-        let s = core::str::from_utf8(&buf[..total_read]).expect("invalid UTF-8");
+        let s = core::str::from_utf8(&maps).expect("invalid UTF-8");
         for line in s.lines() {
             let parts: Vec<&str> = line.split_whitespace().collect();
             if parts.len() < 5 {
```
CvvT, line 319: Why the parsing is outside the loop? Also, what if one line is truncated in one iteration?

CvvT:
It is unclear to me why we need this PR. The proc mapping is read at the very beginning with the assumption that litebox itself does not need much memory before it loads the guest program and thus a fixed-size buffer should suffice. Does that assumption break now?

wdcui:
> It is unclear to me why we need this PR. The proc mapping is read at the very beginning with the assumption that litebox itself does not need much memory before it loads the guest program and thus a fixed-size buffer should suffice. Does that assumption break now?

Over the weekend, I added the support for running the runner and the broker in a single process. In this mode, the broker is loaded before the runner/platform is loaded. When this function is called, there could be way more VMAs from the broker which may cause the file to exceed 8KB.

PR #1336, opened 2026-09-09
Fix Windows guest-entry interrupt gap

CvvT:
An interrupt arriving after shim entry but before `is_in_guest` is set is recorded as pending without redirecting execution. Previously, neither Windows resume path checked that flag before entering the guest, potentially delaying delivery indefinitely. This PR follows Linux platform to check the `interrupt` flag right after setting `is_in_guest`.


wdcui (approved):
LGTM. Thanks!

PR #1340, opened 2026-09-09
Preserve guest floating-point and extended CPU state on Windows

CvvT:
Preserve guest floating-point and SIMD state across syscalls, exceptions, and host interrupts, while ensuring host Rust code runs with Windows ABI floating-point controls.

- For syscall that does not update `rip`, save guest state with XSAVE and restore it with XRSTOR on the fast resume path.
- For exception and interrupt, preserve extended state through Windows context capture and NtContinue.
- For syscall that updates `rip`, save guest state with XSAVE but restore it with NtContinue.


litebox_platform_windows_userland/src/lib.rs
```diff
@@ -405,11 +448,294 @@ fn run_thread_inner(
...
+                mxcsr = in(reg) &raw mut mxcsr,
+                options(nostack, preserves_flags),
+            );
+        }
+        debug_assert_eq!(x87_control_word, HOST_X87_CONTROL_WORD);
+        debug_assert_eq!(mxcsr & HOST_MXCSR_CONTROL_MASK, HOST_MXCSR);
+    }
+}
+
+const XSAVE_LEGACY_SIZE: usize =
```
wdcui, line 486: @CvvT This part of code is more like platform-independent but architecture-dependent. Should we consider making it sharable between linux and windows platforms?

wdcui: How do we handle this in the Linux userland platform?

CvvT: Linux should be similar. I will move the common code when I add the support for Linux userland.

wdcui:
@sangho2 can you help review this PR? Thanks!

sangho2:
Do we need to worry about the below? perhaps out of this PR's scope (AArch64 Linux variant somehow handles it).

https://github.com/microsoft/litebox/blob/d90f306dce7ee01859b076fdc0d09e63eacb51d3/litebox_shim_linux/src/syscalls/signal/x86_64.rs#L97

litebox_platform_windows_userland/src/lib.rs
```diff
@@ -405,11 +450,294 @@ fn run_thread_inner(
...
+            assert!(core::arch::x86_64::__cpuid(0).eax >= 0x0d);
+            let feature_info = core::arch::x86_64::__cpuid(1);
+            assert_eq!(
+                feature_info.ecx & (CPUID_XSAVE | CPUID_OSXSAVE),
+                CPUID_XSAVE | CPUID_OSXSAVE,
+                "XSAVE must be supported and enabled by Windows",
+            );
+            let features = core::arch::x86_64::__cpuid_count(0x0d, 0);
+            let mask = unsafe { core::arch::x86_64::_xgetbv(0) };
+            assert_eq!(mask & 3, 3, "x87 and SSE state must be enabled");
```
sangho2, line 522: does it mean that AVX/AVX2 are optional?

sangho2 (approved):
Overall, looks good to me. We might want to use `XSAVEC` or `XAVEOPT` later, but better to have a correct implementation first.

CvvT:
> Do we need to worry about the below? perhaps out of this PR's scope (AArch64 Linux variant somehow handles it).
> 
> https://github.com/microsoft/litebox/blob/d90f306dce7ee01859b076fdc0d09e63eacb51d3/litebox_shim_linux/src/syscalls/signal/x86_64.rs#L97

Yes, we should, but probably after I also fix Linux userland platform.

litebox_platform_windows_userland/src/lib.rs
CvvT: Yes, it is optional. The code only assumes `XSAVE_FORMAT` that requires x87 and SSE state.

PR #1342, opened 2026-09-10
Fix Linux guest signal exit code reporting

sangho2:
This PR fixes Linux guest signal exit code reporting in the Linux userland runner. LiteBox currently passes `signal + 256` to `std::process::exit()` whereas the shell convention is `signal + 128`.


wdcui (approved):
LGTM. Thanks!

PR #1351, opened 2026-09-11
LVBS: Enable safe TA page table teardown

sangho2:
This PR enables safe TA page table teardown for LVBS/OP-TEE. It aligns the LVBS platform's page table teardown with the `Arc`-based ownership model, deferring page table destruction until the final owner drops it. Although this teardown strategy is unsafe in the general case (e.g., COW/shared mappings), it is safe for the LVBS/OP-TEE runner because it strictly controls session/instance lifecycles and does not share any TA page frames.

### DIFF
--- a/litebox_shim_optee/src/lib.rs
+++ b/litebox_shim_optee/src/lib.rs
@@ -11,7 +11,7 @@ extern crate alloc;
 use crate::loader::elf::ElfLoaderError;
 use crate::syscalls::pta::PseudoTa;
 use aes::{Aes128, Aes192, Aes256};
-use alloc::{sync::Arc, vec};
+use alloc::{boxed::Box, sync::Arc, vec};
 use core::cell::Cell;
 use ctr::Ctr128BE;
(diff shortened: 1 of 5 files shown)


litebox_shim_optee/src/lib.rs
```diff
@@ -285,6 +289,17 @@ impl<Platform: OpteeShimPlatform> Clone for OpteeShim<Platform> {
 }
 
 impl<Platform: OpteeShimPlatform> OpteeShim<Platform> {
+    /// Retains the task page table before the shim is shared.
```
wdcui: I don't follow this comment. Is this "task page table" for a TA or the kernel?

wdcui (approved):
LGTM. Thanks.

litebox_shim_optee/src/lib.rs
sangho2: Let me revise it. Here, a task page table is a TA/user page table. This `retaining` is nothing but for keeping the page table alive.

PR #1385, opened 2026-09-16
Fix O_PATH branch returns

jaybosamiya-ms:


jaybosamiya-ms:
This is just trivial fixups, not sure why I had not done these before, probably was thinking `O_PATH` needed special handling for these from an error perspective, but the right move is just the existing errors anyways

jaybosamiya-ms:
Hmm, indeed, needed some new variants, @wdcui's agent pointed out a few examples:
```
│ Consensus │ Severity │ Location                           │ Finding                                               │
├───────────┼──────────┼────────────────────────────────────┼───────────────────────────────────────────────────────┤
│ Both      │ Medium   │ litebox/src/fs/resolver.rs:752-754 │ lseek() on an O_PATH descriptor returns ESPIPE; Linux │
│           │          │                                    │ semantics require EBADF.                              │
├───────────┼──────────┼────────────────────────────────────┼───────────────────────────────────────────────────────┤
│ Both      │ Medium   │ litebox/src/fs/resolver.rs:955-957 │ getdents64() on an O_PATH directory returns ENOTDIR;  │
│           │          │                                    │ Linux semantics require EBADF.                        │
├───────────┼──────────┼────────────────────────────────────┼───────────────────────────────────────────────────────┤
│ Opus only │ Low      │ litebox/src/fs/resolver.rs:809-811 │ The new truncate() path_only branch is unreachable    │
│           │          │                                    │ because !write_allowed returns first. Consequently,   │
│           │          │                                    │ ftruncate() on an O_PATH descriptor still returns     │
│           │          │                                    │ EACCES instead of EBADF.
```

I've pushed a new change to add the new variants

wdcui (approved):
LGTM. Thanks!

PR #1386, opened 2026-09-16
Directory write check logic sits better in shim

jaybosamiya-ms:
This PR decides to explicitly diverge the LiteBox core file system semantics from POSIX.  It is part of the longer-term plan to move away from the `OFlags` and use a more broader-applicable abstraction there anyways.  Roughly, POSIX requires that `open` of `O_WRONLY` or such on a directory should return an `EISDIR`.  However, this is a shim-level concern, and the core should not be concerned with this sort of semantics.  Thus, I've moved this decision into the shim.


litebox_shim_linux/src/syscalls/file.rs
```diff
@@ -365,10 +365,23 @@ impl<Platform: ShimPlatform> Task<Platform> {
...
+                .fs
+                .fd_file_status(&file)
+                .map_err(Errno::from)?
+                .file_type
+                == litebox::fs::FileType::Directory
+        {
+            files.fs.close(&file).unwrap();
+            return Err(Errno::EISDIR);
+        }
+        Ok(file)
```
jaybosamiya-ms, line 387: Yes, this does mean that any writable `open` of a file suffers an additional status check, but I think the simplification of the semantics and the unblocking of the Windows side (on `ulitebox`) is worth this tiny and rare inefficiency.  Most programs are not constantly opening+closing writable files: usually the write itself is the bottleneck, so I think this is a reasonable tradeoff to make.

jaybosamiya-ms:
Related: #1255 (specifically, see a871fb5a0d0b0811cc103a6ea194eff6a5076bd8) and #894 (afe6cddcdb01230cb63b15f154fbf97be427b652)

wdcui (approved):
LGTM. Thanks.

PR #1395, opened 2026-09-18
Use bottom-up ELF placement policy

sangho2:
This PR replaces the Linux shim loader's top-down fallback placement policy with a bottom-up policy. Currently, if placement of the main ELF at `DEFAULT_LOW_ADDR` fails, the loader searches downward from `TASK_ADDR_MAX` for a suitable gap without accounting for trampoline allocation or `brk`-based heap growth. Instead, this PR places the ELF in the first suitable gap at or above `DEFAULT_LOW_ADDR`.


sangho2:
Since now both Linux and OP-TEE shims require bottom-up page placement, the core `litebox` crate might need to provide an API for it for performance.

wdcui (approved):
LGTM. Thanks.

CvvT:
> Since now both Linux and OP-TEE shims require bottom-up page placement, the core `litebox` crate might need to provide an API for it for performance.

`litebox_shim_windows` also needs this. I'm working on it.

PR #1399, opened 2026-09-21
Verify TAs

athvu:
Add a feature signed-ta-rsa to enable TA signature verification. A verification key can be embedded through the LITEBOX_TA_VERIFY_KEY build variable.
Built-in TAs can either be unsigned .elf or signed .ta files. Dynamic TAs must be signed.


praveen-pk:
Why make this a compile time feature: `signed-ta-rsa`? Why not determine the binary type dynamically, by checking the MAGIC value and take appropriate steps?

litebox_shim_optee/src/lib.rs
```diff
@@ -230,10 +277,10 @@ impl<Platform: OpteeShimPlatform> GlobalState<Platform> {
             Some(ta_bin)
         } else {
             let ta_bin = Self::rpc_get_ta_bin(ta_uuid)?;
-            if !self.store_ta_bin(ta_uuid, &ta_bin) {
+            if !self.store_ta_bin(ta_uuid, &ta_bin, TaSource::Dynamic) {
```
praveen-pk, line 280: fyi, the place to store_ta_bin changed with the open Dynamic TA Support PR (#1213).

So, this should be revisited later

PR #1407, opened 2026-09-22
Fix LinuxUserland::read_maps to handle large /proc/self/maps output

dywongcloud:
As described in [[issue #1328](https://github.com/microsoft/litebox/issues/1328)](https://github.com/microsoft/litebox/issues/1328), `read_maps()` could panic when `/proc/self/maps` exceeded its fixed 8 KiB buffer, sometimes triggered by worker threads starting during the read.

Read into a growable buffer until EOF, retaining partial data if allocation fails. Parse the raw bytes so non-UTF-8 pathnames do not prevent parsing. Add a regression test with thousands of mappings that exceed the old limit.


dywongcloud:
@wdcui 
@sangho2

CvvT:
Thanks for the contribution! However, I'm working on refactoring/redesigning the memory management interfaces and this PR would become unnecessary then.

